PR #66, opened 2015-07-31
Login & Logout

krasa:
Hello,

as you asked on your stream, here's some contribution. 

I don't know if it's any good, I have never really done anything in JavaFX. :-)

That PircBotX might need some attention, it likes to throw exceptions when logging out immediately after login.


Cheyans:
Hmm, with the current branch the failed login hanging brings up a bug in one of my additions, I think your change will resolve this.

micheljung:
Hi. thanks for your contribution, I'll look into it

Cheyans:
Can we decide what we want to do with this?

PR #249, opened 2016-06-18
Feature/179 join via chat

Brutus5000:
- Introduced JoinGameHelper as general utility for joining games from UI side.
- Reworked password dialog
- Connected GameTilesController and GamesTableController to new JoinGameHelper
- Added tests for JoinGameHelper and GameTileController
- Small changes for better test


micheljung:
thxalot!

PR #270, opened 2016-06-27
#183 apply chat filter to new users

### LINKED ISSUE #183: Chat filter is no longer applied for people joining
When the chat user list is filtered by a name in the search field, this filter is no longer applied to people joining afterwards.

I say "no longer" because I'm 99% sure I explicitly implemented this case, and that it worked :-)

Cheyans:
closes #183


micheljung:
Horray! :dancers:

PR #291, opened 2016-07-27
#290 uid outdated and linux support for uid

### LINKED ISSUE #290: uid outdated and linux support for uid
The uid is segfaulting on linux and outdated. Also it's being launched the wrong way.

yenon:
uid outdated and linux support for uid
Also updated windows uid.exe


micheljung:
thanks a lot!

PR #308, opened 2016-08-07
Fix for #298 (IRC reconnect)

### LINKED ISSUE #298: "Reconnect" only affects #aeolus
**Steps to reproduce:**
- join one or more custom channels with other users in it
- click in the bottom right corner on "Chat connected" and then "Reconnect"

**Expected behavior:** the user lists in all open tabs should be renewed
**Observed behavior:** only the #aeolus user list gets renewed

Brutus5000:
Fixes #298.
<br />There is no test. Realistic simulation of connection, disconnection, reconnection with rejoining is not working in the current codebase. Existing tests are working around this with some dirty hacks (not actually simulating a complete IRC server/client communication).


Brutus5000:
Approach no longer works due to recent changes in develop.

PR #309, opened 2016-08-07
Fix for #294 (PlayerTooltip missing)

### LINKED ISSUE #294: Player tooltip is broken in chat user list
It's just empty

Brutus5000:
Uses the existing PlayerCardTooltip, even though it doesn't add much information.

Note: clan tags still show an empty popup!

(Connects to #294 - waffle.io)

### DIFF
--- a/src/main/java/com/faforever/client/chat/ChatUserItemController.java
+++ b/src/main/java/com/faforever/client/chat/ChatUserItemController.java
@@ -276,10 +276,18 @@ public class ChatUserItemController {
       return;
     }
 
-    // FIXME #294
+    // Create the username tooltip
+    PlayerCardTooltipController playerCardTooltipController = applicationContext.getBean(PlayerCardTooltipController.class);
+    playerCardTooltipController.setPlayer(playerInfoBean);
+
+    Tooltip statusTooltip = new Tooltip();
+    statusTooltip.setGraphic(playerCardTooltipController.getRoot());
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/chat/ChatUserItemController.java
```diff
@@ -311,6 +312,20 @@ void onMouseEnterGameStatus() {
   }
 
   @FXML
+  void onMouseEnterUsername() {
+    if (playerInfoBean.getChatOnly()) {
+      return;
+    } else {
```
micheljung: Bei einem `return;` ist das `} else {` überflüssig

src/main/java/com/faforever/client/chat/ChatUserItemController.java
```diff
@@ -311,6 +312,20 @@ void onMouseEnterGameStatus() {
...
+      PlayerCardTooltipController playerCardTooltipController = applicationContext.getBean(PlayerCardTooltipController.class);
+      playerCardTooltipController.setPlayer(playerInfoBean);
+
+      Tooltip statusTooltip = new Tooltip();
+      statusTooltip.setGraphic(playerCardTooltipController.getRoot());
+      Tooltip.install(usernameLabel, statusTooltip);
+    }
```
micheljung: This installs a tooltip on every mouse-over, even if it has already been installed.

Also, there already is a tooltip installed. See `configureTooltip()`

PR #472, opened 2017-01-12
Bug/#459 faf icon not showing

Geosearchef:
Not quite happy with this fix, but as fonts are loaded automatically, I don't know how to do it differently. This way shouldn't be expensive and worked reliably for me.


axel1200:
What about having font defined by attribute in Fxml file font-famliy (should be that attribute)?

.idea/misc.xml
```diff
@@ -7,7 +7,7 @@
       <item index="1" class="java.lang.String" itemvalue="org.junit.Test" />
     </list>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_3" default="false" project-jdk-name="1.8" project-jdk-type="JavaSDK">
```
micheljung: Language level 1.3? NO? :D

.idea/modules.xml
```diff
@@ -4,6 +4,8 @@
     <modules>
       <module fileurl="file://$PROJECT_DIR$/.idea/modules/downlords-faf-client.iml" filepath="$PROJECT_DIR$/.idea/modules/downlords-faf-client.iml" />
       <module fileurl="file://$PROJECT_DIR$/.idea/modules/downlords-faf-client_main.iml" filepath="$PROJECT_DIR$/.idea/modules/downlords-faf-client_main.iml" group="downlords-faf-client" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/downlords-faf-client_main.iml" filepath="$PROJECT_DIR$/.idea/modules/downlords-faf-client_main.iml" group="downlords-faf-client" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/downlords-faf-client_test.iml" filepath="$PROJECT_DIR$/.idea/modules/downlords-faf-client_test.iml" group="downlords-faf-client" />
```
micheljung: Please revert

src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -52,6 +54,11 @@ public void initialize() {
     loginProgressPane.setVisible(false);
     loginErrorLabel.managedProperty().bind(loginErrorLabel.visibleProperty());
     loginErrorLabel.setVisible(false);
+
+    // fixes #459
+    // "re" loads the font for the FAF logo and warning icons on the login page
+    // these are defined in style.css ".icon" but not used if WindowController.maximize() is called at stage initialization
+    Font.loadFont(LoginController.class.getResource("/font/dfc-icons.ttf").toExternalForm(), 10);
```
micheljung: I just realize that this font isn't loaded anywhere anymore, not even in the CSS? This was never intended :) You can load this font in the class `Main` like so:
```
    Font.loadFont(getClass().getResourceAsStream("/font/dfc-icons.ttf"), defaultFontSize);
```

And the Open Sans font should be loaded in the CSS file (I could swear it was there....). However, this, unfortunately, causes issues with SceneBuilder (doesn't render style anymore). I think it was something like this:

```
@font-face {
   src: url("@../font/dfc-icons.ttf");
}
```
or:
```
@font-face {
   src: url("/font/dfc-icons.ttf");
}
```

IIRC one worked in SceneBuilder, while the other worked in the program.

.idea/misc.xml
Geosearchef: No, definetly not.

Issue #306, opened 2016-08-04
Switch to constructor injection


micheljung:
Originally I was using `@Resource` to autowire fields. As part of #307, this was changed to `@Inject`. Next step is to use constructor injection instead of field injection.

Since this is a lot of work and affects almost every file, this should be done gradually.

Example:

```
  private final PreferencesService preferencesService;

  @Inject
  public FaInitGenerator(PreferencesService preferencesService) {
    this.preferencesService = preferencesService;
  }
```

Geosearchef:
circular reference between gameservice and replayservice:

https://github.com/FAForever/downlords-faf-client/blob/develop/src/main/java/com/faforever/client/game/GameServiceImpl.java#L116

https://github.com/FAForever/downlords-faf-client/blob/develop/src/main/java/com/faforever/client/replay/ReplayServiceImpl.java#L88

Geosearchef:
Another circular reference in unused class ForgedAllianceServiceImpl. (see TODO)

PR #476, opened 2017-01-18
Feature/#318 focus game on start

Geosearchef:
Game should now be shown, restored and brought to foreground on game start/click on "game full" notification.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -264,8 +264,8 @@ dependencies {
   }
   compile("com.google.guava:guava:18.0")
   compile("org.apache.commons:commons-compress:1.9")
-  compile("net.java.dev.jna:jna:4.1.0")
-  compile("net.java.dev.jna:jna-platform:4.1.0")
+  compile("net.java.dev.jna:jna:4.2.1")
+  compile("net.java.dev.jna:jna-platform:4.2.1")
   compile("org.eclipse.jgit:org.eclipse.jgit:4.5.0.201609210915-r")
   compile("org.jetbrains:annotations:13.0")
   compile("net.sf.jtidy:jtidy:r938")
--- a/src/main/java/com/faforever/client/fx/PlatformServiceImpl.java
+++ b/src/main/java/com/faforever/client/fx/PlatformServiceImpl.java
@@ -96,7 +96,7 @@ public class PlatformServiceImpl implements PlatformService {
   public String getForegroundWindowTitle() {
     HWND window = User32.INSTANCE.GetForegroundWindow();
 
-    if (window == null) {
+    if (window != null) {
       char[] textBuffer = new char[255];
       User32.INSTANCE.GetWindowText(window, textBuffer, 255);
       return new String(textBuffer).trim();
--- a/src/main/java/com/faforever/client/fx/PlatformServiceImpl.java
+++ b/src/main/java/com/faforever/client/fx/PlatformServiceImpl.java
@@ -5,6 +5,7 @@ import com.sun.jna.platform.win32.WinDef.HWND;
 import com.sun.jna.platform.win32.WinUser;
 import com.sun.jna.platform.win32.WinUser.WINDOWPLACEMENT;
 import javafx.application.HostServices;
+import org.springframework.beans.factory.annotation.Value;
 
 import java.nio.file.Path;
 
@@ -14,6 +15,8 @@ import static org.bridj.Platform.show;
 public class PlatformServiceImpl implements PlatformService {
 
   private final HostServices hostServices;
+  @Value("${forgedAlliance.windowTitle}")
+  private String faWindowTitle;
 
   public PlatformServiceImpl(HostServices hostServices) {
     this.hostServices = hostServices;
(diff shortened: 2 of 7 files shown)


Geosearchef:
I need help testing this feature, since I can only run one instance of the game at a time.

build.gradle
```diff
@@ -244,8 +244,8 @@ dependencies {
   compile 'org.slf4j:slf4j-api:1.7.2'
   compile 'org.apache.commons:commons-compress:1.9'
   compile 'org.apache.httpcomponents:httpclient:4.4.1'
-  compile 'net.java.dev.jna:jna:4.1.0'
-  compile 'net.java.dev.jna:jna-platform:4.1.0'
+  compile 'net.java.dev.jna:jna:4.2.1'
+  compile 'net.java.dev.jna:jna-platform:4.2.1'
```
Geosearchef: needed for gett WINDOWPLACEMENT

src/main/java/com/faforever/client/fx/PlatformServiceImpl.java
```diff
@@ -13,6 +15,8 @@
 public class PlatformServiceImpl implements PlatformService {
 
   private final HostServices hostServices;
+  @Value("${forgedAlliance.windowTitle}")
+  private String faWindowTitle;
```
Geosearchef: This value will currently be null, how do I cause it to be initialized, is this even a good idea? I wouldn't pass the faWindowTitle from every code line where I want to focus the game, since this should be (in my opionon) be the task of the PlatformService, this also applies to the focusGameWindow, startFlashingGameWindow, .... below.

src/main/java/com/faforever/client/fx/PlatformServiceImpl.java
```diff
@@ -14,8 +16,15 @@
 
   private final HostServices hostServices;
 
+  private boolean isOnWindows;
+
+  @Value("${forgedAlliance.windowTitle}")
+  private String faWindowTitle;
```
Geosearchef: This value will currently be null, how do I cause it to be initialized, is this even a good idea? I wouldn't pass the faWindowTitle from every code line where I want to focus the game, since this should be (in my opionon) be the task of the PlatformService, this also applies to the focusGameWindow, startFlashingGameWindow, .... below.

micheljung: I agree with the change.
`PlatformServiceImpl` is instantiated a little differently than other beans (in Main.java), let's clean that up.
I think the result of `getHostServices()` should be added as a bean `hostServices` and the PlatformServiceImpl should get a @Service annotation and have `HostServices` injected.

src/main/java/com/faforever/client/fx/PlatformServiceImpl.java
```diff
@@ -67,11 +127,28 @@ public void stopFlashingWindow(String windowTitle) {
...
+
+    HWND window = User32.INSTANCE.GetForegroundWindow();
+
+    if (window != null) {
+      char[] textBuffer = new char[255];
+      User32.INSTANCE.GetWindowText(window, textBuffer, 255);
+      return new String(textBuffer).trim();
+    } else {
+      return null;
+    }
```
micheljung:
```
  if (window == null) {
     return null;
  }

  char[] textBuffer = new char[255];
  User32.INSTANCE.GetWindowText(window, textBuffer, 255);
  return new String(textBuffer).trim();
```

Reduces nesting and favors "guards" where the method returns on invalid conditions, leaving the "actual" code always at the end.

src/main/java/com/faforever/client/game/GameServiceImpl.java
```diff
@@ -121,6 +122,8 @@
   IceAdapter iceAdapter;
   @Inject
   ModService modService;
+  @Inject
+  PlatformService platformService;
```
micheljung: Have a look at #306

src/main/java/com/faforever/client/game/GameServiceImpl.java
```diff
@@ -492,6 +495,12 @@ private void onGameInfo(GameInfoMessage gameInfoMessage) {
         currentGame.set(game);
       }
     }
+
+    game.statusProperty().addListener((observable, oldValue, newValue) -> {
+      if (oldValue == GameState.OPEN && newValue == GameState.PLAYING && game.getTeams().values().stream().anyMatch(team -> team.contains(currentPlayer.getUsername())) && !platformService.isGameWindowFocused()) {
```
micheljung:
```
if (oldValue == GameState.OPEN
    && newValue == GameState.PLAYING
    && game.getTeams().values().stream().anyMatch(team -> team.contains(currentPlayer.getUsername()))
    && !platformService.isGameWindowFocused()) {
```
for readability

src/main/java/com/faforever/client/game/GameServiceImpl.java
```diff
@@ -493,6 +496,15 @@ private void onGameInfo(GameInfoMessage gameInfoMessage) {
...
     }
+
+    game.statusProperty().addListener((observable, oldValue, newValue) -> {
+      if (oldValue == GameState.OPEN
+          && newValue == GameState.PLAYING
+          && game.getTeams().values().stream().anyMatch(team -> team.contains(currentPlayer.getUsername()))
+          && !platformService.isWindowFocused(faWindowTitle)) {
+        platformService.focusWindow(faWindowTitle);
+      }
+    });
```
Geosearchef: doesn't this add a listener each time the game is updated?

PR #519, opened 2017-05-17
Fixes #507 images for news categories

### LINKED ISSUE #507: Different images for news posts
On the left side of every news post, there's currently the same icon. The idea, however, was to display a different icon based on the news category.

## List of categories to support
The list of categories is not well defined. The ones that should be supported are the following:

- `Server Update`
- `Tournament`
- `FA Update`
- `Lobby Update`
- `Balance`
- `website`
- `Cast`
- `Podcast`
- `featured mods`
- `Development Update`
- `Uncategorized`

Since the letter casing is inconsistent, the implementation needs to be case insensitive.

## Choosing the right category

- If there are multiple categories, the first match according to the order above should be chosen.
- If there are no categories, assume `Uncategorized`
- If all categories are unknown, assume `Uncategorized`

## Images to use

The images are not yet available. Just let all categories point to the current image, I'll take care of it later.

## Implementation tips

- The news is fetched in `LegacyNewsService` from http://direct.faforever.com/news/feed.
- Create an enum for the categories, containing their lowercase string (see how other enums are implemented, like `Faction`)
- Compare the category string from the news item using `toLowerCase(Locale.US)`

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43686202-different-images-for-news-posts?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

axel1200:
Introduces an Enum for NewsCategories

### DIFF
--- a/src/main/java/com/faforever/client/news/LegacyNewsService.java
+++ b/src/main/java/com/faforever/client/news/LegacyNewsService.java
@@ -4,6 +4,7 @@ import com.faforever.client.config.CacheNames;
 import com.faforever.client.config.ClientProperties;
 import com.faforever.client.preferences.PreferencesService;
 import com.google.common.eventbus.EventBus;
+import com.rometools.rome.feed.synd.SyndCategory;
 import com.rometools.rome.feed.synd.SyndEntry;
 import com.rometools.rome.feed.synd.SyndFeed;
 import com.rometools.rome.io.SyndFeedInput;
@@ -82,7 +83,14 @@ public class LegacyNewsService implements NewsService {
       String content = syndEntry.getContents().get(0).getValue();
       Date publishedDate = syndEntry.getPublishedDate();
 
-      result.add(new NewsItem(author, link, title, content, publishedDate));
+      NewsCategory newsCategory = syndEntry.getCategories().stream()
+          .filter(Objects::nonNull)
+          .findFirst()
+          .map(SyndCategory::getName)
+          .map(NewsCategory::fromString)
+          .orElse(NewsCategory.UNCATEGORIZED);
+
+      result.add(new NewsItem(author, link, title, content, publishedDate, newsCategory));
     }
 
     return result;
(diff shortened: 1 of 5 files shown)


axel1200:
Can test enum!

src/main/java/com/faforever/client/news/LegacyNewsService.java
```diff
@@ -81,8 +81,9 @@ private void pollForNews() {
       String title = syndEntry.getTitle();
       String content = syndEntry.getContents().get(0).getValue();
       Date publishedDate = syndEntry.getPublishedDate();
+      NewsCategory newsCategory = NewsCategory.fromString(syndEntry.getCategories().get(0).getName());
```
micheljung: `get(0)` fails if there are no categories. 

```
NewsCategory newsCategory = syndEntry.getCategories().stream()
  .filter(Objects::nonNull)
  .findFirst()
  .map(Category::getName)
  .map(NewsCategory::fromString)
  .orElse(NewsCategory.UNCATEGORIZED)
```
or whatever compiles.

src/main/java/com/faforever/client/news/LegacyNewsService.java
```diff
@@ -81,8 +82,9 @@ private void pollForNews() {
       String title = syndEntry.getTitle();
       String content = syndEntry.getContents().get(0).getValue();
       Date publishedDate = syndEntry.getPublishedDate();
-
-      result.add(new NewsItem(author, link, title, content, publishedDate));
+      NewsCategory newsCategory = syndEntry.getCategories().stream().filter(Objects::nonNull).findFirst().map(SyndCategory::getName)
+          .map(NewsCategory::fromString).orElse(NewsCategory.UNCATEGORIZED);
```
micheljung: Please format properly, like in the example I gave you ;)

```
......stream()
  .filter()
  .findFirst()
  .map()
  .map()
  .orElse()
```

PR #520, opened 2017-05-20
Fixes #505 focuses window on click of notification

### LINKED ISSUE #505: Chat with friends
If a friend joins u get a toast to start a conversation and this works, but it should also open the chat tab otherwise this is useless.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/42559683-chat-with-friends?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

axel1200:
Opens Chat Tab before creating new chat Event, so user does not have to click on the chat tab if he intends to start a conversation

### DIFF
--- a/src/main/java/com/faforever/client/player/FriendOnlineNotifier.java
+++ b/src/main/java/com/faforever/client/player/FriendOnlineNotifier.java
@@ -4,6 +4,8 @@ import com.faforever.client.audio.AudioService;
 import com.faforever.client.chat.InitiatePrivateChatEvent;
 import com.faforever.client.chat.SocialStatus;
 import com.faforever.client.i18n.I18n;
+import com.faforever.client.main.NavigateEvent;
+import com.faforever.client.main.NavigationItem;
 import com.faforever.client.notification.NotificationService;
 import com.faforever.client.notification.TransientNotification;
 import com.faforever.client.util.IdenticonUtil;
@@ -52,7 +54,10 @@ public class FriendOnlineNotifier {
               i18n.get("friend.nowOnlineNotification.title", username),
               i18n.get("friend.nowOnlineNotification.action"),
               IdenticonUtil.createIdenticon(player.getId()),
-              actionEvent -> eventBus.post(new InitiatePrivateChatEvent(username))
+              actionEvent -> {
+                eventBus.post(new NavigateEvent(NavigationItem.CHAT));
+                eventBus.post(new InitiatePrivateChatEvent(username));
+              }
           ));
     }
   }


axel1200:
@micheljung Somethings wrong here, some tests fail, even though i added a line that' not even tested

src/main/java/com/faforever/client/fx/WindowController.java
```diff
@@ -1,7 +1,7 @@
 package com.faforever.client.fx;
 
 import com.faforever.client.theme.UiService;
-import javafx.collections.ObservableList;
+import com.sun.javafx.util.Utils;
```
micheljung: (Almost) all `com.sun` are JVM internal and do not exist on other JVM implementations. Also, such package names may change with any update without prior notice, breaking compatibility ([like it happened to ControlsFX]("com.sun.javafx.Utils" changes to "com.sun.javafx.util.Utils" in 1.8.0_60ea))

This, this import should be removed. Keep the old code or, if easy, copy the code from `Utils.getVisualBounds` into our `JavaFxUtil`

src/main/java/com/faforever/client/notification/TransientNotificationController.java
```diff
@@ -112,6 +113,10 @@ public void onClicked(MouseEvent event) {
     if (event.getButton().equals(MouseButton.SECONDARY)) {
       dismiss();
     } else {
+      if (!StageHolder.getStage().isFocused()) {
+        StageHolder.getStage().setIconified(false);
+        StageHolder.getStage().requestFocus();
```
micheljung: Don't call `StageHolder.getStage()` repeatedly, extract to variable (select, CTRL+ALT+V)

src/main/java/com/faforever/client/player/FriendOnlineNotifier.java
```diff
@@ -51,7 +53,10 @@ public void onUserOnline(UserOnlineEvent event) {
               i18n.get("friend.nowOnlineNotification.title", username),
               i18n.get("friend.nowOnlineNotification.action"),
               IdenticonUtil.createIdenticon(player.getId()),
-              actionEvent -> eventBus.post(new InitiatePrivateChatEvent(username))
+              event1 -> {
```
micheljung: Avoid variable names like `event1`

src/main/resources/i18n/messages.properties
```diff
@@ -144,7 +144,7 @@ clan.leader=Leader
 clan.clanMembers=Members
 clan.messageLeader=Message clan leader
 clan.visitPage=Visit clan website
-chat.privateMessage.playerLeft={0} is now offline
+chat.privateMessage.playerLeft={0} is now offline and left chat
```
micheljung: Isn't that implicitly clear to the user? ;)
"Person X died and does no longer speak"

axel1200: The problem is if you are chatting and the user is your friend what should be quite common then two Notifications are shown both containing the same message.  It therfore looks like a bug. Acctually i did not want to commit that change to this branch, but yeah. So we either change the messanges or make clear only one is shown in this particular case....

axel1200:
**** I manged to delete most of this branch while rebasing

src/main/java/com/faforever/client/fx/WindowController.java
micheljung: threw ;)

micheljung:
Didn't test it, but looks good. Solve conflicts and I'll merge :)

src/main/java/com/faforever/client/fx/WindowController.java
```diff
@@ -60,12 +58,11 @@ public WindowController(UiService uiService) {
...
-    double y1 = stage.getY() + (stage.getHeight() / 2);
-
-    Rectangle2D windowCenter = new Rectangle2D(x1, y1, 1, 1);
-    ObservableList<Screen> screensForRectangle = Screen.getScreensForRectangle(windowCenter);
-    return screensForRectangle.get(0).getVisualBounds();
+    return JavaFxUtil.getScreenForRectangle(
+        new Rectangle2D(stage.getX(),
+            stage.getY(),
+            stage.getWidth(),
+            stage.getHeight())).getVisualBounds();
```
micheljung: _Why_ was this necessary? I can't believe that Java's `Screen.getScreensForRectangle()` doesn't work correctly.

axel1200: I thruw nullpointer Exceptions

axel1200: occaisonally

PR #529, opened 2017-06-06
Fixes #528 by opening an Immediate Notifcation displaying the error...

### LINKED ISSUE #528: Abort login on exceptions
If an exception is thrown during the login, the login screen seems to keep trying to login, even though it will never succeed.

How to reproduce: run the client from IntelliJ without faf-uid.exe
The following `com.github.nocatch.NoCatchException: java.io.IOException: Cannot run program "lib/faf-uid.exe": CreateProcess error=2, Das System kann die angegebene Datei nicht finden` exception is not converted into a ui error message.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47298918-abort-login-on-exceptions?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

axel1200:


src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -427,8 +432,22 @@ private void onSessionInitiated(SessionMessage sessionMessage) {
   }
 
   private void logIn(String username, String password) {
-    String uniqueId = uidService.generate(String.valueOf(sessionId.get()), preferencesService.getFafDataDirectory().resolve("uid.log"));
-    writeToServer(new LoginClientMessage(username, Hashing.sha256().hashString(password, UTF_8).toString(), sessionId.get(), uniqueId, localIp));
+    try {
+      String uniqueId = uidService.generate(String.valueOf(sessionId.get()), preferencesService.getFafDataDirectory().resolve("uid.log"));
+      writeToServer(new LoginClientMessage(username, Hashing.sha256().hashString(password, UTF_8).toString(), sessionId.get(), uniqueId, localIp));
+    } catch (IOException e) {
+      onUIDNotFound(e);
```
micheljung: `IOException` doesn't mean "not found". It could also be access denied, non-readable, failure when reading etc. So the method should be renamed. Thus, the method should be renamed.

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -427,8 +432,22 @@ private void onSessionInitiated(SessionMessage sessionMessage) {
...
+    }
+  }
+
+  @Override
+  protected void onUIDNotFound(Exception e) {
+    if (e.getMessage() == null) {
+      return;
+    }
```
micheljung: We shouldn't swallow exceptions just because there is no message. At least log it, additionally we can display an error to the user just without the message.

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -427,8 +432,22 @@ private void onSessionInitiated(SessionMessage sessionMessage) {
...
+    notificationService.addNotification(new ImmediateNotification(i18n.get("UIDNotFound"), e.getMessage(), Severity.ERROR,
```
micheljung: Whenever displaying an exception message to the user, use `e.getLocalizedMessage()`

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -427,8 +432,22 @@ private void onSessionInitiated(SessionMessage sessionMessage) {
...
+        Collections.singletonList(new ReportAction(i18n, reportingService, e))));
+    logger.error("UID.exe not found", e);
```
micheljung: * It may or may not be called "UID.exe" (and in fact, it's not) so don't assume a file name. Either use the correct name or use an error message like "UID binary could not be found"
* Just because there was an `IOException` doesn't mean "not found"

src/main/java/com/faforever/client/remote/AbstractServerAccessor.java
```diff
@@ -57,4 +57,6 @@ void close() throws IOException {
     stopped = true;
     IOUtils.closeQuietly(dataInput);
   }
+
+  protected abstract void onUIDNotFound(Exception e);
```
micheljung: This method is only called from the implementation itself and should therefore not be declared here, and private.

src/main/resources/i18n/messages.properties
```diff
@@ -33,7 +33,7 @@ main.community.news=News
 main.chat=Chat
 main.play=Play
 main.vault=Vault
-main.units=Units
+main.units=UnitsFUIV`D
```
micheljung: What? :P

axel1200: No idea how that happend XD

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -427,8 +433,22 @@ private void onSessionInitiated(SessionMessage sessionMessage) {
...
+  }
+
+  @VisibleForTesting
+  protected void onUIDNotExecuted(Exception e) {
+    if (e.getMessage() == null) {
+      return;
+    }
+    notificationService.addNotification(new ImmediateNotification(i18n.get("UIDNotExecuted"), e.getMessage(), Severity.ERROR,
+        Collections.singletonList(new ReportAction(i18n, reportingService, e))));
+    logger.error("UID.exe not executed", e);
```
micheljung: In case there is an exception without a message, it never gets logged and we'll never know anything about it (because `if(e.getMessage() == null) return;` above). If you move this log statement to the first line of the method you can solve this problem

src/main/resources/i18n/messages_de_DE.properties
```diff
@@ -566,3 +565,4 @@ updater.downloadingFile=Lade {0} herunter
 updater.patchingFile=Patche {0}
 achievement.stepsFormat={0,number,#}/{1,number,#}
 urlPreviewDescription={0} ({1})
+UIDNotExecuted=UID executable konnte nicht ausgef�hrt werden.
```
micheljung: Im englischen Satz stehen Mehr Informationen :)

src/main/resources/i18n/messages.properties
```diff
@@ -559,3 +559,4 @@ reviewScore.four=4
 reviewScore.three=3
 reviewScore.two=2
 reviewScore.one=1
+UIDNotExecuted=UID binary could not be executed. It might be removed, not found or unaccessable.
```
micheljung: inaccessible
might have been removed

Issue #138, opened 2015-12-11
Support replay IDs in chat


micheljung:
People often enter replay IDs in chat, like `#12345`. Currently there is an IRC bot who recognizes this and posts a link in order to download the replay. Instead of this, the client should turn such IDs into links.

Whenever a user clicks on such a link, the client should open the replay in the replay vault.

themasch:
Is there any way the client could detect its a proper replay id instead of just checking for /#\d+/? 
I guess, if possible, we should prevent creating links in messages like "`I'm the #1 supernoob!`" .

micheljung:
Other than checking for whether a replay actually exists (but that should be the case for about numbers 1-3'000'000, we could have a different format like `replay:1234` but that seems less convenient ;) And I think the false positives are negligible.

axel1200:
Look at my PR i don't know why its not shown here, but i implemented a copy Link button and had replay Links being opened by the client and not the browser.

micheljung:
The issue ID needs to be linked in the PR description in order to show up here. Title is not enough.

andrewtvuong:
Can we close this since it's merged

1-alex98:
well I think it used to work but was removed again in some refactoring

PR #538, opened 2017-06-16
Featur/#95 share replays

axel1200:


axel1200:
I need to test what hapoens with non existing replays🤐 ups that eould be important

axel1200:
I tested it it just does not open, I think that is the correct behavior

axel1200:
I mean normally u should only paste Links that are valid

axel1200:
fixes #138

.gitignore
```diff
@@ -4,12 +4,16 @@
 
 # Generated by gradle
 *.iml
+*.gitignore
```
micheljung: why would we add `*.gitignore` to `.gitignore`? :P

.gitignore
```diff
@@ -4,12 +4,16 @@
...
 
 # User-specific stuff:
 .idea/workspace.xml
 .idea/tasks.xml
 .idea/dictionaries
 .idea/sonarIssues.xml
+.idea/misc.xml
+.idea/modules.xml
+.idea/vsc.xml
```
micheljung: These files should not be ignored as they contain project-specific configuration that should be shared across all developers

src/main/java/com/faforever/client/api/FafApiAccessor.java
```diff
@@ -111,4 +111,7 @@
   Optional<MapVersion> findMapVersionById(String id);
 
   void deleteMapVersionReview(Integer id);
+
+  Game findReplayById(int id);
```
micheljung: This should return `Optional<Game>` as there's no guarantee for a result.

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -198,6 +209,9 @@ public AbstractChatTabController(ClanService clanService, WebViewConfigurer webV
     this.autoCompletionHelper = autoCompletionHelper;
     this.eventBus = eventBus;
     this.countryFlagService = countryFlagService;
+    this.replayService = replayService;
+
+    replayBaseUrl = String.format(clientProperties.getVault().getReplayDownloadUrlFormat(), "");
```
micheljung: This approach works as long as the ID is at the end of the URL, but it might not be. A safer implementation would be to convert the string format into a regular expression that allows extracting the ID, like so:

```
    String quotedReplayUrlFormat = Pattern.quote(clientProperties.getVault().getReplayDownloadUrlFormat());
    replayUrlPattern = Pattern.compile(quotedReplayUrlFormat.replace("%s", "(.*?)"));
```

then, below, you can do:

```

    Matcher replayUrlMatcher = replayUrlPattern.matcher(url);
    if(replayUrlMatcher.matches()) {
      String replayId = replayUrlMatcher.group(1);
      ...
    }
```

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -497,6 +511,34 @@ public void openPrivateMessageTab(String username) {
    * Called from JavaScript when user clicked a URL.
    */
   public void openUrl(String url) {
+    if (url.contains(replayBaseUrl)) {
+      //wrong querry
```
micheljung: What does that misspelled comment mean? ;) Turn it in a clean, helpful comment or, better, remove it and make the code explaining itself.

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -497,6 +511,34 @@ public void openPrivateMessageTab(String username) {
...
+          stage.setWidth(((Region) replayDetailRoot).getWidth());
+          stage.setHeight(((Region) replayDetailRoot).getHeight());
+
+          Stage mainStage = StageHolder.getStage();
+          stage.setX(mainStage.getX());
+          stage.setY(mainStage.getY());
+
+          stage.show();
+        });
+          });
```
micheljung: If you extract the  `thenAccept` block into a method with a descriptive name, the code becomes more readable:

```
      replayService.findById(Integer.parseInt(url.replace(replayBaseUrl, ""))).thenAccept(this::displayReplayDetail);
```

And if you set these in your commit-dialog, the code always gets indented and arranged properly:

image

src/main/java/com/faforever/client/chat/ChatMessage.java
```diff
@@ -1,8 +1,11 @@
 package com.faforever.client.chat;
 
+import lombok.Data;
+
 import java.time.Instant;
 import java.util.Objects;
 
+@Data
```
micheljung: 👍

src/main/java/com/faforever/client/remote/FafService.java
```diff
@@ -141,4 +141,6 @@
   Optional<MapBean> findMapById(String id);
 
   CompletableFuture<Void> deleteMapVersionReview(Review review);
+
+  CompletableFuture<Replay> findReplayById(int id);
```
micheljung: `CompletableFuture<Optional<Replay>>`

src/main/java/com/faforever/client/replay/Replay.java
```diff
@@ -179,11 +184,11 @@ public Temporal getStartTime() {
     return startTime.get();
   }
 
-  public void setStartTime(Temporal startTime) {
+  public void setStartTime(OffsetDateTime startTime) {
```
micheljung: No idea why we have this method, but only the one with `Temporal` needs to exist. We can remove this one.

src/main/java/com/faforever/client/replay/ReplayDownloadTask.java
```diff
@@ -48,7 +48,7 @@ public ReplayDownloadTask(I18n i18n, ClientProperties clientProperties, Preferen
   protected Path call() throws Exception {
     updateTitle(i18n.get("mapReplayTask.title", replayId));
 
-    String replayUrl = getReplayUrl(replayId, clientProperties.getVault().getReplayDownloadUrlFormat());
+    String replayUrl = Replay.getReplayUrl(replayId, clientProperties.getVault().getReplayDownloadUrlFormat());
```
micheljung: 👍

src/main/resources/i18n/messages.properties
```diff
@@ -556,3 +556,7 @@ reviewScore.four=4
 reviewScore.three=3
 reviewScore.two=2
 reviewScore.one=1
+chat.replay=replay
```
micheljung: This doesn't seem to be used anywhere and should be removed (otherwise, it should probably start uppercase)

src/main/resources/i18n/messages.properties
```diff
@@ -556,3 +556,7 @@ reviewScore.four=4
...
+replay.copyLink=Copy link for chat
```
micheljung: I wouldn't call it "for chat" because hyperlinks aren't dedicated to being used in chats

src/test/java/com/faforever/client/chat/AbstractChatTabControllerTest.java
```diff
@@ -69,6 +73,7 @@
 
   private static final long TIMEOUT = 5000;
   private static final String sampleClanTag = "xyz";
+  private static final String testReplayBaseUrl = "http://test.de/replay/%s";
```
micheljung: we should spell `private static final String` ALL_UPPERCASE

.gitignore
axel1200: Probably my ide changed it and I did not notice

axel1200:
Thanks for reviewing

src/main/resources/i18n/messages.properties
```diff
@@ -558,3 +558,6 @@ reviewScore.four=4
 reviewScore.three=3
 reviewScore.two=2
 reviewScore.one=1
+replay.copyLink=Copy link
+menu.help=
+menu.logOut=
```
micheljung: Why would we need empty messages?

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -497,7 +514,44 @@ public void openPrivateMessageTab(String username) {
...
+
+    String replayId = replayUrlMatcher.group(1);
+
+    replayService.findById(Integer.parseInt(replayId))
+        .thenAccept(replay -> Platform.runLater(() -> displayReplyDetail(replay, url)));
+  }
+
+  private void displayReplyDetail(Optional<Replay> replay, String url) {
+    if (!replay.isPresent()) {
+      logger.warn("Replay with url: " + url + "could not be found");
```
micheljung: Use logger placeholders

src/main/java/com/faforever/client/replay/Replay.java
```diff
@@ -127,6 +126,11 @@ public static Replay fromDto(Game dto) {
     return teams;
   }
 
+
+  public static String getReplayUrl(int replayId, String baseUrl) {
+    return String.format(baseUrl, replayId);
```
micheljung: baseUrlFormat

src/main/resources/i18n/messages_de_DE.properties
```diff
@@ -566,3 +566,6 @@ updater.downloadingFile=Lade {0} herunter
 updater.patchingFile=Patche {0}
 achievement.stepsFormat={0,number,#}/{1,number,#}
 urlPreviewDescription={0} ({1})
+
+menu.connect=Verbinde
```
micheljung: Obsolet

PR #540, opened 2017-06-17
Fixes #539 Login Error

### LINKED ISSUE #539: Don't crash if password can not be read
After pull client crashed.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46314635-in-the-current-production-version-the-password-is-not-encoded?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

axel1200:


src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -61,7 +61,14 @@ public void display() {
...
     String username = loginPrefs.getUsername();
-    String password = loginPrefs.getPassword();
+    String password = "";
+    try {
+      password = loginPrefs.getPassword();
+    } catch (Exception e) {
+      loginPrefs.setPassword("");
+      loginPrefs.setAutoLogin(false);
+      logger.warn("Error during password decryption, password was thrown away:", e);
+    }
```
micheljung: We should rather put this into `getPassword` where the *deobfuscation* (not *decryption*) happens, so that  all the hacks stay at the same place and it's much easier to remove later.

axel1200: Ok shell i do that or did u already?

PR #554, opened 2017-06-20
Allow users to configure if they want notifications for non-@-mentions.

### LINKED ISSUE #145: Notify on @name mentioning only
Let the user specify if he wants to get a notification if someone types his name into the chat or only if it's preceded by an @.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29319454-notify-on-name-mentioning-only?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

themasch:
This should fix #145. 

I'm not really happy with the texts and translations and not sure about the check in ChannelTabController.


src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -273,6 +273,10 @@ protected WebView getMessagesWebView() {
 
   @Override
   protected void onMention(ChatMessage chatMessage) {
+    if (preferencesService.getPreferences().getNotification().isNotifyOnAtMentionEnabled()
```
axel1200: Ok I hear NotifyOnAtMention and I thing u are either notified if somebody mentions you with an @ sign or you are not at all. Maybe name it NotifyOnAtMentioningOnly. So the purpose gets clear. In Intelij you can refactor name by selecting the name and pressing (Umschalten+F6)

src/main/resources/i18n/messages.properties
```diff
@@ -579,3 +579,5 @@ reviewScore.four=4
 reviewScore.three=3
 reviewScore.two=2
 reviewScore.one=1
+settings.chat.notifyOnAtMentionOnly=Only notifiy when mentioned via @username.
```
micheljung: "notify"

src/main/resources/i18n/messages_de_DE.properties
```diff
@@ -43,37 +43,37 @@ leaderboard.loading=Lade Rangliste\u2026
 ranked1v1.players=Spieler
 ranked1v1.play=Spiel
 ranked1v1.cancel=Abbrechen
-ranked1v1.choseFactions.title=W�hle eine Fraktion
+ranked1v1.choseFactions.title=W�hle eine Fraktion
```
micheljung: Seems like you introduced some encoding issues in this file

src/main/resources/i18n/messages_ru_RU.properties
```diff
@@ -0,0 +1,2 @@
+settings.chat.notifications=Notifications
+settings.chat.notifyOnAtMentionOnly=Only notifiy when mentioned via @username.
```
micheljung: Don't add such "default" translations; Java will automatically fall back to the default language file if no key is found in the matching language file.

src/main/resources/i18n/messages_de_DE.properties
themasch: Oh damn. How about I fix this by replacing every non-iso-8859-1 character with its unicode escape sequence like described here: https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html? 
Should be more fail-save and "correct" then, isn't it?

axel1200:
@themasch still working on it?

themasch:
@axel1200 sorry, I'm quite busy at the moment. If anyone feels like it I'd be happy to see this one be solved by someone else. I do not know when I have time to fix ist.

1-alex98:
@themasch It is just changing the translations ;) for me that means cloning your fork :(

1-alex98:
@themasch redid thi stuff

Issue #563, opened 2017-07-06
Map upload gives HTTP 400


micheljung:
Probably not a client issue, but that's where people see it.
Can someone figure out what happens? You probably need to test against a locally running API.

micheljung:
Biass sent me an example map https://www.dropbox.com/sh/tjsberr0qzvdqsa/AACzP1a3bty_4R6S88h53CDSa?dl=0

seero11:
[downlords-faf-client.zip](https://github.com/FAForever/downlords-faf-client/files/1132092/downlords-faf-client.zip)

Geosearchef:
I'll take a look and see if I can do sth in the client. Talked to someone who couldn't uploud a map and he solved it by just installing an old version of the java client he had lying around somewhere.

micheljung:
I guess we need to look at the API logs

Geosearchef:
Concerning which API? Java or Python?

Geosearchef:
Finally got database, server and api running locally (some flyaway scripts didn't do their job), am now investigation this.

Geosearchef:
![capture](https://user-images.githubusercontent.com/5748922/28242132-2cec0efa-69a4-11e7-9cf7-be4e329630d5.PNG)
Gonna see if I can reconstruct the missing parameter.

Geosearchef:
Ok, this problem is fixed, now I'm getting 500 server error on local due to missing table map_statistics in db. On production I get
`org.springframework.web.client.HttpClientErrorException: 422 Unprocessable Entity`

Geosearchef:
resolved, next problem: https://github.com/FAForever/faf-java-api/issues/113

micheljung:
We don't close issues before its PR (#570 in this case) has actually been merged into develop

PR #570, opened 2017-07-15
#563 fix map upload bad request

Geosearchef:


src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
```diff
@@ -257,6 +257,7 @@ public void uploadMod(Path file, ByteCountListener listener) {
   @Override
   public void uploadMap(Path file, boolean isRanked, ByteCountListener listener) {
     MultiValueMap<String, Object> multipartContent = createFileMultipart(file, listener);
+    multipartContent.add("metadata", "{\"isRanked\":\"" + isRanked + "\"}");
```
Geosearchef: Maybe a more elegant way to do this? Like using an object mapper as in the API (can't be autowired as it's not a bean in the client). Creating an extra class for this is a bit overkill.

Brutus5000: I'm a bit confused. That doesn't look like a solution to any of the problems I have encountered so far. Can you give specific info which error message you intend to fix here?

Geosearchef: The 400 bad request issue #563 . With this line added I'm able to upload maps perfectly fine.

micheljung:
Also, please Write "Fix #563" into the commit message, so that the related issue gets closed automatically as soon as this PR is merged

Issue #558, opened 2017-06-26
Rating distribution in 1v1


speed2CZ:
So clicking on 1v1 ranked match tab shows this for rating distribution
Not saying too much
![downlords-faf-client_2017-06-26_12-10-04](https://user-images.githubusercontent.com/10947024/27534850-9b6519c6-5a68-11e7-9961-5ea9377ae32b.png)

NiklasKopp:
Cannot reproduce. This is how it currently looks on develop running it with prod configuration:

![image](https://user-images.githubusercontent.com/15572250/30750073-baad1d8c-9fb5-11e7-94ad-03f56fc51158.png)

Looks fine to me.

I would personally prefer ordering by score, so that the positive scores aren't mixed with the negative ones, which would yield the following:

![image](https://user-images.githubusercontent.com/15572250/30750854-daae56f8-9fb7-11e7-8d38-b8e395d95dd3.png)

If you guys like the second version better I could file a pull request for it, it's just a simple one line change. Not sure if that's in the scope of this issue though.

Androish:
Might be better to remove the 0 rank? I am assuming the reason it is so big is that anyone who has no ranked games ends up with that? That being so large "hides" the other scores.

Hmm better idea, not hide it but only show the rating distribution for players that has had more than X(maybe 10?) games.

I Agree the 2nd one is better looking

speed2CZ:
Hiding the 0 rank was the idea. no one cares how many ppl havent played a game yet

NiklasKopp:
This is the result hiding all with fewer than 10 played games. There are some 0 rated players in there, so rank 0 does not mean no games played

![image](https://user-images.githubusercontent.com/15572250/30751840-df95b3c0-9fba-11e7-8d23-53debfc3e34d.png)

Androish:
That looks alot better! :) I guess it might need a small text saying that it is for more than x games

NiklasKopp:
Text is added (en + de)

Might even work with 5 instead of 10

![image](https://user-images.githubusercontent.com/15572250/30752498-bed32a44-9fbc-11e7-8d30-d0196f181854.png)

speed2CZ:
10 games looks better

OCGDFS:
Are the 1v1 games counted as "games played", or global games, or all? because there is still a big deviation from the normal distribution for 0rating/10games.

OCGDFS:
[...]

PR #590, opened 2017-09-22
Feature/#558 rating distribution

NiklasKopp:
Orders the 1v1 rating distribution by score and filters out players with very few games to make it look better #558 

Before: 
image

After:
image

### DIFF
--- a/src/main/java/com/faforever/client/leaderboard/LeaderboardServiceImpl.java
+++ b/src/main/java/com/faforever/client/leaderboard/LeaderboardServiceImpl.java
@@ -17,7 +17,6 @@ import java.util.stream.Collectors;
 @Service
 @Profile("!" + FafClientApplication.PROFILE_OFFLINE)
 public class LeaderboardServiceImpl implements LeaderboardService {
-
   private final FafService fafService;
 
   @Inject
@@ -29,6 +28,7 @@ public class LeaderboardServiceImpl implements LeaderboardService {
   public CompletableFuture<List<RatingStat>> getLadder1v1Stats() {
     return fafService.getLadder1v1Leaderboard()
         .thenApply(entries -> entries.stream()
+            .filter(entry -> entry.gamesPlayedProperty().get() >= MINIMUM_GAMES_PLAYED_TO_BE_SHOWN)
             .collect(Collectors.groupingBy(leaderboardEntry -> (int) leaderboardEntry.getRating() / 100 * 100, Collectors.counting()))
             .entrySet().stream()
             .map(entry -> new RatingStat(entry.getKey(), entry.getValue().intValue()))
--- a/src/main/java/com/faforever/client/leaderboard/LeaderboardServiceImpl.java
+++ b/src/main/java/com/faforever/client/leaderboard/LeaderboardServiceImpl.java
@@ -9,8 +9,10 @@ import org.springframework.stereotype.Service;
 
 import javax.inject.Inject;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 
 @Lazy
@@ -26,13 +28,25 @@ public class LeaderboardServiceImpl implements LeaderboardService {
 
   @Override
   public CompletableFuture<List<RatingStat>> getLadder1v1Stats() {
-    return fafService.getLadder1v1Leaderboard()
-        .thenApply(entries -> entries.stream()
-            .filter(entry -> entry.gamesPlayedProperty().get() >= MINIMUM_GAMES_PLAYED_TO_BE_SHOWN)
-            .collect(Collectors.groupingBy(leaderboardEntry -> (int) leaderboardEntry.getRating() / 100 * 100, Collectors.counting()))
-            .entrySet().stream()
-            .map(entry -> new RatingStat(entry.getKey(), entry.getValue().intValue()))
-            .collect(Collectors.toList()));
+    return fafService.getLadder1v1Leaderboard().thenApply(this::toRatingStats);
+  }
+
+  private List<RatingStat> toRatingStats(List<LeaderboardEntry> entries){
+    Map<Integer, Long> totalCount = countByRating(entries.stream());
+    Map<Integer, Long> countWithoutFewGames = countByRating(entries.stream()
+        .filter(entry -> entry.gamesPlayedProperty().get() >= MINIMUM_GAMES_PLAYED_TO_BE_SHOWN));
+
+    return totalCount.entrySet()
+        .stream()
+        .map(entry -> new RatingStat(
+            entry.getKey(),
+            entry.getValue().intValue(),
+            countWithoutFewGames.getOrDefault(entry.getKey(), 0L).intValue()))
+        .collect(Collectors.toList());
+  }
+
+  private Map<Integer, Long> countByRating(Stream<LeaderboardEntry> entries){
+    return entries.collect(Collectors.groupingBy(leaderboardEntry -> (int) leaderboardEntry.getRating() / 100 * 100, Collectors.counting()));
   }
 
   @Override
--- a/src/main/java/com/faforever/client/leaderboard/LeaderboardServiceImpl.java
+++ b/src/main/java/com/faforever/client/leaderboard/LeaderboardServiceImpl.java
@@ -36,8 +36,7 @@ public class LeaderboardServiceImpl implements LeaderboardService {
     Map<Integer, Long> countWithoutFewGames = countByRating(entries.stream()
         .filter(entry -> entry.gamesPlayedProperty().get() >= MINIMUM_GAMES_PLAYED_TO_BE_SHOWN));
 
-    return totalCount.entrySet()
-        .stream()
+    return totalCount.entrySet().stream()
         .map(entry -> new RatingStat(
             entry.getKey(),
             entry.getValue().intValue(),
--- a/src/main/java/com/faforever/client/leaderboard/LeaderboardServiceImpl.java
+++ b/src/main/java/com/faforever/client/leaderboard/LeaderboardServiceImpl.java
@@ -3,6 +3,7 @@ package com.faforever.client.leaderboard;
 import com.faforever.client.FafClientApplication;
 import com.faforever.client.game.KnownFeaturedMod;
 import com.faforever.client.remote.FafService;
+import com.faforever.client.util.RatingUtil;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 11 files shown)


src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
```diff
@@ -259,9 +259,9 @@ private void updateRatingHint(int rating) {
   @SuppressWarnings("unchecked")
   private void plotRatingDistributions(List<RatingStat> ratingStats, Player player) {
     XYChart.Series<String, Integer> series = new XYChart.Series<>();
-    series.setName(i18n.get("ranked1v1.players"));
+    series.setName(i18n.get("ranked1v1.players", LeaderboardService.MINIMUM_GAMES_PLAYED_TO_BE_SHOWN ));
     series.getData().addAll(ratingStats.stream()
-        .sorted(Comparator.comparingInt(RatingStat::getCount))
+        .sorted(Comparator.comparingInt(RatingStat::getRating).reversed())
```
micheljung: I'd go without `.reversed()` so that low numbers are left and high numbers are right

src/main/resources/i18n/messages.properties
```diff
@@ -61,7 +61,7 @@ coop.leaderboard.numberOfPlayersFormat={0,number,#} players
 coop.leaderboard.couldNotLoad=Co-Op leaderboard could not be loaded.
 coop.couldNotLoad=Co-Op missions could not be loaded: {0}
 
-ranked1v1.players=Players
+ranked1v1.players=Players (with at least {0} games played)
```
micheljung: Good that we tell people and don't just hide the fact that we're filtering :)

NiklasKopp:
I just realised that I'm doing the filtering too early. With the filtering active in the service it affects the ("out of X Players") display too. Will fix that and add a test for it.

src/main/java/com/faforever/client/leaderboard/LeaderboardServiceImpl.java
```diff
@@ -26,13 +28,25 @@ public LeaderboardServiceImpl(FafService fafService) {
...
+    return fafService.getLadder1v1Leaderboard().thenApply(this::toRatingStats);
+  }
+
+  private List<RatingStat> toRatingStats(List<LeaderboardEntry> entries){
+    Map<Integer, Long> totalCount = countByRating(entries.stream());
+    Map<Integer, Long> countWithoutFewGames = countByRating(entries.stream()
+        .filter(entry -> entry.gamesPlayedProperty().get() >= MINIMUM_GAMES_PLAYED_TO_BE_SHOWN));
+
+    return totalCount.entrySet()
+        .stream()
```
micheljung: We keep `.stream()` on the line above

micheljung:
If you like, you can also try marking the rating bar that matches the current player's rating with a different color (e.g. yellow) and label it "You". Or you can set a label node instead that says that.

NiklasKopp:
Good idea, looks like someone already tried that, there is a pseudo-class being set for the bar node with the players score, only the css part was missing.

image

axel1200:
Hi @NiklasKopp really well done, you might consider joining our slack -> https://github.com/FAForever/downlords-faf-client/wiki/Mentor-program ;)

micheljung:
Yeah, that someone was me :) I thought having "You" in the legend would be nice, but I realize now that's not possible. But could you try a `data.setNode(new Label(...))` saying "You" and see how that looks?

NiklasKopp:
@micheljung haha okay^^ Got the label to work, not as easy as I thought but it looks ok, I think.

image

I'm not used to rebasing, (I'd normally just merge the develop branch into mine now) so what I need to do is checkout my feature branch and go "rebase develop"?

@axel1200 thanks, I guess I'll do that as soon as I get to it :)

OCGDFS:
Im rather afraid that players who just dont play ladder are also included in that calculation. Regarding the big deviation from the normal distribution for players with 0 rating, they are more likely to distort the whole thing than the amount of accs with <10 **ladder games**.

speed2CZ:
I dont think that "You" label is even needed there, the different colour says it all

micheljung:
Inactive players aren't included, BUT I couldn't find that the server sets players inactive, so I guess that's the root cause of the high number of players with rating 0.

micheljung:
I set all players inactive that haven't been online for 3 months, how does it look now? :)

Sheeo:
Please clamp displayed rating values to the [0;3000] interval as well -- values outside this range are meaningless.

Androish:
There are more players with 10 games or more with -100 rating. than there are of any single bracket over 1400. So I'd say it is relevant?

Brutus5000:
No. Negative rating simply means, the player has not enough games for a "proper" rating. The high uncertainity makes the value go negative.

Androish:
Ok. But then shuld this not be solved by excluding everyone that does not have enough games for a propper rating? How many games does one need?

micheljung:
It's not correct that values outside [0;3000] are meaningless, and assuming values like 0 and 3000 isn't correct either because these values depend on the TrueSkill parameters and number of skill classes. Negative values also don't mean that there aren't enough games, nor is it caused by high uncertainty.

Please keep it like it is.

http://www.moserware.com/assets/computing-your-skill/The%20Math%20Behind%20TrueSkill.pdf

NiklasKopp:
Looks like this now: 
image

So excluding inactive players didn't really change a lot.

Androish:
Idea from me again. Is there a variable(if not maybe one can be calculated?) that the server keeps how "accurate" the rating is based on the table provided from @micheljung. Maybe sort on that then instead of an arbitrary number of games played?

And its starting to look a bit thin, maybe have a toggle to show only active, or all?

Sorry for just throwing out ideas but no real solutions :) I must say it is looking really good

OCGDFS:
These high amount of 0 rated players are players who havent played a single ladder game, and have 0 rating because of that.
are players set as inactive who havent played a ladder game, or a game in general?

micheljung:
@Androish I guess instead of "sort" you meant "filter"; yes we could filter out people who's uncertainty is higher than X, but I don't think we're actually winning something here. You don't want to disappear from the leaderboard just because the system is uncertain of your skill, and we don't want this graph to be different from the leaderboard either.

@NiklasKopp that's a very good catch, didn't think of it. All players with a rating less than 0 should be rounded down instead of up (-1 falls into the -100 category).

@OCGDFS the "wasn't online for 3 months" was a temporary cleanup for a currently broken system. Eventually I will implement "hasn't played ladder for 3 months" correctly, but this needs another database and software update first (scheduled for 2017-10-23).

NiklasKopp:
Corrected the rounding of the ratings:

image

@micheljung If there is nothing else (after fixing the codacy issue) I guess I'll rebase on develop and force-push this?

src/main/java/com/faforever/client/leaderboard/LeaderboardServiceImpl.java
```diff
@@ -27,6 +27,14 @@ public LeaderboardServiceImpl(FafService fafService) {
   }
 
   @Override
+  public int roundRatingToLowerHundred(double rating) {
+    if(rating < 0 ){
+      rating -= 100;
+    }
+    return (int)rating / 100 * 100;
```
micheljung: Please make sure that you have "Reformat code` enabled in your commit dialog (as well as "Rearrange code" and "Optimize imports"), and apply the formatter retroactively to all files changed in this PR.

And the method shouldn't be opened more than required;
```
@VisibleForTesting
int roundRatingToLowerHundred(double rating)
```

So no need for an interface method and mock implementation.

NiklasKopp: I just installed the Save-Actions Plugin, so that should resolve any future formatting issues (I'm just used to formatting to apply on strg + s ^^)

The roundRatingToLowerHundred-Method is also used in the Ladder1v1Controller to round the player rating to find the matching bar to put the player.

Maybe the whole thing should be moved to the RatingUtil class instead? There is a getRoundedRating() - Method but that's behaving a bit differently: return (rating + 50) / 100 * 100;

micheljung: Oh now I see it, I missed that :) Yeah I guess it should go into the utility class. Maybe call it `getRoundedRatingBy100`, or allow to pass a rounding parameter (50, 100). And while we're at it, we should fix the existing `getRoundedRating` as well.

NiklasKopp: Is there a reason to have these two different kinds of roundings? If not, going with the (fixed) one from the RatingUtil class might make things a lot clearer.

Sheeo:
> It's not correct that values outside [0;3000] are meaningless, and assuming values like 0 and 3000 isn't correct either because these values depend on the TrueSkill parameters and number of skill classes. Negative values also don't mean that there aren't enough games, nor is it caused by high uncertainty.

Yes, it's correct precisely because of our parameters. The rating space is given by the normal distribution with `1500` as the mean and `500` as the deviation (mu and sigma-squared respectively). This puts values outside of the interval [0;3000] _outside_ of the distribution, and hence invalid.

micheljung:
No, it's not. It's not how TrueSkill works. What we are displaying here is the leaderboard **rating**, which is different from the **skill**, as you surely know. The rating is `skill - 3*dev` and TrueSkill doesn't prevent you from having a skill of 100 with a deviation of 50. It's not like the worse you get, the smaller your deviation is. Otherwise, if your skill was 0, your deviation needs to be 0 in order to have a rating of at least 0. But you'll never get a deviation of 0.

Let's look at our current values.

**Initial mean/deviation**
Our *initial* mean is 1500 and our *initial* deviation 500. Those are called **initial** values because that's exactly what they are. It does not mean that those values set the boundaries of our TrueSkill system.

**Beta**
Our beta is set to 240, which is the **skill class width**. Meaning that if one player has 240 more skill points than another, he has a 76% chance of winning. If all of our players were similarly good, everyone would have a 50% chance of winning against any other player. We had exactly 1 skill class and all **skill points** would be about 1400-1600.
However, the more the skill of the worst and best players differ, **the more skill classes** we have. This is something we cannot possibly predict without looking at data! There could be 3, 5, 10, or 20 different skill classes. As a matter of fact, our best active ladder1v1 player has a **skill** of 2457 while our worst active ladder1v1 player has a **skill** of -10 (yes, even that is a sensible value), giving us a difference of 2467 skill points, divided by 240 equals to **10.28** skill classes in FAF.

**A very simple example**
Using the values above, let's take 2 totally new players A and B with skill/dev 1500/500. They play a game, A wins, B loses. We do the math and the result is the following.

A: 1765.51/429.19
B: 1234.49/429.19

Both skills changed by 265.51 and deviation got reduced by 71.81.

Now let's do the math:

**Rating** of A: 1765.51 - 3\*429.19 = **477.94**
**Rating** of B: 1234.49 - 3\*429.19 = **-53.08**

Now, what happens if Player B loses against another player, C, which the exact same stats as B? Player B will drop to 839/311, resulting in a **rating** of -94.

This totally makes sense because the deviation does not get smaller as a player approaches 1500-3*500 or 1500+3*500.

And if you still don't believe me, go to [Microsoft's Official TrueSkill calculator](http://boson.research.microsoft.com/trueskillranking/RankCalculator.aspx?Players=25:8.333:0,25:8.333:1&Ranks=01234567&Draws=WWWWWWW&Probability=10) and calculate the outcome yourself (Spoiler: Bob gets a Rating of -0.909 after playing against the same-skilled Alice).

What they do, however, is `if( trueSkill<0 ) trueSkill = 0;` to display data. Doesn't change the fact that the actual rating is still negative, and that players with a negative rating don't necessarily have the same skill.

PR #602, opened 2017-10-09
Fixes #601

### LINKED ISSUE #601: Application crashes when being connected over virtual Network Interface
The Mac address is used to encrypt the user creditentials, but virtual Network Interfaces do not have one.....

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/50112366-application-crashes-when-being-connected-over-virtual-network-interface?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

axel1200:


axel1200:
Still thinking about the best way to do it, but its important to me that this gets fixed

src/main/java/com/faforever/client/preferences/LoginPrefs.java
```diff
@@ -6,24 +6,17 @@
...
 
   private static final String KEY;
 
   static {
-    try {
-      KEY = new String(NetworkInterface.getByInetAddress(InetAddress.getLocalHost()).getHardwareAddress(), StandardCharsets.US_ASCII);
-    } catch (SocketException | UnknownHostException e) {
-      throw new UnsupportedOperationException("No socket available");
-    }
+    KEY = System.getProperty("user.name");
```
micheljung: The `static {}` block was only necessary because the code threw an exception before. Now you can set the value directly where you declare KEY.

PR #613, opened 2017-10-18
Gets Clan features running again

### LINKED ISSUE #595: Clan Tag dropdown broken
When clicking on the clan Label in Chat certain Information should shown but is not.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/50112371-clan-tag-dropdown-broken?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

axel1200:
Fixes #595

### DIFF
--- a/src/main/java/com/faforever/client/chat/ChatUserItemController.java
+++ b/src/main/java/com/faforever/client/chat/ChatUserItemController.java
@@ -15,6 +15,7 @@ import com.faforever.client.player.PlayerService;
 import com.faforever.client.preferences.ChatPrefs;
 import com.faforever.client.preferences.PreferencesService;
 import com.faforever.client.theme.UiService;
+import com.faforever.client.ui.StageHolder;
 import com.google.common.eventbus.EventBus;
 import javafx.application.Platform;
 import javafx.beans.InvalidationListener;
@@ -91,6 +92,7 @@ public class ChatUserItemController implements Controller<Node> {
   private ChangeListener<String> clanChangeListener;
   private ChangeListener<PlayerStatus> gameStatusChangeListener;
   private InvalidationListener userActivityListener;
+  private ClanTooltipController clanTooltipController;
 
   @Inject
   // TODO reduce dependencies, rely on eventBus instead
@@ -360,9 +362,10 @@ public class ChatUserItemController implements Controller<Node> {
 
   public void onMouseEnteredClanTag() {
     clanService.getClanByTag(player.getClan()).thenAccept(optionalClan -> {
-      if (!optionalClan.isPresent() || clanMenu.getTooltip() != null) {
+      if (!optionalClan.isPresent()) {
         return;
       }
+
       clanMenu.getItems().clear();
 
       Clan clan = optionalClan.get();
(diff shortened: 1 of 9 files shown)


src/main/java/com/faforever/client/chat/ChatUserItemController.java
```diff
@@ -360,10 +362,13 @@ private void onUserActivity() {
...
     clanService.getClanByTag(player.getClan()).thenAccept(optionalClan -> {
-      if (!optionalClan.isPresent() || clanMenu.getTooltip() != null) {
+      if (!optionalClan.isPresent()) {
         return;
       }
-      clanMenu.getItems().clear();
+
+      if (!clanMenu.getItems().isEmpty()) {
+        clanMenu.getItems().clear();
+      }
```
micheljung: Why do we only call `clear()` if it's not empty? Clearing an empty list shouldn't have any effect.

src/main/java/com/faforever/client/clan/Clan.java
```diff
@@ -43,14 +43,15 @@ public Clan() {
 
   public static Clan fromDto(com.faforever.client.api.dto.Clan dto) {
     Clan clan = new Clan();
-    clan.setId(dto.getId());
+    clan.setId(Integer.parseInt(dto.getId()));
```
micheljung: Does anything prevent us from making this clan bean's ID as String as well? Otherwise, this is rather a gamble.

src/main/java/com/faforever/client/clan/ClanTooltipController.java
```diff
@@ -20,16 +20,19 @@
...
+    descriptionTitleLabel.managedProperty().bind(descriptionTitleLabel.visibleProperty());
     descriptionLabel.visibleProperty().bind(descriptionLabel.textProperty().isNotEmpty());
+    descriptionTitleLabel.visibleProperty().bind(descriptionLabel.textProperty().isNotEmpty());
   }
 
   public void setClan(Clan clan) {
     nameLabel.setText(clan.getName());
     // TODO improve formatting
-    membersLabel.setText(clan.getMembers().toString());
+    membersLabel.setText(String.valueOf(clan.getMembers().size()));
```
micheljung: `String.valueOf()` shouldn't be used to display integers since it doesn't consider the user's locale. Use `i18n.number()` instead.

axel1200:
Approved why don't we merge it?
Ok it is not approved anymore because I rebased...

axel1200:
What does it mean if push fails and pr  in Travis does not????

I thought it would mean push was the current version of the branch and pr the one after being merged with develop.... But that is not reasonable here, because the branch is up to date, so when applying this defenition there should not be any difference between the two 🤔

Brutus5000:
You have a concurrency problem (there was some sort of ConcurrentException). I restarted the built and this time it did not occur. But there is a problem somewhere.

micheljung:
I didn't approve it, you dismissed my review? Would've been good if you copied the stacktrace @Brutus5000 , unfortunately I can't find any travis build history of this branch.

axel1200:
So what do I do now? @Brutus5000 Yeah there are always some concurrency problem in the unit tests ... 😬

Brutus5000:
Sorry I thought Travis creates a new build and stores the old log :(

Issue #599, opened 2017-10-07
Game path not recognized


NiklasKopp:
I first thought this was an issue with my development setup, but I could recently reproduce on an other machine using the alpha client exe (0.74) from faforever.com.

Both systems have the following in common: Windows 10, 64 bit and Forged Alliance installed under the Windows 10 default location: `C:\Program Files (x86)\THQ\Gas Powered Games\Supreme Commander - Forged Alliance`
The old python client works on both of these systems.

The Client starts just fine, no sign of anything going wrong, except for some log messages:

 `2017-10-07 11:05:30.223  INFO 2624 --- [JavaFX-Launcher] c.f.c.preferences.PreferencesService     : Game path is not specified or non-existent, trying to detect
2017-10-07 11:05:30.231  INFO 2624 --- [JavaFX-Launcher] c.f.c.preferences.PreferencesService     : Game path could not be detected`

Every click on a "play" button then just does nothing, which renders the client unusable. 

I think there should be at least a notification or a file chooser if the client isn't able to find the game on it's own. 

I could look into that myself since it's fairly easy to reproduce for me

EDIT: Found a possible source of this problem: The MissingGamePathNotifier is registered to the eventbus after the PreferenceService fires the MissingGamePathEvent, so it never gets the event.

EDIT2: You might actually be able to reproduce this by just deleting the client.prefs file and restarting the client.

NiklasKopp:
The `isGamePathValid(Path binPath)` method apparently doesn't get the /bin Path but the path above. When doing `binPath = binPath.respolve("bin")` in there, the game is being detected correctly.

micheljung:
Wow, I'm surprised this hasn't been reported for so long. I never had the game installed in a default location :D

PR #615, opened 2017-10-18
Bugfix/#599 game path not recognized

NiklasKopp:
This fixes the file resolving issue of #599 (missing '/bin' in the path) and also brings up a warning pop-up with a file chooser action on an undetected game path. 
Additionally any events that are posted to the guava eventbus but not being handled are logged at WARN level now.

### DIFF
--- a/src/main/java/com/faforever/client/config/BaseConfig.java
+++ b/src/main/java/com/faforever/client/config/BaseConfig.java
@@ -1,6 +1,9 @@
 package com.faforever.client.config;
 
+import com.google.common.eventbus.DeadEvent;
 import com.google.common.eventbus.EventBus;
+import com.google.common.eventbus.Subscribe;
+import lombok.extern.java.Log;
 import org.springframework.context.MessageSource;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.support.ResourceBundleMessageSource;
@@ -22,6 +25,18 @@ public class BaseConfig {
 
   @Bean
   EventBus eventBus() {
-    return new EventBus((exception, context) -> exception.printStackTrace());
+    EventBus bus = new EventBus((exception, context) -> exception.printStackTrace());
+    bus.register(new DeadEventHandler());
+    return bus;
+  }
+
+  @Log
+  private static class DeadEventHandler {
+    @Subscribe
+    public void onDeadEvent(DeadEvent deadEvent) {
+      Object unhandledEvent = deadEvent.getEvent();
+      log.warning("No event handler registered for event of type " +
+          unhandledEvent.getClass().getSimpleName() + ": " + unhandledEvent);
+    }
   }
 }
--- a/src/main/java/com/faforever/client/config/BaseConfig.java
+++ b/src/main/java/com/faforever/client/config/BaseConfig.java
@@ -3,16 +3,18 @@ package com.faforever.client.config;
 import com.google.common.eventbus.DeadEvent;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
-import lombok.extern.java.Log;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.context.MessageSource;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
 import org.springframework.context.support.ResourceBundleMessageSource;
 
 /**
  * This configuration has to be imported by other configurations and should only contain beans that are necessary to run
  * the application.
  */
-@org.springframework.context.annotation.Configuration
+@Slf4j
+@Configuration
 public class BaseConfig {
 
   @Bean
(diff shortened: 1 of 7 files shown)


src/main/java/com/faforever/client/config/BaseConfig.java
```diff
@@ -1,6 +1,9 @@
 package com.faforever.client.config;
 
+import com.google.common.eventbus.DeadEvent;
 import com.google.common.eventbus.EventBus;
+import com.google.common.eventbus.Subscribe;
+import lombok.extern.java.Log;
```
micheljung: We use `@Slf4j` on class level

src/main/java/com/faforever/client/config/BaseConfig.java
```diff
@@ -22,6 +25,18 @@ MessageSource messageSource() {
...
+    return bus;
+  }
+
+  @Log
+  private static class DeadEventHandler {
+    @Subscribe
+    public void onDeadEvent(DeadEvent deadEvent) {
+      Object unhandledEvent = deadEvent.getEvent();
+      log.warning("No event handler registered for event of type " +
+          unhandledEvent.getClass().getSimpleName() + ": " + unhandledEvent);
```
micheljung: Cool, didn't know about DeadEvent. When you switch to slf4j, make sure to use playerholders `{}` instead of string concatenation and follow the [logging format guidelines](https://github.com/FAForever/java-guidelines/wiki/Coding-Standards#always-decorate-placeholders).

src/main/java/com/faforever/client/config/BaseConfig.java
```diff
@@ -22,6 +25,18 @@ MessageSource messageSource() {
 
   @Bean
   EventBus eventBus() {
-    return new EventBus((exception, context) -> exception.printStackTrace());
+    EventBus bus = new EventBus((exception, context) -> exception.printStackTrace());
```
micheljung: `printStackTrace()` must not be used. Use a logger instead.

src/main/java/com/faforever/client/game/MissingGamePathNotifier.java
```diff
@@ -40,7 +40,7 @@ public void onMissingGamePathEvent(MissingGamePathEvent event) {
     List<Action> actions = Collections.singletonList(
         new Action(i18n.get("missingGamePath.locate"), chooseEvent -> eventBus.post(new GameDirectoryChooseEvent()))
     );
-
-    notificationService.addNotification(new PersistentNotification(i18n.get("missingGamePath.notification"), Severity.WARN, actions));
+    String message = i18n.get("missingGamePath.notification");
+    notificationService.addNotification(new ImmediateNotification(message, message, Severity.WARN, actions));
```
micheljung: The client aims to be as non-intrusive as possible. Immediate notifications should therefore be used sparingly and only in cases where a user's immediate action is required.

Let's say someone only has the client installed for chatting in #aeolus, or maybe he's a client developer who doesn't actually have the game installed but starts it very often. This user gets an annoying notification now, every time he starts the client.

This is why we made it a PersistentNotification.

src/main/java/com/faforever/client/preferences/PreferencesService.java
```diff
@@ -139,11 +139,7 @@ public void postConstruct() throws IOException {
       Files.copy(getClass().getResourceAsStream("/game.prefs"), gamePrefs);
     }
 
-    Path path = preferences.getForgedAlliance().getPath();
-    if (path == null || Files.notExists(path)) {
-      logger.info("Game path is not specified or non-existent, trying to detect");
-      detectGamePath();
-    }
+    // no need to try to detect the game path here, the corresponding handlers are not initialized yet
```
micheljung: Good to know for the review, but let's remove this comment now :)

src/main/java/com/faforever/client/game/MissingGamePathNotifier.java
NiklasKopp: Good point. 
I'll change that back to persistent, but I think we need a very interrupting popup when the player hits the "play" button without a game path set. The persistent notification is easily overlooked and the current client does not give any direct feedback on pressing the "play" button without a game path set: [Ladder1v1Controller.java](https://github.com/FAForever/downlords-faf-client/blob/develop/src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java#L172)
I'm not sure if there are other places with the same problem

axel1200:
By the way @NiklasKopp we are often hanging around in Discord if you have question or whatever. We are mostly in the Developer channel... It is really very helpful sometimes https://discord.gg/Yk55DV

NiklasKopp:
@axel1200 just did that ;) doesn't look like I can join the developer channel just like that though (Niklas # 4882)

micheljung:
@NiklasKopp ping me when you're online so I can give you permission

src/main/java/com/faforever/client/preferences/event/MissingGamePathEvent.java
```diff
@@ -1,5 +1,14 @@
 package com.faforever.client.preferences.event;
 
-public enum MissingGamePathEvent {
-  INSTANCE
+import lombok.AllArgsConstructor;
+import lombok.Value;
+
+@Value
```
micheljung: Oh boy, I see all the classes passing in front of me where I should've used this instead of @Data :)

PR #620, opened 2017-11-01
Feature/#35 aliases

axel1200:
Its not jet tested!!!!
First the scope of the method needs to be changed to write_accountdata in the Api....


src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
```diff
@@ -387,6 +389,11 @@ public void deleteMapVersionReview(Integer id) {
   }
 
   @Override
+  public void changeUsername(String name) {
+    post("users/changeUsername?newUsername=" + name, "", false);
```
axel1200: I need help there

src/main/java/com/faforever/client/chat/UserInfoWindowController.java
```diff
@@ -232,6 +258,11 @@ public void setPlayer(Player player) {
         });
   }
 
+  private void updateNameHistory(Player player) {
+    fafService.getPlayersByIds(Collections.singletonList(player.getId()))
+        .thenAccept(players -> nameHistoryTable.setItems(players.get(0).getNames()));
```
micheljung: We should handle `.exceptionally()` as well. You could use `com.faforever.client.notification.NotificationService#addImmediateErrorNotification`

src/main/java/com/faforever/client/fx/OffsetDateTimeCell.java
```diff
@@ -0,0 +1,23 @@
...
+import com.faforever.client.util.TimeService;
+import javafx.scene.control.TableCell;
+
+import java.time.OffsetDateTime;
+
+public class OffsetDateTimeCell<S> extends TableCell<S, OffsetDateTime> {
+  private final TimeService timeService;
+
+  public OffsetDateTimeCell(TimeService timeService) {
+
```
micheljung: Superfluous newline :P

src/main/java/com/faforever/client/player/Player.java
```diff
@@ -350,4 +360,45 @@ public void updateFromPlayerInfo(com.faforever.client.remote.domain.Player playe
       setAvatarTooltip(player.getAvatar().getTooltip());
     }
   }
+
+  public static class NameRecord {
```
micheljung: Extract to its own file?

src/main/resources/i18n/messages.properties
```diff
@@ -530,6 +534,7 @@ settings.account.steamAccount.prompt=You can link your steam account here.
 settings.account.steamAccount.linkToSteam=Link to steam
 settings.account.username=FAF Username
 settings.account.username.change=Change username
+settings.account.username.error=Your username could not be changed. Please be aware that you can only change it every 30 days. Server reports: {0}
```
micheljung: It's nice to give the user more info, however, the 30 days are server-side configuration and might change; I'd rely on the server message instead (which should contain the explanation what failed).

src/main/java/com/faforever/client/player/UsernameChangeController.java
```diff
@@ -0,0 +1,73 @@
...
+    this.displayPlayer = displayPlayer;
+    ChangeListener<String> changeListener = (observable, oldValue, newValue) -> {
+      if (!userService.getUsername().equals(newValue)) {
+        root.setVisible(false);
+        return;
+      }
+      usernameField.setText(newValue);
+    };
+    changeListener.changed(displayPlayer, null, displayPlayer.get());
+    displayPlayer.addListener(changeListener);
```
micheljung: Whenever you work with listeners, think of memory leaks. As a rule of thumb, whenever a short-lived object adds a listener to a long-lived object, you create a memory leak.

`UsernameChangeController` is of scope prototype, therefore a new bean is created every time needed. It's therefore a short-lived object.
The player object, to which the listener is added, is a longed-lived object. If you create 10 user change controllers, the player object will keep a reference to all of them, preventing them from being garbage collected.

One solution to this problem is to remove the listener when the controller is being destroyed, but that's rather inconvenient.
Another solution, already used in the client, is to use weak references. Weak references do not prevent objects from being garbage collected. But then you need to prevent the listener from being collected too early.

However, I'm not sure if we even need a change listener here. What's the exact use case?

src/main/resources/i18n/messages.properties
```diff
@@ -421,6 +421,10 @@ userInfo.ratingHistory.global=Global
 userInfo.achievements=Achievements
 userInfo.tooltipFormat={0} ({1,number,#})
 userInfo.ratingFormat=Global Rating: {0,number,#}\nLeaderboard Rating: {1,number,#}
+userInfo.nameHistory=Name history
+userInfo.nameHistory.date=Date of change
+userInfo.nameHistory.noHistory=User did not change name
```
micheljung: `The user has never changed their name`

src/main/resources/i18n/messages_de_DE.properties
```diff
@@ -366,6 +366,7 @@ settings.account.steamAccount.prompt=Du kannst dein Steam-Konto hier verkn
 settings.account.steamAccount.linkToSteam=Mit Steam verkn�pfen.
 settings.account.username=FAF Benutzername
 settings.account.username.change=�ndere Benutzername
+settings.account.username.error=Dein Nutzername konnte nicht ge�ndert werden.
```
micheljung: `Benutzername`

src/main/resources/i18n/messages_de_DE.properties
```diff
@@ -574,7 +575,12 @@ game.notValid=Das Spiel wurde nicht gewertet
 game.notRatedYet=Das Rating ist noch nicht verf�gbar
 game.ended=Hat dir das Spiel "{0}" gefallen? Teile deine Erfahrung!
 game.rate=Jetzt bewerten
+userInfo.nameHistory=Namen Historie
+userInfo.nameHistory.date=Zeitpunkt der �nderung
+userInfo.nameHistory.name=Name
+userInfo.nameHistory.noHistory=Benutzer hat seinen Namen nicht ge�ndert
```
micheljung: `Der Benutzer hat seinen Namen nie geändert`

src/main/resources/i18n/messages_de_DE.properties
```diff
@@ -574,7 +575,12 @@ game.notValid=Das Spiel wurde nicht gewertet
 game.notRatedYet=Das Rating ist noch nicht verf�gbar
 game.ended=Hat dir das Spiel "{0}" gefallen? Teile deine Erfahrung!
 game.rate=Jetzt bewerten
+userInfo.nameHistory=Namen Historie
```
micheljung: `Namenshistorie`

micheljung:
If the commit's title wouldn't be "Rework" I'd merge this ;)

axel1200:
done

PR #627, opened 2017-11-05
Sorting of vault

### LINKED ISSUE #626: Sort replay result descending
image

It should display newest replays at the top

axel1200:
Fixes that Transistent Notifications are shown behind the Task Bar(discovered this along side the work). I enabled autofix.

Enables you to sort in the replay and game vault.
Enables you to search for items that contain a string. To do that I internally used the RE (regex) comparator, which I should not have done, however since this is not an Comparator, but embeding the text into "*", I cant do it with the structure you imposed otherwise...

Fixes #626


axel1200:
What Codacy says is not true

Issue #524, opened 2017-05-26
Reinstante old connectivity code


micheljung:
I've been waiting for too long for ICE. I shouldn't have removed the old connectivity code until it was really DONE.
Reinstate the code and remove all ICE code until it's released in the main client. Even if that's tomorrow.

See #413

micheljung:
Done. Branch is not meant to be merged but should continuously be rebased on top of develop until we can finally delete the branch.

PR #643, opened 2017-11-14
Writes Mods and reads them from game.prefs

### LINKED ISSUE #623: Fix and remember selected mods
In the "Create Game" dialog, there is a list of mods:
image

However, selected mods are not yet activated in the game. Their UIDs needs to be written into the `game.prefs`. I thought I implemented this already but apparently, it's not/no longer there, or broken (but I can't even find `game.prefs` in the code).

After this has been implemented, the "Create Game" dialog needs to read the game.prefs in order to figure out which mods had been enabled in the last game and preselect them (mods can also be enabled within the game, and the client should be aware of that). Make sure that the client doesn't fail if no game.prefs yet exists.

axel1200:
Was tested on the #524 Branch.
Fixes #623


micheljung:
Why closed?

PR #645, opened 2017-11-15
Fixes #644

### LINKED ISSUE #623: Fix and remember selected mods

### LINKED ISSUE #644: Create Button Text not updated

axel1200:
Fixes #623
Fixes that games with mods can be joined
Fixes the create game button text
Fixes Matchmaker Message before Login

(cherry picked from commit df1ad11)


axel1200:
Codacy does not Analyse my branch

src/main/java/com/faforever/client/mod/ModServiceImpl.java
```diff
@@ -176,7 +176,11 @@ public void loadInstalledMods() {
 
   @Override
   public CompletableFuture<Void> downloadAndInstallMod(String uid) {
-    return fafService.getMod(uid).thenAccept(mod -> downloadAndInstallMod(mod, null, null));
+    return fafService.getMod(uid).thenAccept(mod -> downloadAndInstallMod(mod, null, null))
+        .exceptionally(throwable -> {
+          logger.warn("Mod for modupdate not found", throwable);
+          return null;
+        });
```
micheljung: Whoever calls this method will now know that it failed (only the first `exceptionally` is ever called) so the game will start even though the mod could not be installed.

The error must, therefore, be handled in the calling code.
Also, the assumption "not found" isn't correct as there can be other reasons why an exception was thrown, like disconnect, server down, disk space full, file system permission, out of memory, and so on.

src/main/java/com/faforever/client/mod/ModServiceImpl.java
```diff
@@ -242,7 +246,7 @@ public void enableSimMods(Set<String> simMods) throws IOException {
 
   @Override
   public boolean isModInstalled(String uid) {
-    return getInstalledUiModsUids().contains(uid) || getInstalledModUids().contains(uid);
+    return getInstalledModUids().contains(uid);
```
micheljung: Are we sure that we don't need check the UI mods as well? Both are mods. Otherwise, we need to rename the method. And I expect that this method is being used from the mod vault as well, so now UI mods will probably never been detected as installed.

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -206,6 +207,12 @@ private void initModList() {
     modListView.getSelectionModel().setSelectionMode(SelectionMode.MULTIPLE);
     modListView.setCellFactory(modListCellFactory());
     modListView.getItems().setAll(modService.getInstalledMods());
+    try {
+      modService.getActivatedSimAndUIMods().forEach(mod -> modListView.getSelectionModel().select(mod));
+    } catch (IOException e) {
+      logger.error("Activated Mods could not be loaded", e);
```
micheljung: `mods`

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -331,6 +338,12 @@ public void onRandomMapButtonClicked() {
   public void onCreateButtonClicked() {
     ObservableList<Mod> selectedMods = modListView.getSelectionModel().getSelectedItems();
 
+    try {
+      modService.overrideActivatedMods(modListView.getSelectionModel().getSelectedItems());
+    } catch (IOException e) {
+      logger.warn("Activated Mods could not be updated", e);
```
micheljung: `mods`

src/main/java/com/faforever/client/game/GameServiceImpl.java
```diff
@@ -229,6 +229,13 @@ public ReadOnlyBooleanProperty gameRunningProperty() {
 
     return modService.getFeaturedMod(game.getFeaturedMod())
         .thenCompose(featuredModBean -> updateGameIfNecessary(featuredModBean, null, featuredModVersions, simModUIds))
+        .thenAccept(aVoid -> {
+          try {
+            modService.enableSimMods(simModUIds);
+          } catch (IOException e) {
+            logger.warn("SimMods were not enabled", e);
```
micheljung: `SimMods could not be enabled`

src/main/java/com/faforever/client/mod/ModServiceImpl.java
axel1200: getInstalledUiModsUids is obsolute because all UI Mods are also in the getInstalledMods....

axel1200: it is not necessary to check if it is in the UI Mods because all UI mods are also mods.... See the code of both methods and you will notice ;)

PR #664, opened 2017-12-11
Shows Ladder Maps

### LINKED ISSUE #650: Show ladder map pool
You can't see when you want to play a ladder game the different maps available for it (unless you do some dirty solution for example search on vault 1v1 ladder replays and find the maps...)

It would be nice to have a preview of all the maps somewhere near the start ladder button (could be a button show map or something). I heard it was possible before but not anymore.
Copy of https://github.com/FAForever/client/issues/889

axel1200:
Fixes #650 misses unit tests


micheljung:
Let's look at this together

axel1200:
So I might need to say some words about what I did.
I took your feedback quite serious and tried long to do it this way. For you to rember, you wanted me to make Events that could be extended and not to store the Event in AbstractViewController(and similar to the Way Android does it). I did make Events that extend each other, but they are still handled similar to my Parameters. There is simply no other way for it to work differently. That is because the onDisplay() callback/Method does only work in the top level Controllers. It does not work in Tabs and to make it work in Tabs and subcontrollers  would be really hard and my ideas on how to make it work would be ugly coding style. So I have the top Controller call the subcontrollers onEvent(). 

I think it is a quite nice mix of what you suggested and what I had done already and more important it does actually work...

axel1200:
Now Codacy tells me to use.equals for enums... ?? @micheljung so whats correct now

PR #673, opened 2018-01-06
Mod info icon's path corrected

### LINKED ISSUE #555: Check/Fix mod preview
Take this mod for instance:

```
name = "Smart Tactical Missiles"
uid = "6BEA1FF0-5D8B-3656-0180-C359C5A5E849"
version = 18
copyright = "2016-2017 Uveso"
description = "Self-acting Tactical Missile Launcher. Targetpriorities: Commander, Experimental, Heavy Assault, Mobile T3 Artillery, Structure, Engineer, mobile unit."
author = "Uveso"
url = "http://forums.faforever.com/viewtopic.php?f=41&t=13068"
icon = "/mods/SmartTacticalMissiles/Tactical.png"
selectable = true
enabled = true
exclusive = false
ui_only = false
requires = { }
requiresNames = { }
conflicts = { }
before = { }
after = { }
```

It references `icon = "/mods/SmartTacticalMissiles/Tactical.png"`, which is the correct way to do it. Other mods would reference `icon = "Tactical.png"` or similar, which is the wrong way to do it.

Check the way the client implementation resolves the thumbnail image, and fix if required.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47298913-check-fix-mod-preview?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

axel1200:
Is that what you meant?
Fixes #555

### DIFF
--- a/src/main/java/com/faforever/client/mod/Mod.java
+++ b/src/main/java/com/faforever/client/mod/Mod.java
@@ -20,6 +20,7 @@ import javafx.collections.ObservableList;
 import org.apache.maven.artifact.versioning.ComparableVersion;
 import org.jetbrains.annotations.Nullable;
 
+import java.io.File;
 import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


micheljung:
Hmm no, what I wanted to say is: If the icon path within the mod info LUA file is correct, the client should resolve it correctly and if it's not correct in the LUA file, the client should display no icon. We won't try to guess the icon path, it's the author's responsibility to do it right.

I'm just not sure whether the client currently resolves the icon using the right way or not, and that's the thing that needs to be checked/fixed :)

axel1200:
Wer'll need to talk abou that...

PR #678, opened 2018-01-06
Resolves async Issue with achievements loading

### LINKED ISSUE #667: Client window out of screen bound
When minimizing the client window the window is gone it seems to be out of the screen. See the stored location of the minimized window is x>1700 but screen is only abut 1300 pixel in width!

[client.txt](https://github.com/FAForever/downlords-faf-client/files/1584399/client.txt)

### LINKED ISSUE #677: Achievments do not load
image
...

axel1200:
Fixes #677

### DIFF
--- a/src/main/java/com/faforever/client/achievements/AchievementServiceImpl.java
+++ b/src/main/java/com/faforever/client/achievements/AchievementServiceImpl.java
@@ -3,14 +3,11 @@ package com.faforever.client.achievements;
 import com.faforever.client.api.dto.AchievementDefinition;
 import com.faforever.client.api.dto.PlayerAchievement;
 import com.faforever.client.config.CacheNames;
-import com.faforever.client.i18n.I18n;
-import com.faforever.client.notification.NotificationService;
 import com.faforever.client.player.PlayerService;
 import com.faforever.client.remote.AssetService;
 import com.faforever.client.remote.FafService;
 import com.faforever.client.remote.UpdatedAchievementsMessage;
-import com.faforever.client.theme.UiService;
-import com.faforever.client.user.UserService;
+import com.google.common.annotations.VisibleForTesting;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import javafx.scene.image.Image;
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/achievements/AchievementItemController.java
```diff
@@ -87,7 +87,7 @@ public void setPlayerAchievement(PlayerAchievement playerAchievement) {
       imageView.setEffect(null);
     }
 
-    if (AchievementType.INCREMENTAL == achievementDefinition.getType()) {
+    if (AchievementType.INCREMENTAL.equals(achievementDefinition.getType())) {
```
micheljung: To compare enums, we always use `==`

See https://stackoverflow.com/a/2937561/513451

src/main/java/com/faforever/client/achievements/AchievementServiceImpl.java
```diff
@@ -58,9 +52,12 @@ public AchievementServiceImpl(UserService userService, FafService fafService,
...
   @Override
   public CompletableFuture<List<PlayerAchievement>> getPlayerAchievements(Integer playerId) {
-    if (Objects.equals(userService.getUserId(), playerId)) {
+    int currentPlayerId = playerService.getCurrentPlayer().orElseThrow(() -> new IllegalStateException("Player has to be set")).getId();
+    if (Objects.equals(currentPlayerId, playerId)) {
       if (readOnlyPlayerAchievements.isEmpty()) {
-        reloadAchievements();
+        CompletableFuture<List<PlayerAchievement>> achievmentsLoadedFuture = new CompletableFuture<>();
+        reloadAchievements(achievmentsLoadedFuture);
+        return achievmentsLoadedFuture;
```
micheljung: `return reloadAchievements();` and have the method create the future.
And in `postConstruct()` you can just ignore the return value.

src/main/java/com/faforever/client/achievements/AchievementItemController.java
axel1200: cool explaination

PR #686, opened 2018-01-10
Delete cache directory on close

### LINKED ISSUE #684: Catch access denied
If the user starts the client as admin and later on starts it as normal user he might get access denied exceptions because he has not all permissions for the game files....

axel1200:
Deletes cache directory on close and start up
Fixes #684


micheljung:
Cache, not cash :P

PR #697, opened 2018-01-16
Fix wrong size of map on upload

### LINKED ISSUE #666: Wrong map size when uploading map
when uploading a map the client is showing the map size, but the size is game units, not km
This above is an example of 5x5km map.

The size needs to be translated into km

km | game units
5x5 | 256x256
10x10 | 512x512
20x20 | 1024x1024
40x40 | 2048x2048
80x80 | 4096x4096

20seven:
Changed width and height to use kilometres instead of pixels.
Fixes #666


axel1200:
it normally is bugfix/ ;) but does not matter

PR #698, opened 2018-01-16
Add option to hide games with mods

### LINKED ISSUE #641: option to hide modded games
in addition to the option to hide private games

20seven:
Added a checkbox to give user the option to hide modded games
Fixes #641

### DIFF
--- a/src/main/java/com/faforever/client/game/CustomGamesController.java
+++ b/src/main/java/com/faforever/client/game/CustomGamesController.java
@@ -12,9 +12,10 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import javafx.application.Platform;
+import javafx.beans.value.ChangeListener;
+import javafx.beans.value.WeakChangeListener;
 import javafx.collections.ObservableList;
 import javafx.collections.transformation.FilteredList;
-import javafx.event.ActionEvent;
 import javafx.scene.Node;
 import javafx.scene.control.Button;
 import javafx.scene.control.CheckBox;
@@ -56,6 +57,11 @@ public class CustomGamesController implements Controller<Node> {
   private final PreferencesService preferencesService;
   private final EventBus eventBus;
   private final I18n i18n;
+  public GameDetailController gameDetailController;
+
+  private GamesTableController gamesTableController;
+  private final ChangeListener<Boolean> showModsColumnChangedListener = (observable, oldValue, newValue)
+      -> gamesTableController.setModsColumnVisibility(newValue);
 
   public ToggleButton tableButton;
   public ToggleButton tilesButton;
@@ -64,10 +70,17 @@ public class CustomGamesController implements Controller<Node> {
   public Pane gameViewContainer;
   public Pane gamesRoot;
   public ScrollPane gameDetailPane;
-  public GameDetailController gameDetailController;
   public ChoiceBox<TilesSortingOrder> chooseSortingTypeChoiceBox;
 
-  private FilteredList<Game> filteredItems;
+  @VisibleForTesting
+  FilteredList<Game> filteredItems;
+
+  public CheckBox showModdedGamesCheckBox;
+  public CheckBox showPasswordProtectedGamesCheckBox;
+  private final ChangeListener<Boolean> filterConditionsChangedListener = (observable, oldValue, newValue) -> updateFilteredItems();
+  private final ChangeListener<Boolean> showPasswordColumnChangedListener = (observable, oldValue, newValue)
+      -> gamesTableController.setPasswordProtectionColumnVisibility(newValue);
+  private GamesTilesContainerController gamesTilesContainerController;
 
   @Inject
   public CustomGamesController(UiService uiService, GameService gameService, PreferencesService preferencesService,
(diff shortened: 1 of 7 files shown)


src/main/resources/i18n/messages_de_DE.properties
```diff
@@ -95,6 +95,7 @@ games.create=Er
 games.create.minRating=Min.
 games.create.maxRating=Max.
 games.showPasswordProtectedGames=Zeige passwortgesch�tzte Spiele
+games.showModdedGames=Zeige modifizierte Spiele
```
1-alex98: Bad translation

src/main/java/com/faforever/client/game/CustomGamesController.java
```diff
@@ -122,15 +123,32 @@ public TilesSortingOrder fromString(String string) {
...
   }
 
-  public void onShowPrivateGames(ActionEvent actionEvent) {
-    CheckBox checkBox = (CheckBox) actionEvent.getSource();
-    boolean selected = checkBox.isSelected();
-    if (selected) {
-      filteredItems.setPredicate(OPEN_CUSTOM_GAMES_PREDICATE);
-    } else {
-      filteredItems.setPredicate(OPEN_CUSTOM_GAMES_PREDICATE.and(gameInfoBean -> !gameInfoBean.getPasswordProtected()));
+    public void onShowPrivateGames(ActionEvent actionEvent) {
```
1-alex98: You should save the state of the Checkbox in the preferences. Cosider binding the checkedProperty of the CheckBox to the property in the preferences. See https://github.com/FAForever/downlords-faf-client/blob/1bee186d0a576fd1dc00cdc10874bc4599f666db/src/main/java/com/faforever/client/preferences/ui/SettingsController.java#L168

src/main/resources/i18n/messages.properties
```diff
@@ -158,6 +158,7 @@ games.create=Create game
 games.create.minRating=Min.
 games.create.maxRating=Max.
 games.showPasswordProtectedGames=Show password protected games
+games.showModdedGames=Show Modded Games
```
1-alex98: Rather "Games with mods"

src/main/resources/i18n/messages_de_DE.properties
1-alex98: Maybe "Zeige Spiele mit Mods" ;)

axel1200:
Maybe write a small/quick unit test

axel1200:
I just had a cool idea bind the visibility property(maybe not the visibilty property I dont know which property you must bind, but i am sure you will find out) of the password/mod column to the selected property of the check boxes or better to the preference properties... ;)

1-alex98:
@micheljung your turn

src/main/java/com/faforever/client/game/CustomGamesController.java
```diff
@@ -96,14 +110,22 @@ public TilesSortingOrder fromString(String string) {
...
     filteredItems = new FilteredList<>(games);
-    filteredItems.setPredicate(OPEN_CUSTOM_GAMES_PREDICATE);
+    showModdedGamesCheckBox.selectedProperty().bindBidirectional(preferencesService.getPreferences().showModdedGamesProperty());
+    showPasswordProtectedGamesCheckBox.selectedProperty().bindBidirectional(preferencesService.getPreferences().showPasswordProtectedGamesProperty());
+
+    updateFilteredItems();
+    preferencesService.getPreferences().showModdedGamesProperty().addListener(new WeakChangeListener<>(filterConditionsChangedListener));
+    preferencesService.getPreferences().showPasswordProtectedGamesProperty().addListener(new WeakChangeListener<>(filterConditionsChangedListener));
+
+    preferencesService.getPreferences().showModdedGamesProperty().addListener(new WeakChangeListener<>(showModsColumnChangedListener));
```
1-alex98: What about the password column?

1-alex98: Its small about we could also make that one disappear no? Did you test if that looks ok if they disappear and come back or if that messes up the column widths?

20seven: yea, I like it better with password column still there as it gives some margins to the border of the table. But latest commit removes pw column as well,. So if you want to check out how it looks, please do

1-alex98:
I like it ❤️

1-alex98:
image

1-alex98:
image

1-alex98:
@20seven Please resolve conflicts by rebasing on orign/develop

src/main/java/com/faforever/client/game/CustomGamesController.java
```diff
@@ -80,6 +93,9 @@ public CustomGamesController(UiService uiService, GameService gameService, Prefe
   }
 
   public void initialize() {
+    gamesTableController = uiService.loadFxml("theme/play/games_table.fxml");
+    gamesTilesContainerController = uiService.loadFxml("theme/play/games_tiles_container.fxml");
```
micheljung: This is the only thing that bugs me about this PR; we're loading both controllers even though we're only displaying one.
The client already uses more memory than I want it to, so if we can find a better solution I'd appreciate that. I didn't take the time yet to take a closer look and if I do, I'd like to discuss it directly in Discord ;)

PR #706, opened 2018-01-17
Check if a user has an id before showing user info

### LINKED ISSUE #705: Showing the user info suggest for ppl that are only chat users
Hide that option for chat only users or those having no id.

axel1200:
Fixes #705

### DIFF
--- a/src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
+++ b/src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
@@ -21,6 +21,7 @@ import com.faforever.client.theme.UiService;
 import com.faforever.client.user.UserService;
 import com.google.common.eventbus.EventBus;
 import javafx.application.Platform;
+import javafx.beans.binding.Bindings;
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
```diff
@@ -126,6 +128,11 @@ ContextMenu getContextMenu() {
 
   public void setPlayer(Player player) {
     this.player = player;
+    if (showUserInfo.visibleProperty().isBound()) {
+      showUserInfo.visibleProperty().unbind();
+    }
```
micheljung: Do we have to explicitly unbind? Doesn't a bind() call just replace the existing bind?

1-alex98: Might be I don't know

1-alex98: unbind is not needed...

PR #708, opened 2018-01-17
Reports via Email

### LINKED ISSUE #707: Repot via Email

axel1200:
Fixes #707


1-alex98:
What do we do with that?

1-alex98:
Do we want this?

micheljung:
We certainly want the event publishing refactoring, but I forgot what the status of the "send email" thing is :) Too much to look through atm :P

src/main/java/com/faforever/client/reporting/SupportServiceImpl.java
```diff
@@ -0,0 +1,108 @@
...
+      "username: {0}\n" +
+      "clientVersion: {1}\n" +
+      "supportType: {2}\n" +
+      "system: {3}\n" +
+      "\n" +
+      "error message:\n" +
+      "{4}\n" +
+      "\n" +
+      "lastLogStatements:\n" +
+      "{5}";
```
Geosearchef, line 51: Are we sure that we wanna put this in code instead of a resource file?

1-alex98: Yeah we wanna put it in an resource file

1-alex98:
@micheljung I need to know if we want the email thingy

micheljung:
The email thingy doesn't really work as we would need it to, so no for now?

PR #716, opened 2018-01-27
Show the correct rating in replay detail

### LINKED ISSUE #704: Wrong kind of ratings in replay detail
In the replay detail view the custom game ratings are shown even if it is a 1v1 game. Also the current Rating is shown and not the one at the time of the game. Change that.... You can calculate the correct rating from the playerstats of tht game ;)

20seven:
Changed teamcard so that it displays the ratings of each player during that particular replay.
Fixes #704

### DIFF
--- a/src/main/java/com/faforever/client/game/PlayerCardTooltipController.java
+++ b/src/main/java/com/faforever/client/game/PlayerCardTooltipController.java
@@ -4,7 +4,6 @@ import com.faforever.client.chat.CountryFlagService;
 import com.faforever.client.fx.Controller;
 import com.faforever.client.i18n.I18n;
 import com.faforever.client.player.Player;
-import com.faforever.client.util.RatingUtil;
 import javafx.scene.Node;
 import javafx.scene.control.Label;
 import javafx.scene.image.ImageView;
@@ -30,11 +29,7 @@ public class PlayerCardTooltipController implements Controller<Node> {
     this.i18n = i18n;
   }
 
-  public void setPlayer(Player player) {
-    setPlayerWithRating(player, RatingUtil.getRoundedGlobalRating(player));
-  }
-
-  public void setPlayerWithRating(Player player, int rating) {
+  public void setPlayer(Player player, int rating) {
     if (player == null) {
       return;
     }
--- a/src/main/java/com/faforever/client/game/PlayerCardTooltipController.java
+++ b/src/main/java/com/faforever/client/game/PlayerCardTooltipController.java
@@ -31,12 +31,16 @@ public class PlayerCardTooltipController implements Controller<Node> {
   }
 
   public void setPlayer(Player player) {
+    setPlayerWithRating(player, RatingUtil.getRoundedGlobalRating(player));
+  }
+
+  public void setPlayerWithRating(Player player, int rating) {
     if (player == null) {
       return;
     }
     countryFlagService.loadCountryFlag(player.getCountry()).ifPresent(image -> countryImageView.setImage(image));
 
-    String playerInfoLocalized = i18n.get("userInfo.tooltipFormat", player.getUsername(), RatingUtil.getRoundedGlobalRating(player));
+    String playerInfoLocalized = i18n.get("userInfo.tooltipFormat", player.getUsername(), rating);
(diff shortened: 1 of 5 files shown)


1-alex98 (dismissed):
Well done

1-alex98:
Did you check if the teamcards still work correctly in the Play Tab?

20seven:
@axel1200 Yeah it did however show the exact rating instead of rounding it. Thought it was fine first but it looks a lot cleaner with rounded ratings in play tab. I'd still want exact rating in replaydetails because it makes the ratingchangelabel more useful imo. So last commit just allows us to define if we want to use exact or rounded ratings when showing teams.

1-alex98 (dismissed):
🔝

1-alex98:
@micheljung your turn

micheljung:
To make sure the code is properly formatted, please enable the following options in the commit dialog:
image

For now, you'll have to manually format them (CTRL+ALT+L) :)

1-alex98:
or  make soft reset and recommit

20seven:
I think that should've fixed the formatting. @micheljung

1-alex98 (dismissed):

src/main/java/com/faforever/client/game/PlayerCardTooltipController.java
```diff
@@ -31,12 +31,16 @@ public PlayerCardTooltipController(CountryFlagService countryFlagService, I18n i
   }
 
   public void setPlayer(Player player) {
+    setPlayerWithRating(player, RatingUtil.getRoundedGlobalRating(player));
+  }
+
+  public void setPlayerWithRating(Player player, int rating) {
```
micheljung: I suggest that we remove `setPlayer()` and introduce an enum `RatingMode` with values `EXACT` and `ROUNDED`.
This can then be passed to this method instead of `rating`, reducing the complexity and putting the rating calculation into one place.

src/main/java/com/faforever/client/game/TeamCardController.java
```diff
@@ -59,25 +59,31 @@ public static void createAndAdd(ObservableMap<? extends String, ? extends List<S
...
       TeamCardController teamCardController = uiService.loadFxml("theme/team_card.fxml");
       teamCardController.setPlayersInTeam(entry.getKey(), players,
-          player -> new Rating(player.getGlobalRatingMean(), player.getGlobalRatingDeviation()));
+          player -> new Rating(player.getGlobalRatingMean(), player.getGlobalRatingDeviation()), true);
       teamsPane.getChildren().add(teamCardController.getRoot());
     }
   }
 
-  public void setPlayersInTeam(String team, List<Player> playerList, Function<Player, Rating> ratingProvider) {
+  public void setPlayersInTeam(String team, List<Player> playerList, Function<Player, Rating> ratingProvider, boolean useRoundedRating) {
```
micheljung: `useRoundedRating` can then be replaced by the new enum. Note also how the enum allows multiple values while boolean only allow two, and how passing `RatingMode.ROUNDED` is more expressive than passing `true`.

src/main/java/com/faforever/client/game/TeamCardController.java
```diff
@@ -59,25 +59,31 @@ public static void createAndAdd(ObservableMap<? extends String, ? extends List<S
...
         continue;
       }
       PlayerCardTooltipController playerCardTooltipController = uiService.loadFxml("theme/player_card_tooltip.fxml");
-      playerCardTooltipController.setPlayer(player);
+      int playerRating = RatingUtil.getRating(ratingProvider.apply(player));
+      if (useRoundedRating) {
+        playerCardTooltipController.setPlayer(player);
+      } else {
+        playerCardTooltipController.setPlayerWithRating(player, playerRating);
+      }
```
micheljung: This `if` assumes that `setPlayer` will round the rating, but that might not be true. Neither is the assumption that `setPlayerWithRating` will not round it.

By passing the new enum parameter, we fix this.

micheljung (changes requested):

micheljung (approved):

PR #723, opened 2018-01-30
Sort news items

### LINKED ISSUE #719: News aren't sorted correctly
One would expect the dates to be in descending order:
image

1-alex98:
Sorts news items by data directly when fetching them.
Fixes #719

### DIFF
--- a/src/main/java/com/faforever/client/news/LegacyNewsService.java
+++ b/src/main/java/com/faforever/client/news/LegacyNewsService.java
@@ -20,6 +20,7 @@ import java.net.URL;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 import java.util.Objects;
@@ -92,7 +93,7 @@ public class LegacyNewsService implements NewsService {
 
       result.add(new NewsItem(author, link, title, content, publishedDate, newsCategory));
     }
-
+    result.sort(Comparator.comparing(NewsItem::getDate).reversed());
     return result;
   }
 }


micheljung (approved):

Brutus5000:
Doesn't this break pinning of news items in Wordpress?

micheljung:
what is "pinning of news items"?

Brutus5000:
A pinned news always stays on top of the list, regardless of the publishing date.

micheljung:
That doesn't sound like anything I want ;)

Brutus5000:
You mean stuff like "Incompatible server change - Update your client" or "New community rules" should not be pinned?

micheljung:
Yes, that's what I think. Pinning is a pretty sure solution to whatever problem it's trying to solve.

Addressing your examples:
"Incompatible server change - Update your client". How long do we pin this? A week? A month? If we pin it for a month, what about the players that log in after a month + 1 day?
This should rather be solved by having a mechanism that allows the client to know whether or not its compatible with the server.

"New community rules". Again, how long do we pin this? What about the users who come online later? Which one will be displayed if we need to pin multiple such news? Who decides what needs to be pinned and what not?
In this example, sending out an email to inform everyone (just like when any service updates its TOS) is a much better solution.

Pinning hides other news, and pinning only allows to artificially display some news longer at the top than others. It's like calling `Thread.sleep()`, hoping that an operation gets done within your timeout. Not very reliable.

Brutus5000:
There are problably a lot more examples of just informing people about important stuff going on. The same questions apply on other solutions like pinning something in the #aeolus topic or spamming it via QAI and yet we use these too. However the website is a public place for it and it get's a lot of other "spam" like tournament notifications etc. so that important stuff can leaves the front page very soon.

Yes, this my not be the solution to all of our problems, but the following questions remains: why do you consume the interface if you don't want to follow it's semantics. The wordpress user creating the news does not know that you ignore the pinning. He is just using the features offered by wordpress. If he'd know that you ignore it, maybe he would perform additional steps (spamming QAI or whatever).

micheljung:
The client doesn't show the channel topic for exactly the same reason; it's a poor man's solution to the underlying problem and there are better ways to achieve the same thing.

I'm consuming this interface because there is no other. The question is rather if the interface should be implemented so sticky items come on top, rather than just setting a flag "sticky" and letting the consumer decide what to do about it. After all, the interface's responsibility is to deliver data, not to decide in which order or how they should be displayed.

And that's exactly what the interface does, if you check http://direct.faforever.com/wp-json/wp/v2/posts you'll see that the sticky post isn't at the top but has a flag "sticky". So no, the sorting issue we fixed does not break the sticky feature :)

Btw. the website doesn't seem to do anything with the sticky flag either.

PR #734, opened 2018-02-01
Integrate Rackover's unit db

### LINKED ISSUE #728: Integrate rackovers unit db
Have a setting where user can change between the two databases(spookies we have integrated currently and rackover's one).
- [x] decide which db to show as a default

1-alex98:
Store cookies via a custom CookieStore, that stores cookies in the preferences.
Fixes #728

### DIFF
--- a/src/main/java/com/faforever/client/preferences/Preferences.java
+++ b/src/main/java/com/faforever/client/preferences/Preferences.java
@@ -5,16 +5,24 @@ import com.faforever.client.game.KnownFeaturedMod;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.IntegerProperty;
 import javafx.beans.property.ListProperty;
+import javafx.beans.property.MapProperty;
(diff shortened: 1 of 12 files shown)


src/main/java/com/faforever/client/preferences/Preferences.java
```diff
@@ -43,6 +49,8 @@
   private final VaultPrefs vaultPrefs;
   private final ListProperty<Pair<String, SortType>> gameListSorting;
   private final ObjectProperty<TilesSortingOrder> gameTileSortingOrder;
+  private final ObjectProperty<UnitDataBaseType> unitDataBaseType;
+  private final HashMap<URI, ArrayList<HttpCookie>> storedCookies;
```
1-alex98: Maybe an observable but what kind of observable

1-alex98: Wrap in Objectproperty or in ObservableMap

PR #740, opened 2018-02-02
Fix the event handler for FriendJoinedGameEvent

### LINKED ISSUE #701: Implement "friend joined game" notification
Log says **2018-01-16 14:49:55.574  WARN 235776 --- [lication Thread] com.faforever.client.config.BaseConfig   : No event handler registered for event of type 'FriendJoinedGameEvent**

Implemt a FriendJoinedGameNotifier, like the FriendOnlineNotifier.

20seven:
Removed an uninitialized sound and made sure that the user isn't notified several times for same event.
Fixes #701

### DIFF
--- a/src/main/java/com/faforever/client/audio/AudioServiceImpl.java
+++ b/src/main/java/com/faforever/client/audio/AudioServiceImpl.java
@@ -142,7 +142,8 @@ public class AudioServiceImpl implements AudioService {
     if (!notificationsPrefs.isFriendJoinsGameSoundEnabled()) {
       return;
     }
-    playSound(friendJoinsGameSound);
+    // FIXME implement
+//    playSound(friendJoinsGameSound);
   }
 
   @Override
--- a/src/main/java/com/faforever/client/player/FriendJoinedGameNotifier.java
+++ b/src/main/java/com/faforever/client/player/FriendJoinedGameNotifier.java
@@ -13,8 +13,6 @@ import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import org.springframework.stereotype.Component;
 
-import java.util.ArrayList;
-import java.util.List;
 import javax.annotation.PostConstruct;
 import javax.inject.Inject;
 
@@ -30,7 +28,6 @@ public class FriendJoinedGameNotifier {
   private final JoinGameHelper joinGameHelper;
   private final PreferencesService preferencesService;
   private final AudioService audioService;
-  private List<FriendJoinedGameEvent> eventList;
 
   @Inject
   public FriendJoinedGameNotifier(NotificationService notificationService, I18n i18n, EventBus eventBus,
@@ -42,7 +39,6 @@ public class FriendJoinedGameNotifier {
     this.joinGameHelper = joinGameHelper;
     this.preferencesService = preferencesService;
     this.audioService = audioService;
-    eventList = new ArrayList<>();
   }
 
   @PostConstruct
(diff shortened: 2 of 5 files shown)


src/main/java/com/faforever/client/audio/AudioServiceImpl.java
```diff
@@ -142,7 +142,8 @@ public void playFriendJoinsGameSound() {
     if (!notificationsPrefs.isFriendJoinsGameSoundEnabled()) {
       return;
     }
-    playSound(friendJoinsGameSound);
+    // FIXME implement
```
1-alex98, line 145: Why exactly did you delete it?

20seven:
```
//    friendJoinsGameSound = loadSound(FRIEND_JOINS_GAME_SOUND);
```

That sound doesn't exist. I figured - since friendOffline, friendOnline, and friendPlaysGame are also sounds not implemented yet - it was either a work in progress or someone decided not to use the same sound as the others that are already implemented.

1-alex98: Hmm ok

1-alex98: Would you make an issue on that, if there is no one yet ;)

20seven: issue made: #746

src/main/java/com/faforever/client/player/FriendJoinedGameNotifier.java
```diff
@@ -30,15 +30,19 @@
   private final JoinGameHelper joinGameHelper;
   private final PreferencesService preferencesService;
   private final AudioService audioService;
+  private List<FriendJoinedGameEvent> eventList;
```
micheljung: Since it's initialized in the constructor and not meant to be changed, this should be `final`.

src/main/java/com/faforever/client/player/FriendJoinedGameNotifier.java
```diff
@@ -49,11 +53,13 @@ void postConstruct() {
...
     Player player = event.getPlayer();
-    Game game = player.getGame();
+    Game game = event.getGame();
 
     audioService.playFriendJoinsGameSound();
 
-    if (preferencesService.getPreferences().getNotification().isFriendJoinsGameToastEnabled()) {
+    if (preferencesService.getPreferences().getNotification().isFriendJoinsGameToastEnabled()
+        && !eventList.contains(event)) {
+      eventList.add(event);
```
micheljung: Since the events are never removed from the list, this creates a memory leak. The notifier keeps a reference to all events, the events have a reference to the players and games, those have further references, preventing all these objects from being garbage collected.

I'm not sure why this was added; since each event is only fired exactly once, the `contains()` check should always return false. What happens if we just remove `eventList`?

src/main/java/com/faforever/client/player/FriendJoinedGameNotifier.java
```diff
@@ -49,11 +53,13 @@ void postConstruct() {
   @Subscribe
   public void onFriendJoinedGame(FriendJoinedGameEvent event) {
     Player player = event.getPlayer();
-    Game game = player.getGame();
+    Game game = event.getGame();
```
micheljung, line 52: Not saying it's wrong, but why did you decide to add the game to the event, instead of reading it from the user?

I do see one reason why this is better, but I'd like to know what yours is :)

micheljung (changes requested):

src/main/java/com/faforever/client/player/FriendJoinedGameNotifier.java
20seven: You're right, should've thought further than that easy fix. 

I put it in because, no, it doesn't fire exactly once. In fact, it fires every time someone joins or leaves a game where you have a friend. It also fired each time any change was made to the game (gametitle change, swapped teams, added mods, etc.). If youre unlucky, you have two or more friends in there and it goes completely haywire.

This is how it looks when i befriended someone hosting a popular Gap-game:
sanchopanda joined game


But I've looked over the code in PlayerServiceImpl and worked out a fix (which makes eventList unnecessary)

src/main/java/com/faforever/client/player/FriendJoinedGameNotifier.java
20seven: Well, I had some issues with how games were set to the player. After fixing that, I still think it's more reasonable and logical that the message about a player joining a game should contain exactly that information; the player and the game.

micheljung (approved):

PR #741, opened 2018-02-02
Fix play button to not be active when no faction is selected (on init)

### LINKED ISSUE #721: "Play" button should not be active when no 1v1 faction is selected
image

This only happens if the view is loaded initially.

20seven:
Fixed so that reader parses the FaValue stored in clientprefs and fixed logic for disabling button in init.
Fixes #721

### DIFF
--- a/src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
+++ b/src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
@@ -138,7 +138,7 @@ public class Ladder1v1Controller extends AbstractViewController<Node> {
     for (Faction faction : EnumSet.of(AEON, CYBRAN, UEF, SERAPHIM)) {
       factionsToButtons.get(faction).setSelected(factions.contains(faction));
     }
-    playButton.setDisable(!factionsToButtons.values().stream().anyMatch(ToggleButton::isSelected));
+    playButton.setDisable(factionsToButtons.values().stream().noneMatch(ToggleButton::isSelected));
 
     preferencesService.addUpdateListener(preferences -> {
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/client/remote/gson/FactionTypeAdapter.java
```diff
@@ -27,6 +27,6 @@ public Faction read(JsonReader in) throws IOException {
       in.nextNull();
       return null;
     }
-    return Faction.fromString(in.nextString());
+    return Faction.fromFaValue(Integer.parseInt(in.nextString()));
```
micheljung: I prefer meaningful strings like "aeon" to be stored in the preferences rather than numbers, because of human readability (vs. magic numbers). Therefore, I suggest using `value.getString()` to write the value instead.

src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
```diff
@@ -138,7 +138,7 @@ public void initialize() {
     for (Faction faction : EnumSet.of(AEON, CYBRAN, UEF, SERAPHIM)) {
       factionsToButtons.get(faction).setSelected(factions.contains(faction));
     }
-    playButton.setDisable(factionsToButtons.values().stream().anyMatch(ToggleButton::isSelected));
+    playButton.setDisable(!factionsToButtons.values().stream().anyMatch(ToggleButton::isSelected));
```
micheljung: As IntelliJ suggests, `!anyMatch()` should be replaced by `noneMatch()` :)
image

micheljung (changes requested):

micheljung:
Thanks! :)

PR #756, opened 2018-02-19
Fix mod download

### LINKED ISSUE #717: Missing mods aren't installed correctly before starting game
image
Should it not have installed that mods for me?

1-alex98:
Make some adjustmennts to get mod downloading work.
Fixes #717


src/main/java/com/faforever/client/legacy/UnixUidService.java
```diff
@@ -17,6 +21,7 @@
   @Override
   public String generate(String sessionId, Path logFile) throws IOException {
     String uidDir = System.getProperty("nativeDir", "lib");
+    Files.setPosixFilePermissions(Paths.get(uidDir, "/faf-uid"), Sets.immutableEnumSet(PosixFilePermission.OTHERS_READ));
```
micheljung: No path separator required/desired in `"/faf-uid"`

src/main/java/com/faforever/client/mod/ModServiceImpl.java
```diff
@@ -163,14 +163,15 @@ private Thread startDirectoryWatcher(Path modsDirectory) throws IOException, Int
...
-  public void loadInstalledMods() {
+  public CompletableFuture<Void> loadInstalledMods() {
     try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(modsDirectory, entry -> Files.isDirectory(entry))) {
       for (Path path : directoryStream) {
         addMod(path);
       }
     } catch (IOException e) {
       logger.warn("Mods could not be read from: " + modsDirectory, e);
     }
+    return completedFuture(null);
```
micheljung: There is no need/benefit in returning a future here since it doesn't execute any asynchronous task

src/main/java/com/faforever/client/mod/ModServiceImpl.java
```diff
@@ -201,14 +207,19 @@ public void loadInstalledMods() {
     }
 
     return taskService.submitTask(task).getFuture()
-        .thenAccept(aVoid -> loadInstalledMods());
+        .thenCompose(aVoid -> loadInstalledMods());
```
micheljung: Changing this from `thenAccept` to `thenCompose` makes no difference. The proper call would have been `thenRun` anyway, since the result of the previous future is not of interest and the called method also doesn't return anything.

src/main/java/com/faforever/client/mod/ModServiceImpl.java
```diff
@@ -181,7 +182,12 @@ public void loadInstalledMods() {
   @SneakyThrows
   @Override
   public CompletableFuture<Void> downloadAndInstallMod(String uid) {
-    return fafService.getMod(uid).thenAccept(mod -> downloadAndInstallMod(mod, null, null));
+    return fafService.getModVersion(uid)
+        .thenCompose(mod -> downloadAndInstallMod(mod, null, null))
+        .exceptionally(throwable -> {
+          logger.warn("Sim mod was not installed", throwable);
```
micheljung: `Sim mod could not be installed` (otherwise, it sounds like "it wasn't but now it is")

src/main/java/com/faforever/client/mod/ModVersion.java
```diff
@@ -0,0 +1,234 @@
...
+import javafx.beans.property.SimpleBooleanProperty;
+import javafx.beans.property.SimpleObjectProperty;
+import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.property.StringProperty;
+import org.apache.maven.artifact.versioning.ComparableVersion;
+
+import java.net.URL;
+import java.time.OffsetDateTime;
+
+public class ModVersion {
```
micheljung: As discussed, rename `Mod` instead.

micheljung (changes requested):

micheljung (approved):

PR #758, opened 2018-02-21
Blacklist coop for git updater

### LINKED ISSUE #752: Coop dowloading broken
How to reproduce? Delete your game data files and try to play coop

1-alex98:
Video Files are not on git but only on the server added to the legacy updater manually, that why we need to use the simple http updater.
Fixes #752


1-alex98:
Should we blacklist it and when we blacklist it we would alsoneed to blacklist other feature mods like phantom x

PR #772, opened 2018-02-24
Better function of create game button

### LINKED ISSUE #720: Fix "Create Game" button in map vault
image

The "Create Game" button should either only be visible if the map is installed OR clicking it should install the map first, before trying to create a game.

Opinions appreciated.

1-alex98:
Make the game selected when the button is clicked and only show it when map is installed.
Fixes #720


micheljung:
Was this intentionally closed unmerged?

PR #775, opened 2018-02-26
Feature/#638 compact and extended

20seven:
Started on this when i was really new here and have now tried to clean it up to follow the project's coding style as I've learned from other issues and such. 

How it looks:
Settings:
chatsettings

Compact chat:
chatcompact

It keeps all the css-classes and formatting of the current chat (called Extended in this). But displays it in a table with configured widths to fit avatars, flags, usernames, etc instead.


src/main/java/com/faforever/client/chat/ChatFormat.java
```diff
@@ -0,0 +1,17 @@
+package com.faforever.client.chat;
+
+public enum ChatFormat {
+  EXTENDED("settings.chat.chatFormat.extended"),
+  COMPACT("settings.chat.chatFormat.compact");
+
+  private String i18nKey;
```
1-alex98: make it private final and annotate with @Getter

src/main/java/com/faforever/client/preferences/ChatPrefs.java
```diff
@@ -64,6 +66,14 @@ public void setTimeFormat(TimeInfo time) {
     this.timeFormat.set(time);
   }
 
+  public ChatFormat getChatFormat() {
+    return this.chatFormat.get();
+  }
+
```
1-alex98, line 72: How about getting the property

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -270,6 +273,23 @@ public void onTimeFormatSelected() {
     preferencesService.storeInBackground();
   }
 
+
+  private void configureChatSetting(Preferences preferences) {
+    chatComboBox.setButtonCell(new StringListCell<>(chatFormat -> i18n.get(chatFormat.getI18nKey())));
+    chatComboBox.setCellFactory(param -> new StringListCell<>(chatFormat -> i18n.get(chatFormat.getI18nKey())));
+    chatComboBox.setItems(FXCollections.observableArrayList(ChatFormat.values()));
+    chatComboBox.setDisable(false);
```
1-alex98: Not needed or if needed then change thst in the fxml

src/main/resources/i18n/messages_de_DE.properties
```diff
@@ -313,6 +313,9 @@ settings.chat.previews=Vorschau
 settings.chat.previewImages=Vorschaubiler f�r Internet-Links anzeigen
 settings.chat.colors=Farben
 settings.chat.timeFormat=Zeitformat
+settings.chat.chatFormat=Chatformat
+settings.chat.chatFormat.extended=Verl�ngert
```
1-alex98: Na "Verlängert "sounds strange, maybe "Erweitert"

1-alex98 (changes requested):
Looks cool I think

src/main/java/com/faforever/client/preferences/ChatPrefs.java
20seven: added

src/main/java/com/faforever/client/preferences/ChatPrefs.java
```diff
@@ -64,6 +66,18 @@ public void setTimeFormat(TimeInfo time) {
...
+  public void setChatFormat(ChatFormat chatFormat) {
+    this.chatFormat.setValue(chatFormat);
+  }
+
+  public ObjectProperty<ChatFormat> getChatFormatProperty() {
+    return chatFormat;
+  }
```
Geosearchef: the getter for properties is just called "chatFormatProperty()" throughout the project afaik. (have a look at the method below)

Geosearchef (changes requested):
nice PR

1-alex98 (dismissed):

src/main/java/com/faforever/client/preferences/ChatPrefs.java
20seven: fixed:)

1-alex98 (dismissed):

src/main/resources/theme/style-webview.css
```diff
@@ -130,6 +136,34 @@ img.country-flag[src=""] {
...
+td#flag {
+  width: 16px;
+}
+
+td#time {
+  width: 40px;
+  padding: 1em;
+  text-align: right;
+}
+
```
micheljung, line 165: I'm not sure I'm happy with assuming an 18ch width for the author name, 99% for text and 40px for time.
These things all need to scale well with different window sizes and zoom levels - do they?

I see two additional solutions:
1. We do it like it was initially, the whole chat is one table so whichever username is the longest will determine the width of the username column. All text will then be aligned the same:
```
Short Name       Hi all
Some long name   Hi!
```

2. Or, like DIscord does, texts are not aligned at all:

image

Which one of the 3 do we prefer? I think I'd go with the DIscord solution.

20seven: The reasoning for the sizes:
author = 18ch. Because the max number of characters for a username is 16. Add the fact that many have clans, i found 18ch was a good comprimise.
However, I am increasing it to 23ch to simply cover the longest name + clan tag (21 would've been enough but because of font choices and spacing it needs another 2 to fully wrap a max-length name).

I had text = 99% just to fill out that column. But it is actually not necessary as it is atm. Will remove.

40px time could (should) be replaced by 5ch for HH:MM = 5 characters as it's more appropriate.

flag = 16px, because flag icons are 16x16
avatars = 44px, because avatars are 40x20. 4 px is just a little extra padding.

All of this scales just fine (within reason). But I'm adding a small fix that is needed if the client is shrunk to minimum size or (ridiculously) too zoomed in, so not to break the table-structure.

I asked when i first began with this on how it should look. I remember posting different layouts in slack. **One of the images was of the discord-kind, which got bad reception**. Essentially, it boiled down to "make it look like the python client chat".

I think there are a significant number of people who stay on the python client _because of_ its chat. Why not make it similar to that? One of their biggest grievances will be gone; they can no longer complain about their favorite chat-format not being supported.

Adding a Discord style chat can be the next step (it would be rather simple now), but getting a chat that most FAF:ers can recognize and feel familiar with should come first.

micheljung: Please don't get me wrong, I'm absolutely pro compact mode (as an option) and I'm glad you're implementing this. I just want it to work well :) If not absolutely required, we should avoid assumptions like "the username is 18 characters at max" or "the clan tag is 3 characters at max" - it should just work well with any length since having these limits today doesn't mean we will have them tomorrow (and we don't want to touch code when it changes).

I checked the Python client and this is how it looks like:
image

so the Python client basically has the same implementation as Discord which makes me wonder how people can like the Python client's implementation but not the Discord's one? I think the Python client's implementation is inferior because it's difficult to distinguish between username and text. This is way better in your implementation.

What do others think, @Brutus5000 @axel1200 @Geosearchef @germanicianus (and you, @20seven)?
* If we have the adjust to the longest username, it works for all username lengths but it has the downside that the column grows and shrinks over time
* If we keep it like implemented in this PR, the column always stays the same but there is lots of space wasted (this gets worse if e.g. in future we allow usernames of 40 characters) and we have to take a guess how wide "works in most cases" (otherwise, username gets cut off or multilined)
* If we do it like in Discord (and in fact, the Python client, just with better visibility of the username) we waste no space and it works for any username length, but chat messages aren't aligned.

Let me know your thoughts.

1-alex98: I dont know, I dont see what I like or dislike about it. I might be the wrong person to ask about design aspects :D

1-alex98: I think @20seven implementation is the most straight forward one, if I think about it. I mean it really supports readability and the amount of space wasted is ok I guess. Of curse only because it is this way now(username length), it won't be always like this, but in this very case I think it is neglectable as username length is not gonna ever change . And if so this is not the only playe where username length is limited(game team card thingy).

Brutus5000: I don'tuse the chat very often. Since a few community members stated, they don't switch to the java client because of the superior chat in python one, I'd go with it's design for opportunistic reasons

germanicianus: I suggest to put the flag right after the user name, orherwise the space may be to wide to collate them easily. For badges this doesn‘t matter so much.

After all the chat can be styled with a theme, correct? So we could provide two different styles - one of them matching the Python client chat style.

PR #778, opened 2018-02-27
Fix not corrupted replays being marked as corrupt

### LINKED ISSUE #763: Replays marked as corrupt
All my replays from 2018 were moved to a folder marked corrupt, plus a few more. They are not corrupt.

1-alex98:
Fixes #763

### DIFF
--- a/src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
+++ b/src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
@@ -326,9 +326,9 @@ public class FafApiAccessorImpl implements FafApiAccessor {
 
   @Override
   public Optional<MapVersion> findMapByFolderName(String folderName) {
-    List<MapVersion> maps = getMany(MAP_ENDPOINT, 1, ImmutableMap.of(
-        "include", "latestVersion,author",
-        "sort", "-updateTime"));
+    List<MapVersion> maps = getMany("/data/mapVersion", 1, ImmutableMap.of(
+        "filter", String.format("filename==\"*%s*\"", folderName),
+        "include", "map,map.statistics,reviews"));
     if (maps.isEmpty()) {
       return Optional.empty();
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
```diff
@@ -326,9 +327,9 @@ public Mod getMod(String uid) {
 
   @Override
   public Optional<MapVersion> findMapByFolderName(String folderName) {
-    List<MapVersion> maps = getMany(MAP_ENDPOINT, 1, ImmutableMap.of(
-        "include", "latestVersion,author",
-        "sort", "-updateTime"));
+    List<MapVersion> maps = getMany("/data/mapVersion", 1, ImmutableMap.of(
+        "filter", MessageFormat.format("filename==\"*{0}*\"", folderName),
```
micheljung: `MessageFormat` is for human-readable strings and it adds locale-specific things like separators in `1'000'000` vs `1,000,000` and such. It's, therefore, not suitable for technical string building; use the "dumb" `String.format()` instead.

src/main/java/com/faforever/client/replay/ReplayServiceImpl.java
```diff
@@ -176,7 +177,7 @@ static String guessModByFileName(String fileName) {
                 replayInfos.add(new Replay(replayInfo, replayFile, featuredMod, mapBean.get()));
               });
         } catch (Exception e) {
-          logger.warn("Could not read replay file {} ({})", replayFile, e.getMessage());
+          logger.warn(MessageFormat.format("Could not read replay file {0}", replayFile), e);
```
micheljung: The logger supports using both, string replacement and exceptions as stack traces;

```
 logger.warn("Could not read replay file '{}'", replayFile, e);
```

micheljung (changes requested):

micheljung (approved):

PR #779, opened 2018-02-27
Fix gametype not shown with local replays

### LINKED ISSUE #732: Game type often not given in local replays screen
image
The games are a mixture of ladder games, FAF beta games and FAF games but very few have the game type present.

1-alex98:
Fixes #732

### DIFF
--- a/src/main/java/com/faforever/client/replay/LoadLocalReplaysTask.java
+++ b/src/main/java/com/faforever/client/replay/LoadLocalReplaysTask.java
@@ -26,7 +26,7 @@ public class LoadLocalReplaysTask extends CompletableTask<Collection<Replay>> {
   @Override
   protected Collection<Replay> call() throws Exception {
     updateTitle(i18n.get("replays.loadingLocalTask.title"));
-    return replayService.getLocalReplays();
+    return replayService.getLocalReplays().get();
   }
 
 }
--- a/src/main/java/com/faforever/client/replay/ReplayServiceImpl.java
+++ b/src/main/java/com/faforever/client/replay/ReplayServiceImpl.java
@@ -7,7 +7,6 @@ import com.faforever.client.game.GameService;
 import com.faforever.client.game.KnownFeaturedMod;
 import com.faforever.client.i18n.I18n;
 import com.faforever.client.map.MapService;
-import com.faforever.client.mod.FeaturedMod;
 import com.faforever.client.mod.ModService;
 import com.faforever.client.notification.Action;
 import com.faforever.client.notification.DismissAction;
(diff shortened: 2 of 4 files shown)


src/main/java/com/faforever/client/replay/ReplayServiceImpl.java
```diff
@@ -161,7 +161,7 @@ static String guessModByFileName(String fileName) {
       for (Path replayFile : directoryStream) {
         try {
           LocalReplayInfo replayInfo = replayFileReader.parseMetaData(replayFile);
-          FeaturedMod featuredMod = modService.getFeaturedMod(replayInfo.getFeaturedMod()).getNow(FeaturedMod.UNKNOWN);
+          FeaturedMod featuredMod = modService.getFeaturedMod(replayInfo.getFeaturedMod()).get();
```
1-alex98: getNow is bad as this is connected to a api call here which might or might not finish in time, we could also use a get with timeout, but the api call has a timeout it self

Geosearchef (dismissed):

src/main/java/com/faforever/client/replay/ReplayServiceImpl.java
micheljung: We shouldn't use getNow() nor get(), but instead use thenAccept(), right?

micheljung (changes requested):

src/main/java/com/faforever/client/replay/ReplayServiceImpl.java
1-alex98: Yeah we should

1-alex98: No we can but we don't need to it is all inside a CompletedFutureTask, so I have to call get at some point anyway ;)

src/main/java/com/faforever/client/replay/LoadLocalReplaysTask.java
```diff
@@ -26,7 +26,7 @@ public LoadLocalReplaysTask(ReplayService replayService, I18n i18n) {
   @Override
   protected Collection<Replay> call() throws Exception {
     updateTitle(i18n.get("replays.loadingLocalTask.title"));
-    return replayService.getLocalReplays();
+    return replayService.getLocalReplays().get();
```
1-alex98, line 29: Now get just shifted further uppwards

micheljung: There has to be a solution for this, let's review it together

1-alex98: ok lets do that

PR #788, opened 2018-03-04
hyperlink channel names in chat

### LINKED ISSUE #786: Create channel links in IRC
Is it possible to turn channel names in IRC into a link you can click to open the respective channel?

Geosearchef:


src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -171,6 +173,7 @@
    */
   private String receiver;
   private Pattern mentionPattern;
+  private Pattern channelPattern;
```
1-alex98: Can it not be final?

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -212,6 +215,7 @@ public AbstractChatTabController(ClanService clanService, WebViewConfigurer webV
     String urlFormat = clientProperties.getVault().getReplayDownloadUrlFormat();
     String[] splittedFormat = urlFormat.split("%s");
     replayUrlPattern = Pattern.compile(Pattern.quote(splittedFormat[0]) + "(\\d+)" + Pattern.compile(splittedFormat.length == 2 ? splittedFormat[1] : ""));
+    channelUrlPattern = Pattern.compile("\\QfafJoinChannel://\\E(#[a-z].+)", CASE_INSENSITIVE);
```
1-alex98: Make Patterns constants nO?

src/main/java/com/faforever/client/chat/ChatController.java
```diff
@@ -215,6 +216,13 @@ private void joinChannel(String channelName) {
     chatService.joinChannel(channelName);
   }
 
+  @Subscribe
+  public void onFocusChannelTab(FocusChannelTabEvent focusChannelTabEvent) {
+    Platform.runLater(() -> {
+      tabPane.getSelectionModel().select(getOrCreateChannelTab(focusChannelTabEvent.getChannel()).getRoot());
+    });
+  }
+
```
1-alex98: We are slowly converting to spring event Bus, so better use applicationService.post for new Events. Also this should be a NavigationEvent so it can be used also from the outside. Best would be an event that creates channel and focuses it, can be used in other contexts as well then

src/main/java/com/faforever/client/chat/event/FocusChannelTabEvent.java
```diff
@@ -0,0 +1,10 @@
+package com.faforever.client.chat.event;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+@Getter
```
1-alex98: Use @Data

1-alex98: Does the same

src/test/java/com/faforever/client/chat/AbstractChatTabControllerTest.java
```diff
@@ -194,7 +196,7 @@ public void testOnSendMessageSuccessful() throws Exception {
     instance.onSendMessage();
 
     verify(chatService).sendMessageInBackground(eq(receiver), eq(message));
-    assertThat(instance.messageTextField().getText(), is(emptyString()));
+    assertTrue(instance.messageTextField().getText().equals(""));
```
1-alex98: Why are you changing this, did it not work?

1-alex98 (changes requested):

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
Geosearchef: no, it cannot due to not being initialized in the constructor but in the initialize() method, of course one could change that, but mentionPattern is also initialized there as it requires some information not available in the constructor.

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
Geosearchef: ?

src/main/java/com/faforever/client/chat/event/FocusChannelTabEvent.java
Geosearchef: No it doesn't, it creates @Setter, @EqualsAndHashcode aswell.

src/test/java/com/faforever/client/chat/AbstractChatTabControllerTest.java
Geosearchef: it does not work on linux using the OpenJDK.

src/main/java/com/faforever/client/chat/ChatController.java
Geosearchef: It already auto creates a TabController which auto joins a channel, what do you mean?

Geosearchef: There is no ApplicationService in spring. Do you mean the one from JavaFX?

Geosearchef: Also just changing one post when there are multiple others in the same class doesn't make a lot of sense. Make an issue for it ;).

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
Geosearchef: they are, this is the constructor.

micheljung: axel1200 is right, this should be a constant like:
```
private static final Pattern CHANNEL_URL_PATTERN = Pattern.compile("\\QfafJoinChannel://\\E(#[a-z].+)", CASE_INSENSITIVE);
```

Otherwise, one pattern will be instantiated per instance of `AbstractChatTabController` even though they are identical. So we can just have a single, static instance.

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -281,6 +285,7 @@ public void setReceiver(String receiver) {
 
   public void initialize() {
     mentionPattern = Pattern.compile("\\b(" + Pattern.quote(userService.getUsername()) + ")\\b", CASE_INSENSITIVE);
+    channelPattern = Pattern.compile("#([a-z][^#]+)\\b", CASE_INSENSITIVE);
```
micheljung: Constant as well. `mentionPattern` isn't because it needs to built with runtime information which is not the case for `channelPattern`

Also, I'd expect this pattern to be equal to what's being used in `channelUrlPattern`: `(#[a-z].+)` (or the other way around)

src/main/java/com/faforever/client/chat/ChatController.java
micheljung: He means `ApplicationEventPublisher`. I do agree we'd rather have all fixed at once instead of mixing things up. However, one issue to fix them all would result in a huge PR with a big potential for many conflicts. Therefore, I'd rather see us fix one location at a time when we're touching it anyway - especially when it's new code.

src/test/java/com/faforever/client/chat/AbstractChatTabControllerTest.java
micheljung: This has something to do with new mockito version, even though I don'd understand how it suddenly happened without breaking builds before.

New API is `isEmptyString()` AFAIK.

micheljung (changes requested):

src/main/java/com/faforever/client/chat/ChatController.java
Geosearchef: This is not the right place to fix it, as this class publishes events to about 10 other classes. Also the Event I'm using extends NaviagteEvent, so if we wanna extend ApplicationEvent in that class, we're gonna have to change a lot.

micheljung:
If you can rebase and resolve the merge conflict, I can merge this

1-alex98 (dismissed):

PR #790, opened 2018-03-04
#789 rackover db language settings

### LINKED ISSUE #789: Auto detect and adapt language for rackover's DB

Geosearchef:


src/main/resources/application.yml
```diff
@@ -26,7 +26,7 @@ faf-client:
 
   unit-database:
     spookiesUrl: http://spooky.github.io/unitdb/#/
-    rackOversUrl: http://90.103.145.211/index.php?settings64=eyJwcmV2aWV3Q29ybmVyIjoiTm9uZSJ9
+    rackOversUrl: http://90.103.145.211/index.php?settings64=
```
1-alex98: Please update

1-alex98 (changes requested):

src/main/java/com/faforever/client/units/UnitsController.java
```diff
@@ -3,31 +3,36 @@
...
 public class UnitsController extends AbstractViewController<Node> {
   private final ClientProperties clientProperties;
   private final PreferencesService preferencesService;
   private final CookieService cookieService;
+  private final I18n i18n;
   public WebView unitsRoot;
 
   @Inject
-  public UnitsController(ClientProperties clientProperties1, PreferencesService preferencesService, CookieService cookieService) {
+  public UnitsController(ClientProperties clientProperties1, PreferencesService preferencesService, CookieService cookieService, I18n i18n) {
```
1-alex98: clientProperties1 should be clientProperties

src/main/java/com/faforever/client/units/UnitsController.java
```diff
@@ -41,7 +46,14 @@ public void onDisplay(NavigateEvent navigateEvent) {
...
     UnitDatabase unitDatabase = clientProperties.getUnitDatabase();
-    unitsRoot.getEngine().load(newValue == UnitDataBaseType.SPOOKY ? unitDatabase.getSpookiesUrl() : unitDatabase.getRackOversUrl());
+
+    String rackoverSettingsEncoded = "";
+    if (newValue == UnitDataBaseType.RACKOVER) {
+      String rackoverSettings = new Gson().toJson(new RackoversDBSettings(i18n.getUserSpecificLocale()));
+      rackoverSettingsEncoded = Base64.getEncoder().encodeToString(rackoverSettings.getBytes());
+    }
+
+    unitsRoot.getEngine().load(newValue == UnitDataBaseType.SPOOKY ? unitDatabase.getSpookiesUrl() : unitDatabase.getRackOversUrl() + rackoverSettingsEncoded);
```
1-alex98: Do not concatenate String format String

1-alex98 (changes requested):

src/main/java/com/faforever/client/units/RackoversDBSettings.java
```diff
@@ -0,0 +1,44 @@
...
+      case "ru":
+        lang = "RU";
+        break;
+      case "tzm":
+        lang = "TZM";
+        break;
+      default:
+        lang = "US";
+    }
+  }
```
micheljung, line 43: What happens if we pass an unknown language to the unit DB? I'd expect it to just fall back to US. In which case we could simply write:

```
locale.getLanguage().toUppercase(Locale.US)
```

Otherwise, I propose a list of supported languages like so:
```
private static final Collection<String> SUPPORTED_LANGUAGES = Arrays.asList("CZ", "DE", ...);
...
String uppercaseLanguage = locale.getLanguage().toUppercase(Locale.US));
return SUPPORTED_LANGUAGES .contains(uppercaseLanguage) ? uppercaseLanguage : "US";
```

src/main/java/com/faforever/client/units/UnitsController.java
```diff
@@ -41,7 +46,14 @@ public void onDisplay(NavigateEvent navigateEvent) {
...
     UnitDatabase unitDatabase = clientProperties.getUnitDatabase();
-    unitsRoot.getEngine().load(newValue == UnitDataBaseType.SPOOKY ? unitDatabase.getSpookiesUrl() : unitDatabase.getRackOversUrl());
+
+    String rackoverSettingsEncoded = "";
+    if (newValue == UnitDataBaseType.RACKOVER) {
+      String rackoverSettings = new Gson().toJson(new RackoversDBSettings(i18n.getUserSpecificLocale()));
+      rackoverSettingsEncoded = Base64.getEncoder().encodeToString(rackoverSettings.getBytes());
+    }
+
+    unitsRoot.getEngine().load(newValue == UnitDataBaseType.SPOOKY ? unitDatabase.getSpookiesUrl() : String.format(unitDatabase.getRackOversUrl(), rackoverSettingsEncoded));
```
micheljung: In the `if` we check if it's Rackover's, here we check if it's SPOOKY's and assume it's Rackover's otherwise.
One of the checks is redundant, and it only works as long as there are only two values.

I suggest we refactor this to use a switch-case with a default case that throws a `ProgrammingException` if `newValue` is an unimplemented enum value.

micheljung (changes requested):

src/main/java/com/faforever/client/units/RackoversDBSettings.java
Geosearchef: The only problem with that approach is that the database (the forged alliance game files) do not use those ISO 639-1 country codes. Also there was another reason why I created a switch statement having todo with ISO codes being subject to change, but can't remember that right now.

Geosearchef:
this is probably ready to be merged (see review comment)

1-alex98:
It needs a rebase and did you look at codacy?

src/main/java/com/faforever/client/units/RackoversDBSettings.java
micheljung: My solution above does the exact same thing as the switch case but it's way shorter :)

1-alex98:
@Geosearchef still doing this?

Rackover:
@Geosearchef yes he is, isn't he? :P

Geosearchef:
not at the moment, pls rebase if you wanna merge, it's one small commit

1-alex98:
U got 1 Month till I close it

PR #795, opened 2018-03-10
Return empty String to avoid NullPointerException

### LINKED ISSUE #794: NullPointer on player having no games
image

1-alex98:
Fixes #794

### DIFF
--- a/src/main/java/com/faforever/client/chat/UserInfoWindowController.java
+++ b/src/main/java/com/faforever/client/chat/UserInfoWindowController.java
@@ -443,6 +443,9 @@ public class UserInfoWindowController implements Controller<Node> {
         int number = object.intValue();
         int numberOfDataPoints = dataPoints.size();
         int dataPointIndex = number >= numberOfDataPoints ? numberOfDataPoints - 1 : number;
+        if (dataPointIndex >= dataPoints.size() || dataPointIndex < 0) {
+          return "";
+        }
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/chat/UserInfoWindowController.java
```diff
@@ -443,7 +443,7 @@ public String toString(Number object) {
         int number = object.intValue();
         int numberOfDataPoints = dataPoints.size();
         int dataPointIndex = number >= numberOfDataPoints ? numberOfDataPoints - 1 : number;
-        return DATE_FORMATTER.format(dataPoints.get(dataPointIndex).getInstant());
+        return dataPoints.get(dataPointIndex) == null ? "" : DATE_FORMATTER.format(dataPoints.get(dataPointIndex).getInstant());
```
micheljung: Actually, the issue #794 is about getting an `ArrayIndexOutOfBoundsException` rather than `NullPointerException` which is caused when `dataPoints` does not contain an index `dataPointIndex` - so the exact same thing still happens with the code above.

The proper fix would be:
```
if(dataPointIndex >= dataPoints.size()) {
  return "";
}
return dataPoints.get(dataPointIndex);
```

Please adjust the issue title and commit message as well.

micheljung (changes requested):

micheljung (approved):

PR #810, opened 2018-03-20
Consider tab to be null, during reconnect...

### LINKED ISSUE #806: Loading symbol shown in wrong place
image
Chat reconnect thingy is also shown in game tab

1-alex98:
Also resolves that auto join channels are not reloaded on reconnect. As a side effect channels that are closed by user before reconnect(means explicit reconnect) are reopened, but I think that's fine, if user decides to reconnect(his fault XD). We could remember the autojoin channels the User left, but I thing reconnect should be a complete restart. "Consider no tab to be selected while reconnect" is happens because channels are removed on disconnect and no tab is selected so we need a null check there,
Fixes #806

### DIFF
--- a/src/main/java/com/faforever/client/chat/ChatController.java
+++ b/src/main/java/com/faforever/client/chat/ChatController.java
@@ -71,7 +71,7 @@ public class ChatController extends AbstractViewController<Node> {
   private void onConnected() {
     connectingProgressPane.setVisible(false);
     tabPane.setVisible(true);
-    noOpenTabsContainer.setVisible(false);
+    noOpenTabsContainer.setVisible(tabPane.getTabs().isEmpty());
   }
 
   private void onConnecting() {
@@ -85,8 +85,6 @@ public class ChatController extends AbstractViewController<Node> {
   }
 
   private void removeTab(String playerOrChannelName) {
-    nameToChatTabController.remove(playerOrChannelName);
-
     if (nameToChatTabController.containsKey(playerOrChannelName)) {
       tabPane.getTabs().remove(nameToChatTabController.remove(playerOrChannelName).getRoot());
     }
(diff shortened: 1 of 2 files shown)


Geosearchef (dismissed):

src/main/java/com/faforever/client/chat/ChatController.java
```diff
@@ -242,7 +242,7 @@ public void onDisplay(NavigateEvent navigateEvent) {
     super.onDisplay(navigateEvent);
     if (!tabPane.getTabs().isEmpty()) {
       Tab tab = tabPane.getSelectionModel().getSelectedItem();
-      nameToChatTabController.get(tab.getId()).onDisplay();
+      Optional.ofNullable(nameToChatTabController.get(tab.getId())).ifPresent(AbstractChatTabController::onDisplay);
```
micheljung: The real question is: How can we have a tab in the UI but no entry in `nameToChatTabController`?
That should not be possible, so here we're just concealing a bug.

Hint: synchronize access to `tabPane.getTabs()` and `nameToChatTabController` on `nameToChatTabController`.

micheljung (changes requested):

src/main/java/com/faforever/client/chat/ChatController.java
1-alex98: it happens if you lose connection

1-alex98: The thing is it is possible that there is no tab selected on reconnect

micheljung: How is it possible that no tab is selected? To my understanding, if there are tabs then there is always exactly one selected.
The current implementation is not thread-safe. We have two lists and we assume that one list contains something the other list does as well, but this might change just after the first check.

Either we need to synchronize, or we need to make sure that all accesses happen in the application thread.

micheljung:
It's still not thread-safe, please see/implement my comment

PR #815, opened 2018-03-22
Bugfix/#782 apply ircfilter on user login

20seven:
1. Applies filter on new chat users logging in.
2. If one applied a PlayerStatus filter (i.e. only show PlayerStatus.PLAYING), it could not be unset. Opted for RadioMenuItems in a togglegroup as more than one status should not be selectable anyway.
3. Filter should apply despite closing the popup, so now shows with standard togglebutton styling whether or not a filter is active/applied.

### DIFF
--- a/src/main/java/com/faforever/client/chat/ChannelTabController.java
+++ b/src/main/java/com/faforever/client/chat/ChannelTabController.java
@@ -37,6 +37,7 @@ import javafx.scene.control.Tab;
 import javafx.scene.control.TextField;
 import javafx.scene.control.TextInputControl;
 import javafx.scene.control.TitledPane;
+import javafx.scene.control.ToggleButton;
 import javafx.scene.input.KeyCode;
 import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.HBox;
@@ -80,7 +81,7 @@ public class ChannelTabController extends AbstractChatTabController {
   private final Map<String, Map<Pane, ChatUserItemController>> userToChatUserControls;
   private final ThreadPoolExecutor threadPoolExecutor;
   private final TaskScheduler taskScheduler;
-  public Button advancedUserFilter;
+  public ToggleButton advancedUserFilter;
   public HBox searchFieldContainer;
   public Button closeSearchFieldButton;
   public TextField searchField;
@@ -315,13 +316,7 @@ public class ChannelTabController extends AbstractChatTabController {
 
     userFilterController = uiService.loadFxml("theme/chat/user_filter.fxml");
     userFilterController.setChannelController(this);
-    userFilterController.getIsFilterAppliedProperty().addListener((observable, oldValue, newValue) -> {
-      if (!oldValue && newValue) {
-        advancedUserFilter.setStyle("-fx-text-fill: -swatch-400");
-      } else if (oldValue && !newValue) {
-        advancedUserFilter.setStyle(null);
-      }
-    });
+    userFilterController.filterAppliedProperty().addListener(((observable, oldValue, newValue) -> advancedUserFilter.setSelected(newValue)));
     filterUserPopup.getContent().setAll(userFilterController.getRoot());
   }
 
@@ -452,7 +447,7 @@ public class ChannelTabController extends AbstractChatTabController {
       if (!userSearchTextField.textProperty().get().isEmpty()) {
         chatUserItemController.setVisible(isUsernameMatch(chatUserItemController));
       }
-      if (userFilterController.getIsFilterAppliedProperty().get()) {
+      if (userFilterController.isFilterApplied()) {
         chatUserItemController.setVisible(userFilterController.filterUser(chatUserItemController));
       }
     }
@@ -586,12 +581,13 @@ public class ChannelTabController extends AbstractChatTabController {
   }
 
   public void onAdvancedUserFilter(ActionEvent actionEvent) {
+    advancedUserFilter.setSelected(userFilterController.isFilterApplied());
     if (filterUserPopup.isShowing()) {
       filterUserPopup.hide();
       return;
     }
 
-    Button button = (Button) actionEvent.getSource();
+    ToggleButton button = (ToggleButton) actionEvent.getSource();
 
     Bounds screenBounds = advancedUserFilter.localToScreen(advancedUserFilter.getBoundsInLocal());
     filterUserPopup.show(button.getScene().getWindow(), screenBounds.getMinX(), screenBounds.getMaxY());
--- a/src/main/java/com/faforever/client/chat/ChannelTabController.java
+++ b/src/main/java/com/faforever/client/chat/ChannelTabController.java
@@ -315,6 +315,13 @@ public class ChannelTabController extends AbstractChatTabController {
 
     userFilterController = uiService.loadFxml("theme/chat/user_filter.fxml");
     userFilterController.setChannelController(this);
+    userFilterController.getIsFilterAppliedProperty().addListener((observable, oldValue, newValue) -> {
+      if (!oldValue && newValue) {
+        advancedUserFilter.setStyle("-fx-text-fill: -swatch-400");
+      } else if (oldValue && !newValue) {
+        advancedUserFilter.setStyle(null);
+      }
+    });
     filterUserPopup.getContent().setAll(userFilterController.getRoot());
   }
 
@@ -445,8 +452,8 @@ public class ChannelTabController extends AbstractChatTabController {
       if (!userSearchTextField.textProperty().get().isEmpty()) {
         chatUserItemController.setVisible(isUsernameMatch(chatUserItemController));
       }
-      if (filterUserPopup.isShowing()) {
-        userFilterController.filterUser(chatUserItemController);
+      if (userFilterController.getIsFilterAppliedProperty().get()) {
+        chatUserItemController.setVisible(userFilterController.filterUser(chatUserItemController));
       }
     }
   }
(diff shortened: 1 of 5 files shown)


src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -33,6 +36,8 @@
   public TextField clanFilterField;
   public TextField minRatingFilterField;
   public TextField maxRatingFilterField;
+  public ToggleGroup gameStatusToggleGroup;
+  private BooleanProperty isFilterApplied = new SimpleBooleanProperty(false);
```
micheljung: * As a convention, we always initialize fields in the constructor, never in the declaration
* Since this field should never change, it should be `final`
* The JavaBean naming convention for boolean fields is `isXXX` where `XXX` is the property name. Therefore, the prefix `is` in the field should be removed, otherwise we'll get `isIsFilterApplied`

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -70,6 +76,10 @@ boolean filterUser(ChatUserItemController chatUserItemController) {
         && isGameStatusMatch(chatUserItemController);
   }
 
+  public BooleanProperty getIsFilterAppliedProperty() {
```
micheljung: The naming convention here is: `isFilterAppliedProperty`

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -124,23 +134,42 @@ boolean isGameStatusMatch(ChatUserItemController chatUserItemController) {
...
+    setOrUnsetGameStatusFilter(LOBBYING);
   }
 
   public void onGameStatusNone(ActionEvent actionEvent) {
-    playerStatusFilter = IDLE;
-    gameStatusMenu.setText(i18n.get("game.gameStatus.none"));
+    setOrUnsetGameStatusFilter(IDLE);
+  }
+
+  private void setOrUnsetGameStatusFilter(PlayerStatus newStatus) {
```
micheljung: "Set or unset" is not very intuitive to understand, a bit magical and thus error-prone.
I'd rather like to see the call like this:

```
if(((Toggle) actionEvent.getTarget()).isSelected()) {
  setGameStatusFilter(IDLE);
} else {
  setGameStatusFilter(null);
}
```

Yes, it's more code and less fancy, but very explicit and thus easy to understand, no magic involved.

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -124,23 +134,42 @@ boolean isGameStatusMatch(ChatUserItemController chatUserItemController) {
...
+    playerStatusFilter = newStatus == playerStatusFilter ? null : newStatus;
+    setGameStatusMenuText(playerStatusFilter);
     filterUsers();
   }
 
+  private void setGameStatusMenuText(PlayerStatus status) {
+    if (status == null) {
+      gameStatusMenu.setText(i18n.get("game.gameStatus"));
+      gameStatusToggleGroup.selectToggle(null);
```
micheljung: If you `return;` at the end of this `if`, you don't need an `else` and can avoid nested blocks and indentation.
Also, a `default` case should always be added to make sure that there is no uncovered enum value - or at least to let us know there is (if someone adds a value, he might not be aware that he needs to adjust this code accordingly, so let's tell him).

```
private void setGameStatusMenuText(PlayerStatus status) {
  if (status == null) {
    gameStatusMenu.setText(i18n.get("game.gameStatus"));
    gameStatusToggleGroup.selectToggle(null);
    return;
  }

  switch (status) {
    case PLAYING:
      gameStatusMenu.setText(i18n.get("game.gameStatus.playing"));
      break;
    case LOBBYING:
       gameStatusMenu.setText(i18n.get("game.gameStatus.lobby"));
      break;
    case IDLE:
      gameStatusMenu.setText(i18n.get("game.gameStatus.none"));
      break;
    default:
      throw new ProgrammingError("Uncovered player status: " + status);
  }
}
```

Oh and since `setGameStatusMenuText` doesn't actually **set** the value given to it but does some calculation based on the given value to eventually **update** the property with some value, I'd rather call it `updateGameStatusMenuText`.

micheljung (changes requested):

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -61,6 +68,7 @@ private void filterUsers() {
         chatUserItemController.setVisible(filterUser(chatUserItemController));
       }
     }
+    filterApplied.set(!maxRatingFilterField.getText().isEmpty() || !minRatingFilterField.getText().isEmpty() || !clanFilterField.getText().isEmpty() || playerStatusFilter != null);
```
1-alex98, line 71: Donot you wanna bind these properties

1-alex98 (changes requested):

src/main/java/com/faforever/client/chat/UserFilterController.java
1-alex98: Ok you dont

1-alex98 (dismissed):

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -445,8 +447,8 @@ private synchronized void onUserJoinedChannel(ChatUser chatUser) {
       if (!userSearchTextField.textProperty().get().isEmpty()) {
         chatUserItemController.setVisible(isUsernameMatch(chatUserItemController));
       }
-      if (filterUserPopup.isShowing()) {
-        userFilterController.filterUser(chatUserItemController);
+      if (userFilterController.isFilterAppliedProperty().get()) {
```
Geosearchef: This somewhat doubled, instead I'd add a isFilterApplied() getter to UserFilterController. (additionally to the property getter)

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -579,12 +581,13 @@ public void addSearchFieldListener() {
   }
 
   public void onAdvancedUserFilter(ActionEvent actionEvent) {
+    advancedUserFilter.setSelected(userFilterController.isFilterAppliedProperty().getValue());
```
Geosearchef: just use property.get() instead of getValue() as above, getValue() returns a Double, get() a double, here the primitive double is expected, won't change any behaviour but would add more consistency.

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -124,23 +136,55 @@ boolean isGameStatusMatch(ChatUserItemController chatUserItemController) {
...
   }
 
   public void onGameStatusNone(ActionEvent actionEvent) {
-    playerStatusFilter = IDLE;
-    gameStatusMenu.setText(i18n.get("game.gameStatus.none"));
+    if (playerStatusFilter == IDLE) {
+      updateGameStatusMenuText(null);
+    } else {
+      updateGameStatusMenuText(IDLE);
+    }
```
Geosearchef: You are doing this 3 times in a row, pls extract the if else logic into a method (e.g. updateGameStatusMenuText).

Geosearchef (changes requested):
good PR, needs some minor cleanup in my opinion, especially the doubled(tripled) code.

src/main/java/com/faforever/client/chat/UserFilterController.java
20seven: I did, but then @micheljung requested this:

>I'd rather like to see the call like this:
>
>if(((Toggle) actionEvent.getTarget()).isSelected()) {
>  setGameStatusFilter(IDLE);
>} else {
>  setGameStatusFilter(null);
>}
>
>Yes, it's more code and less fancy, but very explicit and thus easy to understand, no magic >involved.

Geosearchef: I see now. SetOrUnset... is a very confusing method name, I agree, rather choose sth like update or so. But if @micheljung said so then it should be fine, is a bit overkill but makes it more readable.

src/main/java/com/faforever/client/chat/ChannelTabController.java
Geosearchef: Or just introduce a primtive boolean getter.

Geosearchef (dismissed):

src/main/java/com/faforever/client/chat/UserFilterController.java
micheljung: I agree, having the if-else 3 times is ugly. How about this?

```
updateGameStatusMenuText(playerStatusFilter == IDLE ? null : IDLE);
```

20seven: Fixed

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -70,6 +78,14 @@ boolean filterUser(ChatUserItemController chatUserItemController) {
         && isGameStatusMatch(chatUserItemController);
   }
 
+  public BooleanProperty isFilterAppliedProperty() {
```
micheljung: `filterAppliedProperty`

micheljung (changes requested):
Almost ;)

1-alex98 (dismissed):

src/main/java/com/faforever/client/chat/UserFilterController.java
20seven: Done!:)

Issue #821, opened 2018-03-24
UI crashes when starting it with java.exe because missing install4j


germanicianus:
Workaround is to comment the following line in MainController.java:
noCatch(SplashScreen::hide);

This is the one and only place where install4j is referenced in the code.

1-alex98:
But do we wanna support starting it that way, I would say we dont care

Geosearchef:
We do wanna support starting it without Install4J, but that has very low priority. I had a discussion with downlord on this, we shouldn't invest into linux compatibility, but the client has already come a long way and is only missing packaging steps and some minor fixes to be fully linux compatible. So I'd say we care. If you wanna discuss if we'll do this please move to slack.

micheljung:
Try to load the class via classloader (Class.forName()), if it's there execute the code else don't

PR #838, opened 2018-03-26
  #821 hide install4j splash via reflection

Geosearchef:


src/main/java/com/faforever/client/main/MainController.java
```diff
@@ -338,6 +339,20 @@ public void display() {
     registerWindowListeners();
   }
 
+  private void hideSplashScreen() {
+    try {
+      Class splashScreenClass = Class.forName("com.install4j.api.launcher.SplashScreen");
+      Method hideMethod = splashScreenClass.getDeclaredMethod("hide");
+      hideMethod.invoke(null);
+    } catch (ClassNotFoundException e) {
```
germanicianus: Right, this is the case we want: Catching the case of no available install4j. So only debug message.

src/main/java/com/faforever/client/main/MainController.java
```diff
@@ -338,6 +339,20 @@ public void display() {
...
+      log.debug("No install4j splash screen found to close.");
+    } catch (NoSuchMethodException | IllegalAccessException e) {
```
germanicianus: This would be unexpected, so I suggest to log an error. Same for the InvocationTargetException which means that the called method has thrown an error (like ConnectionException of hide()).

Geosearchef: So you mean error instead of warning?

src/main/java/com/faforever/client/main/MainController.java
Geosearchef: Is this a confirmation that I put enough thought into this? ;)

germanicianus: Yes :D .

src/main/java/com/faforever/client/main/MainController.java
germanicianus: Yes.

src/main/java/com/faforever/client/main/MainController.java
```diff
@@ -338,6 +339,20 @@ public void display() {
     registerWindowListeners();
   }
 
+  private void hideSplashScreen() {
```
germanicianus: I suggest to make variables final wherever possible. Here is some explanation why:
http://www.codeaffine.com/2014/04/28/java-code-style-the-final-decision/
I also suggest to make the method static to make clear that it's standalone and has no object dependencies.

micheljung:
Looks good, thanks for the reviews, saved me time :) To make sure the issue is automatically closed when merging the PR, please write the commit message like this:

```
Hide install4j splash via reflection

Fixes #821
```

Oh and maybe add a comment to the code explaining why this is done. It seems weird for someone who doesn't know the reason behind it.

Geosearchef:
done.

micheljung (approved):

PR #839, opened 2018-03-26
Exclude hidden maps/mods from vault query

### LINKED ISSUE #755: Map vault search includes hidden maps
Query string needs to be extended bei mapVersion.hidden=false criterion in background.
Mod vault probably affected too.

Geosearchef:
Fixes #755

### DIFF
--- a/src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
+++ b/src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
@@ -389,7 +389,7 @@ public class FafApiAccessorImpl implements FafApiAccessor {
   public List<Mod> findModsByQuery(SearchConfig searchConfig, int page, int count) {
     MultiValueMap<String, String> parameterMap = new LinkedMultiValueMap<>();
     if (searchConfig.hasQuery()) {
-      parameterMap.add("filter", searchConfig.getSearchQuery());
+      parameterMap.add("filter", searchConfig.getSearchQuery() + ";latestVersion.hidden==\"false\"");
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/map/MapVaultController.java
```diff
@@ -136,7 +136,7 @@ public void initialize() {
   }
 
   private void searchByQuery(SearchConfig searchConfig) {
-    this.query = searchConfig.getSearchQuery();
+    this.query = searchConfig.getSearchQuery() + ";latestVersion.hidden==\"false\"";
```
micheljung: There are different possible filter dialects, and the field name is subject to change as well.
Hardcoding this here is, therefore, not a future-proof solution. Also, when we display the search query to the user, this part will be left out so he might wonder why he gets different results.

The proper solution will probably include adding a pre-defined condition as a parameter to `com.faforever.client.vault.search.SearchController#buildQuery`. And I'd put the field name into `SearchableProperties` so that they are all at one place and not distributed through the application.

 It's not going to be easy but it will be much more robust.

micheljung (changes requested):

src/main/java/com/faforever/client/map/MapVaultController.java
micheljung: After discussing this with alex, we decided to add the filter in API accessor.
This makes the implementation very easy

micheljung (changes requested):

1-alex98 (approved):

micheljung (approved):

PR #845, opened 2018-03-27
init sentry integration

Cheyans:
This is a first pass implementation, I still need to add the following:
- [ ] Tests
- [x] A way to handle development environments
- [ ] Auto breadcrumbs to handle user action tracking
- [ ] Remove manual error reporting


micheljung:
Wait... are you Aulex or did someone hijack your account? Welcome back :D

build.gradle
```diff
@@ -297,6 +297,7 @@ dependencies {
   compile("com.github.jasminb:jsonapi-converter:${jsonapiConverterVersion}")
   compile("com.github.Brutus5000:BiReUS-JCL:${bireusVersion}")
   compile("com.github.micheljung:i4jruntime:${i4jruntimeVersion}")
+  compile("io.sentry:sentry:1.7.2")
```
micheljung: Version number should be extracted to properties like above

src/main/java/com/faforever/client/reporting/ReportingServiceImpl.java
```diff
@@ -14,8 +18,22 @@
...
+        .setUsername(username)
+        .setId(String.valueOf(userId));
+    context.setUser(userBuilder.build());
+  }
+
   @Override
   public void reportError(Throwable e) {
-    logger.warn("Reporting has not yet been implemented");
+    Sentry.capture(e);
+    LoggerFactory.getLogger((MethodHandles.lookup().lookupClass())).warn("Logging error to sentry: {}", e.toString());
```
micheljung: If you use `@Slf4J` on the class, then you can call `log.warn()`.

However, this should be logged *before* sentry is called. Otherwise, it should be past tense (but I prefer before because if something goes wrong, we don't lose the initial exception). And logging the stack trace is preferred over logging the exception message.

src/main/java/com/faforever/client/reporting/AutoReportedException.java
```diff
@@ -0,0 +1,13 @@
+package com.faforever.client.reporting;
+
+import io.sentry.Sentry;
+import org.slf4j.LoggerFactory;
+
+import java.lang.invoke.MethodHandles;
+
+public class AutoReportedException extends RuntimeException {
+    public AutoReportedException(Throwable cause) {
+      Sentry.capture(cause);
```
micheljung: Constructors should not call external code, especially code that does network requests.

The proper way to handle uncaught exceptions is to configure `Thread.setDefaultUncaughtExceptionHandler` (since there will also be other uncaught exceptions other than the ones created by NoCatch). Also, see `AsyncConfig` where a `SimpleAsyncUncaughtExceptionHandler` is configured.

micheljung (changes requested):

src/main/java/com/faforever/client/reporting/ReportingServiceImpl.java
Cheyans: The proper fix is to remove this method all together, hence the last change listed in the issue op, the point of sentry is to remove all manual exception reporting from the app.

src/main/java/com/faforever/client/reporting/AutoReportedException.java
Cheyans: Sentry already implements the interface
https://github.com/getsentry/sentry-java/blob/38aaffa2d863fdc858cc5f9b24ff2e87d14b5a7a/sentry/src/main/java/io/sentry/SentryUncaughtExceptionHandler.java#L13
I do agree that the constructor shouldn't be making any outbound requests, this was more of a way for me to do end to end testing. The proper solution is to just let all the exceptions propagate up.

Cheyans:
This pr is not in a mergeable state by any means, I'm mostly putting it up here to properly track my progress.

src/main/java/com/faforever/client/config/BaseConfig.java
```diff
@@ -28,8 +29,13 @@ MessageSource messageSource() {
...
   @Bean
   EventBus eventBus() {
-    EventBus bus = new EventBus((exception, context) -> log.warn("Exception in '{}#{}' while handling event: {}",
-        context.getSubscriber().getClass(), context.getSubscriberMethod().getName(), context.getEvent(), exception));
+    EventBus bus = new EventBus((exception, context) -> {
+      log.warn(
+          "Exception in '{}#{}' while handling event: {}",
+          context.getSubscriber().getClass(), context.getSubscriberMethod().getName(), context.getEvent(), exception
+      );
+      Sentry.capture(exception);
```
Cheyans, line 37: need to use `reportingService#reportError`

src/main/java/com/faforever/client/reporting/ReportingServiceImpl.java
micheljung: With a limit of 10k errors per month, I do not think that automated reporting is a good idea

Cheyans: I disagree, there are a variety of rate limiting options in sentry. Also, if we feel that sentry provides enough value we can easily spin up our own instance. https://hub.docker.com/_/sentry/
As noted in their installation process it can store quite a bit of events in about 1.5 tb 
https://docs.sentry.io/server/installation/
I don't think we'll hit anywhere near that and also this might be a real easy way to allow multi user access.

1-alex98:
Closing due to inactivity and because we have bugsnag

PR #854, opened 2018-03-30
Auto-joining chat-channels

### LINKED ISSUE #816: Autojoin chat channels on startup
The python client has an autojoin channels setting:
autojoinchannels
https://github.com/FAForever/client/pull/901

 Seems to be a very appreciated feature.

20seven:
Adding a setting which allows adding multiple chat-channels that will be automatically joined on client startup.
Fixes #816


src/main/java/com/faforever/client/preferences/ChatPrefs.java
```diff
@@ -46,6 +52,7 @@ public ChatPrefs() {
     userToColor = new SimpleMapProperty<>(FXCollections.observableHashMap());
     chatColorMode = new SimpleObjectProperty<>(CUSTOM);
     idleThreshold = new SimpleIntegerProperty(10);
+    autoJoinChannels = new SimpleListProperty<>(FXCollections.observableArrayList(new ArrayList<String>()));
```
micheljung: `new SimpleListProperty<>(FXCollections.observableArrayList());`

src/main/java/com/faforever/client/preferences/ui/AutoJoinChannelsController.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+import javax.inject.Inject;
+
+@Component
+@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+public class AutoJoinChannelsController implements Controller<Node> {
+  private final PreferencesService preferencesService;
+  @FXML
```
micheljung: `@FXML` is only needed for private fields, by convention we use public fields w/o the annotation

src/main/java/com/faforever/client/preferences/ui/AutoJoinChannelsController.java
```diff
@@ -0,0 +1,52 @@
...
+  public TextField channelTextField;
+  public Button addChannelButton;
+  public ListView<String> channelListView;
+  public GridPane autoJoinChannelsSettingsRoot;
+
+  @Inject
```
micheljung: `@Inject` is not required with only one constructor so we omit it.

src/main/java/com/faforever/client/preferences/ChatPrefs.java
```diff
@@ -164,4 +171,16 @@ public void setIdleThreshold(int idleThreshold) {
...
+    return autoJoinChannels.get();
+  }
+
+  public void addAutoJoinChannel(String channelName) {
+    autoJoinChannels.get().add(channelName);
+  }
+
+  public void removeAutoJoinChannel(String channelName) {
+    autoJoinChannels.get().remove(channelName);
+  }
```
micheljung: For list properties in other classes, we don't provide "add" and "remove" so, for the sake of consistency, I wouldn't do it here either.

src/main/java/com/faforever/client/preferences/ui/AutoJoinChannelsController.java
```diff
@@ -0,0 +1,52 @@
...
+  public AutoJoinChannelsController(PreferencesService preferencesService) {
+    this.preferencesService = preferencesService;
+  }
+
+  public void initialize() {
+    channelListView.setOnMouseClicked(event -> {
```
micheljung: By convention, we define the `onMouseClicked` method to be called in the FXML

Geosearchef (dismissed):
looks  good to me

micheljung (changes requested):

Wesmania:
Few things to take note of:
* It's better to use getLanguage since language channels are a better fit than country channels.
* For many languages there won't really be official rooms with moderators, so I don't think there's much point in autojoining those. The Python client currently autojoins #french, #german and #russian, and these have active mods IIRC.
* Python client also has logic for autojoining the #newbie channel if current player has played few games. Maybe we can add this too.

If you'd like to look up Python client's autojoin logic for rough reference / inspiration, it's in `src/chat/channel_autojoiner.py` on develop.

20seven:
@Wesmania yep, language is probably better; changed it.
I also added the three "official" channels, in a way i hope is easy to add onto if needed.

The newbie channel bothered me a bit because I couldn't figure out a way that wouldn't make it all messy. I checked the server and I would consider that a much better place for this. In `lobbyconnection.py` from lines 672:

```python
channels = []
if self.player.mod:
   channels.append("#moderators")

if self.player.clan is not None:
    channels.append("#%s_clan" % self.player.clan)

 if self.player.numGames < 50:
    channels.append("#newbie")
```
I can try and explain my issues with newbie channel in slack (or maybe someone already knows a solution) but I found those two last lines much more convenient. Let me know what you think

1-alex98:
@20seven any ambitions to continue here? Because otherwise I will probably redo it.
Because:
1.Now the py client is gone I would rather take the language local the user set his client and system to and join the channel with the according language code
2. I would not use a pop up but do it in the normal settings, if at all I would just join them and not leave a chance to join any other channels automatically that can be done in a separate issue

Wesmania:
Remember that channels need moderators, so one of the above still holds.

1-alex98:
oh ok

1-alex98:
Well I checked out the branch tested etc and it is perfectly fine. No idea why i wanted to redo it

1-alex98 (dismissed):

1-alex98 (dismissed):

1-alex98:
I rebased and tested it works perfectly

1-alex98:
@Geosearchef @Brutus5000 please have a look

1-alex98 (approved):

Katharsas:
I cannot test this unless its rebased on top of a more recent develop commit.
Edit: I guess is can just rebase this for myself. Do you want me to push a rebased version? Otherwise ill let you deal with the merge conflicts (they probably aren't too bad) : )

1-alex98:
U can happily rebase onto develop and u can also locally rebase it onto something else and not push

1-alex98 (dismissed):

src/main/java/com/faforever/client/chat/OfficialLanguageChannel.java
```diff
@@ -0,0 +1,35 @@
...
+
+
+  OfficialLanguageChannel(String localeLanguage, String channelName) {
+    this.localeLanguage = localeLanguage;
+    this.channelName = channelName;
+  }
+
+  public static Optional<OfficialLanguageChannel> getChannelName(String localeLanguage) {
+    for (OfficialLanguageChannel channel : values()) {
+      if (channel.getLocaleLanguage().equals(localeLanguage)) {
```
Wesmania: Python client also adds players with 'by' locale to #russian, 'uk'  is a candidate too.

Katharsas:
Seems to work good. What about #newbie auto-joining? The following seems to work fine:
```
PlayerService playerService;
[...]
if (playerService.getCurrentPlayer().get().getNumberOfGames() < 50) {
    joinChannel("#newbie");
}
```
Edit: I pushed this change.

1-alex98:
travis failed

Katharsas:
The more i think about it the less i like how many things the `PircBotXChatService` is doing. Maybe i can find and create a separation of concerns here.

1-alex98:
Please do it separately tho.(not in this comit)

1-alex98 (dismissed):

1-alex98 (dismissed):

Katharsas:
Travis failed with some cert error.

1-alex98 (dismissed):

Katharsas (dismissed):

Katharsas:
I need to push some additional stuff.

src/test/java/com/faforever/client/chat/ChatControllerTest.java
```diff
@@ -153,8 +154,7 @@ public void testOnJoinChannelButtonClicked() throws Exception {
     doAnswer(invocation -> {
       MapChangeListener.Change<? extends String, ? extends Channel> change = mock(MapChangeListener.Change.class);
       when(change.wasAdded()).thenReturn(true);
-      // Error here is caused by a bug in IntelliJ
-      when(change.getValueAdded()).thenReturn(new Channel(invocation.getArgument(0)));
+      doReturn(new Channel(invocation.getArgument(0))).when(change.getValueAdded());
```
Katharsas: A test is failing because of this. @axel1200

1-alex98 (approved):

1-alex98 (dismissed):

Katharsas (dismissed):

src/main/java/com/faforever/client/chat/OfficialLanguageChannel.java
```diff
@@ -0,0 +1,35 @@
...
+public enum OfficialLanguageChannel {
+  FRENCH("fr", "#french"),
+  GERMAN("de", "#german"),
+  RUSSIAN("ru", "#russian");
+
+  private final String localeLanguage;
+  private final String channelName;
+
+
+  OfficialLanguageChannel(String localeLanguage, String channelName) {
```
Brutus5000: Use Lomboks @AllArgsConstructor instead (makes the whole constructor obsolete)

src/main/java/com/faforever/client/chat/OfficialLanguageChannel.java
```diff
@@ -0,0 +1,35 @@
...
+  public static Optional<OfficialLanguageChannel> getChannelName(String localeLanguage) {
+    for (OfficialLanguageChannel channel : values()) {
+      if (channel.getLocaleLanguage().equals(localeLanguage)) {
+        return Optional.of(channel);
+      }
+    }
+    return Optional.empty();
+  }
+
+  public String getChannelName() {
```
Brutus5000: Use Lomboks @Getter on class level - makes the getter obsolete

src/main/java/com/faforever/client/chat/OfficialLanguageChannel.java
```diff
@@ -0,0 +1,35 @@
...
+    return channelName;
+  }
+
+  public String getLocaleLanguage() {
```
Brutus5000: Use Lomboks @Getter on class level - makes the getter obsolete

src/main/java/com/faforever/client/preferences/ChatPrefs.java
```diff
@@ -48,6 +56,10 @@ public ChatPrefs() {
     chatColorMode = new SimpleObjectProperty<>(CUSTOM);
     idleThreshold = new SimpleIntegerProperty(10);
     chatFormat = new SimpleObjectProperty<>(ChatFormat.COMPACT);
+    autoJoinChannels = new SimpleListProperty<>(FXCollections.observableArrayList());
+    String localeLanguage = Locale.getDefault().getLanguage();
+    Optional<OfficialLanguageChannel> languageChannel = OfficialLanguageChannel.getChannelName(localeLanguage);
```
Brutus5000: Skip the variable assignment, if it is unused and call ifPresent directly (but make a linebreak for the ifPresent).

Brutus5000 (changes requested):
Just some minor codestyle stuff.
I did not test the functionality. I hope my previous approvers did.

1-alex98:
@Katharsas wanna do this ?

Katharsas:
Sure i think i can do this.

src/main/java/com/faforever/client/chat/OfficialLanguageChannel.java
Katharsas: @axel1200 Should i change the code so that it can deal with multiple locales per language channel? I don't think that would be a big change.

src/main/java/com/faforever/client/chat/OfficialLanguageChannel.java
Katharsas: Im in favor of removing getters entirely. I don't see why they would be needed. static method `getChannelName` can just return the channelName directly, as its name implies.

Katharsas:
I changed the code to address the functionality brought up by @Wesmania (multiple locales) and the review by Brutus. I also removed all getters from the enum because i didn't see the point of having them.

1-alex98:
Please look at https://app.codacy.com/app/FAForever/downlords-faf-client/pullRequest?prid=1492232

Katharsas:
Does selecting a language in the settings currently set my locale's language to that langauge?
For example if my local is not russian, but i set my client to russian in the settings, will `Locale.getDefault().getLanguage()` then return `ru` or mu system locale? I guess not.

Because if we use the locale from the settings instead of the default locale, i can revert the whole "multiple languages per langauge channel" thing from my last commit, because it doesn't make much sense. If people select a language other than english in the client settings then we can assume that they know that language and just put them in the appropriate channel.

This way we avoid the whole "mapping multiple languages to a single languages" thing.

Also, this whole `OfficialLanguageChannel` enum is just a disguised, hardcoded immutable map/multimap that maps channels to languages (or the other way around), and it doesn't need to an enum at all (none of its values are ever specifically in code).

Wesmania:
I imagine some people might want to keep an english interface (e.g. me). If we're tying joining language channels to UI language, we should have an option to specify language channels to join.

Katharsas:
Ok but if somebody specifically wants/selects an english interface and specifically NOT the russian one, would they actually usually want to be autojoined into the russian channel?

Btw. I don't think that the Belarusian code `by` ever worked, because the Belarusian language code is actually `be` (@Wesmania you sure that `by` was used in python?), `BY` is the country code. And do operating system even have Belarusian as a language? Is `be` actually ever returned by the operating system? If somebody sets his operating system to russian, they will be auto-joined anyway, so in that case checking for `be` would also be useless.

The whole purpose of language codes is to solve the problem of getting the language of a user correctly. For example, if an Austrian selects "Austria" as his OS language, then the operating system will return `de` (german) as the language code.
This tells me that Belarusian is actually a different language than russian, because if it was very similar, then Belarus would be `ru-BY` and not `be-BY`. So i don'''t think we should map languages to languages, because language codes should already do that by themselves.
And people can always add language channel to their settings manually to join them independently of their locale.

Katharsas:
I think the optimal way to deal with this would be to have an additional checkbox inside every language option in the language settings to join that language's IRC channel. By default, only the setting is ticked for the selected language. The most important thing is to SHOW that language channels exist. Even if we just show the channel name in the language settings for each language option that would be good (without checkbox).
We then ignore the OS locale completely. Kinda like this:
language_settings

Of course, this will mean that we can only have official language channels for languages for which we have translations, so that may be a problem. So maybe a separate UI in the chat preferecnes?

Wesmania:
Python client has something like that - it has an options menu for all language channels and defaults are set using system locale if not set yet.

1-alex98 (dismissed):

1-alex98:
@Katharsas u still wanna change it or get it done? Well tell me I would merge it now if u want to

Katharsas:
- I want to get rid of "multiple languages per channel" change that i made before.
- I also want to get rid of the over-testing (currently the test iterates through ALL hardcoded language channels and asserts that all of them work, which is over-testing imo), but i cannot get rid of this because i cannot mock an enum to have different values, which is an argument for changing `OfficialLanguageChannel` to be a Hashmap<String, String> instead of an enum. Should i do this?

I don't think that i want to add language channels to the chat preferences. Instead, language channels can be advertised in a potential "join new channel" UI, the same one that comes up when you close all channels. But that would belong in a different new issue.

1-alex98:
Do as u please

Katharsas:
Ok so i didn't remove the multi mapping thing but since its now mapped the other way around, there is no distinction (in code) between allowing multiple locales to map to a channel or not. So i guess its fine.

src/main/java/com/faforever/client/preferences/ChatPrefs.java
```diff
@@ -21,11 +23,21 @@
...
+import java.util.Optional;
 
 import static com.faforever.client.chat.ChatColorMode.CUSTOM;
+import static com.faforever.client.preferences.LanguageChannel.*;
 
 public class ChatPrefs {
 
+  @VisibleForTesting
+  public static final ImmutableMap<String, LanguageChannel> LOCALE_LANGUAGES_TO_CHANNELS = ImmutableMap.<String, LanguageChannel>builder()
+      .put("fr", FRENCH)
```
1-alex98: Are there not constants for that Locale.FRENCH.getLanguage() or something...

src/main/java/com/faforever/client/preferences/LanguageChannel.java
```diff
@@ -0,0 +1,14 @@
+package com.faforever.client.preferences;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@AllArgsConstructor
+@Getter
```
1-alex98, line 7: What about @Value I thinks it makes everything final and adds getter as well as an allargs constucor

1-alex98 (changes requested):

src/main/java/com/faforever/client/preferences/ChatPrefs.java
Katharsas: I don't think there are.
Edit: Maybe there are.

src/main/java/com/faforever/client/preferences/LanguageChannel.java
Katharsas: @Value is not supported for enums, only classes.

src/main/java/com/faforever/client/preferences/ChatPrefs.java
Katharsas: There are for some types. I change String type to Locale.

src/main/java/com/faforever/client/preferences/ChatPrefs.java
```diff
@@ -2,24 +2,42 @@
...
 import static com.faforever.client.chat.ChatColorMode.CUSTOM;
+import static com.faforever.client.preferences.LanguageChannel.*;
 
 public class ChatPrefs {
 
+  @VisibleForTesting
+  public static final ImmutableMap<Locale, LanguageChannel> LOCALE_LANGUAGES_TO_CHANNELS = ImmutableMap.<Locale, LanguageChannel>builder()
+      .put(Locale.FRENCH, FRENCH)
+      .put(Locale.GERMAN, GERMAN)
+      .put(new Locale("ru"), RUSSIAN)
```
1-alex98, line 37: There is no constant for "ru" that seems strange
and what is "be"?

1-alex98 (approved):

src/main/java/com/faforever/client/preferences/ChatPrefs.java
Katharsas: There are no constants for most countries/languages, only for some very specific ones. `be` is Belarusian.

1-alex98: oh ok, so be it

1-alex98:
it still has a ready label remove it if u dont want it merged

Katharsas:
I think it can be merged now.

1-alex98:
lemme review one last time

Geosearchef:
> lemme review one last time

Review one last time in remembrance of the good old days when this thread still spammed my mail inbox every view minutes?

1-alex98 (approved):

1-alex98:
Well turn of mail notifications then they are shit anyway XD

Issue #856, opened 2018-04-06
Create linux archive to be extracted


micheljung:

germanicianus:
I am able to extract and run the built zip archive under Linux - see my comment in issue #906.

PR #857, opened 2018-04-06
Feature/#333 open fafreplay

micheljung:


src/main/java/com/faforever/client/os/FileOpeningHandler.java
```diff
@@ -0,0 +1,54 @@
+package com.faforever.client.os;
+
+import com.faforever.client.replay.ReplayService;
+import com.install4j.api.launcher.StartupNotification;
```
micheljung, line 4: While this will break the ability to use the application without the install4j runtime, we no longer ship or create archives without install4j. See #855 and #856

1-alex98:
You really wann have the 5$ right ;), what if I right click on the file and say open with when the client is already running?

1-alex98 (approved):

src/main/java/com/faforever/client/os/FileOpeningHandler.java
```diff
@@ -0,0 +1,54 @@
...
+  @PostConstruct
+  public void postConstruct() {
+    log.debug("Registering file opening handler: {}", this.getClass().getName());
+    StartupNotification.registerStartupListener(this::onStartup);
+  }
+
+  private void onStartup(String parameters) {
+    log.debug("Handling startup: {}", parameters);
+    if (parameters.split("\" \"").length > 2) {
+      throw new IllegalArgumentException("Can't handle multiple files: " + parameters);
```
1-alex98, line 37: should you not '{}'?

1-alex98: or rather %s

micheljung: Exceptions don't support string formatting. I had to do:
```throw new IllegalArgumentException(String.format("Can't handle multiple files: %s", parameters));```
but with simple concatenations like this, I don't see the benefit. With more/complex parameters, I would agree.

micheljung:
> You really wann have the 5$ right ;), what if I right click on the file and say open with when the client is already running?

Then it will start the replay

PR #868, opened 2018-04-18
Allow user to make games private

### LINKED ISSUE #850: Add "Friends only" option in "Create Game" dialog
Split from #848.

Thy Python client has the "Friends only" option:
image

Labeling as minor issue because it's the impact on the user, not the complexity. I even would say it's cosmetic because a user could provide a password.

1-alex98:
Also resolves memory leaks caused by listeners related to CreateGameController.
Fixes #850

### DIFF
--- a/src/main/java/com/faforever/client/preferences/PreferencesService.java
+++ b/src/main/java/com/faforever/client/preferences/PreferencesService.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.io.Reader;
 import java.io.Writer;
 import java.lang.invoke.MethodHandles;
+import java.lang.ref.WeakReference;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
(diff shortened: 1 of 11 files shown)


src/main/java/com/faforever/client/preferences/PreferencesService.java
```diff
@@ -202,8 +203,18 @@ public void storeInBackground() {
       @Override
       public void run() {
         store();
-        for (PreferenceUpdateListener updateListener : updateListeners) {
-          updateListener.onPreferencesUpdated(preferences);
+        ArrayList<WeakReference<PreferenceUpdateListener>> toBeRemoved = new ArrayList<>();
+        for (WeakReference<PreferenceUpdateListener> updateListener : updateListeners) {
+          PreferenceUpdateListener preferenceUpdateListener = updateListener.get();
+          if (preferenceUpdateListener == null) {
```
1-alex98, line 222: Could I use isEnqued there?

1-alex98:
Codacy is wrong, it is needed to keep object from garbage Collection

src/main/java/com/faforever/client/preferences/PreferencesService.java
micheljung: What's `isEnqued`?

micheljung (approved):

Issue #871, opened 2018-04-21
coop replays not being detected


yenon:
coop replays will not be found on the server when you enter the map vault.

1-alex98:
Also the error message is missleading, we should simply remove replays of behalv of the user

PR #872, opened 2018-04-21
Asks the user whether to remove replay

1-alex98:
Prepares #871


1-alex98:
Are we gonna remove this downloading of the map? Or not ? I think we should leave it in place it is not that bad right? And well how often will a usere delte all maps? And if he does at least most get redownloaded(hidden feature) XD

src/main/resources/i18n/messages.properties
```diff
@@ -445,7 +445,7 @@ corruptedMods.show=Show
 
 corruptedReplayFiles.notification=There are corrupted replays on your disk. They were moved to a separate directory for your convenience.
 corruptedReplayFiles.show=Show
-mapNotFound=Map ''{0}'' could not be found on the server nor on this computer.
+mapNotFound=Map ''{0}'' for local replay could not be found on the server nor on this computer. Do you want to mark it as currupted and move to a seperate folder?
```
micheljung: Nice, but let's also say which replay

src/main/resources/i18n/messages_de.properties
```diff
@@ -498,7 +498,7 @@ map.maxPlayers=Max. Spieler
 map.width=Breite
 map.height=Höhe
 replayServer.unreachable=Replay-Server ist nicht erreichbar, Spiel kann nicht aufgezeichnet werden
-mapNotFound=Die Karte ''{0}'' konnte weder auf dem Server noch auf diesem Computer gefunden werden.
+mapNotFound=Die Karte ''{0}'' für lokales Replay konnte weder auf dem Server noch auf diesem Computer gefunden werden. Dieses Replay als korrupt markieren und in seperaten Ordner verschieben?
```
micheljung: All translations need to be added to poeditor as well, unfortunately, otherwise they will get overridden

micheljung (dismissed):

src/main/resources/i18n/messages_de.properties
1-alex98: On the website?

1-alex98:
So which way do we want to do it?

src/main/resources/i18n/messages_de.properties
1-alex98: ok then lets have them overwritten

1-alex98: And I reenter them on the website as soon as the PR gets merged no?

1-alex98:
I would say it is fine the way it is. What so you wanna do @micheljung

micheljung:
Rebase and we're good

1-alex98:
done

1-alex98:
You seems to have done something else there lately @micheljung

micheljung:
wouldn't know what

PR #879, opened 2018-04-30
Fix for issue #754 "Create Game: Map preview issues".

### LINKED ISSUE #754: Create Game: Map preview issues
It shows the map size in pixels instead of 
km:
image

It's not restricted in height, overlapping with other elements:
image

germanicianus:
Fixes #754

### DIFF
--- a/src/main/java/com/faforever/client/game/CreateGameController.java
+++ b/src/main/java/com/faforever/client/game/CreateGameController.java
@@ -108,8 +108,7 @@ public class CreateGameController implements Controller<Pane> {
   }
 
   public void initialize() {
-    mapPreviewPane.minHeightProperty().bind(mapPreviewPane.widthProperty());
-    mapPreviewPane.maxHeightProperty().bind(mapPreviewPane.widthProperty());
+    mapPreviewPane.prefHeightProperty().bind(mapPreviewPane.widthProperty());
     mapSearchTextField.textProperty().addListener((observable, oldValue, newValue) -> {
       if (newValue.isEmpty()) {
         filteredMapBeans.setPredicate(mapInfoBean -> true);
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -238,7 +237,7 @@ private void setSelectedMap(MapBean newValue) {
         new BackgroundSize(BackgroundSize.AUTO, BackgroundSize.AUTO, false, false, true, false))));
 
     MapSize mapSize = newValue.getSize();
-    mapSizeLabel.setText(i18n.get("mapPreview.size", mapSize.getWidthInPixels(), mapSize.getHeightInPixels()));
+    mapSizeLabel.setText(i18n.get("mapPreview.size", mapSize.getWidthInKm(), mapSize.getHeightInKm()));
```
Geosearchef, line 240: Do we wanna include the term km in mapPreview.size?

Geosearchef (approved):
Not much to say, looks good, are you sure prefHeight his enough? Did you test on smaller monitors (resized the client)?

src/main/java/com/faforever/client/game/CreateGameController.java
germanicianus: I would say yes. Regarding size in km there is:
* mapPreview.size
* mapVault.upload.sizeFormat
* map.size (unsure, unused?)

In general I suggest to include the unit in the message key if the unit is used in the translation.

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -108,8 +108,7 @@ public CreateGameController(FafService fafService, MapService mapService, ModSer
   }
 
   public void initialize() {
-    mapPreviewPane.minHeightProperty().bind(mapPreviewPane.widthProperty());
-    mapPreviewPane.maxHeightProperty().bind(mapPreviewPane.widthProperty());
+    mapPreviewPane.prefHeightProperty().bind(mapPreviewPane.widthProperty());
```
1-alex98, line 111: Is that what we want,? I guess this was meant to avoid resizing the window  🤔

germanicianus: The problem with the implementation before was, that the width can get bigger than the space left in height which means the map will overgrow because minHeight was bound to width. Binding to prefHeight avoids that.

Geosearchef: I can confirm this problem on my 1366x768 resolution laptop.

1-alex98 (approved):

micheljung:
This got fixed in the JFoenix branch, thanks anyways :-)

PR #882, opened 2018-05-01
Add advanced filtering call to username search textfield listener.

### LINKED ISSUE #881: Chat filter set using hopper symbol ignored when typing into filter text field
image

germanicianus:
Fixes #881

### DIFF
--- a/src/main/java/com/faforever/client/chat/ChannelTabController.java
+++ b/src/main/java/com/faforever/client/chat/ChannelTabController.java
@@ -202,12 +202,15 @@ public class ChannelTabController extends AbstractChatTabController {
    */
   private void filterChatUserControlsBySearchString() {
     synchronized (userToChatUserControls) {
-      for (Map<Pane, ChatUserItemController> chatUserControlMap : userToChatUserControls.values()) {
-        for (Map.Entry<Pane, ChatUserItemController> chatUserControlEntry : chatUserControlMap.entrySet()) {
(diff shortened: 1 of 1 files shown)


1-alex98 (dismissed):

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -210,7 +210,11 @@ private void filterChatUserControlsBySearchString() {
         synchronized (chatUserControlMap) {
           for (Map.Entry<Pane, ChatUserItemController> chatUserControlEntry : chatUserControlMap.entrySet()) {
             ChatUserItemController chatUserItemController = chatUserControlEntry.getValue();
```
Geosearchef: not actually part of your PR, but while you're at it, the for loop and `getValue()` call can probably just be replaced by an interation over the valueSet.

(either repalce the for loop with `chatUserControlMap.values()` or maybe even just use the stream API and do):
```
chatUserControlMap.values().forEach(chatUserItemController -> {
    ...
});
```

Geosearchef (dismissed):

src/main/java/com/faforever/client/chat/ChannelTabController.java
germanicianus: Done :) .

1-alex98 (dismissed):

micheljung:
Please rebase

1-alex98 (approved):

PR #916, opened 2018-05-28
replace string matcher (linux related)

Geosearchef:


1-alex98 (approved):

micheljung:
`org.hamcrest.Matchers.isEmptyString()` is deprecated and is to be replaced by `org.hamcrest.Matchers.emptyString()`.

Whatever the issue is on Linux, this is not the way to fix it.

PR #928, opened 2018-06-03
Shows a popup that redirect to the voting app

### LINKED ISSUE #901: Add vault filter sanity check
The replay vault's default search query is searching for an ID. An user expecting to be able to search for a username get's a 400 Bad Request error due to us probably actually sending the string as ID parameter to the server. Maybe we should think about changing that default or at least add some kind of sanity check so users get told what they're doing wrong and do not get confused by the exception.

### LINKED ISSUE #918: Voting

1-alex98:
Fixes #918

### DIFF
--- a/src/main/resources/application-prod.yml
+++ b/src/main/resources/application-prod.yml
@@ -24,3 +24,6 @@ faf-client:
 
   forged-alliance:
     exe-url: http://content.faforever.com/faf/updaterNew/updates_faf_files/ForgedAlliance.exe
+
+  voting:
+    base-url: http://voting.faforever.com/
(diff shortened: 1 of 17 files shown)


src/main/resources/application-prod.yml
```diff
@@ -24,3 +24,6 @@ faf-client:
 
   forged-alliance:
     exe-url: http://content.faforever.com/faf/updaterNew/updates_faf_files/ForgedAlliance.exe
+
+  voting:
+    base-url: http://rk.sytes.net:3001
```
1-alex98: Update

1-alex98:
Too much changed needs a remake

Issue #943, opened 2018-07-13
Able integrating avatars in game lobby.


GodFuper:
Modification of the client to transfer data about the player's avatar in the game lobby for him display. Game lobby uses cached avatars files from /programdata/FAForever/cache/avatars .
#942

PR #942, opened 2018-07-12
Modification of the client in order to transfer the parameters of the avatar in the game lobby.

GodFuper:
Adaptation to avatar in the lobby. Lobby will use name.png avatars.

PR #944, opened 2018-07-13
Send avatar info in game lobby

GodFuper:
Sending player avatar info (name/tooltip) in game lobby for displaying near nickname.

last #943 #942

### DIFF
--- a/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
+++ b/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
@@ -179,18 +179,18 @@ public class LaunchCommandBuilder {
     String localIp = Inet4Address.getLoopbackAddress().getHostAddress();
     if (localGpgPort != null) {
       command.add("/gpgnet");
-      command.add(localIp + ":" + localGpgPort);
+      command.add(String.format("%s:%d", localIp, localGpgPort));
     }
 
 
     if(avatarUrl != null) {
-	  String urlTOfilename = null; // In lobby it is easier to use the name of the avatar file than the link itself.
-	  urlTOfilename = Paths.get(URI.create(avatarUrl).getPath()).getFileName().toString();
-      command.add("/avatarurl");   // If you change the command, you need to remember to change in lobby.
-      command.add(String.valueOf(urlTOfilename)); 
+	  String urlToFilename = null;
+	  urlToFilename = Paths.get(URI.create(avatarUrl).getPath()).getFileName().toString();
+      command.add("/avatarurl");
+      command.add(urlToFilename);
 
       command.add("/avatartooltip");
-      command.add(String.valueOf(avatarTooltip));
+      command.add(avatarTooltip);
     }
 
     if (mean != null) {
--- a/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
+++ b/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
@@ -7,6 +7,7 @@ import org.springframework.util.StringUtils;
 import java.net.Inet4Address;
 import java.net.URI;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -21,6 +22,8 @@ public class LaunchCommandBuilder {
   private Float deviation;
   private String country;
   private String clan;
+  private String avatarUrl;
+  private String avatarTooltip;
   private String username;
   private Integer uid;
   private Path executable;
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
```diff
@@ -169,6 +182,17 @@ public LaunchCommandBuilder rehost(boolean rehost) {
       command.add(localIp + ":" + localGpgPort);
     }
 
+
+    if(avatarUrl != null) {
+	  String urlTOfilename = null; // In lobby it is easier to use the name of the avatar file than the link itself.
```
Geosearchef: formatting off

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
```diff
@@ -169,6 +182,17 @@ public LaunchCommandBuilder rehost(boolean rehost) {
...
+	  urlTOfilename = Paths.get(URI.create(avatarUrl).getPath()).getFileName().toString();
```
Geosearchef: why do you first create and then assign inside the same block?
(multiple times)

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
```diff
@@ -169,6 +182,17 @@ public LaunchCommandBuilder rehost(boolean rehost) {
...
+      command.add("/avatarurl");   // If you change the command, you need to remember to change in lobby.
+      command.add(String.valueOf(urlTOfilename));
```
Geosearchef: remove `String.valueOf`, urlTOfilename is already a string

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
```diff
@@ -169,6 +182,17 @@ public LaunchCommandBuilder rehost(boolean rehost) {
       command.add(localIp + ":" + localGpgPort);
```
Geosearchef: Not part of your changes, but gonna mention it anyways ;P  
`String.format("%s:%d", localIp, localGpgPort)`

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
Geosearchef: also, java convention would be to name this variable `urlToFilename`

Geosearchef (changes requested):
PR looks good, but:

This lead to a long discussion in #game-developement with the result that player's avatar, rating and games need to be checked by each client, either by reading a file/contacting API/server /gpgnet.  Therefore the avatar will not be send by other players which hand it to their client as a command line parameter.

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
GodFuper: @Geosearchef Like I tried to do so, but the Travis (plugin in github) gave an error.

Geosearchef:
I also want to mention that you don't need to recreate your pull request unless you change the branch. You can just force push to your current branch and the PR will update automatically. So if you wanna e.g. implement requested changes you can just commit them while ticking the `amend` box during the process and the force push. Github will automatically detect when something I requested a change on changed and marked the request as processed.

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
Geosearchef: Pls do it again and let me take a look. I can't believe that, maybe there was something else going wrong. If it really happened due to this we need to fix travis, not the code ;).

GodFuper:
@Geosearchef 
Your long discussions in #game-development are at the stage of conversations. "need to be checked by each client, either by reading a file/contacting API/server /gpgnet." The implementation of this may take months or even more. That before, that now not was protection from this. Currently there is no API/server connection. So this is no reason to cancel this development. This is the minimum that will make avatars more attractive. Although I also believe that it is necessary to communicate with the API and the server.

1-alex98:
There seem to be user that actively exploit such security breaches. Eventhough that is really stupid. So we need to do it securely

1-alex98:
Should we every change to the java server we will be able to validate users avatars

GodFuper:
Axel, well, have you already done the modification for validate players?

1-alex98:
downlord did it, not me I reviewed it... but as i said on java-server

GodFuper:
so now let's insert the avatar in the lobby?

1-alex98:
java-server is not deployed and we would also need to add this check once it is deployed

GodFuper:
Well, if my help will be need, write me.

1-alex98:
You can help where you want, there are lots of issues open across all different repos. But deciding to deploy java-server can not be helped I assume. You could help improving it if you wanted

1-alex98:
We will close this. We can reopen if server support for this is up.

PR #947, opened 2018-07-19
Fixes #946 New Gradle version & Eclipse JavaFx fix

### LINKED ISSUE #946: Reduce project setup friction
### Eclipse & JavaFx
Eclipse tries to detect and block access to classes/methods that it deems "non-public" (part of internal code rather than public API). This detection usually works fine but is a known problem for JavaFX. Other methods/classes which it blocks:
- com.sun.webkit.WebPage # setBackGroundColor(int)
- com.sun.glass.ui.Window
- netscape.javascript.JSObject 

This can be solved by explicitly allowing access with gradle eclipse plugin. Gradle Eclipse plugin does not work with the current Gradle version, so a Gradle upgrade would be needed for this.

### Java 10
The current gradlew.bat wrapper (NOT Gradle itself; the wrapper which downloads/upgrades/executes gradle) crashes if executed while Java 10 is default on system (which is the case if Java 10 was installed on the system). 

Since the Gradle wrapper is responsible for selecting the right JDK when executing Gradle, this can not be fixed by passing the right JDK via args. So, a gradle upgrade would be needed to fix this, too.

### Gradle upgrade
Currently investigating.

Pull request will land when done. Can I assign myself to this issue?

Katharsas:
Fixes #946 .
I separated the eclipse stuff into its own build file. Of course it could also stay in the normal build file, or in some folder.
This PR causes Hamcrest version to change from 1.4 to 1.3


Katharsas:
Travis failed with
> The command "curl https://download-keycdn.ej-technologies.com/install4j/install4j_unix_7_0_7.tar.gz | tar xzf -C /tmp" failed and exited with 2 during .

Brutus5000:
There is also a reason in the log:

> 2.16s$ curl https://download-keycdn.ej-technologies.com/install4j/install4j_unix_7_0_7.tar.gz | tar xzf -C /tmp
> tar (child): -C: Cannot open: No such file or directory

Since you are piping in the file, remove the `f` parameter from the `tar` command.

Katharsas:
This PR is not in any way or form related to Travis or install4j or curl.

micheljung (dismissed):

micheljung:
Can you rebase and check if it still works as intended? The project is now OpenJDK 10

Katharsas:
Working on it, sec...
(Btw: I think separating JavaFX into dependencies causes Eclipse to not complain about JavaFX anymore, so thats nice.)

Katharsas:
Ok i have removed the rules for JavaFX (since not necessary anymore) and rebased. Eclipse does not find errors in non-Tests. Gradle compile works fine of course.

Eclipse only shows two compile errors in ChatControllerTest:
- https://github.com/FAForever/downlords-faf-client/blob/develop/src/test/java/com/faforever/client/chat/ChatControllerTest.java#L157
`Description	Resource	Path	Location	Type
The method thenReturn(capture#10-of ? extends Channel) in the type OngoingStubbing<capture#10-of ? extends Channel> is not applicable for the arguments (Channel)	ChatControllerTest.java	/downlords-faf-client/src/test/java/com/faforever/client/chat	line 157	Java Problem`

- https://github.com/FAForever/downlords-faf-client/blob/develop/src/test/java/com/faforever/client/chat/ChatControllerTest.java#L171
`Description	Resource	Path	Location	Type
The method thenReturn(capture#22-of ? extends ChatChannelUser) in the type OngoingStubbing<capture#22-of ? extends ChatChannelUser> is not applicable for the arguments (ChatChannelUser)	ChatControllerTest.java	/downlords-faf-client/src/test/java/com/faforever/client/chat	line 171	Java Problem`

But thats not an issue i guess since its a single test and not related to this PR anyway.

Katharsas:
Actually it seems like this PR can be trashed because the new Eclipse version does not show errors whatsoever (with or wothout this PR).

PR #952, opened 2018-08-03
Fixes Featured Mod not shown

### LINKED ISSUE #732: Game type often not given in local replays screen

1-alex98:
The load local replay called is made async already so the get() is fine otherwise the get() will be called one step ahead. getNow is problematic if the Future is not ready it simply uses unknow if future is not finishes which is the case in most cases.
Fixes #732


1-alex98:
@micheljung again something strange in travis

micheljung (approved):

PR #954, opened 2018-08-04
Rework/Fix game tiles

### LINKED ISSUE #945: "Play" Tab -> Table/Tile issues
Actual Issue:
In Tile view, checking/unchecking "Show password protected" / "show game with mods" does not work. I have to switch to Table view and then back to Tile view to apply the checkbox changes.

Annoying but not an issue:
In Table view, when there are many mods in a lobby, i cannot see all mods - it only shows e.g. "mod1, mod2 and 3 more...". i have to switch to Tile view and then hover over the lobby - very unintuitive.

1-alex98:
Fixes #945

### DIFF
--- a/src/main/java/com/faforever/client/game/GameService.java
+++ b/src/main/java/com/faforever/client/game/GameService.java
@@ -52,9 +52,9 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.InitializingBean;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Service;
-import org.springframework.beans.factory.InitializingBean;
 
 import javax.inject.Inject;
 import java.io.IOException;
@@ -269,7 +269,6 @@ public class GameService implements InitializingBean {
   /**
    * @param path a replay file that is readable by the preferences without any further conversion
    */
-
   public void runWithReplay(Path path, @Nullable Integer replayId, String featuredMod, Integer version, Map<String, Integer> modVersions, Set<String> simMods, String mapName) {
     if (isRunning()) {
       logger.warn("Forged Alliance is already running, not starting replay");
@@ -334,7 +333,6 @@ public class GameService implements InitializingBean {
     return games;
   }
 
-
   public Game getByUid(int uid) {
     Game game = uidToGameInfoBean.get(uid);
     if (game == null) {
@@ -394,7 +392,6 @@ public class GameService implements InitializingBean {
    * Returns the preferences the player is currently in. Returns {@code null} if not in a preferences.
    */
   @Nullable
-
   public Game getCurrentGame() {
     synchronized (currentGame) {
       return currentGame.get();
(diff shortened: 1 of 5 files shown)


src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -625,4 +612,11 @@ private void removeGame(GameInfoMessage gameInfoMessage) {
       uidToGameInfoBean.remove(gameInfoMessage.getUid());
     }
   }
+
+  /**
+   * Can be used as a lock/monitor , so that the game service pauses the adding of new games
+   */
+  public ObservableMap<Integer, Game> getUidToInfoGameBean() {
+    return uidToGameInfoBean;
+  }
```
micheljung: Actually, any access on `uidToGameInfoBean` *must* be synchronized. Otherwise, you run into race conditions. Therefore, exposing this is dangerous and should be avoided.

Also, there are no spaces in front of commas but periods at the end of JavaDoc ;-)

If you need to expose it, you could return an immutable copy. If it's needed to e.g. add listeners to it, a method can be provided that adds listeners.

src/main/java/com/faforever/client/game/GamesTableController.java
```diff
@@ -71,6 +75,9 @@ public GamesTableController(MapService mapService, JoinGameHelper joinGameHelper
     this.preferencesService = preferencesService;
 
     this.selectedGame = new SimpleObjectProperty<>();
+
+    modChangedListener = (observable, oldValue, newValue) -> modsColumn.setVisible(newValue);
+    passwordChangeListener = (observable, oldValue, newValue) -> passwordProtectionColumn.setVisible(newValue);
```
micheljung: `modChangedListener` sounds like a mod has been changed. Better: `showModdedGamesChangedListener`.
And `showPasswordProtectedGamesChangedListener`

micheljung (changes requested):

src/main/java/com/faforever/client/game/GameService.java
1-alex98: I exposed it to use it as a monitor

1-alex98:
@Geosearchef would you review

src/main/java/com/faforever/client/game/GameService.java
micheljung: Please check if there's a better way, this violates the Law of Demeter

src/main/java/com/faforever/client/game/GamesTilesContainerController.java
```diff
@@ -60,15 +64,20 @@ public GamesTilesContainerController(UiService uiService, PreferencesService pre
...
-    gameListChangeListener = change -> Platform.runLater(() -> {
-      synchronized (change) {
+    gameListChangeListener = change -> {
+      Platform.runLater(() -> {
         while (change.next()) {
-          change.getRemoved().forEach(gameInfoBean -> tiledFlowPane.getChildren().remove(uidToGameCard.remove(gameInfoBean.getId())));
+          change.getRemoved().forEach(gameInfoBean -> {
+            boolean remove = tiledFlowPane.getChildren().remove(uidToGameCard.remove(gameInfoBean.getId()));
+            if (!remove) {
+              log.error("Tried to remove game tile that did not exist.");
```
micheljung, line 73: Warn, not error

micheljung (changes requested):

PR #991, opened 2018-10-14
Add MapBean Equal Method

### LINKED ISSUE #990: Resetting game location make all maps display twice
Tell the client again where your game is and you will be shown all maps twice

1-alex98:
Fixes #990

### DIFF
--- a/src/main/java/com/faforever/client/map/MapService.java
+++ b/src/main/java/com/faforever/client/map/MapService.java
@@ -209,7 +209,7 @@ public class MapService {
     try {
       MapBean mapBean = readMap(path);
       pathToMap.put(path, mapBean);
-      if (!installedSkirmishMaps.contains(mapBean) && mapBean.getType() == Type.SKIRMISH) {
+      if (!mapsByFolderName.containsKey(mapBean.getFolderName()) && mapBean.getType() == Type.SKIRMISH) {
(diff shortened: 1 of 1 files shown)


Brutus5000:
We should really use shared code in mordor and dl client... there is an AbstractEntityFX who solves all comparison issues.

1-alex98:
I know but better fix it then leave it. Who knows when we are gonna find time for it

1-alex98:
And chaning the data models is gonna be lots of work still.

micheljung:
While the proposed fix will work, I shy away from relying on equals/hashCode as it generally will cause trouble sooner or later.
Implementing equals/hashCode requires that the chosen identifier works for all use cases. When a new use cases requires it to be touched, it might or might not break other use cases.

Therefore, wherever code relies on it being implemented (like the use of `MapBean` in a `Set`, or the object being compared) I recommend to refactor it so that it explicitly uses the properties of interest.

For instance. Instead of
```java
Set<MapBean> maps;
```
use
```java
Map<String, MapBean> mapsByFolderName;
```
The above allows each use case to separately specify what the identity of an object is.

Could you do that, or at least point out where the problematic code is?

1-alex98:
Like that?

micheljung (dismissed):

PR #1001, opened 2018-10-20
Tutorials

1-alex98:
Before merge tutorials need to be made a featured mod... @Brutus5000 or however does it
 but I can help you doing it if you have any question as I successfully did it on test server


micheljung:
I think your commit message an PR title are wrong

1-alex98:
@micheljung it was correct pretty sure either me or somebody else force pushed this commit message on top of it

src/main/java/com/faforever/client/patch/SimpleHttpFeaturedModUpdaterTask.java
```diff
@@ -66,9 +68,17 @@ protected PatchResult call() throws Exception {
...
+
+        if (featuredModFile.getUrl().contains("tutorial")) {
+          featuredModFile.setUrl(featuredModFile.getUrl().replace("content.faforever.com", "content.test.faforever.com"));
+        }
+
+        String url = featuredModFile.getUrl();
+        downloadService.downloadFile(new URL(url), targetPath, this::updateProgress);
       }
 
+      UpdaterUtil.extractMoviesIfPresent(targetPath, fafDataDirectory);
```
1-alex98, line 76: Semms like we need to unpack movies to different folder according to speed

src/main/java/com/faforever/client/vault/review/Review.java
```diff
@@ -26,7 +26,6 @@ public static Review fromDto(com.faforever.client.api.dto.Review dto) {
     review.setId(dto.getId());
     review.setText(dto.getText());
     review.setScore(Optional.ofNullable(dto.getScore()).map(Byte::intValue).orElse(0));
-    review.setPlayer(Player.fromDto(dto.getPlayer()));
```
1-alex98, line 29: bug this is doubled here

src/main/java/com/faforever/client/api/dto/AbstractEntity.java
```diff
@@ -0,0 +1,18 @@
...
+
+@Getter
+@Setter
+@EqualsAndHashCode(of = "id")
+public abstract class AbstractEntity {
+  @Id
+  protected String id;
+  protected OffsetDateTime createTime;
+  protected OffsetDateTime updateTime;
+}
```
Katharsas, line 18: There seem to be a bunch of other classes that maybe should make use of this to stay consistent with API code? Or do you want to change those when they get touched anyway? Maybe we should have an issue were all classes that should use this eventually are listed.

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
```diff
@@ -36,6 +36,18 @@ public ForgedAllianceService(PreferencesService preferencesService) {
     this.preferencesService = preferencesService;
   }
 
+  public Process startGameOffline(List<String> args) throws IOException {
+    Path executable = getExecutable();
+    List<String> launchCommand = LaunchCommandBuilder.create()
+        .executableDecorator(preferencesService.getPreferences().getForgedAlliance().getExecutableDecorator())
```
Katharsas, line 38: Why not use `#defaultLaunchCommand()` ?

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
```diff
@@ -36,6 +36,18 @@ public ForgedAllianceService(PreferencesService preferencesService) {
...
+        .executable(executable)
+        .additionalArgs(args)
+        .logFile(preferencesService.getFafLogDirectory().resolve("game.log"))
+
+        .build();
```
Katharsas, line 43: So its not possible to have replay for playing tutorials for now? Why not?

Katharsas (commented):
Seems like this code wants to keep tutorials as "offline" for some reason. If this feature (?) would be dropped, the game launch code could be simpler.
- You will need to be online to login anyway, right?
- You will need to be online to download the game patch and tutorial themselves anyway?
- You might like to watch replay of people completing the tutorial so it can be improved and debugged easier?
- The client doesnt even have a GUI to run offline

If we want to make it possible to run the client offline, it should imo be a different PR that makes it possible to run any kind of game (including custom games and coop) offline, not just tutorials, so the game launch code feels weird to me since it makes ONLY tutorials run ONLY offline.

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
1-alex98: I think there are some options i dont wanna have here, like replay server etc.

src/main/java/com/faforever/client/api/dto/AbstractEntity.java
1-alex98: Well in the long run we wanna change to java-commons dtos but that is a BIIIGGGG task

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
Katharsas: Have you looked at the function `defaultLaunchCommand()`? I dont see what you mean.

src/main/java/com/faforever/client/api/dto/AbstractEntity.java
1-alex98: I just copied the calsses from there. I would like to see more usage of AbstractEntity  in the future yes... But chaning beyond this seems out of scope of the PR

Katharsas: Ok

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
1-alex98: will look into it

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
1-alex98: i asked @Speed2 on Zulip

1-alex98: it produces a local replay anyway even if u dont not register a replay server 
image

1-alex98: in the documents folder

1-alex98: it is not that easy to start because it misses out on some info for example what featured mod is needed etc. but would be sufficent for debugging.

1-alex98: u would need to start game by hand to watch :D

1-alex98: So what know? I would say that is sufficent...

PR #1004, opened 2018-10-22
#992 Don't list languages twice when custom ones are installed

### LINKED ISSUE #992: Languages are duplicated
v0.9.0-beta.RC8:
image

micheljung:
Fixes #992

### DIFF
--- a/src/test/resources/i18n/messages_de.properties
+++ b/src/test/resources/i18n/messages_de.properties
@@ -1,0 +1,3 @@
+key=Schlüssel
+tree=Baum
+trees=Bäume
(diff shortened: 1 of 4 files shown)


src/test/resources/i18n/messages_de.properties
```diff
@@ -0,0 +1,3 @@
+key=Schlüssel
```
1-alex98, line 1: I thought we dont wanna define german keys any longer but rather enter them on translation website

1-alex98 (dismissed):

src/test/resources/i18n/messages_de.properties
micheljung: this is for unit testing

PR #1030, opened 2018-11-12
Remove Replay Buffer

### LINKED ISSUE #973: Replay stream is being buffered
Replay server (https://github.com/FAForever/faf-aio-replayserver) needs to have continuous stream from game, as the game sends, otherwise replay server won't be able to stream live replay fluently.
Now, buffer size is 8192. It would be nice, if network stream behavior will be fluent/continuous.
Every 1/10 of second game sends at least 7 bytes of data, it would be nice to have same behavior on java client side, like a simple proxy.

`src/main/java/com/faforever/client/replay/ReplayServerImpl.java:105` has `BufferedOutputStream`. Default buffer size is 8k for jre 1.8 on windows. Also `BufferedOutputStream` has second parameter, buffer size, it might help.

1-alex98:
Fixes #973

### DIFF
--- a/src/main/java/com/faforever/client/replay/ReplayServerImpl.java
+++ b/src/main/java/com/faforever/client/replay/ReplayServerImpl.java
@@ -18,8 +18,8 @@ import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Component;
 
 import javax.inject.Inject;
-import java.io.BufferedOutputStream;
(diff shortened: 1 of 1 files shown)


micheljung:
Thanks and sorry, but I think we should go with #1064 instead :)

1-alex98:
ok no problem it is almost equivalent anyway right

PR #1065, opened 2018-12-04
Wrap game filters when window is too small

### LINKED ISSUE #1034: Unnecessary dots
Are these dots intended in the switch between tiles and list games?

downlords-faf-client_2018-11-13_18-18-21

micheljung:
And don't shrink the table/tiles button to "...". This could be improved
so that either the text is displayed as whole, if there is enough space,
or not at all. A tooltip should be added in both cases.

Fixes #1034


micheljung:
image

1-alex98 (approved):

PR #1067, opened 2018-12-08
Fix executable decorator not being always applied

### LINKED ISSUE #1066: Launch command decorator is not always applied
When you try to launch a local replay on Linux for example, the client will bug out about 'no permission to run file' since the 'wine foo %s' decorator is not being applied. The problem stems from ForgedAllianceService not passing executableDecorator to LaunchCommandBuilder in some cases. The builder is intended to always require it, but the decorator is set to "%s"' rather than null in the constructor, so the check does nothing.

Wesmania:
Fixes #1066.

### DIFF
--- a/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
+++ b/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
@@ -35,10 +35,6 @@ public class LaunchCommandBuilder {
   private boolean rehost;
   private Integer localReplayPort;
 
-  private LaunchCommandBuilder() {
-    executableDecorator = "\"%s\"";
-  }
-
   public static LaunchCommandBuilder create() {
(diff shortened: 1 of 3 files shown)


micheljung (approved):

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
```diff
@@ -36,7 +36,7 @@
   private Integer localReplayPort;
 
   private LaunchCommandBuilder() {
-    executableDecorator = "\"%s\"";
+    executableDecorator = null;
```
micheljung: `null` is default, so this is redundant. Whole constructor can be removed

micheljung (dismissed):

1-alex98 (dismissed):

PR #1071, opened 2018-12-15
Improve updater

micheljung:
The new mechanism reads a file hosted on faforever.com which allows us
to add more options, like `mandatory`, which can be set to force client
updates in future (e. g. when ICE out), and platform-specific URLs in
order to avoid guesses based on the file name.


src/main/java/com/faforever/client/update/CheckForUpdateTask.java
```diff
@@ -2,101 +2,108 @@
...
-  @Inject
-  I18n i18n;
+  private final I18n i18n;
+  private final PreferencesService preferencesService;
 
   private ComparableVersion currentVersion;
 
-  public CheckForUpdateTask() {
+  @VisibleForTesting
+  FileSizeReader fileSizeReader = url -> url
```
Brutus5000, line 34: Shouldn't this be injected as a bean? Then @VisibleForTesting would be obsolete as well.

Brutus5000 (commented):
Looks good to me except for 1 tiny thing.

src/main/java/com/faforever/client/update/CheckForUpdateTask.java
micheljung: hmm well, that would work but it's not a component that should generally be available as a singleton but a very local solution with the sole purpose that I can mock the mechanism in the unit test.

However, there is in fact other code that could use it as well (`com.faforever.client.map.MapService#getFileSize`) which supports your point. I'll add a TODO and refactor this at another time

PR #1093, opened 2019-01-08
Add more validity states required for some coop leaderboard checks

Askaholic:
Related to these proposed db additions:
https://github.com/FAForever/db/pull/187


1-alex98 (dismissed):

1-alex98 (approved):

1-alex98:
SOmehow travis fails maybe you can have a look why...

Askaholic:
It's the same reason why the latest commit on develop (df3473c821ca021f75324d9d4b237fdeb7f2fc42) is failing. There's an unused mock in `LoginControllerTest` which causes the following error:

```
com.faforever.client.login.LoginControllerTest > unnecessary Mockito stubbings FAILED
    org.mockito.exceptions.misusing.UnnecessaryStubbingException
```

My guess is that when @micheljung changed the `LoginController` it did some check before calling other services, and therefore the mocked service is no longer called, which causes the stubbing exception.

PR #1115, opened 2019-02-21
Add Test for ReflectionUtil

Serverfrog:
I try to understand the code, and write some tests.
Found out that the ReflectionUtil, if used incorrectly, instead of throwing a NoSuchFieldException, it will create a Endless Loop. Should that also be fixed right away? Seems like it shouldn't do this but it seems it also be now problem till now

### DIFF
--- a/src/test/java/com/faforever/client/util/ReflectionUtilTest.java
+++ b/src/test/java/com/faforever/client/util/ReflectionUtilTest.java
@@ -1,0 +1,57 @@
+package com.faforever.client.util;
+
+import org.hamcrest.Matchers;
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import static org.hamcrest.Matchers.isA;
+import static org.junit.Assert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.fail;
+
+public class ReflectionUtilTest {
+
+
+  @Rule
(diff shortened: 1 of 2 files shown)


src/test/java/com/faforever/client/util/ReflectionUtilTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Test
+  @Ignore // this will end in a endless loop.
+  public void testGetDeclaredFieldPrivate() throws NoSuchFieldException {
+    /*
+        The Method will first look if the field is inside the given class.
+        If not it will look inside the Superclass.
+        Then, if it's found nothing, it will sets its loop variable (currently the Superclass of the given one),
+        again to the superclass of the given class. This results into a endless loop because every class has Object as
+        superclass (except Object of course) so it will always find Object.class as Superclass and checks like
```
1-alex98: Maybe we should fix this then 🤔

Serverfrog: I think too that it should be. The implemented Code looks like that should not be happening.
But i don't know if the implementation should also get the Class of a Private Field (currently not)

Serverfrog:
Is there a Reason Coveralls don't se that some line are covered? Did i need to setup the Test special for it? Jacocoo On IDEA show the coverage

1-alex98 (approved):

1-alex98:
Can u squash commits?

PR #1119, opened 2019-02-24
u00200 instead of Whitespace in Leaderboard Co-Op

### LINKED ISSUE #1116: u00200 instead of Whitespace in Leaderboard Co-Op
In the Leaderboard of the Co-Op, instead of a Space,, a u00200 is printed
u00200

Serverfrog:
Fixes Text Seperator in languages other than english.
It was missing the Backslash before the unicode definition or was without it (just a space instad of \u0020)

Fixes #1116

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-./gradlew --stacktrace -PjavafxPlatform=linux check
+./gradlew --stacktrace --debug -PjavafxPlatform=linux check
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-./gradlew --stacktrace --debug -PjavafxPlatform=linux check
+./gradlew --stacktrace -PjavafxPlatform=linux check
--- a/src/main/resources/i18n/messages_cs.properties
+++ b/src/main/resources/i18n/messages_cs.properties
@@ -396,7 +396,7 @@ duration.hourMinutes = {0,number,\#}h {1,number,\#}min
 # suppress inspection "TrailingSpacesInProperty"
 vsSeparator = \ vs 
 # suppress inspection "TrailingSpacesInProperty"
-textSeparator = , 
+textSeparator=,\u0020
 secondsAgo = před {0,number,\#} sekundami
 minuteAgo = před {0,number,\#} minutou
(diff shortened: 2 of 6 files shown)


1-alex98:
Travis fails

Serverfrog:
> 
> 
> Travis fails

I will look at it. Seems wierd because that means there is some kind of hardcoded string check for the sperator character from the i18n files.

Test that failes: com.faforever.client.chat.PrivateChatTabControllerTest > onPlayerDisconnected
Locally it runs fine which makes it more wierd (also no string check). Seems like some raise condition that can be triggerd in the test. Can a Travis Build be re-runned?

Stacktrace:
`    com.google.gson.JsonSyntaxException at PrivateChatTabControllerTest.java:89
        Caused by: com.google.gson.stream.MalformedJsonException at PrivateChatTabControllerTest.java:89`

Serverfrog:
--debug Crashes the Build because the log will be to big. To Many Exception that are not saying anything about the Status of the test (Exception don't mean a Failed test)
But the Test runs there fine:

> 23:39:48.020 [DEBUG] [TestEventLogger] com.faforever.client.chat.PrivateChatTabControllerTest > onPlayerDisconnected PASSED

1-alex98:
Hmm set it back if it runs, it is ok, seems to be some kind of race condition

Brutus5000:
Correct me if I'm wrong but I think this is fixed in the translations already? Also we do not change translations here, but via the tool (I don't have a link at hand)

Serverfrog:
> Correct me if I'm wrong but I think this is fixed in the translations already? Also we do not change translations here, but via the tool (I don't have a link at hand)

If its done with a tool, then it was not used correctly or it produces this error. The l18n language files had different variants of space for the key _textSeparator_ which causes that in the english version the u00200 is not existing (because in the file the \ was befor u00200), other languages have the \ missing, other have a simple space put in there.  If there is a tool were the translations are managed and then imported into here then a link to that could help ;)

1-alex98:
Oh damn it i completely forgot about it yes it needs to be done in the translation tool

1-alex98:
If you go into the client under settings (language tab) you have a link to the translation website there you can edit the faulty translation. Tho first of all you need to apply as a translator but it takes only 1 click and downlord approves in seconds normally

1-alex98:
Are u gonna do it @Serverfrog because if not i migtht do it

Issue #1134, opened 2019-03-23
Coop mission name resolving seems to not work correctly


1-alex98:
```
Exception in thread "JavaFX Application Thread" java.lang.IllegalStateException: Map folder name did not match with any coop mission
        at com.faforever.client.coop.CoopController.lambda$initialize$20(CoopController.java:172)
at java.base/java.util.Optional.orElseThrow(Unknown Source)
```

1-alex98:
Lets first display unknow map instead of throwing an exception

Askaholic:
Is there still something to do here?

1-alex98:
The Exception is still in the logs for some missions so yes. Something is wrong there.

Sheikah45:
How do you reproduce this?

1-alex98:
You have luck that somebody hosts missions that have this issue...

1-alex98:
or u try do that yourself

Marc-Spector:
The issue is not relevant. All mission names are displayed correctly.

PR #1148, opened 2019-04-19
Dont throw exception but display unknown mission

1-alex98:
Prepares #1134


src/main/resources/i18n/messages.properties
```diff
@@ -50,6 +50,7 @@ play.coop=Co-Op
 
 leaderboard.failedToLoad=Leaderboard could not be loaded
 leaderboard.loading=Loading leaderboard…
+coop.unknownMission=unknown mission
```
Brutus5000: I'd start with an uppercase letter like in 99% of all other messages.

src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -160,17 +163,22 @@ public void initialize() {
...
       Platform.runLater(() -> {
         missionComboBox.setItems(observableList(coopMaps));
         GamesTableController gamesTableController = uiService.loadFxml("theme/play/games_table.fxml");
-        Function<String, String> coopMissionNameProvider = mapFolderName -> coopMaps.stream()
-            .filter(coopMission -> coopMission.getMapFolderName().equalsIgnoreCase(mapFolderName))
-            .findFirst()
-            .orElseThrow(() -> new IllegalStateException("Map folder name did not match with any coop mission"))
-            .getName();
+
+        Function<String, String> coopMissionNameProvider = (mapFolderName -> {
```
Brutus5000: It's too big now. We should put it into a new function imho.

Brutus5000 (changes requested):

1-alex98:
Please somebody review

1-alex98:
@Geosearchef would u mind reviewing?

micheljung (approved):

PR #1163, opened 2019-04-23
set faf-uid file permission after download (might fix #1162)

### LINKED ISSUE #1162: don't set file permissions of faf-uid at runtime on posix systems
You can't set the uid's permissions at runtime here: https://github.com/FAForever/downlords-faf-client/blob/develop/src/main/java/com/faforever/client/legacy/PosixUidService.java#L25

This makes downlord's client unusable on Posix currently.

This would only work, if you run downlord's client from your home directory or sth. 
Whatever is packaging the release build https://github.com/FAForever/downlords-faf-client/releases/download/v0.10.3-beta/_dfc_unix_0_10_3-beta.tar.gz needs to set the file permissions inside the archive correctly.
We're currently using a dirty hack to set the +x permission during package creation here: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=downlords-faf-client#n48

See  #796 and #1157

muellni:
try to fix  #1162 
I dont have a proper environment to test this, but the downloaded file `native/faf-uid` seems to get proper permissions on my machine during build.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -221,6 +221,11 @@ task downloadUnixUid(type: Download) {
   src "https://github.com/FAForever/uid/releases/download/v${faf_uid_version}/faf-uid"
   dest file("${buildDir}/resources/native/faf-uid")
   onlyIfNewer true
+  doLast {
+    project.exec {
+      commandLine('chmod',  '+x', "${buildDir}/resources/native/faf-uid")
+    }
+  }
 }
 
 task downloadWindowsUid(type: Download) {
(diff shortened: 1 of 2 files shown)


muellni:
I realized I should have targeted the `java-ice-adapter` branch. And the latest release https://github.com/FAForever/downlords-faf-client/releases/download/v0.10.4-beta/dfc__unix_0_10_4-beta.tar.gz has proper permissions set for `native/faf-uid`. I will probably create another PR for this

Issue #1078, opened 2018-12-27
Don't cap local replay list after 300 files


Brutus5000:
Right now there is a hard limit of 300 local replays to show.

https://github.com/FAForever/downlords-faf-client/blob/d5cf40884953a578858c01c3ff0ca9843a82189e/src/main/java/com/faforever/client/replay/ReplayService.java#L170

There needs to be a better we of preventing OutOfMemoryExceptions with some sort of paginated loading.

ch3oh:
If the limit must remain for the near future then at least display the most recent 300 replays.  

For info I have a replay folder containing 4,688 replays from 18/05/2013 through to 24/01/2019 so any upper limit needs to be somewhere north of this.

TheKoopa321:
I agree it would be a great improvement. At least to show the latest instead of the oldest.

You can modify the .class file in downlords-faf-client-0.9.3-beta.jar and replace 300 with 1000 for exemple. I did it and it works fine.

1-alex98:
The thing is we need to do some kind of paging then if we load 4688 replays the client will run out of Memory

EcoNooob:
I'd say the biggest issue is that it shows the first 300 replays instead of the latest 300 replays. The local replay vault is now useless unless i want to watch replays from 2013.

PR #1174, opened 2019-04-28
Load last 300 replays

1-alex98:
Prepares #1078

### DIFF
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -51,6 +51,7 @@ import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


micheljung (approved):

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -167,6 +168,7 @@ static String guessModByFileName(String fileName) {
 
     try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(replaysDirectory, replayFileGlob)) {
       StreamSupport.stream(directoryStream.spliterator(), false)
+          .sorted(Comparator.comparing(path -> ((Path) path).toFile().lastModified()).reversed())
```
micheljung: Whenever you need a method that only exists in `File` (which is old and bad API), an equivalent method exists in `File` (which is the new, good API).
```suggestion
          .sorted(Comparator.comparing(path -> noCatch(() -> getLastModifiedTime(path))))
```

micheljung (changes requested):

PR #1183, opened 2019-05-02
Delete folder of old mod

### LINKED ISSUE #1182: Overwriting old mods does not work correctly
While overwriting the old data might not be completely overwritten if the new file is shorter... This causes desyncs

1-alex98:
Fixes #1182


src/test/java/com/faforever/client/mod/InstallModTaskTest.java
```diff
@@ -4,23 +4,22 @@
...
 import java.nio.file.Files;
+import java.nio.file.Path;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.when;
 
-@RunWith(MockitoJUnitRunner.class)
-public class InstallModTaskTest {
+public class InstallModTaskTest extends AbstractPlainJavaFxTest {
```
1-alex98, line 22: Because setting the title of the task is related to java fx

PR #1184, opened 2019-05-03
Add start time column to live replay tab

Wesmania:
We also sort games from most to least recent.  This really helps gauge
if a given replay is worth watching.

### DIFF
--- a/src/main/java/com/faforever/client/vault/replay/LiveReplayController.java
+++ b/src/main/java/com/faforever/client/vault/replay/LiveReplayController.java
@@ -11,6 +11,7 @@ import com.faforever.client.map.MapService;
 import com.faforever.client.map.MapService.PreviewSize;
 import com.faforever.client.remote.domain.GameStatus;
 import com.faforever.client.theme.UiService;
+import com.faforever.client.util.TimeService;
 import com.google.common.base.Joiner;
 import javafx.application.Platform;
 import javafx.beans.binding.Bindings;
@@ -31,6 +32,8 @@ import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
+import java.time.Instant;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map.Entry;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


Wesmania:
I'm getting some exceptions the way it is now, I'll have to debug it for a bit.

Wesmania:
Issue was with game start time sometimes being null, it should be safeguarded against now.

src/main/resources/i18n/messages_fr.properties
```diff
@@ -622,4 +622,5 @@ tournament.status.openForRegistration = Ouvert
 tournament.status.closedForRegistration = Fermé
 tournament.noStartingDate = Pas de date de début
 tournament.noCompletionDate = Pas de date de fin
-settings.notifications.displayNotification = Afficher les notifications
\ No newline at end of file
+settings.notifications.displayNotification = Afficher les notifications
```
1-alex98: Please remove all translations ... Translations are only done in the translation tool

1-alex98 (changes requested):

src/main/resources/i18n/messages_fr.properties
Wesmania: Got it. Intellij is 'smart' enough to add this stuff automatically.

src/main/java/com/faforever/client/vault/replay/LiveReplayController.java
```diff
@@ -116,4 +131,10 @@ public Node getRoot() {
     return new SimpleStringProperty(Joiner.on(i18n.get("textSeparator")).join(modNames));
   }
 
+  @NotNull
+  private String readableGameStartTime(Instant startTime) {
```
Wesmania: I suppose it should go into TimeService? Any preferences regarding naming / format? Or do we consider this formatting a one-off thing not worth standarizing?

1-alex98: yeah u should

1-alex98: Take something short

src/main/java/com/faforever/client/vault/replay/LiveReplayController.java
```diff
@@ -43,25 +46,35 @@
...
     this.i18n = i18n;
     this.mapService = mapService;
+    this.timeService = timeService;
 
     selectedGame = new SimpleObjectProperty<>();
   }
 
   public void initialize() {
-    initializeGameTable(gameService.getGames().filtered(game -> game.getStatus() == GameStatus.PLAYING));
+    var gameSorter = Comparator.comparing(
```
1-alex98: var got introduced in java 10 and only for local variables. 
To keep code consistent I would like to avoid var all together, also in favor of readability.

1-alex98 (changes requested):

1-alex98:
Please resolve conflict

1-alex98 (dismissed):

1-alex98 (approved):

1-alex98:
Please always write fixes #[issue number] into commit message

1-alex98:
And call the branch feature/#[issuenumber]-[name]

Wesmania:
It didn't have an issue to begin with, though :/ I cba. Do you require all features to be preceded by github issues?

1-alex98:
YES!!!! ALSO for changelog...

1-alex98:
ALWAYS HAVE AN ISSUE

src/main/java/com/faforever/client/vault/replay/LiveReplayController.java
```diff
@@ -43,25 +46,35 @@
...
 
   public void initialize() {
-    initializeGameTable(gameService.getGames().filtered(game -> game.getStatus() == GameStatus.PLAYING));
+    Comparator gameSorter = Comparator.comparing(
+        (Game game) -> game.getStartTime(),
+        Comparator.nullsFirst(Comparator.naturalOrder()))
+        .reversed();
+    initializeGameTable(gameService.getGames()
+                    .filtered(game -> game.getStatus() == GameStatus.PLAYING)
+                    .sorted(gameSorter));
```
micheljung, line 77: This passes a `SortedList` as an `ObservableList` which is then wrapped into a `SortedList`. I doubt that this is the way to go, and IntellIJ even warns you about it :(
There's already `liveReplayControllerRoot.comparatorProperty()`.

Not sure what's the side effect is but this needs to be cleaned up.

1-alex98: @Wesmania use Intelij :D

PR #1185, opened 2019-05-04
Bugfix/#1164 coop replay broken

CyrilPaulus:
This fix allow to correctly play coop replays.
I don't know if it's the right solution thought .

For some reason for the coop replays the map file name is empty in the metadata.
Maybe we should fix the replay files instead ?

### DIFF
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -359,7 +359,7 @@ public class ReplayService {
   }
 
   private void runOnlineReplay(int replayId) {
-    runOnlineReplay(replayId);
+    runOnlineReplay(replayId, null);
   }
 
   private void runOnlineReplay(int replayId, String mapName) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -360,7 +375,7 @@ private void runOnlineReplay(int replayId) {
         });
   }
 
-  private void runFafReplayFile(Path path) throws IOException {
+  private void runFafReplayFile(Path path, String overrideMapName) throws IOException {
```
1-alex98, line 378: Please tell me why is that needed and please make a some java doc /comment on it because it does not seem very obvious

1-alex98 (changes requested):

1-alex98:
Please try to always put one issue into one commit via amend option. And if you format the comitt the following way
[title]
[newline]
[description]
Fixes #[issue number]

Then github closes the issue on merge and also puts the title as title and the description as description in the PR by itself

CyrilPaulus:
Ok noted.

I found another cleaner way to fix the problem.

I'll close this pull request and open another cleaner one with the correct commit naming convention and one commit per issue.

PR #1193, opened 2019-05-05
Invalid lock icon char

### LINKED ISSUE #1167: no Icon for password protected game in tile mode
There is no way to see a game is password protected in tile mode
The password protected game is highligthed in red
pw1
Without the toggle
nopw

Can anyone confirm this?

Somewhat related to #945

CyrilPaulus:
Fixes #1167

It seems the wrong unicode char was being used.

Now it looks like this
image


micheljung:
Thank you for the fix!

PR #1195, opened 2019-05-05
Avoid selecting invalid SortProperty

### LINKED ISSUE #1194: Vault replay search criteria not working
The search criteria are not working for the online replay section of the Vault.

There is a least two problems 

* Search button doesn't work, nothing happens when we click on it.
* Sort by doesn't work, it doesn't expand when we click on it.

Edit:
It seems to happen because somehow the value I had stored in my configuration for the sort field isn't present anymore in the sort choices. I don't know how I managed to do that.

Still I have a fix that could prevent such case, is it necessary ?

CyrilPaulus:
Fixes #1194

### DIFF
--- a/src/main/java/com/faforever/client/vault/search/SearchController.java
+++ b/src/main/java/com/faforever/client/vault/search/SearchController.java
@@ -126,7 +126,14 @@ public class SearchController implements Controller<Pane> {
   public void setSortConfig(ObjectProperty<SortConfig> sortConfigObjectProperty) {
     sortPropertyComboBox.getItems().addAll(searchableProperties.values());
     sortOrderChoiceBox.getSelectionModel().select(sortConfigObjectProperty.get().getSortOrder());
-    sortPropertyComboBox.getSelectionModel().select(searchableProperties.get(sortConfigObjectProperty.get().getSortProperty()));
+
+    String savedSortProperty = searchableProperties.get(sortConfigObjectProperty.get().getSortProperty());
+    
+    if (savedSortProperty == null)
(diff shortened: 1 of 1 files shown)


1-alex98 (approved):

src/main/java/com/faforever/client/vault/search/SearchController.java
```diff
@@ -126,7 +126,14 @@ public void setSearchableProperties(Map<String, String> searchableProperties) {
   public void setSortConfig(ObjectProperty<SortConfig> sortConfigObjectProperty) {
     sortPropertyComboBox.getItems().addAll(searchableProperties.values());
     sortOrderChoiceBox.getSelectionModel().select(sortConfigObjectProperty.get().getSortOrder());
-    sortPropertyComboBox.getSelectionModel().select(searchableProperties.get(sortConfigObjectProperty.get().getSortProperty()));
+
+    String savedSortProperty = searchableProperties.get(sortConfigObjectProperty.get().getSortProperty());
+    
+    if (savedSortProperty == null)
+        savedSortProperty = searchableProperties.values().iterator().next();
```
micheljung, line 133: Please enable automatic source code formatting in the commit dialog.

micheljung: Also, there's no need to explicitly select something. The following code would've been simpler and not abuse the variable:

```suggestion
    if (savedSortProperty != null) {
      sortPropertyComboBox.getSelectionModel().select(savedSortProperty);
    }
```

PR #1197, opened 2019-05-05
Better order of vault search options

### LINKED ISSUE #1169: Better order of vault search options
Currently the filter list for searching stuff on the online replay vault is hard to use:
one small but effective improvement would be to change the order in which the filters are listed

right now the first option a user gets is to search by Player UID -> most ppl don't even know their own uid not to mention anyone elses

would be great to have player name, player rating, game mode and map name on top so users can easily search more easily

1-alex98:
Fixes #1169


src/main/resources/i18n/messages_cs.properties
```diff
@@ -194,7 +194,6 @@ game.gameStatus = Stav hry
 game.gameStatus.playing = Hraje
 game.gameStatus.lobby = Čekání na spuštění hry
 game.gameStatus.none = Žádný
-game.id = ID
```
1-alex98, line 197: Because it needs a new translation

Brutus5000 (approved):

src/main/java/com/faforever/client/query/SearchableProperties.java
```diff
@@ -12,34 +12,31 @@
...
+      .put("playerStats.player.login", "game.player.username")
+      .put("playerStats.player.globalRating.rating", "game.player.globalRating")
+      .put("playerStats.player.ladder1v1Rating.rating", "game.player.ladderRating")
+      .put("featuredMod.technicalName", "featuredMod.technicalName")
+      .put("mapVersion.map.displayName", "game.map.displayName")
+      .put("playerStats.faction", "game.player.faction")
+      .put("playerStats.startSpot", "game.player.startSpot")
+      .put("mapVersion.maxPlayers", "map.maxPlayers")
+      .put("mapVersion.ranked", "game.map.isRanked")
       .put("id", "game.id")
```
Geosearchef, line 24: Are you sure we wanna put id that far down?

Geosearchef (approved):

src/main/java/com/faforever/client/query/SearchableProperties.java
1-alex98: I asked @PhilipJFryFAF for an order. And yes almost no user knows game ids

PR #1199, opened 2019-05-05
Fix no effect of show modded games

### LINKED ISSUE #1198: Sometimes clicking "show moded games" has no effect

1-alex98:
Fixes #1198


src/main/java/com/faforever/client/game/GamesTilesContainerController.java
```diff
@@ -65,7 +62,7 @@ public GamesTilesContainerController(UiService uiService, PreferencesService pre
     };
 
     gameListChangeListener = change -> {
-      Platform.runLater(() -> {
+      JavaFxUtil.assertApplicationThread();
```
1-alex98, line 65: run Later made certain actions to be executed in the wrong order and it is on the Application thread anyways because game updates are only done on the application thread

1-alex98: Race conditions caused a malfunction in checkboxes or them having no effect

Geosearchef (approved):

PR #1203, opened 2019-05-06
Dont store password if no auto login

### LINKED ISSUE #1179: password hash is stored even if autoLogin: false
The hashed password of the user is stored (in the client.perf file) even if `Automatic login` is not requested by the user.

Store the password hash in memory only for this specific scenario.

1-alex98:
Fixes #1179


Geosearchef (approved):
are you sure null is fine (and writable) instead of "".

1-alex98:
yes it is I tested it and JSON as no issues to store a null ... it just does not write the filed

PR #1240, opened 2019-05-08
hide old context menu when new one created

### LINKED ISSUE #1218: User context menu opens multiple times
Right clicking a user multiple times opens multiple context menus.

Geosearchef:
closes #1218


Brutus5000 (commented):
Why would we want to create a new one if we have a reference on the old one? Can we not just move the existing one?

Geosearchef:
right, implemented, I've also switched to a weak reference to not keep referencing all the context menus ever created.

1-alex98:
You tested it?

1-alex98 (approved):

Geosearchef:
yes, of course

Brutus5000 (commented):
Please do not introduce a new codestyle (lombok val) without talking to anyone.

1-alex98:
Did not notice.... @Geosearchef java 10 also offers var but we neither want val nor var...

PR #1254, opened 2019-05-20
use path api for map generator service

Geosearchef:


1-alex98 (changes requested):
Travis fails

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -30,19 +25,16 @@
 
 @Lazy
 @Service
+@Slf4j
 public class MapGeneratorService {
 
-  private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  @Getter
-  private static final String GENERATED_MAP_NAME = "neroxis_map_generator_%s_%d"; // The server expects lower case names
+  public static final String GENERATED_MAP_NAME = "neroxis_map_generator_%s_%d"; // The server expects lower case names
```
Brutus5000: Comment on top of the expression

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -91,14 +82,17 @@ public void postConstruct() {
...
+    log.info("Deleting leftover generated maps...");
 
     if (customMapsDirectory != null && customMapsDirectory.toFile().exists()) {
-      Arrays.stream(customMapsDirectory.toFile().listFiles())
-          .filter(File::isDirectory)
-          .filter(f -> GENERATED_MAP_PATTERN.matcher(f.getName()).matches())
-          .map(File::toPath)
-          .forEach(f -> noCatch(() -> FileUtils.deleteRecursively(f)));
+      try {
+        Files.list(customMapsDirectory)
```
Brutus5000: Files.list should be in the try-with-resources block otherwise the stream will not be closed

Brutus5000 (changes requested):

1-alex98:
@Geosearchef will you rework or shall i?

1-alex98:
@Geosearchef pls stick to branch naming convention and open up an issue ALWAYS

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
1-alex98: No that is not case have a deeper look into what it does and you will see
```
    public static Stream<Path> list(Path dir) throws IOException {
        DirectoryStream<Path> ds = Files.newDirectoryStream(dir);
        try {
            final Iterator<Path> delegate = ds.iterator();

            // Re-wrap DirectoryIteratorException to UncheckedIOException
            Iterator<Path> iterator = new Iterator<>() {
                @Override
                public boolean hasNext() {
                    try {
                        return delegate.hasNext();
                    } catch (DirectoryIteratorException e) {
                        throw new UncheckedIOException(e.getCause());
                    }
                }
                @Override
                public Path next() {
                    try {
                        return delegate.next();
                    } catch (DirectoryIteratorException e) {
                        throw new UncheckedIOException(e.getCause());
                    }
                }
            };

            Spliterator<Path> spliterator =
                Spliterators.spliteratorUnknownSize(iterator, Spliterator.DISTINCT);
            return StreamSupport.stream(spliterator, false)
                                .onClose(asUncheckedRunnable(ds));
        } catch (Error|RuntimeException e) {
            try {
                ds.close();
            } catch (IOException ex) {
                try {
                    e.addSuppressed(ex);
                } catch (Throwable ignore) {}
            }
            throw e;
        }
    }
```

Brutus5000: Where is ds being closed? Only in the catch block.

1-alex98: can u close streams did not know that

1-alex98 (approved):

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Brutus5000: Yes. Stream concepts are older than the Java 8 streams. A network connection is a stream, unix stdin or stdout are streams. These streams don't just close if you reach the end (some never end).

Stream extends BaseStream, BaseStream implements AutoCloseable. AutoCloseable defines the close method. AutoCloseable is the interface that is required to be used in the try-with-resources block.

So: collection streams are automatically cleaned up. File streams not necessarily.

Another example is hibernate streams. These are also not auto-closed as you may want to pass them around.

Brutus5000 (approved):

PR #1257, opened 2019-05-20
Fix isAdmin() breaking if cmd.exe has no output

### LINKED ISSUE #1255: isAdmin() causing client to crash at startup for Windows with broken cmd.exe
I have a case of a forum user that has a stack trace like this:

```
Exception in Application init method
java.lang.RuntimeException: Exception in Application init method
	at javafx.graphics/com.sun.javafx.application.LauncherImpl.launchApplication1(Unknown Source)
	at javafx.graphics/com.sun.javafx.application.LauncherImpl.lambda$launchApplication$2(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.util.NoSuchElementException: No line found
	at java.base/java.util.Scanner.nextLine(Unknown Source)
	at com.faforever.client.util.WindowsUtil.isAdmin(WindowsUtil.java:24)
	at com.faforever.client.FafClientApplication.init(FafClientApplication.java:74)
	... 3 more
```

The forum thread is here: http://forums.faforever.com/viewtopic.php?f=3&t=17576
After some brief discussion it seems user's cmd.exe is borked - it opens, then closes for a split second. I assume that also causes the client to fail to run the few script lines that the client needs. We need to catch this at least.

Do we want to depend on cmd.exe to figure out if the current user is an administrator? Maybe there's a better way, I don't know how much cmd.exe can be relied on.

Wesmania:
Fixes #1255.

### DIFF
--- a/src/main/java/com/faforever/client/util/WindowsUtil.java
+++ b/src/main/java/com/faforever/client/util/WindowsUtil.java
@@ -21,6 +21,9 @@ public class WindowsUtil {
 
       boolean printedErrorlevel = false;
       while (true) {
+        if (!scanner.hasNextLine()) {
+          return false;
+        }
         String nextLine = scanner.nextLine();
         if (printedErrorlevel) {
           int errorlevel = Integer.parseInt(nextLine);


1-alex98:
What about a catch Exception in genral to avoid any unexpected behaviours ?

1-alex98:
Have a look at Codacy and what it complain about

src/main/java/com/faforever/client/util/WindowsUtil.java
```diff
@@ -21,6 +21,8 @@ public static boolean isAdmin() {
 
       boolean printedErrorlevel = false;
       while (true) {
+        if (!scanner.hasNextLine())     /* Unexpected output, bail out */
```
1-alex98: use bracets

1-alex98 (changes requested):

src/main/java/com/faforever/client/util/WindowsUtil.java
1-alex98: and remove comment

1-alex98 (changes requested):

src/main/java/com/faforever/client/util/WindowsUtil.java
Wesmania: Bureaucracy :/ (Maybe if 2-space indents weren't the convention we wouldn't need that rule)

1-alex98 (approved):

Brutus5000 (approved):

norraxx:
why not
```
while(scanner.hasNextLine()) {
```
?

1-alex98:
Because we are only interested in the first one...

PR #1258, opened 2019-05-20
Add dragbar if minimized

### LINKED ISSUE #963: UI - can't move window by bar
On windows 10 can't move client window by grabbing it by top bar 
top_cant_move_only_green_small_line_is_ok

1-alex98:
Fixes #963

@micheljung like this?

### DIFF
--- a/src/main/java/com/faforever/client/main/MainController.java
+++ b/src/main/java/com/faforever/client/main/MainController.java
@@ -100,6 +100,8 @@ public class MainController implements Controller<Node> {
   private static final PseudoClass NOTIFICATION_WARN_PSEUDO_CLASS = PseudoClass.getPseudoClass("warn");
   private static final PseudoClass NOTIFICATION_ERROR_PSEUDO_CLASS = PseudoClass.getPseudoClass("error");
   private static final PseudoClass HIGHLIGHTED = PseudoClass.getPseudoClass("highlighted");
+  @VisibleForTesting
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/client/main/MainController.java
```diff
@@ -186,6 +188,16 @@ public void initialize() {
     gameService.addOnRankedMatchNotificationListener(this::onMatchmakerMessage);
     // Always load chat immediately so messages or joined channels don't need to be cached until we display them.
     getView(NavigationItem.CHAT);
+
+    setListenerToSetExtraDragBarIfMinimized();
+  }
+
+  private void setListenerToSetExtraDragBarIfMinimized() {
```
Geosearchef: better name please, add instead for set and maybe also just call it `listen` instead of addListener

Geosearchef (dismissed):

1-alex98:
@micheljung pls review

Brutus5000:
I noticed a bug, but not sure whether it was introduced with this change. If you drag the title bar in full screen mode then it goes to non-full screen. But in this case the maximize button does not update:

grafik


UPDATE: No happens on develop as well.

Brutus5000 (approved):
I'm not happy with replacing the preferenceService-Mock with the actual implementation in the tests. I see no reason for this.

1-alex98:
We do it else where as well. And your strategy is not replace them all to an actual reference. Because a mock needs a thousand stubs and many things are hard to stub. @micheljung said this was preferable to mocking it

PR #1261, opened 2019-05-20
Fix ask player to review

### LINKED ISSUE #1260: Ask player to review game still does not work 100%

1-alex98:
Don't remove the listener if current game is null. Current game might be null because the player dropped out. Only remove listener if the game closed.
Fixes #1260

### DIFF
--- a/src/main/java/com/faforever/client/game/GameService.java
+++ b/src/main/java/com/faforever/client/game/GameService.java
@@ -125,6 +125,7 @@ public class GameService implements InitializingBean {
   private final ModService modService;
   private final PlatformService platformService;
   private final String faWindowTitle;
+  private final DiscordRichPresenceService discordRichPresenceService;
 
   //TODO: circular reference
   @Inject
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -606,20 +625,14 @@ private void onGameInfo(GameInfoMessage gameInfoMessage) {
...
-        removeGame(gameInfoMessage);
+      removeGame(gameInfoMessage);
+      if (!currentPlayerOptional.isPresent() || !Objects.equals(currentGame.get(), game)) {
         return;
       }
+      synchronized (currentGame) {
+        currentGame.set(null);
+      }
 
-      // Don't remove the game until the current player closed it. TODO: Why?
```
1-alex98, line 614: I made a lot of research in the code and this listener causes a lot of issues and yet the comment that the game should only removed when the game process is closed makes absolutely no sense. The game process closing does not trigger this variable to change. Also for it to change we need to call removeGame(gameInfoMessage); .... So it will get stuck there never removing the current game... The code handeling game from the server is a little buggy as I noticed. I hope the restrucuring rules out most of them

Geosearchef (approved):

PR #1262, opened 2019-05-20
Use correct join game method

### LINKED ISSUE #1259: Discord Integration has some bugs
When a game starts it still says "wating" and joining doe not work properly sating "wrong password".

1-alex98:
The other bug that PLAYING is not displayed or games not removed is fixed in a different PR, that reworks code parsing game changes from the server. See PR #1261...
Fixes #1259

### DIFF
--- a/src/main/java/com/faforever/client/discord/DiscordRichPresenceService.java
+++ b/src/main/java/com/faforever/client/discord/DiscordRichPresenceService.java
@@ -93,6 +93,8 @@ public class DiscordRichPresenceService {
 
       if (game.getStartTime() != null) {
         discordRichPresence.setStartTimestamps(game.getStartTime().toEpochMilli());
+      } else if (game.getStatus() == GameStatus.PLAYING) {
+        discordRichPresence.setStartTimestamps(Instant.now().toEpochMilli());
       }
       DiscordRichPresence update = discordRichPresence.build();
       DiscordRPC.discordUpdatePresence(update);
--- a/src/main/java/com/faforever/client/game/JoinGameHelper.java
+++ b/src/main/java/com/faforever/client/game/JoinGameHelper.java
@@ -1,5 +1,6 @@
 package com.faforever.client.game;
 
+import com.faforever.client.discord.DiscordJoinEvent;
 import com.faforever.client.i18n.I18n;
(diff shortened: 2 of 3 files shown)


Geosearchef (dismissed):

1-alex98:
@micheljung  pls review

src/main/java/com/faforever/client/game/JoinGameHelper.java
```diff
@@ -28,7 +32,8 @@
 import static java.util.Arrays.asList;
 
 @Component
-@Scope
+@Scope(ConfigurableBeanFactory.SCOPE_SINGLETON)
```
Brutus5000: Singleton is the default scope.

src/main/java/com/faforever/client/discord/DiscordRichPresenceService.java
```diff
@@ -93,6 +93,8 @@ public void updatePlayedGameTo(Game game, int currentPlayerId, String currentPla
 
       if (game.getStartTime() != null) {
         discordRichPresence.setStartTimestamps(game.getStartTime().toEpochMilli());
+      } else if (game.getStatus() == GameStatus.PLAYING) {
+        discordRichPresence.setStartTimestamps(Instant.now().toEpochMilli());
```
Brutus5000, line 97: This is confusing me.
a) When would this case happen?
b) As you can not join a running game, the start timestamp is still the start time of the game, isn't it?

Brutus5000 (changes requested):

src/main/java/com/faforever/client/game/JoinGameHelper.java
1-alex98: ok

src/main/java/com/faforever/client/discord/DiscordRichPresenceService.java
1-alex98: Well the start time is sometimes set by the server after the change in game status therefore it is not updated....

Brutus5000: Stupid server issues

Brutus5000 (approved):

PR #1270, opened 2019-05-26
Make separate list for sortable properties

### LINKED ISSUE #1263: Not all search options are ok for sorting
There are some options like "player name" that can not be sorted by remove those from the dialog

1-alex98:
Fixes #1263

### DIFF
--- a/src/main/java/com/faforever/client/query/SortableProperties.java
+++ b/src/main/java/com/faforever/client/query/SortableProperties.java
@@ -5,30 +5,23 @@ import com.google.common.collect.ImmutableMap;
 import java.util.Map;
 
 /**
- * Contains mappings of searchable properties (as the API expects it) to their respective i18n key. The reason
- * the i18n keys are not build dynamically is that it makes it impossible for the IDE to detect which key is used where,
- * breaks its refactor capability, and the actual UI text might depend on the context it is used in. Also, this way
- * i18n keys and API keys are nicely decoupled and can therefore be changed independently.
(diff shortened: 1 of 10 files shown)


1-alex98:
Needs 1 or 2 things cleaned up

src/main/java/com/faforever/client/query/SortableProperties.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.Map;
+
+/**
+ * Contains mappings of searchable properties (as the API expects it) to their respective i18n key. The reason the i18n
+ * keys are not build dynamically is that it makes it impossible for the IDE to detect which key is used where, breaks
+ * its refactor capability, and the actual UI text might depend on the context it is used in. Also, this way i18n keys
+ * and API keys are nicely decoupled and can therefore be changed independently.
+ */
+public class SortableProperties {
```
Brutus5000, line 13: The defined constants are obviously not properties but key mappings since Property is a dedicated class now. (Applies for all 3)

Brutus5000 (changes requested):
Some properties should not be sortable to reduce the list to a more user friendly size:
- descriptions
- mod/map version (we only show the latest anyway)
- folderName
- updatedTime
- mod uid

src/main/java/com/faforever/client/query/SortableProperties.java
1-alex98: You want me to rename constants?

Brutus5000 (approved):

PR #1273, opened 2019-05-27
Do nothing on second search

### LINKED ISSUE #1272: Showing everything twice if search clicked twice in mod vault

1-alex98:
Fix #1272


Brutus5000 (dismissed):
The fix looks good but I think it would be more user-friendly to deactivate the search button while loading.

1-alex98:
true that should be pretty easy as well

Brutus5000 (approved):

PR #1275, opened 2019-06-01
Use vault fallback location when needed

### LINKED ISSUE #1274: 3 cases in which maps directory is not found by game
- Unix
- None ascii characters in username
- OneDrive manages Documents
 in all cases we need to use fallback location

1-alex98:
Fix #1274


1-alex98:
@micheljung @Geosearchef @PhilipJFryFAF  maybe lets discuss a little how to do it ...

What I did now was make a setting where you can switch between ProgrammData and MyDouments vault path... Also if a user will have issues with the normal directory due to OneDrive/username/Linux I made the default to be ProgrammData but informaed him because if he already has a symlink he might wanna leave it has it is....

src/main/java/com/faforever/client/preferences/ForgedAlliancePrefs.java
```diff
@@ -63,13 +70,26 @@ public ForgedAlliancePrefs() {
...
     executionDirectory = new SimpleObjectProperty<>();
+    vaultCheckDone = new SimpleBooleanProperty(false);
+    bindVaultPath();
+  }
+
+  /**
+   * Needs to be called after gson deserialization.
+   */
+  protected void bindVaultPath() {
+    customMapsDirectory.unbind();
```
1-alex98, line 90: The problem here is gson recreating the object with new customMapsDirectory  etc. I wanted to stop gson from deserializing it but failed, if i do so other strange exceptions happen maybe somebody can have a look at it... https://stackoverflow.com/questions/4802887/gson-how-to-exclude-specific-fields-from-serialization-without-annotations

src/main/java/com/faforever/client/preferences/PreferencesService.java
```diff
@@ -175,6 +175,15 @@ public Path getFafDataDirectory() {
     return FAF_DATA_DIRECTORY;
   }
 
+  public Path getSecondaryFallbackVaultLocation() {
+    //Todo: find correct path
+    return FAF_DATA_DIRECTORY.resolve("xys");
```
1-alex98: @PhilipJFryFAF  do u know the correct path

src/main/java/com/faforever/client/vault/SpecialCaseForMapVaultLocationChecker.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.bridj.Platform.isUnix;
+import static org.bridj.Platform.isWindows;
+
+/**
+ * Checks if the user is on linux or has a none ascii username. In this cases we need to use a fallback location for the
+ * vault.
+ */
+@Component
+public class SpecialCaseForMapVaultLocationChecker implements InitializingBean {
+  public static final String ONE_DRIVE_IDENTIFIER = "one";
```
1-alex98: not sure how the onedrive folder is called

1-alex98: still to be discussed

Brutus5000: I suggest matching against this regex: `[Oo]ne[ ]*[Dd]rive`

src/main/java/com/faforever/client/vault/SpecialCaseForMapVaultLocationChecker.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static org.bridj.Platform.isUnix;
+import static org.bridj.Platform.isWindows;
+
+/**
+ * Checks if the user is on linux or has a none ascii username. In this cases we need to use a fallback location for the
+ * vault.
+ */
+@Component
+public class SpecialCaseForMapVaultLocationChecker implements InitializingBean {
```
Brutus5000: I would not make this an InitializingBean. How do you make sure that the window is loaded already?
Also, some bean should be responsible for calling this check. (Maybe the init method of the MainController ?)

Brutus5000 (changes requested):

src/main/java/com/faforever/client/vault/SpecialCaseForMapVaultLocationChecker.java
PhilipJFryFAF: example of one user who had the client store stuff in the onedrive folder `C:\Users\%Username%\OneDrive\Documents\My Games\Gas Powered Games\Supreme Commander Forged Alliance\mods`

src/main/java/com/faforever/client/vault/VaultFileSystemLocationChecker.java
```diff
@@ -0,0 +1,62 @@
...
+    if (forgedAlliance.isVaultCheckDone()) {
+      return;
+    }
+    Action revertAction = new Action(i18n.get("revert.alreadySymlinked"), event -> forgedAlliance.setVaultBaseDirectory(preferencesService.getPrimaryVaultLocation()));
+    if (userName != null && !CharMatcher.ascii().matchesAllOf(userName) && isWindows()) {
+      forgedAlliance.setVaultBaseDirectory(preferencesService.getSecondaryFallbackVaultLocation());
+      PersistentNotification notification = new PersistentNotification(i18n.get("mapDir.changed.noneAscii"), Severity.INFO, Collections.singletonList(revertAction));
+      notificationService.addNotification(notification);
+    } else {
+      String myDocumnetsParentFolder = preferencesService.getPrimaryVaultLocation().getParent().toString();
```
Wesmania: typo

src/main/resources/i18n/messages.properties
```diff
@@ -681,4 +681,9 @@ query.notContains=not contains
 game.couldNotJoin=Could not join game with id id ''{0}''.
 discord.disallowJoins=Disallow players joining via Discord
 replay.couldNotOpen=Replay with id ''{0}'' , could not be opened.
-error.game.notTerminatedCorrectly=The game did not terminate correctly and has still certain files open. Some files from the installed game could not be copied over to FAF but unless you recently modified game files in the installed version you probably do not care. We are still starting the game for you.
\ No newline at end of file
+error.game.notTerminatedCorrectly=The game did not terminate correctly and has still certain files open. Some files from the installed game could not be copied over to FAF but unless you recently modified game files in the installed version you probably do not care. We are still starting the game for you.
+revert.alreadySymlinked=Revert this I got it symlinked
```
Wesmania, line 685: Is this meant to be shown?

src/main/java/com/faforever/client/vault/VaultFileSystemLocationChecker.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public void checkVaultFileSystemLocation() {
+    String userName = new NTSystem().getName();
+    ForgedAlliancePrefs forgedAlliance = preferencesService.getPreferences().getForgedAlliance();
+    if (forgedAlliance.isVaultCheckDone()) {
+      return;
+    }
+    Action revertAction = new Action(i18n.get("revert.alreadySymlinked"), event -> forgedAlliance.setVaultBaseDirectory(preferencesService.getPrimaryVaultLocation()));
+    if (userName != null && !CharMatcher.ascii().matchesAllOf(userName) && isWindows()) {
```
Wesmania, line 46: We're doing almost the same thing in all these branches. We could put logic in an enum that returns default FA directory status (OK, Non-ascii username, OneDrive or unix) and do base dir setup and notification in one place.

ci/build.sh
```diff
@@ -8,4 +8,4 @@ if [ ! -d "$HOME/openjfx/javafx-sdk-11" ]; then
   ls -lah $HOME/openjfx/javafx-sdk-11/lib
 fi
 
-./gradlew --stacktrace -Pversion=${APP_VERSION} -PjavafxPlatform=linux jar
+./gradlew --full-stacktrace -Pversion=${APP_VERSION} -PjavafxPlatform=linux jar
```
Wesmania: Is this relevant?

src/main/java/com/faforever/client/preferences/ForgedAlliancePrefs.java
```diff
@@ -41,12 +42,18 @@
...
+  private final ObjectProperty<Path> vaultBaseDirectory;
+  private transient final ObjectProperty<Path> customMapsDirectory;
+  private transient final ObjectProperty<Path> modsDirectory;
   private final IntegerProperty port;
   private final BooleanProperty autoDownloadMaps;
+  /**
+   * Saves if the client checked for special cases in which it needs to set the fallback vault location. See {@link
+   * com.faforever.client.vault.VaultFileSystemLocationChecker}
+   */
+  private final BooleanProperty vaultCheckDone;
```
Wesmania, line 57: What if user moves stuff to a OneDrive? Is vault path check costly?

src/main/resources/i18n/messages.properties
Brutus5000: The user has the option not to create a symlink in case he already did it manually.

1-alex98: Well if you have a symlink and we change the location you see everything twice

ci/build.sh
1-alex98: No sorry was for debugging and I forgot it thx

src/main/java/com/faforever/client/preferences/PreferencesService.java
```diff
@@ -175,6 +175,14 @@ public Path getFafDataDirectory() {
     return FAF_DATA_DIRECTORY;
   }
 
+  public Path getSecondaryFallbackVaultLocation() {
+    return Paths.get(FAF_DATA_DIRECTORY.toAbsolutePath().toString(), "user", "My Games", "Gas Powered Games", "Supreme Commander Forged Alliance");
+  }
```
Geosearchef, line 189: Is this a secondary vault location or a fallback vault location?
And if it's a secondary fallback location, then the secondary fallback or the fallback of the secondary location? And what about a terti....

src/main/java/com/faforever/client/vault/VaultFileSystemLocationChecker.java
```diff
@@ -0,0 +1,62 @@
...
+import static org.bridj.Platform.isUnix;
+import static org.bridj.Platform.isWindows;
+
+/**
+ * Checks if the user is on linux, has a none ascii username or has One Drive installed. In this cases we need to use a fallback location for the
+ * vault.
+ */
+@Component
+public class VaultFileSystemLocationChecker {
+  public static final Pattern ONE_DRIVE_FOLDER_NAME_IDENTIFIER = Pattern.compile("[Oo]ne[ ]*[Dd]rive");
```
Geosearchef, line 25: Don't we wanna fallback on non ascii name chars aswell?

src/main/java/com/faforever/client/preferences/ForgedAlliancePrefs.java
1-alex98: I am not sure what u mean... If One Drive is installed (fully) it moves Documents to One Dive but the game still uses the old Documents folder but Windows tells the client the folder is in One Drive therefore the game does not find the vault... I tried finding a way of querying the none OneDrive Documents location but because I have no system to try if it works I have not changed that so far. For now we just use the fall back location if the user uses One Drive

src/main/java/com/faforever/client/vault/VaultFileSystemLocationChecker.java
Geosearchef: [ ]* can be replace by *

Geosearchef (changes requested):

src/main/java/com/faforever/client/preferences/PreferencesService.java
1-alex98: Secondary is similar to fall back. You mean I should decide on one or what????

src/main/java/com/faforever/client/vault/VaultFileSystemLocationChecker.java
1-alex98: Whatever @Brutus5000 did that and I think it is better to read like this

src/main/java/com/faforever/client/vault/VaultFileSystemLocationChecker.java
1-alex98: Did not use an enum but used extracted to a Method

src/main/java/com/faforever/client/preferences/PreferencesService.java
1-alex98: Renamed to only Secondary...

1-alex98:
Codacy failing without any issues??? I though I fixed it  😢

Brutus5000 (dismissed):

src/test/java/com/faforever/client/map/generator/MapGeneratorServiceTest.java
```diff
@@ -72,14 +73,13 @@ public void setUp() throws IOException {
...
-    preferences.getForgedAlliance().setCustomMapsDirectory(customMapsDirectory.getRoot().toPath());
+    preferences.getForgedAlliance().setVaultBaseDirectory(vaultBaseDirectory.getRoot().toPath());
     when(preferencesService.getPreferences()).thenReturn(preferences);
 
-
     instance = new MapGeneratorService(applicationContext, preferencesService, taskService);
 
     instance.postConstruct();
-    assertThat(Arrays.asList(Objects.requireNonNull(customMapsDirectory.getRoot().list())), not(contains(testMapNameGenerator)));
+    assertThat(Files.list(customMapsDir).collect(Collectors.toList()), not(contains(testMapNameGenerator)));
```
1-alex98: I need to close this tho

src/main/java/com/faforever/client/main/MainController.java
```diff
@@ -115,6 +116,7 @@
   private final int ratingBeta;
   private final GamePathHandler gamePathHandler;
   private final PlatformService platformService;
+  private final VaultFileSystemLocationChecker vaultFileSystemLocationChecker;
```
Geosearchef, line 119: Long name

src/main/java/com/faforever/client/vault/VaultFileSystemLocationChecker.java
```diff
@@ -0,0 +1,70 @@
...
+  public VaultFileSystemLocationChecker(NotificationService notificationService, PreferencesService preferencesService, I18n i18n) {
+    this.notificationService = notificationService;
+    this.preferencesService = preferencesService;
+    this.forgedAlliance = preferencesService.getPreferences().getForgedAlliance();
+
+    this.i18n = i18n;
+  }
+
+  public void checkVaultFileSystemLocation() {
+    String userName = new NTSystem().getName();
```
Geosearchef: ```System.getProperty("user.name")```

Geosearchef: this would crash on linux

Geosearchef (changes requested):

src/main/java/com/faforever/client/main/MainController.java
1-alex98: Yes and long names are good as long as they are more descriptive than a short name.

PhilipJFryFAF (approved):
can't review it cause i'm unable to build the client with this PR locally but from the description (ability to switch and default folder in programdata for users with onedrive/username issues/linux) it sounds like it's ready

PR #1288, opened 2019-06-11
Rebrand the client

### LINKED ISSUE #1233: Debrand the client
This client is now the official faf client. There's no reason anymore like there was in the first place to integrate a player's username into the client name. Technically that will probably stay, but a lot of users are confused by the client's name.

1-alex98:
Fix #1233


Brutus5000 (changes requested):
Run a full search on the client plz. I already found and there is even more (e.g. in the build.gradle)
InitSessionMessage line 6
messages.properties line 651
application.yaml line 46

Geosearchef:
Revert compiler.xml, modules.xml, faf client prod.xml

1-alex98:
> Revert compiler.xml, modules.xml, faf client prod.xml

No they are broken/wrong since a long time

Geosearchef (approved):

1-alex98:
This needs further investigation on what certain things do

micheljung:
Do you really want to call the client (and everything, like log files) "Official FAF Client" rather than just "FAF Client"?

1-alex98:
Well not sure

Rackover:
FAF client is better than Official Faf Client I think, but thats just my opinion

PR #1289, opened 2019-06-11
Add recommended maps category read from remote preferences

### LINKED ISSUE #1282: "M&M Top Picks" Vault Category
layout image

In the vault section of the Java client, a new category in the maps and mods vault will be added which reads a list of map or mod IDs, generated by the M&M team.

The list will live on the message board and through a shared file (link to be added soon(tm)).

No modifications to the GUI are needed and should look like the rest so there is a good flow.

Once the permission system is restored to the java client, the map and mod id list will exist solely as a source of reference. The team can then use their mod access to create the list from within the client rather need to request updates to the moderation team.

Brutus5000:
(closes #1282)

### DIFF
--- a/src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
+++ b/src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
@@ -254,6 +254,19 @@ public class FafApiAccessorImpl implements FafApiAccessor, InitializingBean {
   }
 
   @Override
+  public List<Map> getMapsById(List<Integer> mapIdList, int count, int page) {
+    String filterCriteria = mapIdList.stream()
+        .map(Object::toString)
+        .collect(Collectors.joining(",", "latestVersion.map.id=in=(", ")"));
+
+    return getPage(MAP_ENDPOINT, count, page, ImmutableMap.of(
+        "include", "statistics,latestVersion,author,versions.reviews,versions.reviews.player",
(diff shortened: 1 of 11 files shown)


src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
```diff
@@ -253,6 +253,19 @@ public Ladder1v1LeaderboardEntry getLadder1v1EntryForPlayer(int playerId) {
...
   }
 
+  @Override
+  public List<Map> getMapsById(List<Integer> mapIdList, int count, int page) {
+    String filterCriteria = mapIdList.stream()
+        .map(Object::toString)
+        .collect(Collectors.joining(",", "latestVersion.map.id=in=(", ")"));
+
+    return getPage(MAP_ENDPOINT, count, page, ImmutableMap.of(
+        "include", "statistics,latestVersion,author,versions.reviews,versions.reviews.player",
```
1-alex98, line 263: use MAP INCLUDES constant

1-alex98 (changes requested):

src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
1-alex98: does not exist

1-alex98 (approved):

PR #1293, opened 2019-06-13
Removed background by default from the style.css

### LINKED ISSUE #1232: Consider switching/disabling background image by default
The background image is the opposite of good and clear UI design and annoys a lot of people.
Maybe we should disable it by default or replace it with some more generic that makes the background a bit more interesting than just one color.

Rackover:
The user can still re-add a background as part of their custom theme if they want.
Closes https://github.com/FAForever/downlords-faf-client/issues/1232


1-alex98:
Did you try it I dont't think that will work out

1-alex98:
This css is overwritten by code in the MainController.class anyways

MainController.class line 451

src/main/java/com/faforever/client/main/MainController.java
```diff
@@ -447,19 +447,16 @@ private void registerWindowListeners() {
 
   private void setBackgroundImage(Path filepath) {
     Image image;
-    if (filepath == null || filepath.toString().isEmpty()) {
-      image = new Image(uiService.getThemeFile("theme/images/login-background.jpg"));
-    } else {
+    if (filepath != null && !filepath.toString().isEmpty()) {
```
1-alex98: else overwrite it otherwise how can a user remove an image

src/main/resources/theme/style.css
```diff
@@ -873,13 +873,6 @@
 
 /***************** Login *****************/
 
-.login-root {
```
1-alex98: Add back in

1-alex98 (changes requested):

src/main/resources/theme/style.css
Rackover: https://material.io/design/

https://www.google.fr/search?ei=Z9sCXcWLM4GWaJDJn5AJ&q=material+design+background+image&oq=material+design+background+image&gs_l=psy-ab.3..0i19l2j0i22i30i19.133099.135599..135705...0.0..1.131.2751.26j6......0....1..gws-wiz.......0i71j0i131j0j0i67j0i131i67j0i3j0i22i30j0i13i5i30i19.ZzHelNZT_94

PR #1300, opened 2019-06-20
Fix notifications shown despite disabled and others not always shown

### LINKED ISSUE #1274: 3 cases in which maps directory is not found by game

1-alex98:
Fix #1274

### DIFF
--- a/src/main/java/com/faforever/client/game/GameService.java
+++ b/src/main/java/com/faforever/client/game/GameService.java
@@ -46,7 +46,6 @@ import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.value.ChangeListener;
 import javafx.beans.value.ObservableValue;
 import javafx.collections.FXCollections;
-import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
 import javafx.collections.ObservableMap;
 import lombok.extern.slf4j.Slf4j;
@@ -184,23 +183,6 @@ public class GameService implements InitializingBean {
     games = FXCollections.observableList(new ArrayList<>(),
         item -> new Observable[]{item.statusProperty(), item.getTeams()}
     );
-    games.addListener((ListChangeListener<Game>) change -> {
-      /* To prevent deadlocks (i.e. synchronization on the game's "teams" and on the google event subscriber), only
-      allow this to run on the application thread. */
-      JavaFxUtil.assertApplicationThread();
-
-      while (change.next()) {
-        change.getRemoved().forEach(game -> eventBus.post(new GameRemovedEvent(game)));
-
-        if (change.wasUpdated()) {
-          for (int i = change.getFrom(); i < change.getTo(); i++) {
-            eventBus.post(new GameUpdatedEvent(change.getList().get(i)));
-          }
-        }
-
-        change.getAddedSubList().forEach(game -> eventBus.post(new GameAddedEvent(game)));
-      }
-    });
     JavaFxUtil.attachListToMap(games, uidToGameInfoBean);
   }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -184,23 +183,6 @@ public GameService(ClientProperties clientProperties, FafService fafService,
     games = FXCollections.observableList(new ArrayList<>(),
         item -> new Observable[]{item.statusProperty(), item.getTeams()}
     );
-    games.addListener((ListChangeListener<Game>) change -> {
```
1-alex98, line 187: This is problematic because when updating teams of games.
1.first the teams are cleared
2. Everything is added back in
Other code relying on the Events do then think all players rejoined the games which is wrong so we better post the events in a different spot, there where we process the server message.

src/main/java/com/faforever/client/player/PlayerService.java
```diff
@@ -97,14 +100,16 @@ public void onGameRemoved(GameRemovedEvent event) {
     Game game = event.getGame();
     ObservableMap<String, List<String>> teams = game.getTeams();
     synchronized (teams) {
-      teams.forEach((team, players) -> updateGamePlayers(players, null));
+      List<String> playersInGame = teams.entrySet().stream().flatMap(stringListEntry -> stringListEntry.getValue().stream()).collect(Collectors.toList());
```
1-alex98, line 103: To be honest this is equivalent to the way it was done before but it is more convenient to do it in one call also for debug purposes

src/main/java/com/faforever/client/player/PlayerService.java
```diff
@@ -139,51 +144,54 @@ private void resetIdleTime(Player playerForUsername) {
...
+  private void updateGamePlayers(List<String> currentPlayers, Game game) {
+    log.info(currentPlayers.toString());
+    currentPlayers.stream()
         .map(this::getPlayerForUsername)
         .filter(Optional::isPresent)
         .map(Optional::get)
         .forEach(player -> {
           resetIdleTime(player);
-          updateGameToPlayer(game, player);
-          if (game == null || game.getStatus() == GameStatus.CLOSED) {
```
1-alex98, line 150: moved that into a method

src/main/java/com/faforever/client/player/PlayerService.java
```diff
@@ -139,51 +144,54 @@ private void resetIdleTime(Player playerForUsername) {
...
     if (game != null && game.getStatus() != GameStatus.CLOSED && playersByGame.get(game.getId()) != null) {
-      playersByGame.get(game.getId()).stream()
-          .forEach(player -> {
-            boolean stillInGame = game.getTeams().entrySet().stream()
-                .anyMatch(entry -> entry.getValue().contains(player.getUsername()));
-            if (!stillInGame) {
-              player.setGame(null);
-            }
-          });
+      List<Player> playersThatLeftTheGame = new ArrayList<>();
```
1-alex98, line 159: There where somethings going wrong here which caused the Friend joined game message not to be shown. I am not too sure what it was in particular, becuse it was more than one thing and I just cleaned everything up a little.

Also I added unit tests to make sure the new implementation is working

Brutus5000 (changes requested):
The question whether or not to show a notification or play a sound should be handled inside the NotificationService and/or the AudioService.
- This simplifies the controllers that use it
- If you add or remove client settings you don't have to touch all the cases

If these services cannot decide it due to the type of the message, the messages need to be enhanced to enable this.

1-alex98:
@Brutus5000 u answered your self the way it is built it is not possible for the notification service to tell which notification is which. We can do refactoring on this but I will not do that now. We can make an issue for that. Also I am not sure how u wanna do it. Those are also completely different notification to start with so different methods are called all over the place.

I woul create an Enum with all relevant information for each notification so the service/controller only calls notificationSerivce.addNotification(PlayerOnlie);

1-alex98:
or maybe not because they have parameters like the player name🤔

src/main/java/com/faforever/client/player/FriendOfflineNotifier.java
```diff
@@ -41,18 +45,24 @@ public void afterPropertiesSet() {
...
         return;
       }
 
-      audioService.playFriendOfflineSound();
-      notificationService.addNotification(
-          new TransientNotification(
-              i18n.get("friend.nowOfflineNotification.title", username), "",
-              IdenticonUtil.createIdenticon(player.getId())
-          ));
+      if (notification.isFriendOfflineSoundEnabled()) {
```
Geosearchef, line 55: wouldn't it be better to do those checks in the audioService methods?

Geosearchef (approved):

src/main/java/com/faforever/client/player/FriendOfflineNotifier.java
1-alex98: I will open up an issue to refactor the whole thing but yes. I just followed the way it was done before and added the missing checks

PR #1302, opened 2019-06-22
Add email login warning

### LINKED ISSUE #1295: If user uses an @ in the username warn him to not use his email to login

Katharsas:
Adds new warning if user tries to login with username looking like email
Fixes #1295

### DIFF
--- a/src/main/java/com/faforever/client/login/LoginController.java
+++ b/src/main/java/com/faforever/client/login/LoginController.java
@@ -7,6 +7,7 @@ import com.faforever.client.config.ClientProperties.Server;
 import com.faforever.client.fx.Controller;
 import com.faforever.client.fx.JavaFxUtil;
 import com.faforever.client.fx.PlatformService;
+import com.faforever.client.i18n.I18n;
 import com.faforever.client.preferences.LoginPrefs;
 import com.faforever.client.preferences.PreferencesService;
 import com.faforever.client.update.ClientConfiguration;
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -44,11 +46,13 @@
 @Slf4j
 public class LoginController implements Controller<Node> {
 
+  private static final Pattern emailRegex = Pattern.compile(".*[@].*[.].*");
```
Brutus5000, line 49: constants go in UPPERCASE_WITH_UNDERSCORES in our codestyle

src/test/java/com/faforever/client/login/LoginControllerTest.java
```diff
@@ -76,4 +85,15 @@ public void testForgotPasswordButtton() throws Exception {
...
+
+  @Test
+  public void testUsernameEmailWarning() {
+    instance.usernameInput.setText("test@example.com");
+    instance.passwordInput.setText("foo");
+    instance.loginButton.fire();
+    verify(i18n).get(LOGIN_WITH_EMAIL_WARNING_KEY);
+    verifyZeroInteractions(userService);
+    assertThat(instance.loginErrorLabel.isVisible(), is(true));
+    assertThat(instance.loginErrorLabel.getText(), is(LOGIN_WITH_EMAIL_WARNING_KEY));
```
Brutus5000, line 97: In positive case we should assert that the loginErrorLabel is not visible.

Brutus5000 (changes requested):

src/test/java/com/faforever/client/login/LoginControllerTest.java
1-alex98: Well this not really his job tho

Brutus5000: What? If the code is wrong and always shows the error message? how do you test it then?

PR #1315, opened 2019-06-25
Remove translation of map size to german

Justify87:
It gets cut off in the client anyway, not showing the size of the map in the preview. the other translations don't have it either.


1-alex98:
Translations need to be done on poeditor

Justify87:
well thanks for nothing, i guess

Issue #1104, opened 2019-02-05
Out Of Memory Error while browsing the map vault


Justify87:
java.util.concurrent.CompletionException: java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(Unknown Source)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(Unknown Source)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(Unknown Source)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(Unknown Source)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.OutOfMemoryError: Java heap space

[downlords-faf-client.zip](https://github.com/FAForever/downlords-faf-client/files/2832237/downlords-faf-client.zip)

I clicked on "Load more" to show more maps. A few secs after the second click in that button the error above showed up.

Brutus5000:
Can you reproduce this? How many times did you click on load more? How much memory has your system?
(By the way this is the reason why the local replay vault is limited to 300 records - in a bad way though)

Justify87:
For reprododuction:
- Go to map vault
- Dont' search for anything
- Click on this (newst maps):
![grafik](https://user-images.githubusercontent.com/10554299/52474053-a3392c80-2b97-11e9-97a5-9b2e7cf4e4c4.png)
- the second clicks on this button will trigger the error
![grafik](https://user-images.githubusercontent.com/10554299/52474114-d1b70780-2b97-11e9-8bba-9a54669e87ce.png)

I have 500GB of SSD ROM and 16GB of RAM
about 10GB of RAM were not in use when this error was thrown and I have plenty of free space on my SSD.

I noticed though that after clicking on this:
![grafik](https://user-images.githubusercontent.com/10554299/52474053-a3392c80-2b97-11e9-97a5-9b2e7cf4e4c4.png)
The whole client already felt sluggish even before the error was thrown.

Brutus5000:
[...]

PR #1320, opened 2019-06-27
Simple pagination for map vault and cleanup

### LINKED ISSUE #1319: Code problems with MapVaultController
Related to looking into #1104 
There are a bunch of problems here:

- Constant `LOAD_MORE_COUNT` is the number of maps loaded for each click of 'more' button for PREDEFINED queries, while `MAX_SEARCH_RESULTS` is the number of maps loaded for each click of 'more' button for CUSTOM queries. For all i see, these two constants could be merged into one, becasue they bassically do the same thing.
- `MAX_SEARCH_RESULTS` is misnamed since it kinda implies that it is the maximum number of search results that can be displayed on a page, which it is not.
- More button visibility (which is important for BOTH custom and predefined queries) relies on `MAX_SEARCH_RESULTS`, but not on `LOAD_MORE_COUNT` which doesn't make sense because the visibility code runs for both custom and predefined queries.
- There is a multithreading ordering issue:
  `enterSearchResultState` is the method that sets 'more' button visibility by calling `searchResultPane.getChildren().size()`. However, `enterSearchResultState` is sometimes called from both background thread and application thread (`JavaFxUtil.assertBackgroundThread();` fails for predefined queries). This means that there is multithreaded read/write to `searchResultPane.getChildren()` which causes this to sometimes report the wrong (not yet updated) number of children to be counted by `searchResultPane.getChildren().size()`, causing more button to show up when it shouldn't.

Katharsas:
- Changes 'more' button into working like a 'next page' button (unloads previous map cards on click) to prevent rising memory and cpu usage.
- Makes sure all code in MapVaultController is called from correct thread
- Removes redundant constant `MAX_SEARCH_RESULTS`

Continues #1104
Fixes #1319

This should make nothing worse, should reduce memory usage dramatically, fix some bugs and pepare for proper pagination UI. So its not really a quickfix anymore even if i put that in the branch name. Sorry^^


src/main/java/com/faforever/client/map/MapVaultController.java
```diff
@@ -249,6 +256,7 @@ private void openMapUploadWindow(Path path) {
   }
 
   public void onRefreshButtonClicked() {
+    JavaFxUtil.assertBackgroundThread();
```
1-alex98: Why should tis run on Backgound thread

src/main/java/com/faforever/client/map/MapVaultController.java
```diff
@@ -313,9 +321,12 @@ public void showMoreOwnedMaps() {
...
   }
 
-  private void appendSearchResult(List<MapBean> maps, Pane pane) {
+  private void replaceSearchResult(List<MapBean> maps, Pane pane) {
     JavaFxUtil.assertBackgroundThread();
 
+    Platform.runLater(() -> {
+      pane.getChildren().clear();
+    });
     ObservableList<Node> children = pane.getChildren();
```
1-alex98: Please just fill this variable in the runLater part because it is needed there

src/test/java/com/faforever/client/map/MapVaultControllerTest.java
```diff
@@ -159,7 +159,7 @@ public void testNotifyPropertyShowLadderInitialized() {
     instance.display(new ShowLadderMapsEvent());
 
     WaitForAsyncUtils.waitForFxEvents();
-    verify(mapService).getLadderMaps(100, 1);
+    verify(mapService).getLadderMaps(50, 1);
   }
```
1-alex98, line 184: Maybe add a unit test that tests if pagination works

src/main/java/com/faforever/client/map/MapVaultController.java
```diff
@@ -58,8 +58,7 @@
 
   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
   private static final int TOP_ELEMENT_COUNT = 7;
-  private static final int LOAD_MORE_COUNT = 100;
-  private static final int MAX_SEARCH_RESULTS = 100;
+  private static final int LOAD_MORE_COUNT = 50;
```
1-alex98: Maybe call it page count because it is not only the number of maps loaded on "more" clicked

1-alex98 (changes requested):

src/main/java/com/faforever/client/map/MapVaultController.java
Katharsas: Its the count of maps per page / per 'more' click. So i could name it 'LOAD_PER_PAGE' maybe?

src/main/java/com/faforever/client/map/MapVaultController.java
Katharsas: Yeah it should not, and the assert fails so the button is actually broken. Need to remove that.

src/main/java/com/faforever/client/map/MapVaultController.java
Katharsas: You want me to remove the duplication of `pane.getChildren?` Filling the list happens in batches so i cannot `clear()` there.

src/main/java/com/faforever/client/map/MapVaultController.java
1-alex98: yeah

src/main/java/com/faforever/client/map/MapVaultController.java
1-alex98: I want 
```
pane.getChildren().clear();
ObservableList<Node> children = pane.getChildren();
```
in line 341

Katharsas: Well its not needed i think (reference assignments are thread-safe), but i think the line `ObservableList<Node> children = pane.getChildren();` should be removed entirely. As far as i can see it does not help to make any code clearer or better, and its at an awkward place should this method get refactored / split up in the future.

src/test/java/com/faforever/client/map/MapVaultControllerTest.java
Katharsas: Ok i will look into that.

1-alex98:
Also did you rename the button to "load next" maybe because "load more" sounds like it loads more instead of replacing current one with those up next

Katharsas:
'Load next' is kinda missing the subject imo (which next?). 'Load more' seems more intuitive for non-technical users to me (becuase in that case the subject is obvious: Load more maps). 

Maybe 'load next page'? But id still argue, that 'page' is kind of a dev/technical/non-intuitive term.

1-alex98:
Well whatever more is not tat bad either

Katharsas:
I still want to add tests btw, just had to fix tests in Eclipse first^^

Katharsas:
Tests added. I have to say i really like Mockito.

src/test/java/com/faforever/client/map/MapVaultControllerTest.java
```diff
@@ -38,8 +41,13 @@
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import static com.faforever.client.map.MapVaultController.*;
+
 public class MapVaultControllerTest extends AbstractPlainJavaFxTest {
 
+//  private static final int TOP_ELEMENT_COUNT = 7;
```
1-alex98: No commented out code we have git there is o need for this

1-alex98 (changes requested):

Katharsas:
While we are at it: Is this comment
https://github.com/FAForever/downlords-faf-client/pull/1320/files#diff-5817aa64094c1068a332e4f37f011d87R54
still relevant? Its seems to me that Elide has fully implemented this, but the getXXX methods are still used. So whats the status on that?

1-alex98:
@Brutus5000 is most into elide he might know.

1-alex98 (dismissed):

src/test/java/com/faforever/client/map/MapVaultControllerTest.java
```diff
@@ -16,19 +17,21 @@
...
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
-import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.nullValue;
+import static org.hamcrest.Matchers.is;
```
Geosearchef: Why? Is replacing that intended?

Geosearchef (dismissed):

src/test/java/com/faforever/client/map/MapVaultControllerTest.java
Katharsas: No, ill revert this.

Katharsas:
I think this is just Travis having this issue again.

1-alex98:
Not travis it is some of our tests

1-alex98:
rebased

Brutus5000:
I have no clue what that comment is supposed to mean. The underlying API calls are very different I don't see any harmonization potential.

Brutus5000 (dismissed):
I might be my personal opinion, but if you go for pagination you should offer a previous page button along with the next page (`more...`) button.

1-alex98:
Well this would be the next steep I guess

1-alex98 (dismissed):

Brutus5000:
I'm not approving this. The current state is a loss of functionality and user experience until there is a back button.

1-alex98:
ok

Katharsas:
> I'm not approving this. The current state is a loss of functionality and user experience until there is a back button.

The previous state is that people's vault crashes when they go for the second or third 'more' click. So i would not say that it is a loss of functionality, its more of a (imo positive) trade (going back a page is not really helpful if you cannot go forward properly).
And you can always just go back to the start page and reiterate the pages.

1-alex98:
We can still add the previous button should be really easy i think

Katharsas:
If you add a previous button, you need to split the more button from the front page showroom to stay a 'more' button there but become a 'next' button on the actual pages (because the showroom is not the first page of anything). Sure, still easy, but his PR was meant as a kind of important bugfix for vault crashes (which btw are only reproducable in the actual officially distributed client release), not a UI change.

If we want a 'proper' pagination UI, there are more things that should also be touched/changed. Showing the current search on all pagination pages, direct links to intermediate pages, ect...

src/main/java/com/faforever/client/map/MapVaultController.java
```diff
@@ -367,8 +380,11 @@ private void displayMaps(List<MapBean> maps) {
   }
 
   public void onLoadMoreButtonClicked() {
-    moreButton.setVisible(false);
-    loadingLabel.setVisible(true);
+
```
Katharsas: Can this method be removed? I guess it can.

src/main/java/com/faforever/client/map/MapVaultController.java
```diff
@@ -385,6 +401,18 @@ public void onLoadMoreButtonClicked() {
         });
   }
 
+  public void onLoadPreviousButtonClicked() {
+    currentPage--;
+
```
Katharsas: Im gonna remove these empty lines.

Katharsas (commented):
@Brutus5000 axel added a back button.

1-alex98 (approved):

1-alex98:
@Brutus5000 better now?

Issue #1298, opened 2019-06-20
JavaFX "Error: JavaFX runtime components are missing..."


Katharsas:
When i use either AdoptOpenJDK 10 or 11 and try to build and start the client locally, i get the following error, like some others:
`Error: JavaFX runtime components are missing, and are required to run this application`

Potential cause, citiation from https://stackoverflow.com/questions/52569724/javafx-11-create-a-jar-file-with-gradle :
> This error comes from sun.launcher.LauncherHelper in the java.base module. The reason for this is that the Main app extends Application and has a main method. If that is the case, the LauncherHelper will check for the javafx.graphics module to be present as a named module:
`Optional<Module> om = ModuleLayer.boot().findModule(JAVAFX_GRAPHICS_MODULE_NAME);`
> If that module is not present, the launch is aborted. Hence, having the JavaFX libraries as jars on the classpath is not allowed in this case.

I don't really understand fully what is going on here, but the proposed second fix works:
Add a new java class with a main method and define that class as entry point for the application. This class must not extend javafx's Application. It just simply calls the Application implementation like this:
```
package com.faforever.client;

public class Main {
	public static void main(String[] args) {
		FafClientApplication.main(args);
	}
}
```

I can make a pull request if you want (tell me).

Katharsas:
Ok, this is kinda funny.
This fix works **ONLY** if i use an Eclipse configuration to directly start the Main classes main method. However, `gradle bootRun` is still broken (same error as before). So the fix kinda works, and kinda not.

(i did change 
```
jar {
  manifest {
    attributes "Main-Class": "com.faforever.client.FafClientApplication",
        "Implementation-Title": "Downlord's FAF Client",
        "Implementation-Version": version,
        "Implementation-Vendor": "Downlord"
  }
}
```
to 
```
jar {
  manifest {
    attributes "Main-Class": "com.faforever.client.Main",
        "Implementation-Title": "Downlord's FAF Client",
        "Implementation-Version": version,
        "Implementation-Vendor": "Downlord"
  }
}
```
)
Seems like I actually need to fully understand that stackoverflow page to know whats going on precisely : (

1-alex98:
Well putting it on the classpath is kind of cheating we should be able to also put it on the modulepath with gradle

Katharsas:
Ok so when i add
```
bootRun {
    classpath sourceSets.main.runtimeClasspath
    main = "com.faforever.client.Main"
}
```
[...]

PR #1325, opened 2019-06-29
Allow using JDKs without bundled JavaFX

Katharsas:
**I need somebody with intelliJ Ultimate to test the run configurations.**

Changes the main class (entrypoint) so that it no longer is the same class that extends Application.
- New main method inside com.faforever.client.Main, updated all references
- Explicitly sets entry point for bootRun in build.gradle
- Removes additional ambigous main method from WindowsUtil (strictly this only needed if entry point for boot run is not set, but i guess i remove this just in case)
- Renames original main method to avoid confusion

Workaround for #1298

### DIFF
--- a/.idea/runConfigurations/FafClientApplication.xml
+++ b/.idea/runConfigurations/FafClientApplication.xml
@@ -3,7 +3,7 @@
     <envs>
       <env name="spring.profiles.active" value="prod" />
     </envs>
-    <option name="MAIN_CLASS_NAME" value="com.faforever.client.FafClientApplication" />
+    <option name="MAIN_CLASS_NAME" value="com.faforever.client.Main" />
     <module name="downlords-faf-client.main" />
     <option name="VM_PARAMETERS" value="-DnativeDir=build/resources/native -Dprism.dirtyopts=false -Xms128m -Xmx512m -XX:MinHeapFreeRatio=15 -XX:MaxHeapFreeRatio=33 -XX:+HeapDumpOnOutOfMemoryError -XX:+UseStringDeduplication -javaagent:webview-patch/build/libs/webview-patch.jar" />
     <method v="2">
--- a/build.gradle
+++ b/build.gradle
@@ -132,9 +132,14 @@ configurations {
   codacy
 }
 
+
+bootRun {
+    main = "com.faforever.client.Main"
+}
+
 jar {
   manifest {
-    attributes "Main-Class": "com.faforever.client.FafClientApplication",
+    attributes "Main-Class": "com.faforever.client.Main",
         "Implementation-Title": "Downlord's FAF Client",
         "Implementation-Version": version,
         "Implementation-Vendor": "Downlord"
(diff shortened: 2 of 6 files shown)


1-alex98 (dismissed):

1-alex98:
So u ran a project wide search for the Main class and replaced it?

Katharsas:
Yeah basically. Also since i renamed the original main method, if this breaks anything, it should break pretty obviously. At least i hope so^^

Brutus5000:
Does this enable Java 11 usage?

Katharsas:
Yes. Last time i tested it worked fine with AdoptOpenJDK 10 and 11

Geosearchef (dismissed):
I don't like this, but I've stumbled over that problem myself and haven't found any other solution aswell.

Brutus5000 (commented):
This throws annoying IntelliJ Ultimate errors: `Run Configuration Error: Not a valid Spring Boot application class` (I can still run it though)

Katharsas:
Are you ok with that? Since i don't have Intellij Ultimate i cannot test what changes to run configurations do.

Can we maybe just remove the run configurations alltogether and instead put the VM parameters into the gradle file?^^ Imo profiles should generally be part of the IDE-independent gradle build anyway. I don't see a reason why we should use Intellij specific profile definitions.

1-alex98:
Well the think is the Main class is no longer a Spring Application thingy we should change that indeed

heelix:
I've got ultimate, if you are lookin g for a test monkey.

Katharsas:
Well i can just copy the `@SpringBootApplication` annotation to the new main class and HOPE that that fixes it and doesn't break anything else.

1-alex98:
Yeah Brutus might also know what is needed for a Spring Application

Katharsas:
@Brutus5000 can you check if you still get the error?

Katharsas:
@heelix You could try if you get the problem described by Brutus or if it was fixed.

Brutus5000:
@Katharsas IntelliJ errors have vanished

1-alex98 (dismissed):

Katharsas:
Same error happens on Linux as well (at least for me on Ubuntu LTS after installing JDK 11), and we don't ship a JDK with the Linux install, so im in favour of getting this merged finally.
@axel1200

Brutus5000:
I'd be in favor of merging it *after* releasing 1.0

1-alex98 (dismissed):

Katharsas:
> 
> 
> I'd be in favor of merging it _after_ releasing 1.0

I don't agree.
I was trying to fix #1462 and already would have needed this PR twice just for that issue (1. because sometimes i need to use gradle without an IDE for example for debugging and 2. because i need to test later JavaFX versions to check if they fix/change the Tooltip implementation).
Sorry, but it is becoming too frustrating to always apply this PR onto anything i do so i have stopped any work on this Repo now until this is merged.

PR #1326, opened 2019-06-29
Fix tests in Eclipse

### LINKED ISSUE #1324: Running gradle tests inside Eclipse fails
Due to bugs / problems with Spring Boot that were fixed in new minor versions of Spring Boot.

Katharsas:
- Bumps Spring boot minor version from 2.1.0-RC to 2.1.6
- makes standard output visible when running test from Eclipse or executing gradle from console (**this is not stricly necessary and i can remove this. Its just nice to be able to see logging output when tests run**).

Fixes #1324


build.gradle
```diff
@@ -122,6 +122,12 @@ jacocoTestReport {
   }
 }
 
+test {
+  testLogging {
+    showStandardStreams = true
```
1-alex98, line 127: What is this for?

gradle.properties
```diff
@@ -3,7 +3,7 @@ javafxPlatform=unspecified
 faf_ice_adapter_version=2.3.0
 faf_uid_version=4.0.4
 propdepsVersion=0.0.7
-springBootVersion=2.1.0.RC1
```
1-alex98, line 6: @micheljung any reason we used an RC here?

1-alex98 (dismissed):

gradle.properties
```diff
@@ -3,7 +3,7 @@ javafxPlatform=unspecified
...
+springBootVersion=2.1.6.RELEASE
```
Brutus5000, line 6: One does not simply switch the spring boot version to fix tests.
If you want to update dependencies, please do at least all of the spring ones at once.

Brutus5000 (changes requested):

build.gradle
Katharsas: If this is not set, anything that goes to standard outout (like System.out or logging) or standard err is not shown in the console during `gradle test`.

Maybe IntelliJ changes this internally so that you can always see logging output, but other environments don't, or when running from gradle wrapper. Note that the full logs still are available in a html report when you run `gradle test` without this, but its annoying to look those up (especially if you only run a single test class or method).

gradle.properties
Katharsas: Right, i now upped the oauth autoconfigure as well to the same version.

1-alex98:
travis fails

Brutus5000 (dismissed):

Katharsas:
My local test run fine (except the one that i fixed in the other PR of course). Maybe its Travis fault?^^

Brutus5000:
Does not look like a Travis error. Might be an async test that randomly fails?

com.faforever.client.chat.PircBotXChatServiceTest > testAddOnChatDisconnectedListener FAILED
    java.lang.AssertionError at PircBotXChatServiceTest.java:261

Katharsas:
`assertTrue(latch.await(TIMEOUT, TIMEOUT_UNIT));`
That thimeout is set to 30 seconds. seems like Travis is running REAL slow sometimes^^. Oh well one of the travis runs succeeded now. Can i re-run travis myself btw?

On the other hand i can just squash my commits and let it run again anyway.
Edit: Yeah this time it worked.

1-alex98 (approved):

Brutus5000 (approved):

PR #1327, opened 2019-06-29
Fix LaunchCommandBuilderTest for paths with space

### LINKED ISSUE #1323: LaunchCommandBuilderTest fails if project path contains space
during test method `testCommandFormat()`.

This is because (unlike the `defaultBuilder` instance in the same test) it fails to wrap the `%s` part of the format string in double quotation marks. While i could sinply add quotation marks in the test, i think it would be better idea if the `%s` would not have to be wrapped in quotation marks in the first place (the `LaunchCommandBuilder` can do that by itelf just fine), so that there is no possibility to even forget this.

It seems like the regex can properly deal with redundant double quotations which could come from exising scripts that already use double quotations (like [here](https://github.com/tatsujb/installFAFscript/blob/2e21412a41a98d79ddb71c795bb28453123923b4/faf.sh#L750)), probably best to specifically add a unit test for that.

Edit: Looks like it cannot properly deal with triple double quotation marks.
Edit2: Ok looks like fixing that is pretty easy.

Katharsas:
- Changes executableDecorator argument string to not need to wrap executable placeholder with quotation marks
- Changed regex so that redundant quotation marks are not a problem for that placeholder or any other part of the string

Fixes #1323

### DIFF
--- a/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
+++ b/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
@@ -15,7 +15,7 @@ import java.util.regex.Pattern;
 
 public class LaunchCommandBuilder {
 
-  private static final Pattern QUOTED_STRING_PATTERN = Pattern.compile("([^\"]\\S*|\".+?\")\\s*");
+  private static final Pattern QUOTED_STRING_PATTERN = Pattern.compile("([^\"]\\S*|\"+.+?\"+)\\s*");
 
   private Float mean;
   private Float deviation;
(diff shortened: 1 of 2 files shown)


Katharsas:
Whats going on with Travis? I DID create three pull requests at the same time, maybe it timed out because of that?

1-alex98:
We have an issue with travis some builds time out for unknown reasons but work if restarted. Tho I can not say which test times out from the log so i will need to turn up log level for that...

1-alex98 (approved):
Looks fine for me. But would like somebody to test it on Linux

Katharsas:
@tatsujb Maybe?

1-alex98:
@Geosearchef also has ide set up on linux

Brutus5000 (approved):

PR #1329, opened 2019-06-29
Remove background by default

1-alex98:


Katharsas:
I guess its not worth the trouble to migrate existing preferences from `*useNoImage` to empty string?

1-alex98:
no... and it would make no difference

Brutus5000 (approved):

PR #1335, opened 2019-07-02
Add channel topics

### LINKED ISSUE #1234: add back the channel topic
so TD can actually ping stuff in the channel (eg : challonge link, streams etc)

1-alex98:
Fix #1234

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -20,3 +20,4 @@ coverallsGradlePluginVersion=2.8.2
 jnaVersion=5.0.0
 jacocoVersion=0.8.2
 discordRpcVersion=1.6.2-jna
+install4jHomeDir=C:\\Program Files\\install4j7
--- a/gradle.properties
+++ b/gradle.properties
@@ -20,4 +20,3 @@ coverallsGradlePluginVersion=2.8.2
 jnaVersion=5.0.0
 jacocoVersion=0.8.2
 discordRpcVersion=1.6.2-jna
-install4jHomeDir=C:\\Program Files\\install4j7
--- a/src/main/java/com/faforever/client/chat/ChannelTabController.java
+++ b/src/main/java/com/faforever/client/chat/ChannelTabController.java
@@ -1,7 +1,9 @@
 package com.faforever.client.chat;
 
 import com.faforever.client.audio.AudioService;
+import com.faforever.client.chat.event.ChannelTopicChangedEvent;
 import com.faforever.client.fx.JavaFxUtil;
+import com.faforever.client.fx.PlatformService;
 import com.faforever.client.fx.WebViewConfigurer;
 import com.faforever.client.i18n.I18n;
 import com.faforever.client.notification.NotificationService;
@@ -34,6 +36,8 @@ import javafx.event.ActionEvent;
 import javafx.event.Event;
 import javafx.geometry.Bounds;
 import javafx.scene.control.Button;
+import javafx.scene.control.Hyperlink;
+import javafx.scene.control.Label;
 import javafx.scene.control.ListView;
 import javafx.scene.control.SplitPane;
 import javafx.scene.control.Tab;
@@ -45,6 +49,7 @@ import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.HBox;
 import javafx.scene.layout.VBox;
 import javafx.scene.paint.Color;
+import javafx.scene.text.TextFlow;
 import javafx.scene.web.WebView;
 import javafx.stage.Popup;
 import javafx.stage.PopupWindow;
@@ -65,6 +70,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static com.faforever.client.chat.ChatColorMode.DEFAULT;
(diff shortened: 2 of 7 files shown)


Katharsas:
It looks a bit weird for me, since it moves down the user list:
chat
But this is nothing serious, could live with that^^
Maybe make font size same as chat? Seems bigger (maybe its just the spacing idk).

1-alex98:
It was supposed to move down user list.... I can also make it not to

1-alex98:
The text has the standard size of the client

Katharsas:
Ok. Maybe somebody should just clean up the topic message itself. One single link to one page which has links to everything else should be fine. But that is not related to this issue.

1-alex98:
I also said to @PhilipJFryFAF we will only enable it if it is not misused for links

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -235,6 +252,35 @@ public void setChannel(Channel channel) {
...
+    JavaFxUtil.addListener(channel.topicProperty(), new WeakInvalidationListener(channelTopicListener));
+  }
+
+  private void updateChannelTopic() {
+    boolean hasTopic = channel.getTopic() != null;
+    topicPane.setVisible(hasTopic);
+    if (!hasTopic) {
+      return;
+    }
+    String topic = channel.getTopic() == null ? "" : channel.getTopic();
```
Katharsas: Unnecessary check? If it was null, it would have returned above i think.

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -235,6 +252,35 @@ public void setChannel(Channel channel) {
     searchFieldContainer.visibleProperty().bind(searchField.visibleProperty());
     closeSearchFieldButton.visibleProperty().bind(searchField.visibleProperty());
     addSearchFieldListener();
+    topicPane.managedProperty().bind(topicPane.visibleProperty());
+    updateChannelTopic();
+    JavaFxUtil.addListener(channel.topicProperty(), new WeakInvalidationListener(channelTopicListener));
+  }
+
+  private void updateChannelTopic() {
+    boolean hasTopic = channel.getTopic() != null;
```
Katharsas: I don't think there is a way to set a topic to `null` from the IRC side, only to empty string (?), so i suspect that an empty string should also be interpreted as "no topic".
`channel.getTopic()` returns the string from the `TopicEvent`, whose [JavaDoc ](http://pircbotx.github.io/pircbotx/latest/apidocs/org/pircbotx/hooks/events/TopicEvent.html)is not clear on this.

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -235,6 +252,35 @@ public void setChannel(Channel channel) {
...
+            link.setOnAction(event -> platformService.showDocument(word));
+            topicText.getChildren().add(link);
+          } else {
+            topicText.getChildren().add(new Label(word + " "));
+          }
+        });
+  }
+
+  @Subscribe
+  public void onChannelTopicChangedEvent(ChannelTopicChangedEvent channelTopicChangedEvent) {
```
Katharsas: What is this needed for? I see no code that actually sends these topic events, and you should already get all topic changes through the `Channel` objects themselves (which in turn get their update from IRC `TopicEvent`), right?

Katharsas (changes requested):

gradle.properties
```diff
@@ -19,4 +19,5 @@ javafxVersion=11
 coverallsGradlePluginVersion=2.8.2
 jnaVersion=5.0.0
 jacocoVersion=0.8.2
-discordRpcVersion=1.6.2-jna
\ No newline at end of file
+discordRpcVersion=1.6.2-jna
+install4jHomeDir=C:\\Program Files\\install4j7
```
Wesmania: Leftover local setting?

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -75,6 +81,10 @@
 @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
 public class ChannelTabController extends AbstractChatTabController {
   private static final String USER_CSS_CLASS_FORMAT = "user-%s";
+  /**
+   * Taken from https://stackoverflow.com/questions/163360/regular-expression-to-match-urls-in-java
+   */
+  private static final Pattern URL_REGEX_PATTERN = Pattern.compile("^(https?|ftp|file)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]");
```
Wesmania: Side note, py client has this stupid feature that allows people to post links to games by starting them with `faflive://` or `fafgame://`. Not advocating its return, but it's a cool piece of trivia.

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -235,6 +249,35 @@ public void setChannel(Channel channel) {
...
+  private void updateChannelTopic() {
+    boolean hasTopic = channel.getTopic() != null;
+    topicPane.setVisible(hasTopic);
+    if (!hasTopic) {
+      return;
+    }
+    String topic = channel.getTopic() == null ? "" : channel.getTopic();
+    Arrays.stream(topic.split("\\s"))
+        .forEach(word -> {
+          if (URL_REGEX_PATTERN.matcher(word).matches()) {
```
Wesmania, line 264: Is there any code that finds URLs already? Does chat area do that?

gradle.properties
1-alex98: outdated

src/main/java/com/faforever/client/chat/ChannelTabController.java
1-alex98: ehm how does this relate?

src/main/java/com/faforever/client/chat/ChannelTabController.java
1-alex98: yes but it is done in Java Script

1-alex98: moved it to PlatformService

src/main/java/com/faforever/client/chat/ChannelTabController.java
Wesmania: Just wanted to mention it, because it's really useless and nice ^^

Katharsas:
Yeah looks fine to me now.

PR #1364, opened 2019-07-25
Display ladder queue pop time, fix ladder notifications

### LINKED ISSUE #1361: Display ladder queue pop time / Fix ladder 1v1 available match popup
The last server update (v1.1) changed the matchmaking system such that now matchmaking happens on a scheduled interval, instead of whenever a player starts searching.

We should add timer somewhere on the ladder1v1 tab that counts down the time until the next pop so people know when it's going to happen. This should help remove some of the mystery behind ranked mode. 

It also means that the logic for sending a "Ladder match available" notification needs to be adjusted a little bit so that you don't constantly get spammed by notifications while sitting in queue.

The `matchmaker_info` message now contains an additional field which is the UTC time (in ISO format) of the next queue pop:
```python
    {
        'command': 'matchmaker_info',
        'queues': [
            {
                'queue_name': 'ladder1v1',
                'queue_pop_time': '2019-07-01T16:53:21+00:00',
                'boundary_80s': [],
                'boundary_75s': []
            }
        ]
    }
```

Katharsas:
Fixes #1361

Note:
~Could maybe somebody else make this a bit less ugly (spacing and positioning of the timer UI label)? My Scenebuilder doesn't work right now and i am kinda out of the loop with fxml. I tried to put the timer next to the play button~ (fixed)


Katharsas:
And we still need to figure out if the server can just send the first event on login (will the Ladder1v1Controller class always be instantiated by then?) or if the client must request it.

This PR should disable any ladder match notifications while player is searching, too.

Edit: Not sure about text. "Next round" or "Next match" could be a good description.

@Askaholic

Katharsas:
This should be good now. I opted for no text unless somebody wants one.
ladder_timer
(also shown when not searching, unless logged in recently)

src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
```diff
@@ -155,6 +167,39 @@ public void initialize() {
...
+        }
+      }
+    });
+    
+    Timeline queuePopTimeUpdater = new Timeline(new KeyFrame(javafx.util.Duration.seconds(1), (ActionEvent event) -> {
+      if (nextQueuePopTime != null) {
+        Instant now = Instant.now();
+        Duration timeUntilPopQueue = Duration.between(now, nextQueuePopTime);
+        if (!timeUntilPopQueue.isNegative()) {
+          String formatted = String.format("%2d:%02d",
```
1-alex98: This format string needs to go into message.properties and it needs to use different format(Chinese time is for example not displayed with : )

1-alex98 (changes requested):

src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
Katharsas: Ok. Java does not provide locale based duration formatting, thats why this is a string formatter and not a time formatter, and why i assumed that its useless to try to format this per locale. So i guess for chinese durations only the numbers and maybe separator change, but not the minutes/seconds structure of  duration?

Katharsas: And if we are dealing with i18n anyway, i may add a "Next round: " label as well. Btw. How would a translator know that `%2d:%02d` is supposed to be `minutes : seconds`, should I comment this in the properties file?
Edit: maybe i can give a clue through the message key name.

1-alex98: No u can do proper formating in the message.properities file  as well. 
image Do something like that

1-alex98: then do `i18n.get("timeFormat", minutes, seconds);`

Katharsas: Where can i read about this syntax?
Edit: Nevermind, found it.

Katharsas:
ladder_timer_2

src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
```diff
@@ -155,6 +167,39 @@ public void initialize() {
 
     JavaFxUtil.addListener(playerService.currentPlayerProperty(), (observable, oldValue, newValue) -> Platform.runLater(() -> setCurrentPlayer(newValue)));
     playerService.getCurrentPlayer().ifPresent(this::setCurrentPlayer);
+    
+    gameService.addOnRankedMatchNotificationListener(message -> {
+      if (message.getQueues() == null) {
+        return;
+      }
+      for (MatchmakerMessage.MatchmakerQueue matchmakerQueue : message.getQueues()) {
+        if (!Objects.equals("ladder1v1", matchmakerQueue.getQueueName())) {
```
Brutus5000: put "ladder1v1" into a constant

src/main/java/com/faforever/client/rankedmatch/MatchmakerMessage.java
```diff
@@ -12,13 +12,16 @@
   public static class MatchmakerQueue {
 
     private String queueName;
+    @SerializedName("queue_pop_time")
+    private String queuePopTime;
```
Brutus5000, line 15: Is this really a string? I assume it's a timestamp here? So Gson should be able to convert it into some java.date class? (I only know how to do it with Jackson)

src/main/resources/i18n/messages.properties
```diff
@@ -78,6 +78,7 @@ ranked1v1.play=Play
 ranked1v1.cancel=Cancel
 ranked1v1.choseFactions.title=Select factions
 ranked1v1.choseFactions.description=When the match starts, one of your selected factions will be assigned to you.
+ranked1v1.queuePopTimer=Next round in {0,number,0}:{1,number,00}
```
Brutus5000: Isn't it "next matching" ?

src/test/java/com/faforever/client/remote/ServerAccessorImplTest.java
```diff
@@ -206,6 +206,8 @@ public void testConnectAndLogIn() throws Exception {
     assertThat(result.getMessageType(), is(FafServerMessageType.WELCOME));
     assertThat(result.getId(), is(playerUid));
     assertThat(result.getLogin(), is(username));
+
+    instance.disconnect();
```
Brutus5000, line 212: I don't know what this does (especially in a test), but when you expect it to do something you should assert or verify it. (Also applies to all following calls)

Brutus5000 (changes requested):
I was not able to run it yet, as I can't run it on MacOs right now. More stuff might follow.

src/test/java/com/faforever/client/remote/ServerAccessorImplTest.java
Katharsas: This just cleans up after the test, not really part of the test itself (if missing, the teardown method killing the sockets will cause an EOFException). And I cant put it in teardown because its only needed in some tests (those that connect at the beginning).

src/main/resources/i18n/messages.properties
Katharsas: Maybe thats better, gonna ask around a bit. Gonna put that in if nobody objects.

src/main/java/com/faforever/client/rankedmatch/MatchmakerMessage.java
Katharsas: Well its a fromatted ISO datetime. Don't know if gson can do that automatically.

src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
Katharsas: Ok  it think MainController uses this string as well, will replace it there aswell.

src/main/resources/i18n/messages.properties
Katharsas: So we (kinda) settled on `Matching in x:xx` without the "Next".

src/main/java/com/faforever/client/rankedmatch/MatchmakerMessage.java
Katharsas: Ok i just looked it up, GSON cannot deal with java.time by default, so we would have to use something like https://github.com/gkopff/gson-javatime-serialisers or use our own serializer. Don't know if its worth it for one occurence.

Geosearchef: You can afaik configure gson when using a builder for it to configure how ro serialize specific classes.

src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
Katharsas: I think i can make this an enum, GSON can apparently deserialize straight into an enum. Gonna put it into MatchmakerQueue.java i think.

src/main/java/com/faforever/client/rankedmatch/MatchmakerMessage.java
1-alex98: I think our gson has somthing that can deserialize dates....

src/main/java/com/faforever/client/rankedmatch/MatchmakerMessage.java
```diff
@@ -11,26 +11,42 @@
 
   public static class MatchmakerQueue {
 
-    private String queueName;
+    private QueueName queueName;
+    @SerializedName("queue_pop_time")
```
1-alex98: Remove this it is not needed due to FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES set with Gson

src/main/java/com/faforever/client/rankedmatch/MatchmakerMessage.java
1-alex98: ok we did not but maybe we should set some TypeAdapter

1-alex98 (changes requested):

src/main/java/com/faforever/client/rankedmatch/MatchmakerMessage.java
Katharsas: There is `LocalDate`, `LocalDateTime`, `OffsetDateTime`, `ZonedDateTime` and `Instant` in `java.time` (at least). And there is no straight-forward 1 to 1 mapping between these times and date formats. So you would at least need to be able to specify the format for every `java.time` member.
Or you need to assume that every future JSON object will come in the exact same datetime format (ISO), which imo is a questionable assumption.

So, im not convinced it is worth it to think about any of this. String gives full control and full resposibility to the reader of that DTO to know the exact format and that might be a good thing.

1-alex98: Well ok

src/main/java/com/faforever/client/rankedmatch/MatchmakerMessage.java
Katharsas: kk

Askaholic:
Maybe you can test this out on the test server and see if it works. https://github.com/FAForever/server/pull/489

1-alex98 (approved):

1-alex98:
We did ;)

1-alex98:
@Askaholic

1-alex98:
oh yeah but not this PR...

1-alex98 (approved):

1-alex98:
If you run this version on the test server with the newly deployed py server that send the remaining time straight away, the following happens. 

```
2019-08-29 20:49:51.644  WARN 12668 --- [      Thread-12] c.f.c.remote.AbstractServerAccessor      : Error while handling server message: {"command":"matchmaker_info","queues":[{"queue_name":"ladder1v1","queue_pop_time":"2019-08-29T18:52:18.246558+00:00","boundary_80s":[],"boundary_75s":[]}]}
```

```

java.lang.NullPointerException: null
	at com.faforever.client.main.MainController.onMatchmakerMessage(MainController.java:334) ~[classes/:na]
	at com.faforever.client.remote.FafServerAccessorImpl.lambda$parseServerObject$1(FafServerAccessorImpl.java:437) ~[classes/:na]
	at java.base/java.lang.Iterable.forEach(Iterable.java:75) ~[na:na]
	at com.faforever.client.remote.FafServerAccessorImpl.parseServerObject(FafServerAccessorImpl.java:437) ~[classes/:na]
	at com.faforever.client.remote.FafServerAccessorImpl.onServerMessage(FafServerAccessorImpl.java:406) ~[classes/:na]
	at com.faforever.client.remote.AbstractServerAccessor.blockingReadServer(AbstractServerAccessor.java:44) ~[classes/:na]
	at com.faforever.client.remote.FafServerAccessorImpl$1.call(FafServerAccessorImpl.java:252) [classes/:na]
	at com.faforever.client.remote.FafServerAccessorImpl$1.call(FafServerAccessorImpl.java:224) [classes/:na]
	at javafx.graphics/javafx.concurrent.Task$TaskCallable.call(Task.java:1425) [javafx.graphics:na]
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264) [na:na]
	at javafx.graphics/javafx.concurrent.Service.lambda$executeTask$6(Service.java:725) [javafx.graphics:na]
	at java.base/java.security.AccessController.doPrivileged(Native Method) ~[na:na]
	at javafx.graphics/javafx.concurrent.Service.lambda$executeTask$7(Service.java:724) [javafx.graphics:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1135) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[na:na]
	at java.base/java.lang.Thread.run(Thread.java:844) ~[na:na]

```

PR #1378, opened 2019-08-02
Reduce unnecessary unit test log spam

### LINKED ISSUE #1377: Unit tests log spam
Unit tests spam the logs.
- A significant part of the spam seems to come from `MapServiceTest`. It throws an exception for every official map in the game for every test in that class because the `MapService` cannot find any of those maps whenever it is initialized (which happens in the setup of every map).
- Quite a bit of stuff comes from chat related tests
- Sometimes i18n methods aren't mocked properly and throw exceptions in UI thread (which may not cause tests to fail).
- Some EOFExceptions are fixed by https://github.com/FAForever/downlords-faf-client/pull/1364
- Chat tests: see https://github.com/FAForever/downlords-faf-client/issues/1391

### LINKED ISSUE #1379: Wrong player stats shown for "Games played"
The function `plotGamesPlayedChart` in `UserInfoWindowController` uses the wrong player object (it uses `playerService.getCurrentPlayer()` instead of its own `player` class member). This results in the "Games played" statistics always showing the games of the current player, not of the selected player.

### LINKED ISSUE #1385: AsyncConfig's destroy method fails on shutdown
Apparently because it tries to use a bean method during shutdown (?).
Log when not silent (which logger is by default):
```
21:00:31.821 [JavaFX Application Thread] INFO org.springframework.beans.factory.support.DisposableBeanAdapter - Invocation of destroy method failed on bean with name 'asyncConfig'
org.springframework.beans.factory.BeanCreationNotAllowedException: Error creating bean with name 'taskExecutor': Singleton bean creation not allowed while singletons of this factory are in destruction (Do not request a bean from a BeanFactory in a destroy method implementation!)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:208)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:318)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:199)
	at org.springframework.context.annotation.ConfigurationClassEnhancer$BeanMethodInterceptor.resolveBeanReference(ConfigurationClassEnhancer.java:394)
	at org.springframework.context.annotation.ConfigurationClassEnhancer$BeanMethodInterceptor.intercept(ConfigurationClassEnhancer.java:366)
	at com.faforever.client.config.AsyncConfig$$EnhancerBySpringCGLIB$$e87c2cec.taskExecutor(<generated>)
	at com.faforever.client.config.AsyncConfig.destroy(AsyncConfig.java:55)
	at org.springframework.beans.factory.support.DisposableBeanAdapter.destroy(DisposableBeanAdapter.java:258)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.destroyBean(DefaultSingletonBeanRegistry.java:571)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.destroySingleton(DefaultSingletonBeanRegistry.java:543)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.destroySingleton(DefaultListableBeanFactory.java:1036)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.destroySingletons(DefaultSingletonBeanRegistry.java:504)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.destroySingletons(DefaultListableBeanFactory.java:1029)
	at org.springframework.context.support.AbstractApplicationContext.destroyBeans(AbstractApplicationContext.java:1057)
	at org.springframework.context.support.AbstractApplicationContext.doClose(AbstractApplicationContext.java:1026)
	at org.springframework.context.support.AbstractApplicationContext.close(AbstractApplicationContext.java:975)
	at com.faforever.client.config.FafClientApplicationTest.lambda$0(FafClientApplicationTest.java:36)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:514)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:428)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:427)
	at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
	at com.sun.glass.ui.win.WinApplication._runLoop(Native Method)
	at com.sun.glass.ui.win.WinApplication.lambda$runLoop$3(WinApplication.java:174)
	at java.base/java.lang.Thread.run(Thread.java:844)
```

### LINKED ISSUE #1386: DownloadMapBeanTaskTest fails when run alone
When you ONLY run this test, `testCall` fails. Test should probably extend `AbstractPlainJavaFxTest`.
Seems like JavaFX ToolKit state is global across all test classes / not reset for for every test class (this may be a TestFx issue).

Katharsas:
### MapService
I simply made the list of official maps mockable. I see no reason for them to be an enum, in fact being an immutable list seems to make the code simpler (it removes the need for the weird string-to-enum-map).

However i would like for that list to be final. To make that happen, the maps would need to be stored in a dependency of `MapService` (a good place is probably `ForgedAlliancePrefs`), so that they can be properly Mocked. But not sure about this, opinions wanted. I have not done that for now.
(see #1377)

### Deliberate Exceptions
Added new class `FakeTestException` that works like a normal exception but doesn't print a stacktrace to replace the `new Exception("fake test exception")` lines.

### Chat tests
There is something really interesting/fishy going on in test `testOnChatMessageUnfocusedTriggersNotification` in class `PrivateChatTabControllerTest` (see `TODO`). The test only throws exceptions when other tests of that class run as well. Help wanted.

### Threadpool cleanup
The `AsyncConfig` class tried to shutdown one of the two threadpool singletons with a `destroy` method which failed (only visible during `FeaturedModUpdaterConfigTest`, silent during normal execution) before shutdown (see #1385). This now fixed and the test class was renamed.

---
Fixes #1377 
Fixes #1379 (this bug increased log spam, and fix was easy)
Fixes #1385
Fixes #1386

### DIFF
--- a/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
+++ b/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
@@ -63,7 +63,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
-import java.util.concurrent.CompletionException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
--- a/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
+++ b/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
@@ -18,6 +18,7 @@ import com.faforever.client.theme.UiService;
 import com.faforever.client.ui.StageHolder;
 import com.faforever.client.uploader.ImageUploadService;
 import com.faforever.client.user.UserService;
+import com.faforever.client.util.ConcurrentUtil;
 import com.faforever.client.util.IdenticonUtil;
 import com.faforever.client.util.TimeService;
 import com.google.common.annotations.VisibleForTesting;
@@ -421,7 +422,7 @@ public abstract class AbstractChatTabController implements Controller<Tab> {
       messageTextField.setDisable(false);
       messageTextField.requestFocus();
     }).exceptionally(throwable -> {
-      throwable = throwable instanceof CompletionException ? throwable.getCause() : throwable;
+      throwable = ConcurrentUtil.unwrapIfCompletionException(throwable);
       logger.warn("Message could not be sent: {}", text, throwable);
       notificationService.addNotification(new ImmediateErrorNotification(
           i18n.get("errorTitle"), i18n.get("chat.sendFailed"), throwable, i18n, reportingService)
@@ -443,7 +444,7 @@ public abstract class AbstractChatTabController implements Controller<Tab> {
           messageTextField.requestFocus();
         })
         .exceptionally(throwable -> {
-          throwable = throwable instanceof CompletionException ? throwable.getCause() : throwable;
+          throwable = ConcurrentUtil.unwrapIfCompletionException(throwable);
           // TODO onDisplay error to user somehow
           logger.warn("Message could not be sent: {}", text, throwable);
           messageTextField.setDisable(false);
--- a/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
+++ b/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
@@ -62,6 +62,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.CompletionException;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -420,6 +421,7 @@ public abstract class AbstractChatTabController implements Controller<Tab> {
       messageTextField.setDisable(false);
       messageTextField.requestFocus();
     }).exceptionally(throwable -> {
+      throwable = throwable instanceof CompletionException ? throwable.getCause() : throwable;
       logger.warn("Message could not be sent: {}", text, throwable);
       notificationService.addNotification(new ImmediateErrorNotification(
           i18n.get("errorTitle"), i18n.get("chat.sendFailed"), throwable, i18n, reportingService)
@@ -441,6 +443,7 @@ public abstract class AbstractChatTabController implements Controller<Tab> {
           messageTextField.requestFocus();
         })
         .exceptionally(throwable -> {
+          throwable = throwable instanceof CompletionException ? throwable.getCause() : throwable;
           // TODO onDisplay error to user somehow
           logger.warn("Message could not be sent: {}", text, throwable);
           messageTextField.setDisable(false);
--- a/src/main/java/com/faforever/client/chat/UserInfoWindowController.java
+++ b/src/main/java/com/faforever/client/chat/UserInfoWindowController.java
@@ -10,7 +10,6 @@ import com.faforever.client.api.dto.PlayerEvent;
 import com.faforever.client.domain.RatingHistoryDataPoint;
 import com.faforever.client.events.EventService;
 import com.faforever.client.fx.Controller;
-import com.faforever.client.fx.JavaFxUtil;
 import com.faforever.client.fx.OffsetDateTimeCell;
 import com.faforever.client.game.KnownFeaturedMod;
 import com.faforever.client.i18n.I18n;
@@ -278,7 +277,6 @@ public class UserInfoWindowController implements Controller<Node> {
         });
   }
 
-  @SuppressWarnings("unchecked")
   private void plotFactionsChart(Map<String, PlayerEvent> playerEvents) {
     int aeonPlays = playerEvents.containsKey(EVENT_AEON_PLAYS) ? playerEvents.get(EVENT_AEON_PLAYS).getCurrentCount() : 0;
     int cybranPlays = playerEvents.containsKey(EVENT_CYBRAN_PLAYS) ? playerEvents.get(EVENT_CYBRAN_PLAYS).getCurrentCount() : 0;
@@ -307,7 +305,6 @@ public class UserInfoWindowController implements Controller<Node> {
     Platform.runLater(() -> factionsChart.getData().addAll(winsSeries, lossSeries));
   }
 
-  @SuppressWarnings("unchecked")
   private void plotUnitsByCategoriesChart(Map<String, PlayerEvent> playerEvents) {
     int airBuilt = playerEvents.containsKey(EVENT_BUILT_AIR_UNITS) ? playerEvents.get(EVENT_BUILT_AIR_UNITS).getCurrentCount() : 0;
     int landBuilt = playerEvents.containsKey(EVENT_BUILT_LAND_UNITS) ? playerEvents.get(EVENT_BUILT_LAND_UNITS).getCurrentCount() : 0;
@@ -320,7 +317,6 @@ public class UserInfoWindowController implements Controller<Node> {
     )));
   }
 
-  @SuppressWarnings("unchecked")
   private void plotTechBuiltChart(Map<String, PlayerEvent> playerEvents) {
     int tech1Built = playerEvents.containsKey(EVENT_BUILT_TECH_1_UNITS) ? playerEvents.get(EVENT_BUILT_TECH_1_UNITS).getCurrentCount() : 0;
     int tech2Built = playerEvents.containsKey(EVENT_BUILT_TECH_2_UNITS) ? playerEvents.get(EVENT_BUILT_TECH_2_UNITS).getCurrentCount() : 0;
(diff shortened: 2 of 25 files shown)


src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -420,6 +421,7 @@ private void sendMessage() {
       messageTextField.setDisable(false);
       messageTextField.requestFocus();
     }).exceptionally(throwable -> {
+      throwable = throwable instanceof CompletionException ? throwable.getCause() : throwable;
```
Brutus5000: Amen. This bothered me for ages :D

Brutus5000: I believe this could be used in other services as well. Maybe make it a method in some Util class? If we have one?
In this class it's already used twice.

Brutus5000 (changes requested):
The official maps are constant. Don't put them into a separate class, they belong into the MapService. If you need to make the variable final, then declare it in the constructor. I see no need that this has to be mocked away? It's constant! If there is an issue with the tests, the tests have the issue, not the list.

Please fix codacy issues.

You add some more mocks to some tests. Why? The tests are not failing without it, so it must be unnecessary or the test is lacking asserts.

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
Katharsas: Would it really be worth it to extract a single, very simple ternary operator? The assignment will stay anyway and you will need to jump into a util function when reading just to understand this simple line of code.

Brutus5000: Well, from all the stacktraces I have seen so far, we should always unwrap the CompletionException and only occasionally do it, so I think a function called `unwrapCompletionException` would be used in many places and is easier to read than the ternary operator. But that is just my opinion.

Katharsas: k, i guess `ConcurrentUtil` i a good class for that method.

Katharsas:
@Brutus5000 
I am trying to make the log for a full test run usable / readable. Right now its not usable, because it contains thousands of exceptions. These exceptions are thrown by tests sometimes because for example a certain method is not probably mocked or because some files do not exist that would exist in a normal client launch but not in the test.

So when i add any mocks in this PR, it actually fixes exceptions that are printed to the log but do not influence test failure / success. About the 'value' of fixing test exceptions (and thus being able to read the log) even if they don't affect test success:
Since i have started this work, i have already uncovered two real bugs that happen during normal client execution, and im not even halfway through all the tests yet (i'm working through the log output of all tests).
Btw i think some of the exceptions only don't cause tests to fail because they are not thrown in the test thread itself. They would fail some tests if they were thrown in the same thread.

About why i need to be able to mock official maps (which means that i cannot use constructor assignment as long as the maps are not passed into the constructor), see https://github.com/FAForever/downlords-faf-client/issues/1377
I don't think the problem is inside `MapServiceTest`. Loading official maps is something that the `MapService` does on initialization and thats why it also happens in tests, unless we somehow mock the filesystem contents, which i don't think is feasable.

src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -198,8 +200,8 @@ protected Void call() {
         try (Stream<Path> customMapsDirectoryStream = list(customMapsDirectory)) {
           List<Path> mapPaths = new ArrayList<>();
           customMapsDirectoryStream.collect(toCollection(() -> mapPaths));
-          Arrays.stream(OfficialMap.values())
-              .map(map -> officialMapsPath.resolve(map.name()))
+          officialMaps.stream()
+              .map(mapName -> officialMapsPath.resolve(mapName))
```
Brutus5000: This can be a function reference now

src/main/java/com/faforever/client/config/AsyncConfig.java
```diff
@@ -46,9 +47,18 @@ public ExecutorService taskExecutor() {
   public TaskScheduler taskScheduler() {
     return new ThreadPoolTaskScheduler();
   }
-
-  @Override
-  public void destroy() {
-    taskExecutor().shutdownNow();
+  
+  @Bean
+  public DestructionAwareBeanPostProcessor threadPoolShutdownProcessor() {
```
Brutus5000, line 55: Maybe some logging in this method?

Brutus5000 (changes requested):

src/main/java/com/faforever/client/config/AsyncConfig.java
Katharsas: ha, i had logging and thought "meh, who cares about shutdown" so i removed it xD
Can put it back in of course.

Katharsas:
Progress report: We are down to 10k lines from 20k xD
We should be able to get under 3k i hope.

Katharsas:
@Brutus5000
There are some new code changes that shoud be reviewed as well (last 5 commits)
There is a single Codacy problem left (empty catch clause) which does not really apply.

Katharsas:
Additional Issues/PRs that fix log spam (not fixed in this PR):
- https://github.com/FAForever/downlords-faf-client/pull/1364 cleans up some EOFExceptions in ServerAccessorImplTest
- https://github.com/FAForever/downlords-faf-client/pull/1390 fixes IllegalStateExceptions
- chat tests create a lot of spam (multiple thousand lines), see https://github.com/FAForever/downlords-faf-client/issues/1391

There seems to be a lot of exception spam that **only happens in a Travis test run**, but not in a local run. Tons of these appear in the Travis log:
```
WARN com.faforever.client.remote.FafServerAccessorImpl - Lost connection to FAF server, trying to reconnect in 3s
    java.net.ConnectException: Connection refused (Connection refused)
    	at java.base/java.net.PlainSocketImpl.socketConnect(Native Method)
    	at java.base/java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:400)
    	at java.base/java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:243)
    	at java.base/java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:225)
    	at java.base/java.net.SocksSocketImpl.connect(SocksSocketImpl.java:402)
    	at java.base/java.net.Socket.connect(Socket.java:591)
    	at java.base/java.net.Socket.connect(Socket.java:540)
    	at java.base/java.net.Socket.<init>(Socket.java:436)
    	at java.base/java.net.Socket.<init>(Socket.java:213)
    	at com.faforever.client.remote.FafServerAccessorImpl$1.call(FafServerAccessorImpl.java:238)
    	at com.faforever.client.remote.FafServerAccessorImpl$1.call(FafServerAccessorImpl.java:225)
    	at javafx.concurrent.Task$TaskCallable.call(Task.java:1425)
    	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
    	at javafx.concurrent.Service.lambda$executeTask$6(Service.java:725)
    	at java.base/java.security.AccessController.doPrivileged(Native Method)
    	at javafx.concurrent.Service.lambda$executeTask$7(Service.java:724)
    	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1135)
    	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
    	at java.base/java.lang.Thread.run(Thread.java:844)
```
Cannot reproduce these.

src/main/java/com/faforever/client/chat/UserInfoWindowController.java
```diff
@@ -333,18 +329,16 @@ private void plotTechBuiltChart(Map<String, PlayerEvent> playerEvents) {
...
       int ladderGamesCount = leaderboardEntryBean.getGamesPlayed();
-      int custonGamesCount = currentPlayer.getNumberOfGames();
+      int custonGamesCount = player.getNumberOfGames();
       Platform.runLater(() -> gamesPlayedChart.setData(FXCollections.observableArrayList(
           new PieChart.Data(i18n.get("stats.custom"), custonGamesCount),
           new PieChart.Data(i18n.get("stats.ranked1v1"), ladderGamesCount)
       )));
     })).exceptionally(throwable -> {
-      log.warn("Leaderboard entry could not be read for current player: " + currentPlayer.getUsername(), throwable);
+      log.warn("Leaderboard entry could not be read for player: " + player.getUsername(), throwable);
```
Brutus5000, line 341: Better usage would be
`log.warn("Leaderboard entry could not be read for player: {}", player.getUsername(), throwable);`

Brutus5000 (approved):

PR #1390, opened 2019-08-12
Fail tests on property listener exceptions

### LINKED ISSUE #1389: Exceptions from property listener code are ignored
1. Create a JavaFX property in a `@Test`
2. Add change listener which throws exception
3. Change the property
(everything happens in the same thread)

The test will succeed even though it should not. If you throw a `new Error()` instead of an Exception, the test fails like it should.

This is because of the following code in `com.sun.javafx.binding.ExpressionHelper` which is responsible for calling the change listeners:
```
try {
    curChangeList[i].changed(observable, oldValue, currentValue);
} catch (Exception e) {
    Thread.currentThread().getUncaughtExceptionHandler().uncaughtException(Thread.currentThread(), e);
}
```
which seems to cause tests to just print the exceptions and then continue as if nothing happened (does real application behave in the same way? that would be bad).

This makes it impossible to actually catch the Exception and validate that it was thrown, and it makes it impossible to clean up the test log by catching expected exceptions.

Edit:
The problem is described [here](https://lestard.eu/2015/exception_in_listener/) with a solution:
`Thread.currentThread().setUncaughtExceptionHandler((thread, exception) -> fail("Exception was thrown", exception));`

~~However we should still make sure that the actual client does not have this problem.~~
Edit 2:
The client has the same "problem": Adding the following code
```
BooleanProperty prop = new SimpleBooleanProperty(true);
prop.addListener((a,b, c) -> {
    throw new IllegalStateException("HI");
});
prop.set(false);
```
anywhere will not cause the client to crash, no matter which thread throws this.

Katharsas:
Causes tests to always fail if they call code that throws exceptions in the same thread (without this PR, property change listener code does not).

Fixes (only for tests!) #1389

### DIFF
--- a/src/test/java/com/faforever/client/test/AbstractPlainJavaFxTest.java
+++ b/src/test/java/com/faforever/client/test/AbstractPlainJavaFxTest.java
@@ -32,6 +32,9 @@ public abstract class AbstractPlainJavaFxTest extends ApplicationTest {
 
   public AbstractPlainJavaFxTest() {
     root = new Pane();
+    Thread.currentThread().setUncaughtExceptionHandler((thread, e) -> {
+      throw (RuntimeException) e;
+    });
   }
 
   @Override
--- a/src/test/java/com/faforever/client/chat/ChannelTabControllerTest.java
+++ b/src/test/java/com/faforever/client/chat/ChannelTabControllerTest.java
@@ -153,8 +153,8 @@ public class ChannelTabControllerTest extends AbstractPlainJavaFxTest {
   public void testSetChannelTopic() {
     Channel channel = new Channel("name");
     channel.setTopic("topic https://example.com/1");
+    
     Platform.runLater(() -> instance.setChannel(channel));
-
     WaitForAsyncUtils.waitForFxEvents();
 
     verify(chatService).addUsersListener(eq("name"), any());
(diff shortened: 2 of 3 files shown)


Katharsas:
Codacy complains about catch again, shouldnt be an issue

src/test/java/com/faforever/client/test/AbstractPlainJavaFxTest.java
```diff
@@ -32,6 +32,9 @@
 
   public AbstractPlainJavaFxTest() {
     root = new Pane();
+    Thread.currentThread().setUncaughtExceptionHandler((thread, e) -> {
+      throw (RuntimeException) e;
```
1-alex98, line 36: Should we not log them???

1-alex98 (approved):

src/test/java/com/faforever/client/test/AbstractPlainJavaFxTest.java
Katharsas: This code makes it so that tests fail like as if you throw any other exception inside a test case (which will automatically cause the test framework to log the exception and fail the test). So logging here would produce double stacktrace.

1-alex98: I read the issue now and cool I really hated that behaviour

PR #1397, opened 2019-08-18
Fix circular dependency ReplayService <-> GameService

### LINKED ISSUE #1396: Circular dependency GameService <-> ReplayService
Makes builds deployed via installer crash on startup.

Katharsas:
**Makes installers work again.**
Split ReplayService:
- `ReplayFileService` for the important stuff that has to do with handling replay files and starting replays; maybe this name should be changed back to `ReplayService`
- `ReplayRemoteService`; i really don't know why these 1-line-methods even need to exist if they just wrap `ReplayServer` and `FafService` but ok, the `GameService` just wanted to use these.

Untangle ReplayServer / GameService

Fixes #1396

### DIFF
--- a/src/main/java/com/faforever/client/replay/ReplayServerImpl.java
+++ b/src/main/java/com/faforever/client/replay/ReplayServerImpl.java
@@ -181,7 +181,6 @@ public class ReplayServerImpl implements ReplayServer {
   }
 
   private void finishReplayInfo(Supplier<Game> onGameInfoFinished) {
-    // game needs to running or finished
     Game game = onGameInfoFinished.get();
 
     replayInfo.updateFromGameInfoBean(game);
--- a/src/main/java/com/faforever/client/replay/ReplayServerImpl.java
+++ b/src/main/java/com/faforever/client/replay/ReplayServerImpl.java
@@ -2,7 +2,6 @@ package com.faforever.client.replay;
 
 import com.faforever.client.config.ClientProperties;
 import com.faforever.client.game.Game;
-import com.faforever.client.game.GameService;
 import com.faforever.client.i18n.I18n;
 import com.faforever.client.notification.Action;
 import com.faforever.client.notification.NotificationService;
@@ -30,6 +29,7 @@ import java.net.SocketException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Supplier;
 
 import static com.github.nocatch.NoCatch.noCatch;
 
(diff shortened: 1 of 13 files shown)


Katharsas:
Ok seems like the client doesnt even start locally now.

Katharsas:
Ok so now there is still a circular dependency between `GameService` and `ReplayServer` (since `GameService` depends on `ReplayRemoteService`).

Does anybody have an idea how to refactor this?
`ReplayServer` needs to call only one method `public Game getByUid(int uid)`. But i have not clear understanding of what exactly is inside `uidToGameInfoBean` and when, so is see no clear way to refactor.

Katharsas:
Btw since i replaced member injection with constructor injection in this branch the circular dependencies will now cause every startup to fail, including local ones, so anybody can reproduce.

Katharsas:
First circular dependency (solved):
Thinking about renaming `ReplayFileService` back to `ReplayService` to keep git history more readable and prevent a bunch of variable renaming otherwise needed. Opinions?

Second circular dependency:
The ReplayServer circle can be resolved by passing game supplier function to ReplayServer instead of ReplayServer querying GameService. Not pretty but i don't see alternatives since it is called from a separate thread inside ReplayServer which makes other solutions not pretty as well. Opinions?

Once ReplayServer is changed, this issue is solved and client installer can be built (see branch `hotfix/homepage-download`, where i implemented above suggestion).

1-alex98:
> First circular dependency (solved):
> Thinking about renaming `ReplayFileService` back to `ReplayService` to keep git history more readable and prevent a bunch of variable renaming otherwise needed. Opinions?
> 
> Second circular dependency:
> The ReplayServer circle can be resolved by passing game supplier function to ReplayServer instead of ReplayServer querying GameService. Not pretty but i don't see alternatives since it is called from a separate thread inside ReplayServer which makes other solutions not pretty as well. Opinions?
> 
> Once ReplayServer is changed, this issue is solved and client installer can be built (see branch `hotfix/homepage-download`, where i implemented above suggestion).

You will need to explain that to me maybe discord?

1-alex98 (approved):

src/main/java/com/faforever/client/replay/ReplayServerImpl.java
```diff
@@ -179,12 +176,13 @@ private void recordAndRelay(int uid, ServerSocket serverSocket, @Nullable Output
...
     log.debug("FAF has disconnected, writing replay data to file");
-    finishReplayInfo();
+    finishReplayInfo(onGameInfoFinished);
     replayFileWriter.writeReplayDataToFile(replayData, replayInfo);
   }
 
-  private void finishReplayInfo() {
-    Game game = gameService.getByUid(replayInfo.getUid());
+  private void finishReplayInfo(Supplier<Game> onGameInfoFinished) {
+    // game needs to running or finished
```
1-alex98: Remove

1-alex98 (changes requested):

1-alex98:
build it and tested it

1-alex98 (approved):

PR #1408, opened 2019-08-29
Use Default Decorator

### LINKED ISSUE #1317: missing command line switch
If a user deletes the "%s" in the command line switch field in the client settings they effectively break the client -> following error will appear upon hosting/joining a game after that:

```
java.util.concurrent.CompletionException: com.github.nocatch.NoCatchException: java.io.IOException: Cannot run program "/init" (in directory "C:\ProgramData\FAForever\bin"): CreateProcess error=2, The system cannot find the file specified
```

Please don't allow users to break the client by removing the content of the command line switch field

1-alex98:
Fix #1317

### DIFF
--- a/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
+++ b/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
@@ -2,6 +2,7 @@ package com.faforever.client.fa;
 
 import com.faforever.client.game.Faction;
 import com.faforever.client.preferences.ForgedAlliancePrefs;
+import com.google.common.base.Strings;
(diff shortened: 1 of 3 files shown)


Katharsas:
Btw. i think i changed stuff some time ago so you don't need the `"` around `%s` anymore.

1-alex98:
I thought so as well but it was not

Katharsas:
What?
Tests at least work without `"`, see LaunchCommandBuilderTest line 20.
and see ` @Test
  public void testCommandFormatWithSpaces()`

1-alex98:
Sorry I thought the thing I was trying to solve here. I missread your comment :D

PR #1410, opened 2019-08-29
Use https

### LINKED ISSUE #1407: Consistently use https
There are place in production where there is used http for content.faforever.com

1-alex98:
Fix #1407

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ buildscript {
   repositories {
     mavenCentral()
     maven { url 'https://maven.ej-technologies.com/repository' }
-    maven { url 'http://repo.spring.io/plugins-release' }
+    maven { url 'https://repo.spring.io/plugins-release' }
   }
 
   dependencies {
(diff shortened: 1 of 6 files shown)


Katharsas:
There are more usages of http in the other application-xxx.yaml files that i don't see a reason to not change as well. And while you are at it maybe switch the remaining two usages of http in the build.gradle and the ci scripts^^

1-alex98:
ehm no in test they need to stay http because often https does not work on test

Katharsas:
application-local.yml contains for example
`http://content.faforever.com/faf/updaterNew/updates_faf_files/ForgedAlliance.exe`
and
```
vault:
    base-url: http://content.faforever.com
```
Why can these not be http?
Same goes for other yml files.

1-alex98:
ok true but test.faforever does not always have a working certificate

1-alex98:
Anything I missed? Made a project wide search

Katharsas (approved):
Looks good.

PR #1413, opened 2019-08-29
Fix MatchmakerMessage rating boundaries

### LINKED ISSUE #1412: Regression: MatchmakerMessage NullpointerException
Removing annotations
```Java
@SerializedName("boundary_75s")
private List<RatingRange> boundary75s;
@SerializedName("boundary_80s")
private List<RatingRange> boundary80s;
```
causes these lists to be null always, because apparently an underscore before a number is not removed to match variable names.

Katharsas:
Fixes #1412


Katharsas:
@axel1200 this is the regression i was talking about. It should be in the next RC.
Seems like a non-related test randomly failed?

1-alex98 (approved):

1-alex98:
merging when test pass

PR #1423, opened 2019-09-04
Fix missing uninstall calls for tooltips

### LINKED ISSUE #1421: Potential Tooltip leak
`Tooltip.install` gets called in some places without corresponding `Tooltip.uninstall`.

Katharsas:
Fixes #1421

### DIFF
--- a/src/main/java/com/faforever/client/game/GamesTilesContainerController.java
+++ b/src/main/java/com/faforever/client/game/GamesTilesContainerController.java
@@ -48,7 +48,6 @@ public class GamesTilesContainerController implements Controller<Node> {
   private GameTooltipController gameTooltipController;
   private Tooltip tooltip;
 
-  @Inject
   public GamesTilesContainerController(UiService uiService, PreferencesService preferencesService) {
     this.uiService = uiService;
     this.preferencesService = preferencesService;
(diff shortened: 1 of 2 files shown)


1-alex98:
Lets see if that was it but I can imagine so. There where lots of tooltips in memory dump

src/main/java/com/faforever/client/game/GamesTilesContainerController.java
```diff
@@ -68,7 +67,11 @@ public GamesTilesContainerController(UiService uiService, PreferencesService pre
       JavaFxUtil.assertApplicationThread();
         while (change.next()) {
           change.getRemoved().forEach(gameInfoBean -> {
-            boolean remove = tiledFlowPane.getChildren().remove(uidToGameCard.remove(gameInfoBean.getId()));
+            Node card = uidToGameCard.remove(gameInfoBean.getId());
+            if (card != null) {
+              Tooltip.uninstall(card, tooltip);
+            }
+            boolean remove = tiledFlowPane.getChildren().remove(card);
```
Brutus5000: The following three lines should go into the card!=null block
The else block should also throw an error in the log.
(We don't know how many times it actually occurs and that would help I guess)

Brutus5000 (changes requested):

Brutus5000 (approved):

1-alex98 (approved):

1-alex98:
@Katharsas did you check if tooltips are removed in chat as well?

1-alex98:
I think the tooltips are never removed but i think the difference is that the tooltip is not held by an external Controller

1-alex98:
So should be fine

Katharsas:
We basically made a search for `*.install` in all sources and made sure that every install has corresponding uninstall. Don't remember if we found chat stuff.

PR #1425, opened 2019-09-05
Prevent autologin in case of mandatory update

### LINKED ISSUE #1424: Mandatory client updates don't work if auto-login is active
The user sees the warning for a second, then he is logged in anyway.

Brutus5000:
Fixes #1424

### DIFF
--- a/src/main/java/com/faforever/client/login/LoginController.java
+++ b/src/main/java/com/faforever/client/login/LoginController.java
@@ -17,6 +17,7 @@ import com.faforever.client.update.DownloadUpdateTask;
 import com.faforever.client.update.UpdateInfo;
 import com.faforever.client.update.Version;
 import com.faforever.client.user.UserService;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.jfoenix.controls.JFXButton;
 import javafx.application.Platform;
@@ -60,6 +61,8 @@ public class LoginController implements Controller<Node> {
   private final ClientProperties clientProperties;
   private final I18n i18n;
   private final ClientUpdateService clientUpdateService;
+  private CompletableFuture<Void> initializeFuture;
+  private Boolean loginAllowed;
 
   public Pane loginFormPane;
   public Pane loginProgressPane;
@@ -81,7 +84,8 @@ public class LoginController implements Controller<Node> {
   public TextField apiBaseUrlField;
   public JFXButton serverStatusButton;
 
-  private CompletableFuture<UpdateInfo> updateInfoFuture;
+  @VisibleForTesting
+  CompletableFuture<UpdateInfo> updateInfoFuture;
 
   public LoginController(
       UserService userService,
@@ -163,23 +167,31 @@ public class LoginController implements Controller<Node> {
 
 
     if (clientProperties.isUseRemotePreferences()) {
-      preferencesService.getRemotePreferencesAsync()
-          .thenApply(clientConfiguration ->
-          {
+      initializeFuture = preferencesService.getRemotePreferencesAsync()
+          .thenApply(clientConfiguration -> {
             String minimumVersion = clientConfiguration.getLatestRelease().getMinimumVersion();
             if (minimumVersion != null && Version.shouldUpdate(Version.getCurrentVersion(), minimumVersion)) {
-              showClientOutdatedPane(minimumVersion);
+              loginAllowed = false;
+              Platform.runLater(() -> showClientOutdatedPane(minimumVersion));
+            } else {
+              loginAllowed = true;
             }
             return clientConfiguration;
           })
           .thenAccept(clientConfiguration -> {
-            Endpoints defaultEndpoint = clientConfiguration.getEndpoints().get(0);
-            environmentComboBox.getItems().addAll(clientConfiguration.getEndpoints());
-            environmentComboBox.getSelectionModel().select(defaultEndpoint);
+            Platform.runLater(() -> {
+              Endpoints defaultEndpoint = clientConfiguration.getEndpoints().get(0);
+              environmentComboBox.getItems().addAll(clientConfiguration.getEndpoints());
+              environmentComboBox.getSelectionModel().select(defaultEndpoint);
+            });
           }).exceptionally(throwable -> {
-        log.warn("Could not read remote preferences");
-        return null;
-      });
+            log.warn("Could not read remote preferences");
+            loginAllowed = true;
+            return null;
+          });
+    } else {
+      loginAllowed = true;
+      initializeFuture = CompletableFuture.completedFuture(null);
     }
   }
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -124,7 +124,7 @@ jacocoTestReport {
 
 test {
   testLogging {
-    showStandardStreams = true
+    events("passed", "skipped", "failed")
```
Katharsas: The problem with only having this is that while it will tell you that a test fails it will print almost no additional information like: The message of the exception that may have caused a fail, its stacktrace, cause etc.
I suggest adding some more options so we have this:
```
    events("passed", "skipped", "failed")
    showExceptions true
    exceptionFormat "full"
    showCauses true
    showStackTraces true
    showStandardStreams = false
```
This makes sure that when an exception causes a test to fail, that exception gets properly printed. If some developer looks at this he will know why standard output is not shown locally and can enable it by changing `showStandardStreams ` temporarily. So i'd say that makes it self-documented.

And we still get rid of any logs that don't cause test failure.
(I took those options from a gradle forum post and tested them, they seem to work great)

src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -163,23 +168,31 @@ public Endpoints fromString(String string) {
...
-        log.warn("Could not read remote preferences");
-        return null;
-      });
+            log.warn("Could not read remote preferences");
+            loginAllowed = true;
+            return null;
+          });
+    } else {
+      initializeFuture = CompletableFuture.completedFuture(null);
+      loginAllowed = true;
```
Katharsas, line 193: I guess this is garantueed to execute in the same thread and before the display method. But paranoid me would still would invert the order of these two statements.

src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -225,13 +238,17 @@ public void display() {
...
+      Assert.notNull(loginAllowed, "loginAllowed must be set here or you mixed up the futures.");
+
+      if (loginAllowed && loginPrefs.getAutoLogin() && !isNullOrEmpty(username) && !isNullOrEmpty(password)) {
+        login(username, password, true);
+      } else if (isNullOrEmpty(username)) {
+        usernameInput.requestFocus();
+      } else {
+        passwordInput.requestFocus();
+      }
+    });
```
Katharsas, line 253: It looks like you should add an exception handler to this. Otherwise any exceptions, including the `Assert.notNull...` above will get swallowed right?

Katharsas (changes requested):

src/main/java/com/faforever/client/login/LoginController.java
Brutus5000: Well the Assert is there to at least give some information in the logs if the application is in an invalid state. But there is no proper recovery for that. What would I do in the exceptionally case?

Katharsas: The Assert is fine, but right now it wont get logged if it throws. If i remember this correctly, if a CompletableFuture throws an exception asynchronously, that will just go into the Threadpool Nirvana unless you always register an exception handler that prints any exceptions.

Katharsas: Which is a horrible thing in general because there are porbably tons of code like this that don't register any exception handlers so all of these exceptions dont get logged. Unless there is some global handler which i don't think exists (definitely not by default, and i made a quick test with this code, seems like there is no global handler).

build.gradle
1-alex98: I liked the way it way with showStandardStreams = true

it spamed the log but it finally printed needed information and if u know where to search for the right info u will find it

Brutus5000: I can live with it if we can suppress it for successful tests

Katharsas: Showing standard streams only for failure can be done it seems but is very ugly and not worth it imo:
https://stackoverflow.com/questions/36558340/during-gradle-test-show-standard-out-and-err-only-for-failed-tests 

@axel1200 I kinda agree with Brutus, searching though Travis output is rather annoying currently, and since most Exceptions will now actually fail our tests and therefore get printed i think its fine to disable standard streams by default, as long as its super easy to change locally.

1-alex98: I dont have a problem with it... I search for FAILED and there are 8 ones i find i skip the first 5 till i get to the failed test

src/main/java/com/faforever/client/login/LoginController.java
Brutus5000: Fixed

src/main/java/com/faforever/client/login/LoginController.java
Brutus5000: Okay, I now log it as error and enable login instead

build.gradle
Brutus5000: I removed it for now from the PR. But I see this differently. The logs are garbage with shitload of exceptions flying for passing tests.

1-alex98 (approved):

build.gradle
1-alex98: They are even worse if the Exception is not in them :D

src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -225,13 +237,20 @@ public void display() {
...
     autoLoginCheckBox.setSelected(isAutoLogin);
 
-    if (loginPrefs.getAutoLogin() && !isNullOrEmpty(username) && !isNullOrEmpty(password)) {
-      login(username, password, true);
-    } else if (isNullOrEmpty(username)) {
-      usernameInput.requestFocus();
-    } else {
-      passwordInput.requestFocus();
-    }
+    initializeFuture.thenAccept(aVoid -> {
```
1-alex98: this should be thenRun

1-alex98 (changes requested):

src/main/java/com/faforever/client/login/LoginController.java
1-alex98: as aVoid is not used

build.gradle
Katharsas: Reading the exceptions of for example half-broken chat test over and over does not help in everyday work. If someone wants to fix them, he can just switch `showStandardStreams` to true locally. And we should probably make an issue for that. And somebody should probably go and fix it.

src/main/java/com/faforever/client/login/LoginController.java
Katharsas: Ok.
I think it was ok/better the way before (if you had just added an `exceptionally` lamda after the `thenRun`). Because if for example any of the other code in that lamda throws a different exception, it will still not get logged.
And in the future, we should have a global exception handler that makes sure that exceptions at least get logged. Then we could remove the `exceptionally` handlers everywhere in the coder where they just catch/log exceptions.

1-alex98 (approved):

PR #1428, opened 2019-09-07
Fix toggle selection

### LINKED ISSUE #1427: Show if table or tile is selected
Visit Issue hunt...

1-alex98:
Fix #1427

### DIFF
--- a/src/main/resources/theme/play/custom_games.fxml
+++ b/src/main/resources/theme/play/custom_games.fxml
@@ -19,13 +19,12 @@
 <?import javafx.scene.layout.StackPane?>
 <?import javafx.scene.layout.VBox?>
 <StackPane xmlns:fx="http://javafx.com/fxml/1" fx:id="gamesRoot" alignment="TOP_CENTER"
-           xmlns="http://javafx.com/javafx/10.0.1" fx:controller="com.faforever.client.game.CustomGamesController">
+           xmlns="http://javafx.com/javafx/10.0.2" fx:controller="com.faforever.client.game.CustomGamesController">
     <children>
         <GridPane fx:id="gamesGridPane">
             <columnConstraints>
                 <ColumnConstraints hgrow="ALWAYS" minWidth="10.0"/>
-                <ColumnConstraints hgrow="SOMETIMES" minWidth="300.0"
-                                   prefWidth="300.0" fx:id="sidePaneColumn"/>
+                <ColumnConstraints fx:id="sidePaneColumn" hgrow="SOMETIMES" minWidth="300.0" prefWidth="300.0"/>
             </columnConstraints>
             <rowConstraints>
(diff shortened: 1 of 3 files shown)


Katharsas (changes requested):
This breaks my client layout slightly:
Without PR:
css_tiles_1
With PR:
css_tiles_2

You may need to set Windows Scaling to 125 or 115 percent to reproduce, not sure.

1-alex98:
This makes not much sense tho because all i did was chnage background colour and spacing

src/main/resources/theme/play/custom_games.fxml
```diff
@@ -86,7 +87,7 @@
                     </content>
                 </ScrollPane>
                 <AnchorPane fx:id="gameViewContainer" maxHeight="1.7976931348623157E308"
-                            maxWidth="1.7976931348623157E308" GridPane.rowIndex="1" prefWidth="10000"/>
+                            maxWidth="1.7976931348623157E308" prefWidth="10000" GridPane.rowIndex="1"/>
```
1-alex98, line 90: Maybe this causes issue tho i did not set this this was scene builder...

Katharsas: I don't see the difference, do you mean the order?

Katharsas:
Let me check again against the base between this branch and develop instead of develop. Maybe some other change on develop causes this.

src/main/resources/theme/play/custom_games.fxml
1-alex98: oh no i thought it added prefWidth

Katharsas:
No its definitely this commit causing the layout change. Probably because the additional highlighting adds some space here and there which throws of the rest of the layout.
Edit: Oh wait.
I think i know whats happening.

Katharsas:
The problem is horizontal resizing. If you make the width of the client smaller and smaller, at some point the tile settings will be split into two lines. This looks horrible in all versions of the client (as in the screen). More than 2 lines is basically unusable.
Your change adds some horizontal space to the toggle buttons, which means that now when i am in fullscreen, the tile options won't fit into one line (as you can see in the screenshot, the sorting option flowed to the second line), whereas before it did barely fit in one line.

So i think its just too many options without enough horizontal space which is made slightly worse by this change.

1-alex98:
Well if i dont add the spacing it looks awful believe me

Katharsas:
Its really a different issue i think (the multi line layout), so im fine with this. Havent looked at the code yet.

Katharsas:
I want to test this again, can you rebase?

Katharsas:
The weird thing is that in the table view elements resize perfectly. Only in the tiles view do they not.

Katharsas:
Ok found i pushed a fix for the resize problem.

Katharsas (approved):

PR #1433, opened 2019-09-09
Fix nullpointer

1-alex98:


1-alex98:
happens if u move mouse over empty table row

src/main/java/com/faforever/client/game/GamesTableController.java
```diff
@@ -215,6 +214,9 @@ protected void updateItem(Game game, boolean empty) {
       }
     });
     row.setOnMouseEntered(event -> {
+      if (row.getItem() == null) {
+        return;
+      }
       Game game = row.getItem();
       gameTooltipController.setGame(game);
```
Brutus5000, line 221: Do we need to set the game to null here as well?

1-alex98: true

1-alex98: Acctaually no

1-alex98: Their is only one tooltip for all rows and it is only installed on rows which are not empty. So doing nothing here is just fine. Also before i added the nullcheck it was just fine the only issue is that this Exception spams the log

1-alex98: I mean the tooltip wont be shown on empty rows any way

Katharsas (approved):

src/main/java/com/faforever/client/game/GamesTableController.java
Brutus5000: But the game can't be cleaned up as long as it is referenced. I know this is not a leak, but shouldn't we remove unused references?

1-alex98: Well it is only always one game at a time and the reference is refreshed once the user hover over the next game. The issue cleaning the whole thing is some work and also performance wise it is better the way it is when returning. imagine i need to clear the whole UI that also requires CPU load and GPU load. I mean we should remove unused references but foe this one being temporary and only one object at a time it seems reasonable not to.

Brutus5000: Ok

Brutus5000 (approved):

PR #1434, opened 2019-09-09
Fix assertion by setting launch before state

1-alex98:


Brutus5000 (changes requested):
You are changing behavior without adapting a test and it doesn't break a unit test --> There is a unit test missing.

1-alex98:
Well there was no test for the order of them...

1-alex98:
Should i add one?

Brutus5000:
Yes, we should add it 😄

Brutus5000 (approved):

PR #1436, opened 2019-09-09
Add platform run later

1-alex98:


Geosearchef (approved):

src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
```diff
@@ -261,7 +259,7 @@ public void onFactionButtonClicked() {
   }
 
   private void setCurrentPlayer(Player player) {
-    playerRatingListener = ratingObservable -> updateRating(player);
+    playerRatingListener = ratingObservable -> Platform.runLater(() -> updateRating(player));
```
Katharsas, line 262: There are only two calls to `setCurrentPlayer()`, one of which already wraps the entire call in Platform.runLater (remove i guess?).

Katharsas: Or i guess you could remove it here and add it there,

Brutus5000 (approved):

src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
1-alex98: Well but this is a listener so not on the same thread than setCurentPlayer

Katharsas: Oh right.

PR #1459, opened 2019-10-04
Opt in for beta updates

### LINKED ISSUE #1458: Beta opt in
Option for the user to opt in for beta pre releases released on github.com

1-alex98:
Fixes #1458

### DIFF
--- a/src/main/java/com/faforever/client/config/ClientProperties.java
+++ b/src/main/java/com/faforever/client/config/ClientProperties.java
@@ -152,5 +152,6 @@ public class ClientProperties {
     private String applicationId;
     private String smallImageKey;
     private String bigImageKey;
+    private String discordPrereleaseFeedbackChannelUrl;
   }
 }
--- a/src/main/java/com/faforever/client/main/MainController.java
+++ b/src/main/java/com/faforever/client/main/MainController.java
@@ -34,7 +34,7 @@ import com.faforever.client.remote.domain.RatingRange;
 import com.faforever.client.theme.UiService;
 import com.faforever.client.ui.StageHolder;
 import com.faforever.client.ui.tray.event.UpdateApplicationBadgeEvent;
-import com.faforever.client.update.ClientUpdateService;
+import com.faforever.client.user.event.LoggedInEvent;
 import com.faforever.client.user.event.LoggedOutEvent;
 import com.faforever.client.user.event.LoginSuccessEvent;
 import com.faforever.client.vault.VaultFileSystemLocationChecker;
@@ -76,6 +76,7 @@ import javafx.stage.StageStyle;
 import javafx.util.Duration;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.config.ConfigurableBeanFactory;
+import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
@@ -106,19 +107,18 @@ public class MainController implements Controller<Node> {
   private static final PseudoClass HIGHLIGHTED = PseudoClass.getPseudoClass("highlighted");
   @VisibleForTesting
   protected static final PseudoClass MAIN_MINIMIZED = PseudoClass.getPseudoClass("minimized");
-  private final Cache<NavigationItem, AbstractViewController<?>> viewCache = CacheBuilder.newBuilder().build();
+  private final Cache<NavigationItem, AbstractViewController<?>> viewCache;
   private final PreferencesService preferencesService;
   private final I18n i18n;
   private final NotificationService notificationService;
   private final PlayerService playerService;
   private final GameService gameService;
-  private final ClientUpdateService clientUpdateService;
   private final UiService uiService;
   private final EventBus eventBus;
   private final GamePathHandler gamePathHandler;
   private final PlatformService platformService;
   private final VaultFileSystemLocationChecker vaultFileSystemLocationChecker;
-  private final ClientProperties clientProperties;
+  private final ApplicationEventPublisher applicationEventPublisher;
   private String mainWindowTitle;
   private int ratingBeta;
   public Pane mainHeaderPane;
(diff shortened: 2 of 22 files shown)


src/main/java/com/faforever/client/config/ClientProperties.java
```diff
@@ -41,6 +42,14 @@
...
   }
 
+  @Data
+  public static class GitHubRepo {
+    /**
+     * Api URL to the client GitHub Repo
+     */
+    private String apiUrl;
+  }
+
```
Brutus5000: I would add this to the remote preferences / client configuration to also taking into account  a potential renaming of the repository (#1288)

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -134,14 +135,15 @@
   public CheckBox disallowJoinsCheckBox;
   public JFXToggleButton secondaryVaultLocationToggleButton;
   public Button autoJoinChannelsButton;
+  public JFXToggleButton prereleaseToggleButton;
```
Brutus5000: I would use ToggleButton instead. Always use the most generic one. (Also applies to line 136)

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -292,6 +295,13 @@ public void initialize() {
       preferences.getForgedAlliance().setVaultBaseDirectory(vaultBaseDirectory);
     });
 
+    prereleaseToggleButton.selectedProperty().bindBidirectional(preferences.prereleaseCheckEnabledProperty());
+    prereleaseToggleButton.selectedProperty().addListener((observable, oldValue, newValue) -> {
+      if (!oldValue && newValue) {
```
Brutus5000: Is this safe for NullPointerExceptions?

src/main/java/com/faforever/client/update/CheckForBetaUpdateTask.java
```diff
@@ -0,0 +1,78 @@
...
+    headers.add("Accept", "application/vnd.github.v3+json");
+    HttpEntity<String> entity = new HttpEntity<>(null, headers);
+
+    ResponseEntity<List<GitHubRelease>> response = restTemplate.exchange(clientProperties.getGitHubRepo().getApiUrl() + PATH_FOR_RELEASE, HttpMethod.GET, entity, new ParameterizedTypeReference<List<GitHubRelease>>() {
+    });
+    List<GitHubRelease> responseBody = response.getBody();
+    if (responseBody == null || responseBody.isEmpty()) {
+      return null;
+    }
+    GitHubRelease latestRelease = responseBody.get(0);
```
Brutus5000: 1. Shouldn't the RestTemplate come via DI?
2. If we need a new one, create it in the constructor
3. The whole code of setting up the restTemplate call should be extracted into a separate method for better readability.

src/main/java/com/faforever/client/update/ClientUpdateService.java
```diff
@@ -8,13 +8,18 @@
   /**
    * Returns information about an available mandatory update. Returns {@code null} if no update is available.
```
Brutus5000: I guess this should be renamed as well?

src/main/java/com/faforever/client/update/UpdateInfo.java
```diff
@@ -12,5 +12,6 @@
   private URL url;
   private int size;
   private URL releaseNotesUrl;
+  private boolean prerelease;
```
Brutus5000, line 15: preRelease?

Brutus5000 (changes requested):

src/main/java/com/faforever/client/config/ClientProperties.java
1-alex98: ok

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
1-alex98: Well i autogenerate those with Intelij. Is it really that bad? I would like to keep using this feature. I mean I could have also used Checkable then but not sure why i should to that in this very case.

src/main/java/com/faforever/client/update/UpdateInfo.java
1-alex98: it is one word!

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
Brutus5000: I would do it, as it makes easier to drop JFoenix if you want to (e.g. different themes).

1-alex98:
fitting config can be fund here https://content.test.faforever.com/dfc-config.json

src/main/java/com/faforever/client/main/MainController.java
```diff
@@ -476,7 +476,11 @@ private void enterLoggedInState() {
     getMainScene().setContent(mainRoot);
     getMainScene().setMoveControl(mainRoot);
 
-    clientUpdateService.checkForRegularUpdateInBackground();
+    if (preferencesService.getPreferences().isPrereleaseCheckEnabled()) {
+      clientUpdateService.checkForRegularUpdateInBackground();
+    } else {
+      clientUpdateService.checkForBetaUpdateInBackground();
```
Geosearchef: Isn't one method with a boolean parameter a better choice? Why create 2 methods?
And why is the preference read in a UI class, shouldn't the clientUpdateService just read the property itself?

src/main/java/com/faforever/client/update/CheckForBetaUpdateTask.java
```diff
@@ -0,0 +1,82 @@
...
+    );
+  }
+
+  private GitHubAssets getAssetOfFileWithEnding(GitHubRelease latestRelease, String ending) {
+    return latestRelease.getAssets().stream()
+        .filter(asset -> asset.getName().contains(ending))
+        .findFirst()
+        .orElseThrow();
+  }
+}
```
Geosearchef, line 82: Isn't this the same code as in the regular Update task?

Geosearchef (changes requested):

src/main/java/com/faforever/client/main/MainController.java
1-alex98: Well this is after Logging in. But yeah we could have an event here...

src/main/java/com/faforever/client/update/CheckForBetaUpdateTask.java
1-alex98: no the regular update task does not use GitHub at all!

Brutus5000 (dismissed):

src/main/java/com/faforever/client/config/ClientProperties.java
```diff
@@ -152,5 +152,6 @@
     private String applicationId;
     private String smallImageKey;
     private String bigImageKey;
+    private String discordPrereleaseFeedback;
```
Brutus5000: Is this a channel name? A URL? Please name variable properly

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -508,5 +519,9 @@ public void onUseNoBackgroundImage(ActionEvent actionEvent) {
     preferencesService.getPreferences().getMainWindow().setBackgroundImagePath(null);
     preferencesService.storeInBackground();
   }
+
+  public void openDiscordFeedback() {
```
Brutus5000: openDiscordFeedbackChannel?

Brutus5000 (changes requested):

Issue #1455, opened 2019-10-03
Code cleanup of Spring components


Brutus5000:
* [ ]  Remove all `@Inject` and `@Autowired` annotations from the main modules (not the tests!). They are obsolete and no longer required.
* [ ]  Replace hand-written constructors for these components with Lombok's `@AllArgsConstructor` or `@RequiredArgsConstructor` (where applicable - does not work in rare cases where the constructor is doing more than just assigning `private final` variables)
* [ ]  Replace `@PostConstruct` and `@PreDestroy` methods -> add the interface implementations for `InitializingBean` and `DisposableBean` and move the code to the interface method.

marcoebbinghaus:
I am working on this. Feel free to assign me.

Brutus5000:
@marcoebbinghaus Parts of it are already done in https://github.com/FAForever/downlords-faf-client/pull/1457

marcoebbinghaus:
@Brutus5000 okay, I've already finished/completed this locally. How should I go on now? Wait for #1457 and see if I can add something on top of it?

PR #1460, opened 2019-10-07
#1455 code cleanup of Spring components

marcoebbinghaus:
I removed @Inject whereever possible and replaced it with @RequiredArgsConstructor. Same for @PostConstruct and @PreDestroy (replaced by implementing InitializingBean and DisposableBean).

There were lots of classes which did field initialization inside of the constructor. I moved those initializations to afterPropertiesSet or to the field declaration. If there was other stuff done inside the constructor I moved that to afterPropertiesSet() if possible.

The only classes that stil have an @Inject-annotated constructor are those classes who are calling a a super constructor in their @Inject-annotated constructor.

I aligned the tests according to the code changes.

If there is another PR already feel free to take whatever you may need from my PR on top of the other PR.


src/main/java/com/faforever/client/FafClientApplication.java
```diff
@@ -87,7 +88,7 @@ public void init() {
     Font.loadFont(FafClientApplication.class.getResourceAsStream("/font/dfc-icons.ttf"), 10);
     JavaFxUtil.fixTooltipDuration();
 
-    applicationContext = new SpringApplicationBuilder(FafClientApplication.class)
+    applicationContext = new SpringApplicationBuilder(FafClientApplication.class, ForgedAlliancePrefs.class)
```
1-alex98: Why is that done?

src/main/java/com/faforever/client/chat/AutoCompletionHelper.java
```diff
@@ -21,18 +20,16 @@
...
 
   private TextInputControl boundTextField;
-  private EventHandler<KeyEvent> keyEventHandler;
+  private EventHandler<KeyEvent> keyEventHandler = keyEvent -> {
+    if (!keyEvent.isControlDown() && keyEvent.getCode() == KeyCode.TAB) {
+      keyEvent.consume();
+      autoComplete();
+    }
+  };
   private Function<String, Collection<String>> completionProposalGenerator;
```
1-alex98: make final

src/main/java/com/faforever/client/chat/AutoCompletionHelper.java
```diff
@@ -21,18 +20,16 @@
...
 
-  @Inject
   public AutoCompletionHelper(Function<String, Collection<String>> completionProposalGenerator) {
```
1-alex98: Replace by @RequiredArgsConstructor

1-alex98 (changes requested):

1-alex98:
Also we are not sure if it is a good idea to move everything to afterPropertiesSet. While that is ok for probbaly 99% of cases. If afterPropertiesSet would rely on each other which could be then it might cause errors, because there is no garanteed calling order of afterPropertiesSet.

1-alex98:
We would prefer initilating them on declaration if you would like to replace as many constructors as possible...

src/main/java/com/faforever/client/FafClientApplication.java
marcoebbinghaus: Without that there is no Bean found for ForgedAlliancePrefs, which is needed in the MapService. 

Possible alternative: undo that change, make field 'forgedAlliancePreferences' in MapService non-final and set it inside afterPropertiesSet() (--> forgedAlliancePreferences = preferencesService.getPreferences().getForgedAlliance();). (Not possible to do that on field declaration as the injected preferencesService is needed here).

I changed it to the latter, please have a look at my new commit and see if it is the better approach.

src/main/java/com/faforever/client/chat/AutoCompletionHelper.java
marcoebbinghaus: done, see my new commit please.

src/main/java/com/faforever/client/chat/AutoCompletionHelper.java
marcoebbinghaus: done too, see my new commit please.

marcoebbinghaus:
I don't get what exactly should be done now regarding afterPropertiesSet / variable initialization on declaration. 

When removing @Inject-annotated constructors for replacing them by @RequiredArgsConstructor I did two things for cases where more stuff than assigning parameters to fields happened inside the constructor:

- move everything possible (field initialization) to field declaration
- move the rest to afterPropertiesSet

So 

> We would prefer initilating them on declaration

this is already done. Have a look at MapService for example. I moved 

`pathToMap = new HashMap<>();`

to field declaration.

while moving 

`Vault vault = clientProperties.getVault();`

to afterPropertiesSet. This cannot be done on field declaration as it is using `clientProperties` which is injected and therefore would throw NPE if moved to field declaration.

Could you show an example that you don't like and show how it should be solved instead?

src/main/java/com/faforever/client/FafClientApplication.java
1-alex98: ForgedAlliancePrefs should not be registered with Spring it is retrived via the preferenceService

1-alex98:
ok here my example(just an example) of why it could possibly break stuff:

Say Service A does some initialization and needs Method C of Service B for that. When doing it in constructor everything works. When doing it in afterProperrtiesSet Service A might call Method C before Service A is completely initialized and cause a crash. Even worse this might result in funny race conditions. So we would like initialization to stay in constcutor. If that means that there are less of them replaced so be it. Work that has nothing to do with initalization like startup tasks can still be done in afterProperrtiesSet .

src/main/java/com/faforever/client/achievements/AchievementService.java
```diff
@@ -105,6 +97,8 @@ public Image getImage(AchievementDefinition achievementDefinition, AchievementSt
   @Override
   public void afterPropertiesSet() {
     fafService.addOnMessageListener(UpdatedAchievementsMessage.class, updatedAchievementsMessage -> reloadAchievements());
+    playerAchievements = FXCollections.observableArrayList();
```
1-alex98: move to declaration

src/main/java/com/faforever/client/achievements/AchievementService.java
```diff
@@ -105,6 +97,8 @@ public Image getImage(AchievementDefinition achievementDefinition, AchievementSt
...
+    readOnlyPlayerAchievements = FXCollections.unmodifiableObservableList(playerAchievements);
```
1-alex98: move in constructor i guess

src/main/java/com/faforever/client/api/FafApiAccessorImpl.java
```diff
@@ -120,6 +109,10 @@ private static String rsql(Condition<?> eq) {
   @Override
   public void afterPropertiesSet() {
     eventBus.register(this);
+    templateBuilder = unconfiguredTemplateBuilder
```
1-alex98, line 112: this could stay. I can garantee that this does no harm

src/main/java/com/faforever/client/chat/MockChatService.java
```diff
@@ -39,40 +39,34 @@
...
+  @Override
+  public void afterPropertiesSet() {
+    eventBus.register(this);
     connectionState = new SimpleObjectProperty<>(ConnectionState.DISCONNECTED);
     unreadMessagesCount = new SimpleIntegerProperty();
 
     onChatMessageListeners = new ArrayList<>();
     channelUserListListeners = new HashMap<>();
 
     timer = new Timer(true);
```
1-alex98: all to declaration

src/main/java/com/faforever/client/chat/PircBotXChatService.java
```diff
@@ -141,33 +142,15 @@
...
-
+  @Override
+  public void afterPropertiesSet() {
+    eventBus.register(this);
     connectionState = new SimpleObjectProperty<>(ConnectionState.DISCONNECTED);
     eventListeners = new ConcurrentHashMap<>();
     channels = observableHashMap();
     chatChannelUsersByChannelAndName = observableMap(new TreeMap<>(String.CASE_INSENSITIVE_ORDER));
     unreadMessagesCount = new SimpleIntegerProperty();
     identifiedFuture = new CompletableFuture<>();
```
1-alex98: to declarations

src/main/java/com/faforever/client/chat/PircBotXChatService.java
```diff
@@ -141,33 +142,15 @@
...
-  }
-
-  @Override
-  public void afterPropertiesSet() {
-    eventBus.register(this);
     fafService.addOnMessageListener(SocialMessage.class, this::onSocialMessage);
```
1-alex98: I think u need to add constructor back here much of this is initialization

src/main/java/com/faforever/client/fa/OnGameFullNotifier.java
```diff
@@ -38,24 +38,12 @@
...
-    this.eventBus = eventBus;
-    this.faWindowTitle = clientProperties.getForgedAlliance().getWindowTitle();
-    this.gameService = gameService;
-  }
+  private final ClientProperties clientProperties;
+  private String faWindowTitle;
 
   @Override
   public void afterPropertiesSet() {
+    this.faWindowTitle = clientProperties.getForgedAlliance().getWindowTitle();
```
1-alex98: set up and hence should be done in constructor

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
```diff
@@ -97,6 +85,7 @@ public void afterPropertiesSet() {
     fafService.addOnMessageListener(GameLaunchMessage.class, this::updateLobbyModeFromGameInfo);
     fafService.addOnMessageListener(DisconnectFromPeerMessage.class, message -> iceAdapterProxy.disconnectFromPeer(message.getUid()));
     fafService.addOnMessageListener(IceServerMessage.class, message -> iceAdapterProxy.iceMsg(message.getSender(), message.getRecord()));
+    iceAdapterProxy = newIceAdapterProxy();
```
1-alex98: depends on what the method depends on either in constructor or if possible in declaration

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -563,6 +512,30 @@ public void onRehostRequest(RehostRequestEvent event) {
...
+
+      ChangeListener<GameStatus> statusChangeListener = generateGameStatusListener(newValue);
+      JavaFxUtil.addListener(newValue.statusProperty(), statusChangeListener);
+      statusChangeListener.changed(newValue.statusProperty(), newValue.getStatus(), newValue.getStatus());
+    });
+
+    games = FXCollections.observableList(new ArrayList<>(),
+        item -> new Observable[]{item.statusProperty(), item.getTeams()}
+    );
+    JavaFxUtil.attachListToMap(games, uidToGameInfoBean);
```
1-alex98: add constructor back

src/main/java/com/faforever/client/main/MainController.java
```diff
@@ -603,6 +584,12 @@ public void onLinksAndHelp() {
     root.requestFocus();
   }
 
+  @Override
+  public void afterPropertiesSet() throws Exception {
+    this.mainWindowTitle = clientProperties.getMainWindowTitle();
```
1-alex98: back to Constructor

src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -108,27 +109,21 @@
...
+  }
 
+  private static URL getPreviewUrl(String mapName, String baseUrl, PreviewSize previewSize) {
+    return noCatch(() -> new URL(format(baseUrl, previewSize.folderName, urlFragmentEscaper().escape(mapName).toLowerCase(Locale.US))));
+  }
+
+  @Override
+  public void afterPropertiesSet() {
+    eventBus.register(this);
+    forgedAlliancePreferences = preferencesService.getPreferences().getForgedAlliance();
```
1-alex98: back to constructor

src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
```diff
@@ -375,4 +356,10 @@ private void addNodeOnTopOfBar(XYChart.Data<String, Integer> data, Node nodeToAd
   public void showLadderMaps(ActionEvent actionEvent) {
     eventBus.post(new ShowLadderMapsEvent());
   }
+
+  @Override
+  public void afterPropertiesSet() throws Exception {
+    youLabel = new Text(i18n.get("ranked1v1.you"));
```
1-alex98: this is does belong in the initialize method because it is java fx related

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -504,4 +478,15 @@ public void ping() {
     }
     writeToServer(PingMessage.INSTANCE);
   }
+
+  @Override
+  public void afterPropertiesSet() throws Exception {
+    addOnMessageListener(NoticeMessage.class, this::onNotice);
```
1-alex98, line 484: this is ok, because it is 100% not needed by others

src/main/java/com/faforever/client/units/CookieService.java
```diff
@@ -19,24 +20,24 @@
...
 
   
   public void setUpCookieManger() {
     CookieManager manager = new CookieManager(new MyCookieStore(), CookiePolicy.ACCEPT_ALL);
     CookieHandler.setDefault(manager);
   }
 
+  @Override
+  public void afterPropertiesSet() throws Exception {
+    Preferences preferences = preferencesService.getPreferences();
```
1-alex98: constructor

1-alex98 (changes requested):

src/main/java/com/faforever/client/FafClientApplication.java
1-alex98: There is a huge difference between the two ways :D The one (preferenceService) loades data from a json for settings and the other just creates a new instance(Spring)

1-alex98: Also the way to add Beans to Spring is to add @Component to the class and not this

src/main/java/com/faforever/client/achievements/AchievementService.java
marcoebbinghaus: as you commented above, this can be done in declaration.

src/main/java/com/faforever/client/chat/PircBotXChatService.java
marcoebbinghaus: i moved initialization to declaration. the rest that is left in afterPropertiesSet was in PostConstruct before I changed the file, so I think the rest might stay here.

marcoebbinghaus:
Thank you for the clarification. I recreated the constructors where necessary, put initialization to declaration and removed afterPropertiesSet where no longer needed.

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -120,26 +121,37 @@
...
   private final PlatformService platformService;
-  private final String faWindowTitle;
   private final DiscordRichPresenceService discordRichPresenceService;
   private final ReplayServer replayServer;
 
   @VisibleForTesting
   RatingMode ratingMode;
 
+  private ObservableList<Game> games;
+  private String faWindowTitle;
```
1-alex98: Should stay final

src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -87,48 +87,66 @@
...
   private final EventBus eventBus;
+  private ForgedAlliancePrefs forgedAlliancePreferences;
 
-  private final String mapDownloadUrlFormat;
-  private final String mapPreviewUrlFormat;
-  private final ForgedAlliancePrefs forgedAlliancePreferences;
-
-  private Map<Path, MapBean> pathToMap;
+  private String mapDownloadUrlFormat;
+  private String mapPreviewUrlFormat;
```
1-alex98: Should stay final

src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -87,48 +87,66 @@
...
+  }
+
+  private static URL getPreviewUrl(String mapName, String baseUrl, PreviewSize previewSize) {
+    return noCatch(() -> new URL(format(baseUrl, previewSize.folderName, urlFragmentEscaper().escape(mapName).toLowerCase(Locale.US))));
+  }
 
+  @Override
+  public void afterPropertiesSet() {
+    eventBus.register(this);
+    forgedAlliancePreferences = preferencesService.getPreferences().getForgedAlliance();
```
1-alex98: doubled now

src/main/java/com/faforever/client/remote/MockFafServerAccessor.java
```diff
@@ -61,31 +61,21 @@
...
-  public MockFafServerAccessor(TaskService taskService, NotificationService notificationService, I18n i18n, EventBus eventBus) {
-    timer = new Timer("LobbyServerAccessorTimer", true);
-    messageListeners = new HashMap<>();
-    connectionState = new SimpleObjectProperty<>();
-    this.taskService = taskService;
-    this.notificationService = notificationService;
-    this.i18n = i18n;
-    this.eventBus = eventBus;
-  }
+  private ObjectProperty<ConnectionState> connectionState = new SimpleObjectProperty<>();
```
1-alex98: final

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -58,18 +59,9 @@
   private Replay replay;
   private Consumer<Replay> onOpenDetailListener;
   private I18n i18n;
-  private InvalidationListener reviewsChangedListener;
+  private InvalidationListener reviewsChangedListener = observable -> populateReviews();
```
1-alex98: final

src/test/java/com/faforever/client/config/FafClientApplicationTest.java
```diff
@@ -30,6 +31,7 @@ public void testDoesItSmoke() throws Exception {
         context.getBeanFactory().registerSingleton("hostService", mock(PlatformService.class));
         context.getBeanFactory().registerSingleton("stage", new Stage());
         context.getBeanFactory().registerSingleton("uidService", mock(UidService.class));
+        context.register(ForgedAlliancePrefs.class);
```
1-alex98: noooooooo! Do not register the Preference classes they need to be obtained by calling preferenceService.getpreferences.ForgedAlliancePrefs();

Spring creates a new instance if u register it. an empty one but the those are the settings loaded fom a json file. 

And further if u would need to register it u would need to annotate it with @Component (like the ForgedAlliancePrefs class) 

ForgedAlliancePrefs can not be Injected by Spring or better it should not be

src/test/java/com/faforever/client/game/GameServiceTest.java
```diff
@@ -133,9 +137,7 @@
   public void setUp() throws Exception {
     junitPlayer = PlayerBuilder.create("JUnit").defaultValues().get();
 
-    ClientProperties clientProperties = new ClientProperties();
-
-    instance = new GameService(clientProperties, fafService, forgedAllianceService, mapService,
+    instance = new GameService(new ClientProperties(), fafService, forgedAllianceService, mapService,
```
1-alex98: No need for changing this why do u change it?

src/test/java/com/faforever/client/remote/ServerAccessorImplTest.java
```diff
@@ -123,13 +127,13 @@ public void setUp() throws Exception {
...
-    ClientProperties clientProperties = new ClientProperties();
-    clientProperties.getServer()
-        .setHost(LOOPBACK_ADDRESS.getHostAddress())
-        .setPort(fafLobbyServerSocket.getLocalPort());
-
-    instance = new FafServerAccessorImpl(preferencesService, uidService, notificationService, i18n, clientProperties, reportingService);
+    final Server server = mock(Server.class);
+    when(server.getHost()).thenReturn(LOOPBACK_ADDRESS.getHostAddress());
+    when(server.getPort()).thenReturn(fafLobbyServerSocket.getLocalPort());
+    when(clientProperties.getServer()).thenReturn(server);
```
1-alex98: We dont wanna change that

1-alex98 (changes requested):

src/test/java/com/faforever/client/config/FafClientApplicationTest.java
marcoebbinghaus: I see and I understand it. I fixed that for FafClientApplication but missed to align the test accordingly.

src/test/java/com/faforever/client/game/GameServiceTest.java
marcoebbinghaus: I can no longer remember. I had to do this as a result of a change I made in production code. But as it seems this is no longer needed (obviously because one of the reverts I had to do). I changed it back to the original implementation before my changes.

src/test/java/com/faforever/client/remote/ServerAccessorImplTest.java
marcoebbinghaus: okay, reverted.

src/main/java/com/faforever/client/chat/MockChatService.java
```diff
@@ -39,40 +39,28 @@
...
+  private Timer timer = new Timer(true);
+  private Collection<Consumer<ChatMessage>> onChatMessageListeners = new ArrayList<>();
+  private Map<String, Channel> channelUserListListeners = new HashMap<>();
+  private ObjectProperty<ConnectionState> connectionState = new SimpleObjectProperty<>(ConnectionState.DISCONNECTED);
+  private IntegerProperty unreadMessagesCount = new SimpleIntegerProperty();
 
   @Override
   public void afterPropertiesSet() {
     eventBus.register(this);
+
```
1-alex98: remove

1-alex98 (changes requested):
Thanks a lot

1-alex98 (approved):

PR #1463, opened 2019-10-12
Tooltip memory leak fix

### LINKED ISSUE #1462: Potential Tooltip memory problem
Current speculation:

If a tooltip is installed on a certain game row/tile (because that game was the last game the user hovered over to show a tooltip), then:
Whenever a lobby game changes (any change, inlcuding team composition / position, ...), additional nodes are evicted from scene graph but cannot be garbage collected until user hovers over another game to show another tooltip. If the game gets closed, the "leaked" nodes will not increase but also not get collected until user hovers over different game.

This means:
- Long lobby sim potentially creates lots of node garbage that is not collected by GC

Data:
single player in lobby, 100 slot/observer changes => 400 direct Labels nodes: 6 in scenegraph, 396 dead and not collectable by GC (were in scenegraph at some point of time, will become collectable once new row is hovered)

Katharsas:
Fixes #1462

### DIFF
--- a/src/main/java/com/faforever/client/fx/JavaFxUtil.java
+++ b/src/main/java/com/faforever/client/fx/JavaFxUtil.java
@@ -129,7 +129,7 @@ public final class JavaFxUtil {
    */
   public static Tooltip createCustomTooltip(Node content) {
     Tooltip tooltip = new Tooltip();
-    PopupWindowHelper.getContent(tooltip).setAll(content);
+    PopupWindowHelper.getContent(tooltip).setAll(content);// effectively calls tooltip.getContent.setAll()
     return tooltip;
   }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/faforever/client/fx/JavaFxUtil.java
```diff
@@ -129,7 +129,7 @@ public static void fixTooltipDuration() {
    */
   public static Tooltip createCustomTooltip(Node content) {
     Tooltip tooltip = new Tooltip();
-    PopupWindowHelper.getContent(tooltip).setAll(content);
+    PopupWindowHelper.getContent(tooltip).setAll(content);// effectively calls tooltip.getContent.setAll()
```
1-alex98: remove

1-alex98 (changes requested):

1-alex98 (approved):

Brutus5000 (approved):

1-alex98 (approved):

PR #1467, opened 2019-10-14
Init ladder queue time

### LINKED ISSUE #1466: Implement initial matchmaker_info request
Related to #1361 
Allows pop time queue to be shown immediately.
Requires https://github.com/FAForever/server/pull/489

Katharsas:
Fixes #1466

### DIFF
--- a/src/main/java/com/faforever/client/game/GameService.java
+++ b/src/main/java/com/faforever/client/game/GameService.java
@@ -27,7 +27,7 @@ import com.faforever.client.player.Player;
 import com.faforever.client.player.PlayerService;
 import com.faforever.client.preferences.NotificationsPrefs;
 import com.faforever.client.preferences.PreferencesService;
-import com.faforever.client.rankedmatch.MatchmakerMessage;
+import com.faforever.client.rankedmatch.MatchmakerInfoMessage;
 import com.faforever.client.remote.FafService;
 import com.faforever.client.remote.ReconnectTimerService;
 import com.faforever.client.remote.domain.GameInfoMessage;
@@ -480,8 +480,8 @@ public class GameService implements InitializingBean {
     return game;
   }
 
-  public void addOnRankedMatchNotificationListener(Consumer<MatchmakerMessage> listener) {
-    fafService.addOnMessageListener(MatchmakerMessage.class, listener);
+  public void addOnRankedMatchNotificationListener(Consumer<MatchmakerInfoMessage> listener) {
+    fafService.addOnMessageListener(MatchmakerInfoMessage.class, listener);
   }
 
   public CompletableFuture<Void> startSearchLadder1v1(Faction faction) {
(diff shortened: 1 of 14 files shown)


1-alex98 (commented):
I can test this on prod?

Katharsas:
No, see issue for required server changes.

Geosearchef (dismissed):

Katharsas:
Works on test.

1-alex98:
@Askaholic did we deployed the needed server change for this

Askaholic:
@axel1200 No it is not deployed yet. It's commit https://github.com/FAForever/server/commit/cc8030143a24b335d3682fa5a0cf69207454738b

Askaholic:
This change has been deployed in [FAForever/server - v1.3](https://github.com/FAForever/server/tree/v1.3.1).

1-alex98:
Is it still deployed with all this everting back to old versions

Askaholic:
You're right. It has been rolled back. I'll comment again when I fix the server issues for good.

Katharsas:
Rebased to fix merge conflicts.

Askaholic:
BTW this is deployed now, although I think ideally we would send the message after login and the client would just not throw it away. Leave it sitting in a message queue somewhere until the ladder tab is opened.

1-alex98:
So should I test or merge or not?

Katharsas:
@Askaholic 
I think this being implemented as a request (like it is right now) is more flexible and, less complex and less error prone.  For example it would allow the client to close & unload the ladder tab and then resend the request to recreate it when needed instead of it always existing.

I see no reason why this information being sent should be specifically bound to login time and why the client should not have the freedom to ask  for it whenever it wants.

Askaholic:
The thing is I'm not a big fan of adding server (and protocol) complexity just for convenience in the client implementation. I'd rather send it at login so it works just like the game list and just like the player list. The client can load and unload the ladder tab all it wants, the queue pop time will be the same until the server sends a new info message. 

It feels like we're letting the client implementation intrude into the protocol by doing this. We're not adding this message because it makes sense for the protocol. Obviously it's not a huge complexity, but it's something that will build up over time, and once it's added it gets very hard to remove.

Katharsas:
With "unloading" i mean forgetting everything about ladder state and having to reinitialize that information from the server.

I guess I just dislike the general protocol architecture then (im used to web world, where generally the client requests what it wants and the server is mostly just a dumb state store, which works generally quite well but is of course very different). But it makes sense that you want to keep the protocol consistent.

1-alex98:
ok an this means for me ?

Katharsas:
Merge. Can still be changed later if somebody wants to build a message queue.

1-alex98 (approved):

1-alex98:
Merging if test run through

PR #1476, opened 2019-10-17
hide ice adapter info window, bump adapter to v2.4.1

Geosearchef:


1-alex98:
But what if ice adpter does not terminate how do i close it (Taskmanager ??)

1-alex98:
Also ho do i turn it on?

Geosearchef:
The ice adapter now has a tray icon, that signals it is still running and when clicked opens the info window able to kill it and show the debug window. (which can also send notifications to your system)

1-alex98:
ok cool

1-alex98 (approved):

Katharsas (commented):
~~I dont see a tray icon appearing (Win 10)~~
Ok i do get the tray icon (and a nice windows notification when it starts). However clicking it does nothing.

Geosearchef:
Is java fx available?

1-alex98:
Probably not he uses jdk 11

Katharsas (commented):
Yeah. Whatever i guess thats not really client issue anyway. You might need to think about this though if switching to JDK 11 becomes a thing. And for the Linux people that dont use the Oracle JDK.

Katharsas (approved):

1-alex98:
Yeah @Geosearchef we need to either apply the workarround we did in the client or u go modular

Issue #1404, opened 2019-08-23
Crash on start


SamuelBadr:
On startup, I get the splash screen for a few seconds and afterwards this error:

https://gist.github.com/Samuel3008/fc36246f9c1609ec835d5acbde65c2a4

What I tried:

- This is on a fresh installation of Windows 10 (I literally did not install anything but Steam and SupCom FA beforehand).
- I do not have a `%appdata%\Forged Alliance Forever` folder.

There are many issues like mine, but all solutions involve deleting some `.prefs` file(s) and it seems I don't have any of those.

Katharsas:
The client crashes with "Access is denied." exception in this line:
`GPG_FA_PATH = Paths.get(Shell32Util.getFolderPath(ShlObj.CSIDL_PERSONAL), "My Games", "Gas Powered Games", "Supreme Commander Forged Alliance");`

This means it doesn't have the rights to or is otherwise prevented by your system from accessing
`C:\Users\<user>\Documents\my games\Gas Powered Games\Supreme Commander Forged Alliance`

SamuelBadr:
This did it, thank you!

I'm running Windows inside of a VM (Parallels Desktop) which automatically shares the host user folder (`~`) with Windows'. Turning that off allowed the client to start normally.

PR #1483, opened 2019-10-19
Catching the shell32 and invalidPath errors

### LINKED ISSUE #1405: Show error message to user instead of crashing...
...if a needed directory cannot be accessed at startup, see https://github.com/FAForever/downlords-faf-client/issues/1404
Or make the log more descriptive at least.

adamrmelnyk:
* The error at https://gist.github.com/Samuel3008/fc36246f9c1609ec835d5acbde65c2a4
  is thrown by the Shell32UTil class when it can't access a folder
* Potential error in Paths.get as well when the file cannot be accessed
* Added a wrapper to catch the errors

Fixes #1405

### DIFF
--- a/src/main/java/com/faforever/client/preferences/ForgedAlliancePrefs.java
+++ b/src/main/java/com/faforever/client/preferences/ForgedAlliancePrefs.java
@@ -3,6 +3,7 @@ package com.faforever.client.preferences;
 import com.faforever.client.preferences.gson.ExcludeFromGson;
 import com.sun.jna.platform.win32.Shell32Util;
 import com.sun.jna.platform.win32.ShlObj;
+import com.sun.jna.platform.win32.Win32Exception;
 import javafx.beans.binding.Bindings;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.ObjectProperty;
@@ -10,11 +11,17 @@ import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.property.SimpleStringProperty;
 import javafx.beans.property.StringProperty;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.logging.log4j.util.Strings;
 
 import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
+import static javax.swing.JOptionPane.showMessageDialog;
+
+@Slf4j
 public class ForgedAlliancePrefs {
 
   public static final Path GPG_FA_PATH;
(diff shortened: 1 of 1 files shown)


1-alex98:
not acceptable... it only delays the problem and client gets unfunctional

1-alex98:
Either we need to find out why this Exception happens or we find a more fail save way to do it

1-alex98:
Well we found the issue in the referenced issue https://github.com/FAForever/downlords-faf-client/issues/1404 . We can probably not do much apart from showing a better description of the error

Katharsas:
I think right now (without this PR) a swing window opens and shows a convoluted, non-helpful exception and stacktrace. All that is needed is that this window instead shows a nice error message (in english only if not i18n is not started) which includes the path that is the problem. And log a nicer exception (which includes the problematic path as well) to the log.

Not sure how to test this though, maybe need to play around with file security, maybe @adamrmelnyk you have an idea?

adamrmelnyk:
> not acceptable... it only delays the problem and client gets unfunctional

@axel1200 Not having file access is difficult to recover from. Wouldn't you need to restart the application?

>Either we need to find out why this Exception happens or we find a more fail save way to do it

I have a better explaination in the ticket about what's happening but this is more of a work in progress about what we should be displaying. I didn't want to go too crazy without some feedback as to how it would be displayed etc.

1-alex98:
I dont now but just catching it does not seem a solution to me

adamrmelnyk:
I'm open to trying other solutions if you have another way in mind or any suggestions.

My thinking for catching and displaying an error is that the file being inaccessible can't be solved by the application. The user will need to change that themselves so displaying the error can at least have an idea about how to go about diagnosing the problem.

adamrmelnyk:
@Katharsas I we can write some unit tests for this but it's tricky because the error is thrown by a static method, so to make that happen and at least test our code path we either we wrap the `Shell32Util.getFolderPath` method and mock that method, or we use a different library like PowerMokito which I assume does something similar.

This will be easy to test once you have it built though, you can just lock down the permissions on that file. I don't have a windows machine, so I can't reproduce that outside of code. Let me know what else I can do here so we can move forward.

Katharsas:
Yeah persmissions should work. Ill come back to this and review if I have some more time , don't right now, sorry.

Katharsas:
Ok i had some time to look at this.

1. >- Potential error in Paths.get as well when the file cannot be accessed.

   I would remove that. Ìf `Paths.get` throws an exception, it will have a nice error message already so i dont see the benefit of catching it (and the PR is inconsistent because it only does this for Linux paths).

2. > This will be easy to test once you have it built though, you can just lock down the permissions on that file. I don't have a windows machine, so I can't reproduce that outside of code. Let me know what else I can do here so we can move forward.

   I just tried and i can not reproduce the `Win32Exception` by locking down the permissions (i removed all permissions for my normal user on the folder "My Documents"). So unless we find a way to do that, we cannot procede. Btw. i don't like to work with PRs that contain code that only works in theory. All PR code should be tested and working before it is reviewed. This needs to be coded/continued by somebody who can test this on windows.

4. If you print error to log, print path to the log as well (just like for dialog). Btw. i think you are using the `log.error(formatString, arg1, arg2)` by accident and actually want `log.error(message, throwable)`

5. I dont think its necessary to show any dialogs. Just throwing a new exception with better message should be better and in addition you dont want to return from any function as if nothing happened (just `throw new UncheckedIOException(message, cause)` and let spring show the exception for you). Right now in case of failure, the PR code will probably set all important paths to empty string and then happily continue with tons of errors instead of cleanly exiting by throwing RuntimeException.

1-alex98:
U got one month till I close it

PR #1496, opened 2019-11-03
#1399 Sort games by average rating

Brutus5000:


norraxx (commented):
I think, we would like to sort games not only by rating.
We have tile and table, and in table is more sorting, than only by rating.

Brutus5000:
Hu? This adds an additional option to sort by average rating in the tile view + a new average rating column in the table view. This doesn't remove any existing sorting options.

1-alex98 (commented):
Looks good still need to test it

1-alex98 (approved):

1-alex98:
nooooo 
XD

Issue #1442, opened 2019-09-16
Map download issue on Linux


Katharsas:
A bug was reported where Maps fail to download on Linux since the RC 1.0

Katharsas:
Ok, so there is an exception thrown by `MapService # addSkirmishMap(...)` that isn't printed anywhere:
```
java.lang.NullPointerException: null
	at java.base/java.nio.file.Files.provider(Files.java:100) ~[na:na]
	at java.base/java.nio.file.Files.isDirectory(Files.java:2228) ~[na:na]
	at com.faforever.client.map.MapService.readMap(MapService.java:258) ~[main/:na]
	at com.faforever.client.map.MapService.addSkirmishMap(MapService.java:237) ~[main/:na]
	at com.faforever.client.map.MapService.lambda$downloadAndInstallMap$17(MapService.java:492) ~[main/:na]
	at com.github.nocatch.NoCatch.noCatch(NoCatch.java:49) ~[nocatch-1.1.jar:na]
	at com.github.nocatch.NoCatch.noCatch(NoCatch.java:34) ~[nocatch-1.1.jar:na]
	at com.faforever.client.map.MapService.lambda$downloadAndInstallMap$18(MapService.java:492) ~[main/:na]
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:714) ~[na:na]
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506) ~[na:na]
	at java.base/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2073) ~[na:na]
	at com.faforever.client.task.CompletableTask.lambda$new$2(CompletableTask.java:18) ~[main/:na]
	at com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86) ~[javafx-base-11-linux.jar:na]
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:238) ~[javafx-base-11-linux.jar:na]
	at com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:191) ~[javafx-base-11-linux.jar:na]
	at com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:58) ~[javafx-base-11-linux.jar:na]
[...]

PR #1497, opened 2019-11-04
workaround map downlodaing issue on linux

betrok:
There are many maps in the vault, where the case of the archive name (always lowercase?) does not coincide with name of the directory in it. This leads to #1442.

Here is a workaround for case sensitive filesystems.

Maybe repacking all the maps would be a cleaner solution...


src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -412,6 +413,21 @@ public Path getPathForMap(String technicalName) {
...
 
+  public Path getPathForMapInsensitive(String approxName) {
+    String normalizedName = approxName.toLowerCase();
+    try (DirectoryStream<Path> stream = Files.newDirectoryStream(customMapsDirectory)) {
+      for (Path entry : stream) {
+        if (entry.getFileName().toString().toLowerCase().equals(normalizedName)) {
+          return entry;
+        }
+      }
+    } catch (IOException e) {
```
1-alex98: lets simply not catch it at all, returning nul only delays the issue and probably results in a NPE, just use 

```suggestion
noCatch(()->{
     //Your code
});
```

src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -485,7 +501,7 @@ public void evictCache() {
     }
 
     return taskService.submitTask(task).getFuture()
-        .thenAccept(aVoid -> noCatch(() -> addSkirmishMap(getPathForMap(folderName))));
```
1-alex98: What about the old method ? Are those all places where this issue exists

1-alex98 (changes requested):

1-alex98:
I would be happy if u join our chat service Zulip, send me your email and I will invite u.
alexander.von.trostorff@gmail.com

src/main/java/com/faforever/client/map/MapService.java
betrok: At first glance we know exact name of the directory everywhere else.

src/main/java/com/faforever/client/map/MapService.java
betrok: Some code actually uses null as a missing map indicator, hm...

betrok:
My public email should be available in the github profile. It can be used for an invite if necessary.

src/main/java/com/faforever/client/map/MapService.java
1-alex98: ok then

betrok:
I use client build with these changes for a week already and it seems to work fine enough.
Can someone take a look and maybe merge it?

1-alex98 (approved):

1-alex98:
tested on Windows works

PR #1500, opened 2019-11-04
Exponential Backoff

### LINKED ISSUE #1499: Reconnect loop is far to quick
The 3 seconds reconnect DDoses the server is the server is down for a while and comes back up

1-alex98:
Fixes #1499

### DIFF
--- a/src/main/java/com/faforever/client/game/GameService.java
+++ b/src/main/java/com/faforever/client/game/GameService.java
@@ -28,6 +28,7 @@ import com.faforever.client.preferences.NotificationsPrefs;
 import com.faforever.client.preferences.PreferencesService;
 import com.faforever.client.rankedmatch.MatchmakerMessage;
 import com.faforever.client.remote.FafService;
+import com.faforever.client.remote.ReconnectTimerService;
 import com.faforever.client.remote.domain.GameInfoMessage;
 import com.faforever.client.remote.domain.GameLaunchMessage;
 import com.faforever.client.remote.domain.GameStatus;
@@ -127,6 +128,7 @@ public class GameService implements InitializingBean {
   private final PlatformService platformService;
   private final DiscordRichPresenceService discordRichPresenceService;
   private final ReplayServer replayServer;
+  private final ReconnectTimerService reconnectTimerService;
 
   @VisibleForTesting
   RatingMode ratingMode;
@@ -156,7 +158,8 @@ public class GameService implements InitializingBean {
                      ModService modService,
                      PlatformService platformService,
                      DiscordRichPresenceService discordRichPresenceService,
-                     ReplayServer replayServer) {
+                     ReplayServer replayServer,
+                     ReconnectTimerService reconnectTimerService) {
     this.fafService = fafService;
     this.forgedAllianceService = forgedAllianceService;
     this.mapService = mapService;
@@ -173,6 +176,7 @@ public class GameService implements InitializingBean {
     this.platformService = platformService;
     this.discordRichPresenceService = discordRichPresenceService;
     this.replayServer = replayServer;
+    this.reconnectTimerService = reconnectTimerService;
 
     faWindowTitle = clientProperties.getForgedAlliance().getWindowTitle();
     uidToGameInfoBean = FXCollections.observableMap(new ConcurrentHashMap<>());
@@ -200,6 +204,7 @@ public class GameService implements InitializingBean {
         item -> new Observable[]{item.statusProperty(), item.getTeams()}
     );
     JavaFxUtil.attachListToMap(games, uidToGameInfoBean);
+    JavaFxUtil.addListener(gameRunning, (observable, oldValue, newValue) -> reconnectTimerService.setGameRunning(newValue));
   }
 
   @NotNull
--- a/src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
+++ b/src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
@@ -85,9 +85,8 @@ import javafx.beans.property.ReadOnlyObjectProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.concurrent.Task;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.compress.utils.IOUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.cache.annotation.Cacheable;
@@ -101,7 +100,6 @@ import org.springframework.util.StringUtils;
 
 import java.io.IOException;
 import java.io.OutputStream;
-import java.lang.invoke.MethodHandles;
 import java.net.Socket;
 import java.net.URL;
 import java.time.Duration;
@@ -122,11 +120,10 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 @Component
 @Profile("!" + FafClientApplication.PROFILE_OFFLINE)
 @RequiredArgsConstructor
+@Slf4j
 public class FafServerAccessorImpl extends AbstractServerAccessor implements FafServerAccessor,
     InitializingBean, DisposableBean {
 
-  private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  private static final long RECONNECT_DELAY = 3000;
   private Gson gson = new GsonBuilder()
       .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
       .registerTypeAdapter(VictoryCondition.class, VictoryConditionTypeAdapter.INSTANCE)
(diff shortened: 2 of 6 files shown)


src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -200,6 +204,7 @@ public GameService(ClientProperties clientProperties,
         item -> new Observable[]{item.statusProperty(), item.getTeams()}
     );
     JavaFxUtil.attachListToMap(games, uidToGameInfoBean);
+    JavaFxUtil.addListener(gameRunning, (observable, oldValue, newValue) -> reconnectService.setGameRunning(newValue));
```
Brutus5000: I'm a bit confused. We have two services talking to each other. No UI involeved. Why are we using JavaFxUtil here? We should keep out JavaFx stuff wherever possible, don't we?

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -164,6 +165,8 @@
   private Socket fafServerSocket;
   private CompletableFuture<List<Avatar>> avatarsFuture;
   private CompletableFuture<List<IceServer>> iceServersFuture;
+  @Setter
```
Brutus5000: I would suggest to change this state via Events on the message bus.

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -247,15 +250,17 @@ protected Void call() throws Exception {
 
             logger.info("FAF server connection established");
             Platform.runLater(() -> connectionState.set(ConnectionState.CONNECTED));
+            reconnectService.resetConnectionFailures();
```
Brutus5000: This also belongs in the reconnect service (and could be also be triggered by events on the message bus?)

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -247,15 +250,17 @@ protected Void call() throws Exception {
...
             blockingReadServer(fafServerSocket);
           } catch (IOException e) {
             Platform.runLater(() -> connectionState.set(ConnectionState.DISCONNECTED));
             if (isCancelled()) {
               logger.debug("Connection to FAF server has been closed");
             } else {
-              logger.warn("Lost connection to FAF server, trying to reconnect in " + RECONNECT_DELAY / 1000 + "s", e);
-              Thread.sleep(RECONNECT_DELAY);
+              logger.warn("Lost connection to Server", e);
+              reconnectService.incrementConnectionFailures();
```
Brutus5000: The reconnect service should track that itself

src/main/java/com/faforever/client/remote/ReconnectService.java
```diff
@@ -0,0 +1,63 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+
+import java.util.Random;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+@Service
+@Slf4j
+public class ReconnectService {
```
Brutus5000: I am unclear about responsibilities here. I would assume a reconnect service is responsible for invoking the reconnect. But as it seems it is merely a simple util.

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -247,15 +250,17 @@ protected Void call() throws Exception {
...
 
             blockingReadServer(fafServerSocket);
           } catch (IOException e) {
             Platform.runLater(() -> connectionState.set(ConnectionState.DISCONNECTED));
             if (isCancelled()) {
               logger.debug("Connection to FAF server has been closed");
             } else {
-              logger.warn("Lost connection to FAF server, trying to reconnect in " + RECONNECT_DELAY / 1000 + "s", e);
-              Thread.sleep(RECONNECT_DELAY);
+              logger.warn("Lost connection to Server", e);
```
Brutus5000: Can we switch to `@Slf4J` in this class?

Brutus5000 (changes requested):

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
Geosearchef: An event here could be really useful for other cases later on aswell.

src/main/java/com/faforever/client/remote/ReconnectService.java
```diff
@@ -0,0 +1,63 @@
...
+    connectionFailures++;
+  }
+
+  @VisibleForTesting
+  long getReconnectTimeOut() {
+    if (gameRunning) {
+      return RECONNECT_DELAY_FACTOR;
+    }
+    double timeToWait = RECONNECT_DELAY_FACTOR * Math.pow(2, connectionFailures);
+    timeToWait = timeToWait > RECONNECT_DELAY_MAX ? RECONNECT_DELAY_MAX : timeToWait;
```
Geosearchef: Use Math.min()

src/main/java/com/faforever/client/remote/ReconnectService.java
```diff
@@ -0,0 +1,63 @@
...
+   * The maximum time somebody will ever wait to reconnect. Though don't worry in the case you are in a game it will
+   * always be RECONNECT_DELAY_FACTOR .
+   */
+  @VisibleForTesting
+  static final long RECONNECT_DELAY_MAX = 80000;
+  /**
+   * The time we start with waiting till we reconnect. After failures the time will increase exponentially.
+   */
+  private static final long RECONNECT_DELAY_FACTOR = 3000;
+  private static final Random RECONNECT_RANDOM = new Random(System.currentTimeMillis());
```
Geosearchef: No need to initialize with system time. Also no constant, although final, pls do not write in caps.

src/main/java/com/faforever/client/remote/ReconnectService.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @VisibleForTesting
+  long getReconnectTimeOut() {
+    if (gameRunning) {
+      return RECONNECT_DELAY_FACTOR;
+    }
+    double timeToWait = RECONNECT_DELAY_FACTOR * Math.pow(2, connectionFailures);
+    timeToWait = timeToWait > RECONNECT_DELAY_MAX ? RECONNECT_DELAY_MAX : timeToWait;
+    return (long) (RECONNECT_RANDOM.nextFloat() * timeToWait);
```
Geosearchef: Consider just using Math.random(), it's static and has an internal LCPRNG.

Geosearchef (changes requested):

src/main/java/com/faforever/client/remote/ReconnectService.java
1-alex98: Why and of cause I know what this means 😛

src/main/java/com/faforever/client/remote/ReconnectService.java
1-alex98: SURE this is a constant.... THAT_IS_HOW_IT_IS_DONE

1-alex98: U are right about init tho

src/main/java/com/faforever/client/game/GameService.java
1-alex98: It is for thread safety. Won't harm and should always be used

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
1-alex98: as u please

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
1-alex98: I put it on the event bus then

src/main/java/com/faforever/client/remote/ReconnectService.java
1-alex98: Renamed it

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
1-alex98: ok

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
1-alex98: actually no, events are bad here as they are async and create the need for synchronization

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
1-alex98: actually no, events are bad here as they are async and create the need for synchronization

1-alex98:
@Brutus5000 if u want this in next release tell me if it is ok? Or what to change

Brutus5000:
I don't get your point of problems with the async events as the whole connectivity topic is async (or at least should be? disconnects can happen anytime). But apart from that it looks fine.

1-alex98:
hard to explain but it is an issue here

PR #1507, opened 2019-11-12
Local replay vault watching/reloading

### LINKED ISSUE #1141: Local replays not updating till client restarted
Can we at least have a refresh button for local replays as I don't think any local replays of games in the current session are shown. Client needs to be restarted.

### LINKED ISSUE #1506: Local Replay Vault not refreshed after games played
**Describe the bug**
After the local replay vault has been loaded once, it is never refreshed to include additional games. The only way to refresh it seems to be to restart the client, which does work reliably to display all recent replays.

**To Reproduce**
1. Open client
2. Show local replay vault
3. Play a game
4. Show local replay vault 

**Expected behavior**
The replays of all recent games should be displayed.

**OS / Version**
Windows 10, Client v1.0.0

andreasley:
Fixes #1141.

This patch implements lazy loading of the local replay vault and automatic updates when replays have been added to or removed from the local replay directory.

- Uses WatchEvents to detect changes in the local replay directory
- Works for both played games and manually added/removed replay files
- Updates the list accordingly and respects previous sorting
- Replays from played games are now written to the cache directory first and then moved atomically  to the replays directory to prevent parsing it before it is written
- No changes to UI

Test coverage is not great. What would you like to have tested exactly?

It's been 10 years or so since I've written any Java, so be gentle. ;)

### DIFF
--- a/src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
+++ b/src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
@@ -1,11 +1,11 @@
 package com.faforever.client.main.event;
 
 import com.faforever.client.replay.Replay;
-import lombok.Data;
+import lombok.Value;
 
 import java.util.Collection;
 
-@Data
+@Value
 public class LocalReplaysChangedEvent {
   private final Collection<Replay> newReplays;
   private final Collection<Replay> deletedReplays;
--- a/src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
+++ b/src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
@@ -2,12 +2,27 @@ package com.faforever.client.main.event;
 
 import com.faforever.client.replay.Replay;
 import lombok.Value;
+import org.springframework.context.ApplicationEvent;
 
 import java.util.Collection;
 
-@Value
-public class LocalReplaysChangedEvent {
+public class LocalReplaysChangedEvent extends ApplicationEvent {
+
   private final Collection<Replay> newReplays;
   private final Collection<Replay> deletedReplays;
+
+  public LocalReplaysChangedEvent(Object source, Collection<Replay> newReplays, Collection<Replay> deletedReplays) {
+    super(source);
+    this.newReplays = newReplays;
+    this.deletedReplays = deletedReplays;
+  }
+
+  public Collection<Replay> getNewReplays() {
+    return newReplays;
+  }
+
+  public Collection<Replay> getDeletedReplays() {
+    return deletedReplays;
+  }
 }
 
--- a/src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
+++ b/src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
@@ -1,0 +1,13 @@
+package com.faforever.client.main.event;
+
+import com.faforever.client.replay.Replay;
+import lombok.Data;
+
+import java.util.Collection;
+
+@Data
+public class LocalReplaysChangedEvent {
+  private final Collection<Replay> newReplays;
+  private final Collection<Replay> deletedReplays;
+}
+
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -39,7 +39,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.eclipse.jgit.util.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.InitializingBean;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.context.event.EventListener;
@@ -50,12 +49,10 @@ import org.springframework.web.util.UriComponentsBuilder;
 import java.io.ByteArrayInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.io.RandomAccessFile;
 import java.lang.invoke.MethodHandles;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.nio.channels.FileLock;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -68,6 +65,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
@@ -184,15 +182,8 @@ public class ReplayService {
       Path fullPathToReplay = preferencesService.getReplaysDirectory().resolve(path);
 
       if (watchEvent.kind() == ENTRY_CREATE) {
-        try {
-          loadLocalReplay(fullPathToReplay)
-              .thenAccept(newReplay -> {
-                newReplays.add(newReplay);
-              });
-        } catch (Exception e) {
-          logger.warn("Failed to load local replay file '{}'", path, e);
-        }
-
+        tryLoadingLocalReplay(fullPathToReplay)
+            .thenAccept(newReplay -> newReplays.add(newReplay));
       } else if (watchEvent.kind() == ENTRY_DELETE) {
         Optional<Replay> existingReplay = localReplays.stream().filter(replay -> replay.getReplayFile().compareTo(fullPathToReplay) == 0).findFirst();
         if (existingReplay.isPresent()) {
@@ -264,14 +255,7 @@ public class ReplayService {
     List<CompletableFuture<Replay>> replayFutures = StreamSupport.stream(directoryStream.spliterator(), false)
         .sorted(Comparator.comparing(path -> noCatch(() -> Files.getLastModifiedTime((Path) path))).reversed())
         .limit(MAX_REPLAYS)
-        .map( replayFile -> {
-          try {
-            return loadLocalReplay(replayFile);
-          } catch (Exception e) {
-            // do nothing; notification should already be shown to user
-            return null;
-          }
-        })
+        .map( replayFile -> tryLoadingLocalReplay(replayFile))
         .filter(Objects::nonNull)
         .collect(Collectors.toList());
 
@@ -285,7 +269,7 @@ public class ReplayService {
   }
 
   @Async
-  private CompletableFuture<Replay> loadLocalReplay(Path replayFile) throws Exception  {
+  private CompletableFuture<Replay> tryLoadingLocalReplay(Path replayFile)  {
     try {
       LocalReplayInfo replayInfo = replayFileReader.parseMetaData(replayFile);
 
@@ -302,7 +286,7 @@ public class ReplayService {
     } catch (Exception e) {
       logger.warn("Could not read replay file '{}'", replayFile, e);
       moveCorruptedReplayFile(replayFile);
-      throw e;
+      return CompletableFuture.failedFuture(e);
     }
   }
 
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -254,21 +254,23 @@ public class ReplayService {
       noCatch(() -> createDirectories(replaysDirectory));
     }
 
-    DirectoryStream<Path> directoryStream = Files.newDirectoryStream(replaysDirectory, replayFileGlob);
-    List<CompletableFuture<Replay>> replayFutures = StreamSupport.stream(directoryStream.spliterator(), false)
-        .sorted(Comparator.comparing(path -> noCatch(() -> Files.getLastModifiedTime((Path) path))).reversed())
-        .limit(MAX_REPLAYS)
-        .map( replayFile -> tryLoadingLocalReplay(replayFile))
-        .filter(e -> !e.isCompletedExceptionally())
-        .collect(Collectors.toList());
 
-    CompletableFuture[] replayFuturesArray = replayFutures.toArray(new CompletableFuture[replayFutures.size()]);
-    return CompletableFuture.allOf(replayFuturesArray)
-        .thenApply(ignoredVoid ->
-            replayFutures.stream()
-                .map(future -> future.join())
-                .collect(Collectors.toList()));
+    try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(replaysDirectory, replayFileGlob)) {
+      List<CompletableFuture<Replay>> replayFutures = StreamSupport.stream(directoryStream.spliterator(), false)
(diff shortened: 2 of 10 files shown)


src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -111,6 +127,85 @@
...
+      localReplays.clear();
+      localReplays.addAll(replays);
+      eventBus.post(new LocalReplaysChangedEvent(replays, new ArrayList<Replay>()));
+    });
+
+    try {
+      Optional.ofNullable(directoryWatcherThread).ifPresent(Thread::interrupt);
+      directoryWatcherThread = startDirectoryWatcher(replaysDirectory);
+    } catch (IOException e) {
+      logger.debug("Failed to start watching the local replays directory");
```
Brutus5000: Should be a warn at least, otherwise it will not end up in logs. Also append the exception as parameter to the logger, so it will print the stacktrace.

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -111,6 +127,85 @@
...
+
+    LoadLocalReplaysTask loadLocalReplaysTask = applicationContext.getBean(LoadLocalReplaysTask.class);
+    taskService.submitTask(loadLocalReplaysTask).getFuture().thenAccept( replays -> {
+      localReplays.clear();
+      localReplays.addAll(replays);
+      eventBus.post(new LocalReplaysChangedEvent(replays, new ArrayList<Replay>()));
+    });
+
+    try {
+      Optional.ofNullable(directoryWatcherThread).ifPresent(Thread::interrupt);
```
Brutus5000, line 148: I am not entirely sure, when would this happen? Is this a regular case?

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -156,35 +251,59 @@ static String guessModByFileName(String fileName) {
...
-    }
+    DirectoryStream<Path> directoryStream = Files.newDirectoryStream(replaysDirectory, replayFileGlob);
+    List<CompletableFuture<Replay>> replayFutures = StreamSupport.stream(directoryStream.spliterator(), false)
+        .sorted(Comparator.comparing(path -> noCatch(() -> Files.getLastModifiedTime((Path) path))).reversed())
+        .limit(MAX_REPLAYS)
+        .map( replayFile -> {
+          try {
+            return loadLocalReplay(replayFile);
+          } catch (Exception e) {
+            // do nothing; notification should already be shown to user
```
Brutus5000: Before doing nothing, trace-log it instead

src/test/java/com/faforever/client/vault/replay/ReplayVaultControllerTest.java
```diff
@@ -78,24 +74,22 @@ public void testGetRoot() throws Exception {
 
   @Test
   public void testLoadLocalReplaysInBackground() throws Exception {
-    LoadLocalReplaysTask task = mock(LoadLocalReplaysTask.class);
-    when(task.getFuture()).thenReturn(CompletableFuture.completedFuture(Arrays.asList(
+
+    var replays = Arrays.asList(
```
Brutus5000: We don't use var in this codebase.

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -156,35 +251,59 @@ static String guessModByFileName(String fileName) {
...
+            return null;
+          }
+        })
+        .filter(replay -> replay != null)
```
Brutus5000: Objects::requireNonNull

Brutus5000 (changes requested):
Did not look into that composing of CompletableFutures. The whole joining futures thing seems a bit odd, but I'm no expert on it and don't have the time to dig deeper into it right now.

src/main/java/com/faforever/client/replay/ReplayService.java
andreasley: In theory, never. In practice, some tests or future code might try to call the method multiple times. This line prevents multiple parallel threads with a watcher for this directory. I've borrowed this line from where another watcher was already implemented in the client.

src/main/java/com/faforever/client/replay/ReplayService.java
andreasley: The method `loadLocalReplay` already logs the exception, that's why I omitted it here.

src/main/java/com/faforever/client/replay/ReplayService.java
andreasley: Not the same thing though; this line filters out entries that are null from a stream. I'm personally not a big fan of chained method calls, but streams sometimes don't easily allow another way. A stream was already used here and I tried to limit my changes to a minimum for my first PR. :)

Brutus5000: Sorry wrong method. Objects::nonNull does the same as the lambda

src/main/java/com/faforever/client/replay/ReplayService.java
andreasley: Fixed.

src/test/java/com/faforever/client/vault/replay/ReplayVaultControllerTest.java
andreasley: Fixed.

src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
```diff
@@ -0,0 +1,13 @@
+package com.faforever.client.main.event;
+
+import com.faforever.client.replay.Replay;
+import lombok.Data;
+
+import java.util.Collection;
+
+@Data
+public class LocalReplaysChangedEvent {
```
norraxx: Does data holder class represents `event`?

norraxx: I mean maybe we could give here better naming.

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
```diff
@@ -104,8 +114,22 @@ public void initialize() {
...
+      loadLocalReplaysInBackground();
+      isDisplayingForFirstTime = false;
+    }
 
-    loadLocalReplaysInBackground();
+    super.onDisplay(navigateEvent);
+  }
+
+  protected void loadLocalReplaysInBackground() {
+    // TODO: Display information that view contents are loading
```
norraxx: Is something "blocking" us to not to have TODO? I mean, if we leave that todo here, this will stay for ages :-)

src/test/java/com/faforever/client/replay/ReplayServiceTest.java
```diff
@@ -122,13 +139,17 @@
...
+  @Mock
+  private EventBus eventBus;
+  @Mock
+  private ExecutorService executorService;
 
   @Before
   public void setUp() throws Exception {
     MockitoAnnotations.initMocks(this);
 
     instance = new ReplayService(new ClientProperties(), preferencesService, replayFileReader, notificationService, gameService,
```
norraxx, line 151: I would like to rename that `instance`... For example to that `replayService`

norraxx: Same to below, but it's just nice to have :-)

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
```diff
@@ -53,12 +59,13 @@
...
 
 import static java.util.Arrays.asList;
+import static java.util.Arrays.setAll;
 
 @Component
 @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
 @RequiredArgsConstructor
 // TODO reduce dependencies
-public class ReplayVaultController implements Controller<Node> {
+public class ReplayVaultController extends AbstractViewController<Node> {
```
norraxx, line 59: 👍

src/main/java/com/faforever/client/replay/ReplayService.java
norraxx: Ok, then, why do we need to raise exception from `loadLocalReplay`? Maybe just ommit it, if everything is logged and "fixed"?

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -111,6 +127,85 @@
...
+          loadLocalReplay(fullPathToReplay)
+              .thenAccept(newReplay -> {
+                newReplays.add(newReplay);
+              });
+        } catch (Exception e) {
+          logger.warn("Failed to load local replay file '{}'", path, e);
+        }
+
+      } else if (watchEvent.kind() == ENTRY_DELETE) {
+        Optional<Replay> existingReplay = localReplays.stream().filter(replay -> replay.getReplayFile().compareTo(fullPathToReplay) == 0).findFirst();
```
norraxx: IMHO too long line.

norraxx (commented):
For me just "nice to have" small things.

norraxx:
Good job! ;-)

src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
andreasley: Not sure what you mean. The `@Data` annotation is part of Lombok and automatically generates getters/setters etc.

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
andreasley: :D 
I do believe that small incremental improvements are better than big changesets. That said, I'd promise to look at that TODO once the PR is merged. :)

src/test/java/com/faforever/client/replay/ReplayServiceTest.java
andreasley: This one is actually not from me. :) In all tests on this project, the respective object is stored in a property named `instance`.

norraxx: Yes, i know.

src/main/java/com/faforever/client/replay/ReplayService.java
andreasley: Right you are.

src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
Brutus5000: To be more precise, the event is a value type right? (Create once, read only)
In this case @Value would be the better lombok annotation as it makes the variables final and the object immutable.

Geosearchef (dismissed):
please squash commits

src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
andreasley: Changed.

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
andreasley: Fixed.

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -155,36 +245,52 @@ static String guessModByFileName(String fileName) {
...
-
-              mapService.findByMapFolderName(replayInfo.getMapname())
-                  .thenAccept(mapBean -> replayInfos.add(new Replay(replayInfo, replayFile, featuredMod, mapBean.orElse(null))));
-            } catch (Exception e) {
-              logger.warn("Could not read replay file '{}'", replayFile, e);
-              moveCorruptedReplayFile(replayFile);
-            }
-          });
-    }
+    DirectoryStream<Path> directoryStream = Files.newDirectoryStream(replaysDirectory, replayFileGlob);
```
Brutus5000: Why was try-with-resources removed? Now we have an unclosed stream.

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -111,6 +125,82 @@
   private final FafService fafService;
   private final ModService modService;
   private final MapService mapService;
+  private final EventBus eventBus;
```
1-alex98: We try to not use EventBus any longer but use ApplicationPublisher from Spring for new changes / features etc.

1-alex98 (changes requested):

src/main/java/com/faforever/client/replay/ReplayService.java
andreasley: @Brutus5000 said the opposite in Zulip. Which one is it? :)

1-alex98: EventBus is deprecated

1-alex98: If he said the opposite he got it the wrong way arround

andreasley: I misremembered; @Brutus5000 said to use Spring. @Geosearchef and @norraxx, however, seemed to prefer EventBus. I really don't mind either way, just want to be sure we're all on the same page.
I can't find anything about `EventBus` being deprecated. Where did you hear that?

1-alex98: Sorry i meant unstable/Beta 
image

1-alex98: I think it was even marked with unstable in older guava versions, also we wanna stay with Spring. In the beginning tho nobody knew Spring had an Event System as well.

andreasley: I've replaced `EventBus` with Spring methods for this PR.
Since Spring's event system works a bit differently, I had to remove the `@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)` from the ReplayVaultController, because otherwise, Spring would instantiate a new controller whenever the Event was triggered.

Brutus5000 (changes requested):

src/main/java/com/faforever/client/replay/ReplayService.java
andreasley: An oversight. Thanks for noticing!

Brutus5000 (approved):

src/main/java/com/faforever/client/main/event/LocalReplaysChangedEvent.java
```diff
@@ -2,12 +2,27 @@
 
 import com.faforever.client.replay.Replay;
 import lombok.Value;
+import org.springframework.context.ApplicationEvent;
 
 import java.util.Collection;
 
-@Value
-public class LocalReplaysChangedEvent {
+public class LocalReplaysChangedEvent extends ApplicationEvent {
```
1-alex98, line 9: Extending Application Event is not needed

1-alex98: I mean u can for sure it won't hurt, we did not do this in any other cases :thinking: not sure whats best here

1-alex98 (approved):

Issue #1292, opened 2019-06-12
Improve localization of server messages


Askaholic:
Client issue for FAForever/server#444.

In an effort to improve localization of server events, we should be sending more json descriptions and less English text. Of course this will require the client to handle such messages appropriately.

See the server issue for more details.

1-alex98:
The java server has a number for all server error messages, if you do the same we can someday translate then all by number

Askaholic:
Hm. IDK if I'm a fan of error code numbers, but I can certainly mention them on the github issue if that helps integrate things on the client side.

Brutus5000:
Had a recent talk at work with a frontend developer at work. He said rather than a "random" error number, we should just use a unique translation key, as he wouldn't need to add an additional mapping from magic number to translation key.
Same applies to the client here. We should use translation keys inside the message (if required then in addition to the error id).

micheljung:
@Brutus5000 I disagree: error codes are not only very common but also super easy and 100% fail-safe. The same can't be said about translation keys; they're subject to change, call for a naming convention that will always fall short at some point, they're [easy to mess up](https://github.com/FAForever/downlords-faf-client/blob/develop/src/main/resources/i18n/messages.properties#L511) and once they need to be changed for whatever reason, they're API-breaking.
Error codes have none of these issues.

PR #1512, opened 2019-11-24
Issue/#1292 Localize authentication_failed messages

Askaholic:
Related #1292.

Add localization for some messages that are currently hard coded strings in the lobby server.

Managed to to get these working with my server PR https://github.com/FAForever/server/pull/505. I plan on adding more messages in a separate PR after I get more comfortable with how all the pieces of the code work together.

### DIFF
--- a/src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
+++ b/src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
@@ -484,21 +484,23 @@ public class FafServerAccessorImpl extends AbstractServerAccessor implements Faf
   }
 
   private String translateAuthenticationMessage(AuthenticationFailedMessage message) {
-    ArrayList<String> key = new ArrayList(Arrays.asList("login", "error"));
-    ArrayList<Object> args = new ArrayList();
-
     String context = message.getContext();
-    if (context != null) {
-      key.add(context);
-      args.add(clientProperties.getWebsite().getSteamLinkUrl());
-
-      String result = message.getResult();
-      if (context.equals("policy") && result != null) {
-        key.add(result);
-        args.add(result);
-      }
+    if (context == null) {
+      // Fallback to displayaing server side text
+      String messageText = message.getText();
+      return messageText != null ? messageText: i18n.get("login.error");
     }
 
+    ArrayList<String> key = new ArrayList(Arrays.asList("login", "error", context));
+    ArrayList<Object> args = new ArrayList(Arrays.asList(clientProperties.getWebsite().getSteamLinkUrl()));
+
+    String result = message.getResult();
+    if (context.equals("policy") && result != null) {
+      key.add(result);
+      args.add(result);
+    }
+
+    // Resolve the most detailed message we can
     Object[] argsArray = args.toArray();
     while(key.size() > 1) {
       try {
@@ -508,8 +510,8 @@ public class FafServerAccessorImpl extends AbstractServerAccessor implements Faf
       }
     }
 
-    String messageText = message.getText();
-    return messageText != null ? messageText: i18n.get("login.error");
+    // Our localization file is messed up
+    return "login.error";
   }
 
   private void onFafLoginSucceeded(LoginMessage loginServerMessage) {
--- a/src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
+++ b/src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
@@ -499,9 +499,10 @@ public class FafServerAccessorImpl extends AbstractServerAccessor implements Faf
       }
     }
 
+    Object[] argsArray = args.toArray();
     while(key.size() > 1) {
       try {
-        return i18n.get(String.join(".", key), args.toArray());
+        return i18n.get(String.join(".", key), argsArray);
       } catch(NoSuchMessageException e) {
         key.remove(key.size() - 1);
       }
--- a/src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
+++ b/src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
@@ -474,24 +474,39 @@ public class FafServerAccessorImpl extends AbstractServerAccessor implements Faf
   private void dispatchAuthenticationFailed(AuthenticationFailedMessage message) {
     fafConnectionTask.cancel();
 
-    String text;
+    String text = translateAuthenticationMessage(message);
 
+    loginFuture.completeExceptionally(new LoginFailedException(text));
+    loginFuture = null;
+  }
+
+  private String translateAuthenticationMessage(AuthenticationFailedMessage message) {
     String context = message.getContext();
     if (context == null) { context = ""; }
 
     switch (context) {
       case "denied":
-        text = i18n.get("login.deniedError");
-        break;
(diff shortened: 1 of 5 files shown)


src/main/java/com/faforever/client/remote/domain/AuthenticationFailedMessage.java
```diff
@@ -1,18 +1,18 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
 public class AuthenticationFailedMessage extends FafServerMessage {
 
   private String text;
+  private String context;
+  // Policy server result when context == "policy"
```
1-alex98: java doc please

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -471,10 +471,42 @@ private void onServerPing() {
...
     fafConnectionTask.cancel();
-    loginFuture.completeExceptionally(new LoginFailedException(message.getText()));
+
+    String text = translateAuthenticationMessage(message);
+
+    loginFuture.completeExceptionally(new LoginFailedException(text));
     loginFuture = null;
   }
 
+  private String translateAuthenticationMessage(AuthenticationFailedMessage message) {
```
1-alex98, line 486: So this ends in a lot of lines if we do this for every mesage. What I would expect would be a unique identifying string/number for the message. Which in return can be found in the messages of the client maybe with a prefix For Example:
Server message has a field messageId= 147
-> i18n.get("serverMessage.147", message.getParameters());
-> serverMessage.147 = Login failed, you are banned for {0} days
Translating hence can be done in one line

1-alex98 (changes requested):

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
Askaholic: Ok I see. I'll have to think about the best way to do this so it still makes sense for the protocol. The only thing is it doesn't handle config parameters like the website url which don't need to be sent by the server.

Askaholic: I don't think I would use numbers, but I might do something like `i18n.get(String.format("authenticationMessage.%s", message.getContext()), /* parameters here */)`

1-alex98: Just saying the here presented solution seems to be a bad idea. Also what we definitely need is a fallback string that is sent if the server message is not found in message.properties

Askaholic: Ok. I've made it more general where it uses the value of the message identifier as the localization key. So for instance localization of a policy server error for "fraudulent" response would use key `login.error.policy.fraudulent`.

If the given key is not found, it will try to fall back by popping off the last identifier in the chain until either it finds a key, or it hits `login.error` which returns a very generic error message. So if for instance the key `login.error.policy.fraudulent` didn't exist, it would try  `login.error.policy` which does exist, and returns a generic error about a policy violation.

1-alex98: sounds better

1-alex98: ok but still this is specific code for "login",  so we need to copy this for everything else right. Tho I understand that if u wanna add specific args from the clientProperties thats hard to realize. So if u want to do it this way that's fine with me.

1-alex98 (approved):

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
Askaholic: Well when I do the next message type I can refactor the bits I need.

1-alex98:
Still need to test this

Askaholic:
I have tested it using the test server (and the PR I made for that). I tried to write some tests, but I don't think I was going about it the right way. JUnit 4 doesn't have very good support for testing that a function throws a specific exception, and I'm thinking that's probably not the best way to do it anyways. Suggestions on how to write the tests are welcome

1-alex98:
I meant test in run and see if it breaks

1-alex98:
Just test the positive cases then and u can still catch the Exceptiona and see if it is what u expect

1-alex98:
@Askaholic is this deployed server sided?

Askaholic:
No, I was going to wait for this client PR to be merged first since it will continue to work with the old server messages.

1-alex98:
ok so I merge this?

1-alex98:
next time make a branch in this repo. I always need to add your fork remote otherwises

1-alex98:
grafik
I won't merge this if that's the message we will get. We should fallback to the or email or password is wrong message.

Askaholic:
How did you produce that? It's supposed to prefer any message sent by the server over the generic error message:
```Java
return messageText != null ? messageText: i18n.get("login.error");
```

1-alex98:
I checked out your branch and entered wrong credidentials

1-alex98 (changes requested):
chnages requested see above

1-alex98:
So shell I try again?

1-alex98:
lemme se

Askaholic:
What is the status of this PR? Have you tested it @axel1200?

1-alex98:
@Askaholic u realise that if we switch to api being the one that grants login that gets obsolete

Askaholic:
The client will still need to authenticate with the server. And not all users are going to choose to use an API token. We still will support both methods of login.

1-alex98:
Yes but the primary login will be the api if it fails then the error message of the api will be shown

Askaholic:
In response to your previous comment, you will still want to localize those error messages from the API.

I think I have a better way of localizing server messages in general. How about we add an `i18n_key` field to the existing `notice` message? That will require a lot less rewriting and be much more flexible in general. We can still send English text as a fallback too.

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -473,10 +476,44 @@ private void onServerPing() {
...
+    Object[] argsArray = args.toArray();
+    while(key.size() > 1) {
+      try {
+        return i18n.get(String.join(".", key), argsArray);
+      } catch(NoSuchMessageException e) {
+        key.remove(key.size() - 1);
+      }
+    }
+
+    // Our localization file is messed up
```
1-alex98, line 513: What does this mean. Please define a proper fallback then...

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -473,10 +476,44 @@ private void onServerPing() {
...
+    return "login.error";
```
1-alex98, line 514: i18n.get(...) right?

1-alex98 (changes requested):

1-alex98:
Has conflicts needs to be rebased

Sheikah45:
What is the status of this pr?

Sheikah45:
Closing as it is stale

PR #1513, opened 2019-11-26
Replay parsing efficiency improvement (wrong branch)

andreasley:
Currently, `.fafreplay` files are combined from two lines of text:

1. Metadata, encoded to JSON and without any linebreaks
2. FA replay data (binary), encoded as a Base64 string

Previously, parsing the replay and reading the metadata would read in the whole file, requiring lots of memory. This patch uses a NIO stream to only read the first line, thereby reducing the required amount of memory and possibly accelerate parsing.


andreasley:
(wrong branch)

PR #1514, opened 2019-11-26
Replay parsing efficiency improvement

andreasley:
Currently, `.fafreplay` files are combined from two lines of text:

1.  Metadata, encoded to JSON and without any linebreaks
2.  FA replay data (binary), encoded as a Base64 string

Previously, parsing the replay and reading the metadata would read in the whole file, requiring lots of memory. This patch uses a NIO stream to only read the first line, thereby reducing the required amount of memory and possibly accelerate parsing.

### DIFF
--- a/src/main/java/com/faforever/client/replay/ReplayFileReaderImpl.java
+++ b/src/main/java/com/faforever/client/replay/ReplayFileReaderImpl.java
@@ -19,6 +19,9 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Stream;
+
+import static com.github.nocatch.NoCatch.noCatch;
 
 @Lazy
 @Component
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/replay/ReplayFileReaderImpl.java
```diff
@@ -34,16 +37,22 @@ public ReplayFileReaderImpl() {
   @SneakyThrows
   public LocalReplayInfo parseMetaData(Path replayFile) {
     logger.debug("Parsing metadata of replay file: {}", replayFile);
-    List<String> lines = Files.readAllLines(replayFile);
-    return gson.fromJson(lines.get(0), LocalReplayInfo.class);
+    Optional<String> metadata = Files.lines(replayFile).findFirst();
+    if (metadata.isPresent()) {
+      return gson.fromJson(metadata.get(), LocalReplayInfo.class);
+    }
+    throw new IOException(String.format("Failed to extract metadata from replay file: {}", replayFile));
```
Brutus5000: Instead of the if use  map(metadata -> gson.fromJson...).orElseThrow(new IOException...)
Also we should put Files.lines into a try-with-resources to ensure closing of the file

src/main/java/com/faforever/client/replay/ReplayFileReaderImpl.java
```diff
@@ -34,16 +37,22 @@ public ReplayFileReaderImpl() {
...
   @SneakyThrows
   public byte[] readRawReplayData(Path replayFile) {
     logger.debug("Reading replay file: {}", replayFile);
-    List<String> lines = Files.readAllLines(replayFile);
-    return QtCompress.qUncompress(BaseEncoding.base64().decode(lines.get(1)));
+    Optional<String> replayData = Files.lines(replayFile).skip(1).findFirst();
+    if (replayData.isPresent()) {
+      return QtCompress.qUncompress(BaseEncoding.base64().decode(replayData.get()));
+    }
+    throw new IOException(String.format("Failed to extract replay data from replay file: {}", replayFile));
```
Brutus5000: same as above

Brutus5000 (changes requested):
Good catch.

src/main/java/com/faforever/client/replay/ReplayFileReaderImpl.java
```diff
@@ -34,16 +37,22 @@ public ReplayFileReaderImpl() {
...
 
   @Override
   @SneakyThrows
   public byte[] readRawReplayData(Path replayFile) {
     logger.debug("Reading replay file: {}", replayFile);
-    List<String> lines = Files.readAllLines(replayFile);
-    return QtCompress.qUncompress(BaseEncoding.base64().decode(lines.get(1)));
+    Optional<String> replayData = Files.lines(replayFile).skip(1).findFirst();
+    if (replayData.isPresent()) {
+      return QtCompress.qUncompress(BaseEncoding.base64().decode(replayData.get()));
```
norraxx: Does this `QtCompres.qUncompress` decompress the stuff right?
https://github.com/norraxx/faf-scfa-replay-api/blob/master/source/utils.py#L21

Do we need there shift by 4 bytes?

andreasley: Hm. The previous code didn't do that either. I suspect the 4 bytes are the linebreak (CRLF).

src/main/java/com/faforever/client/replay/ReplayFileReaderImpl.java
andreasley: Fixed.

src/main/java/com/faforever/client/replay/ReplayFileReaderImpl.java
andreasley: Fixed, but I think readability suffers quite a bit with this style.

Brutus5000 (approved):

1-alex98 (approved):

1-alex98:
For future work it would be cool if u got:
- An issue to whatever you do only then will it show up in the auto generated change log
- Write "Fixes #issuenumber" into your commits so it gets closed on merge
- Call you branch bugfix/#issuenumber-yourname so GitHub knows which branch belongs to which issue, and developers as well, or feature/#... if it is a feature
- Try to have one commit per issue so backlog is clean (try to amend commits)

1-alex98:
tested works

PR #1519, opened 2019-11-29
Installer gives all Windows Users full permissions to ProgramData\FAForever 

### LINKED ISSUE #1501: Log access denied
With the release of version 1.0.1 a lot of ppl complained about not being able to start the client. The issue is the log file can not be accessed( only with admin rights) and client fails to launch. They all updated from 1.0.0 it seems

andreasley:
Most likely fixes https://github.com/FAForever/downlords-faf-client/issues/1501

With this change, the installer now adds full permissions (read/write/execute) for all Windows Users to C:\ProgramData\FAForever to make sure that the directory can be used by all the system's users (regardless if they have admin permissions).
The permissions are inherited by all subfolders/files.

### DIFF
--- a/downlords-faf-client.install4j
+++ b/downlords-faf-client.install4j
@@ -16,9 +16,12 @@
   </application>
   <files keepModificationTimes="false" missingFilesStrategy="warn" globalExcludeSuffixes="" defaultOverwriteMode="4" defaultUninstallMode="0" launcherOverwriteMode="3" defaultFileMode="644" defaultDirMode="755">
     <filesets />
-    <roots />
+    <roots>
+      <root id="506" fileset="" location="${installer:sys.programDataDir}" />
+    </roots>
     <mountPoints>
       <mountPoint id="23" root="" location="" mode="755" />
+      <mountPoint id="507" root="506" location="FAForever" mode="755" />
     </mountPoints>
     <entries>
       <dirEntry mountPoint="23" file="./build/resources/native" overwriteMode="4" shared="false" fileMode="755" uninstallMode="0" overrideFileMode="true" overrideOverwriteMode="false" overrideUninstallMode="false" entryMode="subdir" subDirectory="natives" excludeSuffixes="" dirMode="755" overrideDirMode="false">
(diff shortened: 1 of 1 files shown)


betrok:
Will inheritance work with files created by the client itself?
And why should user files be stored in a system-wide directory at all?

andreasley:
> Will inheritance work with files created by the client itself?

Yes; both existing and new ones.

> And why should user files be stored in a system-wide directory at all?

In my opinion, they really shouldn't be. But my suggestions to store them in the user's `AppData` have so far not found much appreciation, so maybe it was an intentional decision to share this data between users.

Katharsas:
The client configuration is already stored in APPDATA, however there are probably more things that should be in APPDATA and not in ProgramData (thinking about logs and personal local replays for example). However, multiple users can benefit from a shared and and game file cache, so you probably don't want to put everything in APPDATA.
There is also the fact that the python client uses the ProgramData folder in the same way as the Java client (which is nice for people tht have to regularly switch for other reasons), but that should no longer be relevant since pythong client is not supported anymore.

1-alex98:
NICE I like

1-alex98:
I will now build an installer and test this

1-alex98:
It would make it easier in the feature if u joined your ZULIP if u have not and commit to this repo instead of a fork just saying. And tanks by the way. @andreasley

andreasley:
> It would make it easier in the feature if u joined your ZULIP if u have not and commit to this repo instead of a fork just saying. And tanks by the way. @andreasley

I've been participating in Zulip for three weeks. :)
If you give me access, I'll gladly contribute directly to this repository.

1-alex98:
go test https://github.com/FAForever/downlords-faf-client/releases/tag/v1.0.1-RC10

andreasley:
> go test https://github.com/FAForever/downlords-faf-client/releases/tag/v1.0.1-RC10

You might want to tag the next release 1.0.2, because a few people (including me) have the previous 1.0.1 release installed.

1-alex98:
oh yeah true, good idea

1-alex98:
works

1-alex98 (approved):

Issue #1518, opened 2019-11-29
1v1 ranked match matchmaking has a buggy CPU usage/consumption.


xponen:
**Describe the bug**
100% CPU usage when searching for 1v1 ranked match.

**To Reproduce**
First, use Window Taskmanager to check CPU usage and also constraint FAF-client to a single core (this avoid Window(TM) from masking the CPU usage by dividing it by the number of cores), then press "Play" button in 1v1 ranked match's Tab.

**Expected behavior**
I expect the FAF-client to be more efficient for the sake of the game. The game is CPU demanding, it would be very bad if the lobby interfere with the game by stealing CPU resources.

**Screenshots**
![25 percent of 4 core is 100 percent](https://user-images.githubusercontent.com/2525365/69861118-a2ae8580-12d2-11ea-9b34-1d5654fe61ac.jpg)
*happen the instant Play is pressed. In this instance I quickly take Screenshot before a game starting.

**OS**
Windows 10

**Wanna have the bug fixed quickly?**
In the meantime I isolate FAF-client to 1 core hoping it won't interfere with the game.

Katharsas:
Possibly related https://github.com/FAForever/downlords-faf-client/issues/1371

1-alex98:
duplicate #1371

PR #1520, opened 2019-12-08
Slightly improved ladder counter efficiency

andreasley:
Might help with https://github.com/FAForever/downlords-faf-client/issues/1518
Previously, the counter and the corresponding label would be updated 60 times/second. With this patch, this is reduced to once a second, which has no downsides in this case.


1-alex98 (approved):

1-alex98:
this wont fix the progress bar using soo much cpu but still it is a good thing to do

PR #1525, opened 2019-12-15
Show panel if needed permission are given

### LINKED ISSUE #1524: Moderation panel not shown since new permission system
Implement new system

1-alex98:
Fixes #1524

### DIFF
--- a/src/main/java/com/faforever/client/api/FafApiAccessor.java
+++ b/src/main/java/com/faforever/client/api/FafApiAccessor.java
@@ -14,10 +14,10 @@ import com.faforever.client.api.dto.Ladder1v1Map;
 import com.faforever.client.api.dto.Map;
 import com.faforever.client.api.dto.MapVersion;
 import com.faforever.client.api.dto.MapVersionReview;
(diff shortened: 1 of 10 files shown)


1-alex98:
So we will just go with a permission for kicking users which the server ignores/does not know. But when we giv it to all mods(the moderation) group that should be fine. Well some other time the server can be changed to respect the new Permission System.

1-alex98:
TODO: Use correct string from @Brutus5000 PR https://github.com/FAForever/server/pull/509/files

1-alex98:
In sync with server PR only that server pr is not ready yet

1-alex98:
@Brutus5000 I need a review here... I made u the required db migration for the server

1-alex98:
The server access token stuff will be placed on top of this. Because this also needs to know stuff about current user.

Brutus5000 (changes requested):
The MeResult should be taken from faf-java-commons (needs update)

1-alex98:
The server PR got merged

1-alex98:
I wanna make the switch to faf comoons in a different PR

PR #1527, opened 2019-12-17
Remove deprecated preference variable and eliminate map directory cac…

### LINKED ISSUE #1526: Error on start up
User can not start game he gets an error
Log: 
https://github.com/vicblack11/FAF/commit/490f7f6515ce44c260bd27dd75cc8fe457816ebe

1-alex98:
…hing

Fixes #1526

### DIFF
--- a/src/main/java/com/faforever/client/map/MapService.java
+++ b/src/main/java/com/faforever/client/map/MapService.java
@@ -49,7 +49,6 @@ import javax.inject.Inject;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.net.URL;
-import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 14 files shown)


1-alex98:
@betrok u cached  the prefence value hence new users could not create games any longer, because the value was null for them at the start of teh application :D but dont worry I will quickly hotfix it

src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -184,15 +182,14 @@ private void tryLoadMaps() {
       // TODO notify user
     }
 
-    customMapsDirectory = mapsDirectory;
```
1-alex98, line 187: there is the bad caching

betrok:
Hmm... I don't think caching was the root of the problem. In fact it was around there before my changes.
Maps must be reloaded when the path changes, and this variable would be updated when this happens.

May be I will investigate later but idk when :/

1-alex98:
U are right actually

1-alex98:
No idea now it works before it did not

PR #1530, opened 2019-12-17
Remove player choice from live replay button

### LINKED ISSUE #1375: Remove player selection when selecting live replay
With the new replay server, player selection does not matter - any choice will get you the same merged replay. The UI bit can be removed, and you can replace player name with any nonempty string when accessing the replay.

Wesmania:
Current live replay server merges all incoming streams into one,
therefore there's no need to select a specific player when launching a
live replay. Remove the selection from the UI, but still pick and send
some player's name in case the replay server finds it useful.

Fixes #1375.

### DIFF
--- a/src/main/java/com/faforever/client/vault/replay/WatchButtonController.java
+++ b/src/main/java/com/faforever/client/vault/replay/WatchButtonController.java
@@ -4,44 +4,34 @@ import com.faforever.client.config.ClientProperties;
 import com.faforever.client.fx.Controller;
 import com.faforever.client.game.Game;
 import com.faforever.client.i18n.I18n;
-import com.faforever.client.player.Player;
-import com.faforever.client.player.PlayerService;
 import com.faforever.client.replay.ReplayService;
 import com.faforever.client.util.TimeService;
 import javafx.animation.KeyFrame;
 import javafx.animation.Timeline;
 import javafx.scene.Node;
-import javafx.scene.control.MenuButton;
-import javafx.scene.control.MenuItem;
+import javafx.scene.control.Button;
 import javafx.util.Duration;
-import org.jetbrains.annotations.NotNull;
 import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 import org.springframework.util.Assert;
 
 import java.time.Instant;
-import java.util.Collection;
-import java.util.List;
-import java.util.Optional;
-import java.util.stream.Collectors;
 
 @Component
 @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
 public class WatchButtonController implements Controller<Node> {
   private final ReplayService replayService;
-  private final PlayerService playerService;
   private final ClientProperties clientProperties;
   private final TimeService timeService;
 
-  public MenuButton watchButton;
+  public Button watchButton;
   private Game game;
   private I18n i18n;
   private Timeline delayTimeline;
 
-  public WatchButtonController(ReplayService replayService, PlayerService playerService, ClientProperties clientProperties, TimeService timeService, I18n i18n) {
+  public WatchButtonController(ReplayService replayService, ClientProperties clientProperties, TimeService timeService, I18n i18n) {
     this.replayService = replayService;
-    this.playerService = playerService;
     this.clientProperties = clientProperties;
     this.timeService = timeService;
     this.i18n = i18n;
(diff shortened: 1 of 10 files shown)


src/main/java/com/faforever/client/vault/replay/WatchButtonController.java
```diff
@@ -55,30 +51,27 @@ public void initialize() {
     delayTimeline.setCycleCount(Timeline.INDEFINITE);
 
     watchButton.setDisable(true);
+    watchButton.setOnAction(event -> {
+      var player = getAnyPlayer(game);
+      if (player.isPresent()) {
+        replayService.runLiveReplay(game.getId(), player.get().getId());
+      }
```
Brutus5000: This can be optimized:

```
getAnyPlayer(game)
.ifPresent(player -> replayService.runLiveReplay(game.getId(), player.getId()))
```

Brutus5000 (changes requested):
Looks good otherwise.

src/main/java/com/faforever/client/vault/replay/WatchButtonController.java
```diff
@@ -55,30 +51,25 @@ public void initialize() {
     delayTimeline.setCycleCount(Timeline.INDEFINITE);
 
     watchButton.setDisable(true);
+    watchButton.setOnAction(event -> {
+      getAnyPlayer(game)
+        .ifPresent(player -> replayService.runLiveReplay(game.getId(), player.getId()));
```
1-alex98: Why don't we remove the plyer id from the whole request then instead of just removing the selection

1-alex98 (changes requested):

src/main/java/com/faforever/client/vault/replay/WatchButtonController.java
Wesmania: I'm hesitant to remove this information in case it ever becomes useful again for the replay server. We have to send some string to the game anyway, and doing it like this is a smaller change.

Wesmania: If we wanted to, we could of course put whatever string we want here, like current user's name.

1-alex98: Well this is like leaving comments/commented out code in code and i dont like that... If we ever need it again we can revert this every commit u do here or at least look into it on how to do it and copy paste

1-alex98: Thats what we have git for

Wesmania: In that case I can just shove current user's name there instead, sure.

1-alex98: I mean what would be the use case selecting the user now that the replay sever decides which replay stream is best for us

1-alex98: Removing the rest of the code reduces complexity and right now it must be really strange to read why would the client do somthing like that

1-alex98: I would say lets do it clean and remove it all

Wesmania: Hypothetically the replay server could match that username to whatever writer stream it has, like the legacy server did. OTOH, it would have to deal with a name not matching anything anyway, so I guess it doesn't matter either way.

Wesmania: Alright, I'll do that.

1-alex98: And use a dummy string if the game needs such a string

src/main/java/com/faforever/client/vault/replay/WatchButtonController.java
1-alex98: u and your Optionals :D

Wesmania:
Player id is still left in the discord game launch message, but I dare not remove it so as not to break any APIs. Apparently clients communicate between one another here, so I'm definitely not touching it.

src/main/resources/theme/vault/replay/watch_button.fxml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 
-<?import javafx.scene.control.MenuButton?>
-<MenuButton xmlns:fx="http://javafx.com/fxml/1" fx:id="watchButton" disable="true" maxHeight="1.7976931348623157E308" maxWidth="1.7976931348623157E308" minHeight="-Infinity" minWidth="-Infinity" mnemonicParsing="false" text="%game.watchButtonFormat" xmlns="http://javafx.com/javafx/8.0.141" fx:controller="com.faforever.client.vault.replay.WatchButtonController" />
+<?import javafx.scene.control.Button?>
+<Button xmlns:fx="http://javafx.com/fxml/1" fx:id="watchButton" disable="true" maxHeight="1.7976931348623157E308" maxWidth="1.7976931348623157E308" minHeight="-Infinity" minWidth="-Infinity" mnemonicParsing="false" text="%game.watchButtonFormat" xmlns="http://javafx.com/javafx/8.0.141" fx:controller="com.faforever.client.vault.replay.WatchButtonController" />
```
axel1200: Use JFXButton

1-alex98 (changes requested):
nice thx

1-alex98:
I can do the discord thing

1-alex98:
btw u can also use a branch on this repo

1-alex98 (dismissed):

Wesmania:
Won't the discord change break things if an old and new client version interact? At first glance it looks like exchanging some json structures between clients.

1-alex98:
no at all this feature is not activated anyway XD, because we would need to apply for it and i did not do so yet

1-alex98 (approved):

Brutus5000 (approved):

PhilipJFryFAF:
@Wesmania how can users see which players are playing in a live replay with the user list gone?

Wesmania:
That didn't cross my mind. I guess we'll have to add a column for that?

Wesmania:
What would be the best way to do that? Shoving all player names into the player column makes it really wide. Should I show tooltips for cut off text? Or add a tooltip on the player column and only show player number? Make it a dropdown menu?

1-alex98:
@PhilipJFryFAF to late I merged it XD

PR #1542, opened 2020-01-01
Properly handle futures when the local replay directory has changed

### LINKED ISSUE #1541: Local replays sometimes fail to appear
The list of local replays should be reloaded automatically, starting with version 1.1.
For some reason, this doesn't always work, e.g. sometimes, new replays don't appear in the list.

andreasley:
Fixes #1541.

Previously, the code failed to properly wait on completion of futures that load replays. Depending on the timing, the new local replays may have been added correctly or not. This PR fixes that.

### DIFF
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -132,7 +132,7 @@ public class ReplayService {
   private final ExecutorService executorService;
   private Thread directoryWatcherThread;
   private WatchService watchService;
-  private List<Replay> localReplays = new ArrayList<Replay>();
+  protected List<Replay> localReplays = new ArrayList<Replay>();
 
   public void startLoadingAndWatchingLocalReplays() {
     Path replaysDirectory = preferencesService.getReplaysDirectory();
@@ -177,7 +177,8 @@ public class ReplayService {
     return thread;
   }
 
-  private void onLocalReplaysWatchEvent(WatchKey key) {
+  @VisibleForTesting
+  protected void onLocalReplaysWatchEvent(WatchKey key) {
     List<CompletableFuture<Replay>> newReplaysFutures = new ArrayList<CompletableFuture<Replay>>();
     Collection<Replay> deletedReplays = new ArrayList<Replay>();
     for (WatchEvent<?> watchEvent : key.pollEvents()) {
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -178,15 +178,14 @@ public class ReplayService {
   }
 
   private void onLocalReplaysWatchEvent(WatchKey key) {
-    Collection<Replay> newReplays = new ArrayList<Replay>();
+    List<CompletableFuture<Replay>> newReplaysFutures = new ArrayList<CompletableFuture<Replay>>();
     Collection<Replay> deletedReplays = new ArrayList<Replay>();
     for (WatchEvent<?> watchEvent : key.pollEvents()) {
       Path path = (Path) watchEvent.context();
       Path fullPathToReplay = preferencesService.getReplaysDirectory().resolve(path);
 
       if (watchEvent.kind() == ENTRY_CREATE) {
-        tryLoadingLocalReplay(fullPathToReplay)
-            .thenAccept(newReplay -> newReplays.add(newReplay));
+        newReplaysFutures.add(tryLoadingLocalReplay(fullPathToReplay));
       } else if (watchEvent.kind() == ENTRY_DELETE) {
         Optional<Replay> existingReplay = localReplays
             .stream()
@@ -201,8 +200,21 @@ public class ReplayService {
       }
     }
 
-    localReplays.addAll(newReplays);
-    publisher.publishEvent(new LocalReplaysChangedEvent(this, newReplays, deletedReplays));
+    CompletableFuture[] replayFuturesArray = newReplaysFutures.toArray(new CompletableFuture[newReplaysFutures.size()]);
+    CompletableFuture<List<Replay>> newReplaysFuture = CompletableFuture.allOf(replayFuturesArray)
+        .thenApply(ignoredVoid ->
+            newReplaysFutures.stream()
+                .map(future -> future.join())
+                .collect(Collectors.toList())
+    );
+
+    try {
+      List<Replay> newReplays = newReplaysFuture.get();
+      localReplays.addAll(newReplays);
+      publisher.publishEvent(new LocalReplaysChangedEvent(this, newReplays, deletedReplays));
+    } catch (Exception e) {
+      logger.warn("Failed to load new local replays ({})", e.getMessage());
+    }
   }
 
   @VisibleForTesting
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -50,7 +50,6 @@ import org.springframework.stereotype.Service;
 import org.springframework.web.util.UriComponentsBuilder;
 
 import java.io.ByteArrayInputStream;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.net.URI;
@@ -68,10 +67,10 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutorService;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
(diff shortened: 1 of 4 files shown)


axel1200:
unit tests?

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -201,8 +200,21 @@ private void onLocalReplaysWatchEvent(WatchKey key) {
       }
     }
 
-    localReplays.addAll(newReplays);
-    publisher.publishEvent(new LocalReplaysChangedEvent(this, newReplays, deletedReplays));
+    CompletableFuture[] replayFuturesArray = newReplaysFutures.toArray(new CompletableFuture[newReplaysFutures.size()]);
+    CompletableFuture<List<Replay>> newReplaysFuture = CompletableFuture.allOf(replayFuturesArray)
+        .thenApply(ignoredVoid ->
```
1-alex98, line 205: thenRun

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -201,8 +200,21 @@ private void onLocalReplaysWatchEvent(WatchKey key) {
...
+            newReplaysFutures.stream()
+                .map(future -> future.join())
```
1-alex98: CompletableFuture
        .allOf

1-alex98 (changes requested):

1-alex98:
Think about whether u could maybe add unit tests

src/main/java/com/faforever/client/replay/ReplayService.java
andreasley: Why?

src/main/java/com/faforever/client/replay/ReplayService.java
andreasley: What?

1-alex98: To join CompletedFutures use CompletableFuture.allOf()

src/main/java/com/faforever/client/replay/ReplayService.java
1-alex98: Sorry i am wrong... but there is a method that takes a lambda without the parameter ignoredVoid. Use this method instead not totally sure whats it called thenCombine maybe

1-alex98: ok whatever ignore me

src/main/java/com/faforever/client/replay/ReplayService.java
andreasley: Which is exactly what I'm doing. Hence my confusion. :)

1-alex98: Should read the code closer sorry lemme check again

1-alex98:
I made a lot of shitty comments but somehting is wrong here what u are doing seems to be overly complecated

andreasley:
> 
> 
> I made a lot of shitty comments but somehting is wrong here what u are doing seems to be overly complecated

It seems overly complicated to me, too, but that may just be because it's Java. ;)
Seriously though, wherever `CompletableFuture.allOf` is used in the codebase, it's exactly like this.
But if there's a better way, I'd love to learn.

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -201,8 +201,21 @@ private void onLocalReplaysWatchEvent(WatchKey key) {
...
```
1-alex98: use method reference
```suggestion
                CompletableFuture::join
```

1-alex98 (changes requested):

1-alex98:
Apart from this it is totally fine what u do it seems.
One more thing is that The @Async annotaion in line 291 needs to be removed it is just bullshit there. It only works if it is called via a Spring proxy Object which is obviously not the case here (not your fault) just noticed when I looked at the code. I did the two things for u and gonna make a quick commit. Thank you for your work 👍

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -205,7 +202,8 @@ protected void onLocalReplaysWatchEvent(WatchKey key) {
     CompletableFuture<List<Replay>> newReplaysFuture = CompletableFuture.allOf(replayFuturesArray)
         .thenApply(ignoredVoid ->
             newReplaysFutures.stream()
-                .map(future -> future.join())
+                .map(CompletableFuture::join)
+                .filter(Objects::nonNull)
```
1-alex98, line 208: Well we wanna avoid having null objects in list because it crashes UI

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -275,21 +273,21 @@ static String guessModByFileName(String fileName) {
       List<CompletableFuture<Replay>> replayFutures = StreamSupport.stream(directoryStream.spliterator(), false)
           .sorted(Comparator.comparing(path -> noCatch(() -> Files.getLastModifiedTime((Path) path))).reversed())
           .limit(MAX_REPLAYS)
-          .map( replayFile -> tryLoadingLocalReplay(replayFile))
+          .map(this::tryLoadingLocalReplay)
```
1-alex98, line 278: Just a quick way of writting it

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -275,21 +273,21 @@ static String guessModByFileName(String fileName) {
...
           .filter(e -> !e.isCompletedExceptionally())
           .collect(Collectors.toList());
 
       CompletableFuture[] replayFuturesArray = replayFutures.toArray(new CompletableFuture[replayFutures.size()]);
       return CompletableFuture.allOf(replayFuturesArray)
           .thenApply(ignoredVoid ->
               replayFutures.stream()
-                  .map(future -> future.join())
+                  .map(CompletableFuture::join)
+                  .filter(Objects::nonNull)
```
1-alex98, line 287: again see top

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -300,14 +298,14 @@ static String guessModByFileName(String fileName) {
       return CompletableFuture.allOf(featuredModFuture, mapBeanFuture).thenApply(ignoredVoid  -> {
         Optional<MapBean> mapBean = mapBeanFuture.join();
         if (!mapBean.isPresent()) {
-          throw new CompletionException(new FileNotFoundException());
+          logger.warn("Could not find map for replay file '{}'", replayFile);
```
1-alex98, line 303: We certainly dont wanna throw here if we do then the whole loacl replay vault breaks, also there are legimit cases where a replay has no map. One is coop the maps are part of the game mod and hence are not found locally nor on the server

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -300,14 +298,14 @@ static String guessModByFileName(String fileName) {
...
         }
-        return new Replay(replayInfo, replayFile, featuredModFuture.join(), mapBean.get());
+        return new Replay(replayInfo, replayFile, featuredModFuture.join(), mapBean.orElse(null));
```
1-alex98, line 305: Let's just say the map is null

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -300,14 +298,14 @@ static String guessModByFileName(String fileName) {
...
       });
     } catch (Exception e) {
       logger.warn("Could not read replay file '{}'", replayFile, e);
       moveCorruptedReplayFile(replayFile);
-      return CompletableFuture.failedFuture(e);
+      return CompletableFuture.completedFuture(null);
```
1-alex98, line 310: Don't throw others wise replay vault stays empty, return null which is gonna be filtered out

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
```diff
@@ -183,9 +177,12 @@ public OffsetDateTime fromString(String string) {
       protected void updateItem(MapBean map, boolean empty) {
         super.updateItem(map, empty);
 
-        if (empty || map == null) {
+        if (empty) {
           setText(null);
           setGraphic(null);
+        } else if (map == null) {
+          setGraphic(null);
+          setText(i18n.get("map.unknown"));
```
1-alex98, line 185: Let's tell te user we don't have this map

1-alex98 (dismissed):

1-alex98:
So the local replay vault had quite some issues esspecially with coop replays those made the whole local replay vault to not display

1-alex98:
@andreasley maybe read and complain if you don't like something

1-alex98 (dismissed):

andreasley:
> @andreasley maybe read and complain if you don't like something

@axel1200 Looks good to me.
Your comments looked like statements, not questions, so I didn't feel compelled to respond. :)

1-alex98:
No they are not questions, but still donot hesitate to ask if anything is unclear or u disagree.

1-alex98:
btw we are on discord if u wanna have a chatter :D

PR #1544, opened 2020-01-03
Display accurate replay durations in vault

Askaholic:
Uses replays tick numbers (extracted by the new replay server) to display accurate in game durations of replays. These times can be significantly shorter than the real time calculated using `startTime` and `endTime`. The accurate times will be favored over the real times when displaying the replay cards, and both times are displayed on the replay details. Also added tooltips to the labels on the replay details pane to make it even more clear what the icons represent.

image

The new icon is the timer icon with a little "play" icon (triangle within a circle) on the side. It ended up being pretty hard to see in the client because it's so small, so I may change it to just the timer with the triangle inside instead of the little minute hand. Let me know what you think.

Edit:
Ok, here's how that looks:
image


src/main/java/com/faforever/client/replay/Replay.java
```diff
@@ -219,6 +222,19 @@ public void setEndTime(Temporal endTime) {
...
+  @Nullable
+  public Integer getReplayTicks() {
+    return replayTicks.get();
+  }
+
+  public void setReplayTicks(Integer replayTicks) {
+    this.replayTicks.set(replayTicks);
+  }
+
+  public ObjectProperty<Integer> replayTicksProperty() {
```
norraxx, line 234: just for my information, where is that object property used?

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -98,9 +98,15 @@ public void setReplay(Replay replay) {
         .average()
         .ifPresent(averageRating -> ratingLabel.setText(i18n.number((int) averageRating)));
 
-    durationLabel.setText(Optional.ofNullable(replay.getEndTime())
-        .map(endTime -> timeService.shortDuration(Duration.between(replay.getStartTime(), endTime)))
-        .orElse(i18n.get("notAvailable")));
+    Integer replayTicks = replay.getReplayTicks();
+    if (replayTicks != null) {
+      durationLabel.setText(timeService.shortDuration(Duration.ofMillis(replayTicks * 100)));
```
norraxx, line 103: isn't millis 1000?

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -142,7 +144,18 @@ public void initialize() {
...
     copyButton.setText(i18n.get("replay.copyUrl"));
+
+    dateLabel.setTooltip(new Tooltip(i18n.get("replay.dateTooltip")));
+    timeLabel.setTooltip(new Tooltip(i18n.get("replay.timeTooltip")));
+    modLabel.setTooltip(new Tooltip(i18n.get("replay.modTooltip")));
+    durationLabel.setTooltip(new Tooltip(i18n.get("replay.durationTooltip")));
+    replayDurationLabel.setTooltip(new Tooltip(i18n.get("replay.replayDurationTooltip")));
+    playerCountLabel.setTooltip(new Tooltip(i18n.get("replay.playerCountTooltip")));
+    ratingLabel.setTooltip(new Tooltip(i18n.get("replay.ratingTooltip")));
+    qualityLabel.setTooltip(new Tooltip(i18n.get("replay.qualityTooltip")));
```
norraxx, line 157: 👍

src/main/java/com/faforever/client/replay/Replay.java
Askaholic: I didn't use it anywhere. I just included it because the other attributes seemed to have them. IDK what an ObjectProperty even does.

src/main/java/com/faforever/client/replay/ReplayCardController.java
Askaholic: Yes but there are 10 ticks per second. So `10/s * 100 = 1000/s`

src/main/java/com/faforever/client/replay/Replay.java
norraxx: @axel1200 ? :-)

src/main/java/com/faforever/client/replay/ReplayCardController.java
norraxx: Oh! Math hidden!! :-) how about to show it as constant with explanation? Otherwise it's not visible on first sight

src/main/java/com/faforever/client/replay/Replay.java
1-alex98: Yes u do it correctly

1-alex98: it is an object where u can listen for changes

1-alex98 (dismissed):
unit tests

1-alex98:
unit tests maybe? Btw reminder for myself i need to test this

1-alex98 (dismissed):

src/main/java/com/faforever/client/replay/Replay.java
Geosearchef: It's an observable object, following the observer pattern from software engineering. You can add a listener that gets notified on changes. The property, which is introduced by javafc id a special implementation of this observer pattern allowing properties to be bound to others (supporting converters) therefore updating each other/ UI components when changed.

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -98,9 +98,15 @@ public void setReplay(Replay replay) {
...
+      ((Label) durationLabel.getGraphic()).setText("");
```
norraxx, line 104: Maybe we should use the unicode format ala `\u0055`??

norraxx (approved):

src/main/java/com/faforever/client/replay/ReplayCardController.java
Askaholic: That's normally what I would do, but it's like this in the rest of the code so I was just trying to keep it consistent.

1-alex98:
Always create an issue so it shows up in the change log

PR #1557, opened 2020-01-12
GameService: Register events in the Bean life-cycle

craigjbass:
Ideally, constructors should not have side-effects.
This means there are no surprises when constructing them e.g. in test.


1-alex98:
Well kind of makes sense. never thought about it obviously :D

1-alex98 (approved):

craigjbass:
yo I can't actually merge this tho lol #unauthorized

PR #1570, opened 2020-01-31
Add support for new arguments in GameLaunchMessage

### LINKED ISSUE #1281: Accept game launch configuration from server as in GW
For team matchmaking we will be adding configuration options to the `game_launch` command. The client will need to check if these args are present and set the appropriate command line arguments to FA.

For argument names see this PR: https://github.com/FAForever/server/pull/428/files#diff-51c4c921447c6af4a36cbf90068e05c3

Brutus5000:
(fixes #1281)

Server side changes to be found here: https://github.com/FAForever/server/pull/517

### DIFF
--- a/src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
+++ b/src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
@@ -13,7 +13,6 @@ import com.faforever.client.fa.relay.event.GameFullEvent;
 import com.faforever.client.fa.relay.event.RehostRequestEvent;
 import com.faforever.client.fa.relay.ice.event.GpgGameMessageEvent;
 import com.faforever.client.fa.relay.ice.event.IceAdapterStateChanged;
-import com.faforever.client.game.KnownFeaturedMod;
 import com.faforever.client.os.OsUtils;
 import com.faforever.client.player.Player;
 import com.faforever.client.player.PlayerService;
@@ -262,7 +261,7 @@ public class IceAdapterImpl implements IceAdapter, InitializingBean, DisposableB
       case DEFAULT_LOBBY:
         iceAdapterProxy.setLobbyInitMode("normal");
         break;
-      case NO_LOBBY:
+      case AUTO_LOBBY:
         iceAdapterProxy.setLobbyInitMode("auto");
         break;
     }
@@ -292,11 +291,7 @@ public class IceAdapterImpl implements IceAdapter, InitializingBean, DisposableB
   }
 
   private void updateLobbyModeFromGameInfo(GameLaunchMessage gameLaunchMessage) {
-    if (KnownFeaturedMod.LADDER_1V1.getTechnicalName().equals(gameLaunchMessage.getMod())) {
-      lobbyInitMode = LobbyMode.NO_LOBBY;
-    } else {
-      lobbyInitMode = LobbyMode.DEFAULT_LOBBY;
-    }
+    lobbyInitMode = gameLaunchMessage.getInitMode();
   }
 
   @Override
--- a/src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
+++ b/src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
@@ -175,10 +175,10 @@ public class IceAdapterImpl implements IceAdapter, InitializingBean, DisposableB
         log.warn("Forcing ice adapter relay connection");
       }
 
-      if (clientProperties.isShowIceAdapterDebugWindow()) {
+//      if (clientProperties.isShowIceAdapterDebugWindow()) {
         cmd.add("--debug-window");
         cmd.add("--info-window");
-      }
+//      }
 
       try {
         ProcessBuilder processBuilder = new ProcessBuilder();
--- a/src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
+++ b/src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
@@ -13,6 +13,7 @@ import com.faforever.client.fa.relay.event.GameFullEvent;
 import com.faforever.client.fa.relay.event.RehostRequestEvent;
 import com.faforever.client.fa.relay.ice.event.GpgGameMessageEvent;
 import com.faforever.client.fa.relay.ice.event.IceAdapterStateChanged;
+import com.faforever.client.game.KnownFeaturedMod;
 import com.faforever.client.os.OsUtils;
 import com.faforever.client.player.Player;
 import com.faforever.client.player.PlayerService;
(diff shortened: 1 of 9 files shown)


src/main/java/com/faforever/client/remote/domain/GameLaunchMessage.java
```diff
@@ -1,50 +1,33 @@
...
+   */
   private List<String> args;
   private int uid;
   private String mod;
   private String mapname;
+  @NonNull
+  private String name;
+  private int expectedPlayers;
+  private int team;
+  private int mapPosition;
```
Askaholic: What does this actually get used for?

src/main/java/com/faforever/client/remote/domain/GameLaunchMessage.java
```diff
@@ -1,50 +1,33 @@
...
+  /**
+   * Stores game launch arguments, like "/ratingcolor d8d8d8d8", "/numgames 236".
+   */
   private List<String> args;
   private int uid;
   private String mod;
   private String mapname;
+  @NonNull
+  private String name;
+  private int expectedPlayers;
```
Askaholic: So this would be the number of players when the lobby is full? I don't think the server implements it correctly yet. It'll be easy to fix though.

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -442,8 +442,8 @@ public void addOnRankedMatchNotificationListener(Consumer<MatchmakerMessage> lis
             .thenRun(() -> {
               // TODO this should be sent by the server!
```
Geosearchef: we should remove the todo

src/main/java/com/faforever/client/remote/domain/GameLaunchMessage.java
Geosearchef: I think it's the slot you spawn in.

Geosearchef (dismissed):

src/main/java/com/faforever/client/remote/domain/GameLaunchMessage.java
Askaholic: Yea but the PR doesn’t reference it anywhere

Brutus5000: Indeed. I just copied the stuff from the old GW branch and did not cross-check.
In ladder the map positions somehow work out of the box.
But if you have more than 2 players and you need sorting (left vs right or top vs bottom) you will somehow need to solve that.

src/main/java/com/faforever/client/remote/domain/GameLaunchMessage.java
Brutus5000: Yes, the autolobby starts the game if the number of expectedPlayers is connected. Until then you are in the connecting stage.

src/main/java/com/faforever/client/game/GameService.java
Brutus5000: No. We still lack the team number and whatever the players argument does.

1-alex98:
Failing CI

1-alex98:
Seems like this is not yet deployed on the server

src/main/java/com/faforever/client/remote/domain/GameLaunchMessage.java
```diff
@@ -1,50 +1,33 @@
...
+  private int team;
+  private int mapPosition;
+  private Faction faction;
+  private LobbyMode initMode;
```
Askaholic, line 29: Can we add the `numgames` parameter as a dedicated field in this message too instead of passing it via `args`?

1-alex98: I would say yes why not

Askaholic: Shall we call it `num_games`?

Geosearchef:
We need to:
- Add /mapposition parameter to args based on mapPosition in GameService
- Implement map position in lua code (auto lobby, read from command line in own player info, which is sent to host, validate and interpret as host)

src/main/java/com/faforever/client/remote/domain/GameLaunchMessage.java
Askaholic: I decided not to do this for now

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
```diff
@@ -175,10 +175,10 @@ public void onGpgGameMessage(GpgGameMessageEvent event) {
         log.warn("Forcing ice adapter relay connection");
       }
 
-      if (clientProperties.isShowIceAdapterDebugWindow()) {
+//      if (clientProperties.isShowIceAdapterDebugWindow()) {
```
1-alex98: No commented out code ever

1-alex98 (changes requested):

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
```diff
@@ -175,10 +176,10 @@ public void onGpgGameMessage(GpgGameMessageEvent event) {
         log.warn("Forcing ice adapter relay connection");
       }
 
-//      if (clientProperties.isShowIceAdapterDebugWindow()) {
```
1-alex98: no commented out code

1-alex98 (changes requested):

1-alex98 (approved):
tho i think the server first needs to be deployed right?

Askaholic:
Yes, the version that is deployed right now is not sending the right numbers for `expected_players`.

src/main/java/com/faforever/client/game/GameService.java
Askaholic: Team number and players are being sent now. The players argument is just a number that tells the lobby it should wait until that many players have joined before launching the game.

Askaholic:
The required server PR has been deployed as of server `v1.4.1`. Let's remove that `TODO` comment and merge this.

An additional note which will need to be tackled in a later PR is that currently ladder rating is used whenever init mode is autolobby. We'll need to change this for TMM so that the rating type is also sent by the server. Right now in TMM games it will display your ladder rating.

src/main/java/com/faforever/client/game/GameService.java
1-alex98: So u are argueing for removing it?

1-alex98 (dismissed):

1-alex98:
Nobody is working on TMM right?

1-alex98:
Because I think @Geosearchef quit doing it...

1-alex98:
@Askaholic which TODO do u mean the "Should be sent by server" one?

Askaholic:
Yes, that one. Should remove the todo and probably delete the `else` statement too as that was fallback for a possible transition period.

1-alex98:
Can u maybe do that? Or I will someday

1-alex98 (dismissed):

PR #1573, opened 2020-02-09
Fixes #1562

### LINKED ISSUE #1562: When private games disabled, they still show up in coop list without indicator.
When "Show private games" is enabled, games show correctly in coop list.
But when disabled, private games in coop list still show up (but without lock symbol so if you cannot distinguish between private and non-private games).

1-alex98:


src/main/java/com/faforever/client/game/GamesTableController.java
```diff
@@ -84,10 +84,10 @@ public Node getRoot() {
   }
 
   public void initializeGameTable(ObservableList<Game> games) {
-    initializeGameTable(games, null);
+    initializeGameTable(games, null, true);
   }
 
-  public void initializeGameTable(ObservableList<Game> games, Function<String, String> coopMissionNameProvider) {
+  public void initializeGameTable(ObservableList<Game> games, Function<String, String> coopMissionNameProvider, Boolean listenToFilterPreferences) {
```
Geosearchef: use primitive boolean

Geosearchef (changes requested):

Brutus5000 (dismissed):
Apart from Geosearchefs suggestion, looks good.

Katharsas:
Ok so the coop table does not react to either the private or mod flag anymore right? Looks good.

1-alex98:
Yes

Katharsas (dismissed):
Ok.

PR #1574, opened 2020-02-09
Fixes #1569 kill ice adapters on update

### LINKED ISSUE #1569: Update installation fails if ICE adapter is still open
**Describe the bug**
Sometimes the ICE adapter is still running after the client has been closed.
In that case the installation throws an error message that it can not create a file.

**To Reproduce**
Try to keep the ICE adapter alive? - then reinstall

**Expected behavior**
Installation works.

**Screenshots**
image

**OS**
Windows

Idea for solution: Maybe the install4j installer can check that no ICE adapter process is running before installation.

Reported by visionik.

Visit Issue hunt...

1-alex98:
Need to test this tho

### DIFF
--- a/downlords-faf-client.install4j
+++ b/downlords-faf-client.install4j
@@ -126,6 +126,36 @@
                 </serializedBean>
                 <condition />
               </action>
+              <action name="Kill left over ice adapters" id="621" customizedId="" beanClass="com.install4j.runtime.beans.actions.control.RunScriptAction" enabled="true" commentSet="false" comment="" actionElevationType="elevated" rollbackBarrier="false" rollbackBarrierExitCode="0" multiExec="false" failureStrategy="1" errorMessage="">
+                <serializedBean>
+                  <java class="java.beans.XMLDecoder">
+                    <object class="com.install4j.runtime.beans.actions.control.RunScriptAction">
+                      <void property="script">
+                        <object class="com.install4j.api.beans.ScriptProperty">
+                          <void property="value">
+                            <string>ProcessBuilder builder = new ProcessBuilder(
+     "cmd.exe", "/c", "for /f \"tokens=1\" %i in ('jps -m ^| find \"faf-ice-adapter\"') do ( taskkill /F /PID %i )");
+Process p = builder.start();
+p.waitFor(10, java.util.concurrent.TimeUnit.SECONDS);
+Thread.sleep(4000);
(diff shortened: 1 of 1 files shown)


downlords-faf-client.install4j
```diff
@@ -699,6 +699,30 @@ return console.askOkCancel(message, true);
...
+              <action name="Kill left over ice adapters" id="621" customizedId="" beanClass="com.install4j.runtime.beans.actions.control.RunScriptAction" enabled="true" commentSet="false" comment="" actionElevationType="elevated" rollbackBarrier="false" rollbackBarrierExitCode="0" multiExec="false" failureStrategy="1" errorMessage="">
+                <serializedBean>
+                  <java class="java.beans.XMLDecoder">
+                    <object class="com.install4j.runtime.beans.actions.control.RunScriptAction">
+                      <void property="script">
+                        <object class="com.install4j.api.beans.ScriptProperty">
+                          <void property="value">
+                            <string> try{            
+     Runtime rt = Runtime.getRuntime();
+     Process proc = rt.exec("for /f \"tokens=1\" %i in ('jps -m ^| find \"faf-ice-adapter\"') do ( taskkill /F /PID %i )");
```
Geosearchef, line 741: I don't like that, but I don't think there is an alternative.

Geosearchef (dismissed):

downlords-faf-client.install4j
1-alex98: Well I suspect it won't hurt right. Also I need to test that first to

Geosearchef (dismissed):
Why are there two?

1-alex98:
uninstaller and installer

Geosearchef:
if this were normal java code I'd say replace the hardcoded `4000` in the sleep function with `Duration.ofSeconds(4).toMillis()` but as we have no imports here and would have to use the fully qualified name I'd keep it that way.

1-alex98:
@Geosearchef so approve it?

Geosearchef (approved):
I had my finger hovering over the approve button yesterday, but I didn't feel qualified enough to approve such a "hacky"/controversial change. 😅

1-alex98:
Well I built the installer and it works. That's what counts right. Dont like the idea either but it works so fuck it XD Would prefer the ice adpater to close better. It should exit if there is no connection to game for like a few minutes I guess.

PR #1584, opened 2020-02-19
Fix error when joining private games from Discord

iczero:
When attempting to join private games, there is an error due to not running in the ui thread.


1-alex98 (approved):

1-alex98:
Please create an issue next time and link it to your PR so change shows in changelog

Issue #1100, opened 2019-01-23
Periodically verify game location settings


Askaholic:
I recently had the issue that my game location was configured wrong because it used to be installed on one drive, but then was moved to a different drive. The `Supreme Commander Forged Alliance` folder still existed, but it only had the `maps` folder left. Everything else including `bin` got deleted.

The result is that when you try to launch a game, the client does it's usual update and download map stuff, but then just stops with no sign of an error or anything. (It logs the exception which is how I discovered that this is what was happening).

It would be nice if:
  * a). There were some error handling for File IO (because lets be honest, we can't seriously expect file io to never fail)
  * b). There were an error message indicating that the game path was set incorrectly or was no longer valid. Checks for this error could happen on startup and notify the user to fix their settings, or even be triggered from the error handling in part a. Better than just displaying an error, this could even grey out the host/join game buttons so that the user isn't fooled into thinking everything is ok.

To my pleasant surprise the client already implements some validity checks when setting the game path initially which would have been part c.

Katharsas:
Maybe same: https://forums.faforever.com/viewtopic.php?f=3&t=17694

PR #1595, opened 2020-02-29
Always ask for game path

### LINKED ISSUE #1355: prevent users from using the wrong game path (again)
had a user this week on discord who couldn't run the game
he managed to select vanilla supcom for the base game install instead of forged alliance

please don't allow users to select the vanilla supcom install directory (tested myself and the client doesn't throw an error when trying to do so in 0.10.7)

### LINKED ISSUE #1594: Player not always asked if he has not set the game path
Visit Issue hunt...

1-alex98:
#1594


src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -312,6 +319,12 @@ public ReadOnlyBooleanProperty gameRunningProperty() {
       return completedFuture(null);
     }
 
+    if (!preferencesService.isGamePathValid()) {
+      CompletableFuture<Path> gameDirectoryFuture = new CompletableFuture<>();
+      eventBus.post(new GameDirectoryChooseEvent(gameDirectoryFuture));
+      return gameDirectoryFuture.thenCompose(path -> joinGame(game, password));
+    }
+
```
Geosearchef: This code is duplicated 6 times, I'd extract the first 2 lines to a method returning the gameDirectoryFuture.
```
if (!preferencesService.isGamePathValid()) {
    return ensureGameDirectorySet().thenAccept(p -> joinGame()/hostGame/whatever);
}
```

Geosearchef (dismissed):

src/main/java/com/faforever/client/game/GameService.java
1-alex98: true

1-alex98: Acctually no it depends on the method look closely to the differences

1-alex98: There is not much to refactor

Askaholic:
While you are working on this can you take a look at #1100 and #1355. Those problems may be solvable or solved by similar changes.

1-alex98:
Well they are related yes but seperate issues

1-alex98:
#1100 kind of gets fixed by that because it checks with every game start which kind of is periodically

1-alex98:
Before the check was only done sometimes depending on how u started the game

1-alex98:
The check is also triggered if the game path is set but invalid

1-alex98:
Also fixes #1355

Slothologist:
The hashes created by the "sha256OfFile" function differ from the actual hashes (at least on linux). 
Because of that I never run into the "vanilla exe detected" case.
I fixed that in https://github.com/FAForever/downlords-faf-client/compare/bugfix/%231594-ask-for-game...Slothologist:bugfix/%231594-ask-for-game

It would maybe be better to check more explicitly like
```
If calculatedHash == FA.exe
    yay
else if calculatedHash == vanilla.exe
    oh no, you selected the wrong folder
else
    wtf this should not happen
```

1-alex98:
Well there are a ton of different FA version that are supported / ok I think so getting all the hashes might be hard @Slothologist I am not an expert on this but the steam version differs from the retail one but both are supported and therefore might have different hashes and who knows which other retail versions exist out there

1-alex98:
@Slothologist are u on Zulip? You wanna become part of Zulip(chat service)? We somehow need to test your thing on windows as well right?

1-alex98:
Write me an email if u want to alexander.von.trostorff@gmail.com

axel1200:
does not close for me

Slothologist:
> @Slothologist are u on Zulip? You wanna become part of Zulip(chat service)? We somehow need to test your thing on windows as well right?

Thanks for the Zulip invite, I am now part of that.
Furthermore, I tried your version out in Windows and the hash the function provides is identical to the one I got on Linux. So something like xng1gM.... for the FA exe (which is wrong). 

I see the problem with acquiring the hashes of all possible versions, but atm we are only filtering for the steam vanilla version.
Maybe we could just print a warning if an unknown exe is presented? Like "Alright, we seem to have no knowledge about this exe, make sure that you have selected the right folder, but go ahead, we won't stop you"?

1-alex98:
I think the way it is now is good(in your branch). For most ppl it will do and it is also a clear improvement. I rather have zero false positive errors or warnings and stay with it like it is now.

1-alex98:
Merged in your changes @Slothologist

1-alex98:
tested again works out well

1-alex98:
The hash is defined in the properties file and it is a list should there ever be other hashes we can simply add them by adding one line. So everything covered .;)

PR #1604, opened 2020-03-04
Wrap text in settings.fxml

### LINKED ISSUE #1600: Text wrapping
Not every line of text (string) is set to wrapping on window resize.

**!** = is not doing anything with text at all

# Settings
## General
* Remember my last opened tab
* **!** Unit database
* Disallow players joining...
## Chat
**Every string.**
## Notifications
**Every string.**
## Forged Alliance Forever
* Force relay connection
* Download missing maps automatically

I don't think there's anything to wrap in Account and Appearance tabs. Language is fine.

1-alex98:
Fixes #1600

### DIFF
--- a/src/main/resources/theme/settings/settings.fxml
+++ b/src/main/resources/theme/settings/settings.fxml
@@ -17,15 +17,16 @@
 <?import javafx.scene.layout.*?>
 <?import java.lang.String?>
 <VBox xmlns:fx="http://javafx.com/fxml/1" fx:id="settingsRoot" maxHeight="1.7976931348623157E308" prefHeight="640.0"
-      prefWidth="800.0" styleClass="decorated-window" xmlns="http://javafx.com/javafx/10.0.2"
+      prefWidth="800.0" styleClass="decorated-window, settings" xmlns="http://javafx.com/javafx/11.0.1"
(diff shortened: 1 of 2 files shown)


src/main/resources/theme/settings/settings.fxml
```diff
@@ -38,7 +39,7 @@
                                 <RowConstraints minHeight="10.0" prefHeight="50.0" vgrow="SOMETIMES"/>
                                 <RowConstraints minHeight="10.0" prefHeight="50.0" vgrow="SOMETIMES"/>
                                 <RowConstraints minHeight="10.0" prefHeight="50.0" vgrow="SOMETIMES"/>
-                                <RowConstraints minHeight="10.0" prefHeight="50.0" vgrow="SOMETIMES"/>
+                                <RowConstraints minHeight="10.0" vgrow="SOMETIMES"/>
```
Brutus5000, line 42: I don't think this was intentional?

1-alex98: it was, before the text only grew to 50 px know this stretches if the text needs to wrap more than 2 lines. I did this there because it can have potentially long texts

PR #1609, opened 2020-03-05
Change "Ø Rating" to "Avg. Rating"

msgerbs:
Change "Ø Rating" to "Avg. Rating" on the game list because this symbol is not commonly known to mean "Average" in some English-speaking countries. Should probably change in other localizations as well but I don't know how other languages would commonly abbreviate or otherwise represent "Average" so I don't want to change it myself and make it worse.


1-alex98:
Next time please reference the issue by writing "Fixes #number" in your commit

1-alex98:
localization is not done here but in poeditor

PR #1625, opened 2020-03-17
Uid not found in folder with space

### LINKED ISSUE #1624: unix uid not found in folder with space
**OS**
Linux 

Visit Issue hunt...

1-alex98:
Also sets the executable flag when downloading the uid executable.
Fixes #1624

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -212,6 +212,9 @@ task downloadUnixUid(type: Download) {
   src "https://github.com/FAForever/uid/releases/download/v${faf_uid_version}/faf-uid"
   dest file("${buildDir}/resources/native/faf-uid")
   onlyIfNewer true
+  doLast {
+    file("${buildDir}/resources/native/faf-uid").setExecutable(true, false)
+  }
 }
 
 task downloadWindowsUid(type: Download) {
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -212,6 +212,9 @@ task downloadUnixUid(type: Download) {
   src "https://github.com/FAForever/uid/releases/download/v${faf_uid_version}/faf-uid"
   dest file("${buildDir}/resources/native/faf-uid")
   onlyIfNewer true
+  doLast {
+    file("${buildDir}/resources/native/faf-uid").setExecutable(true, false)
```
Geosearchef, line 216: there should already exist some java code doing that when the uid executable is called

Geosearchef (approved):

build.gradle
1-alex98: Well that code does not work then

Slothologist: > there should already exist some java code doing that when the uid executable is called

Does not seem to be the case, see https://github.com/FAForever/downlords-faf-client/blob/develop/src/main/java/com/faforever/client/legacy/PosixUidService.java
Would be weird to do this every time that exe is called.

PR #1627, opened 2020-03-17
Preferences file read after deletion

### LINKED ISSUE #1626: Error about preferences broken blocking user from launching client
**OS**
Linux

Visit Issue hunt...

1-alex98:
Fixes #1626

### DIFF
--- a/src/main/java/com/faforever/client/preferences/PreferencesService.java
+++ b/src/main/java/com/faforever/client/preferences/PreferencesService.java
@@ -43,8 +43,8 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.security.MessageDigest;
 import java.security.DigestInputStream;
+import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.text.MessageFormat;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/preferences/PreferencesService.java
```diff
@@ -181,11 +182,16 @@ public static void configureLogging() {
...
+   * @return true if the file was deleted
    */
-  private void deleteFileIfEmpty() throws IOException {
+  private boolean deleteFileIfEmpty() throws IOException {
     if (Files.size(preferencesFilePath) == 0) {
       Files.delete(preferencesFilePath);
+      preferences = new Preferences();
+      return true;
     }
+    return false;
```
Geosearchef, line 194: `else {
    return false;
}`

Geosearchef (approved):

src/main/java/com/faforever/client/preferences/PreferencesService.java
1-alex98: no why???

Geosearchef: cause it's ten times more readable. No need to use the "abort" case here.

1-alex98: we do this code style all over the application

PR #1641, opened 2020-03-29
Keep context menu open

### LINKED ISSUE #1429: Users online/offline state bumps the chatlist, closes context menu.
**Describe the bug**
if you have a context menu open in the userlist (right click to add friend etc) and someone goes offline/comes online. the menu closes. during active periods this can make it very difficult to add friends etc. 

furthermore the list moves slightly (which may be the cause) this can also make clicking on the intended person difficult during busy times

at the very least i dont want the context menu to close. prefer if it also listed the relevant username, with flag preferred.

it's very triggering.

1-alex98:
Fixes #1429

Having the window as the owner makes it not close it now is open until it loses focus


Geosearchef (approved):
nice!

PR #1646, opened 2020-03-29
#1645-add-faction-icon-replay-detail

### LINKED ISSUE #1645: Add faction logo next to name in replay detail view
Just like in the Python client you should see the faction of a player in the replay overview.

FemtoZetta:
fixes #1645
-added coloured icons next to the name in the replay detail overview
-also fixed long names getting cut off and not being able to see the rating
-also fixed rating change being shown as "..." when there were long names


src/main/java/com/faforever/client/game/PlayerCardTooltipController.java
```diff
@@ -48,9 +53,44 @@ public Node getRoot() {
...
   }
+
+  private void setFactionIcon(Faction faction) {
+    if (faction == null) {
+      return;
+    }
+
+    switch (faction) {
+      case AEON:
+        factionIcon.setVisible(true);
```
Askaholic: You have this same line in every case. How about moving it after the `switch`.

src/main/java/com/faforever/client/game/PlayerCardTooltipController.java
```diff
@@ -48,9 +53,44 @@ public Node getRoot() {
...
+        break;
+      case UEF:
+        factionIcon.setVisible(true);
+        factionIcon.setText("\uE904");
+        factionIcon.getStyleClass().add("uef");
+        break;
+      default:
+        factionImage.setVisible(true);
+        factionImage.setImage(randomImage);
+        break;
```
Askaholic: Alex asked me about this earlier. I think in normal FAF having faction 5 means random, but I'm not sure about nomads. I would recommend you fire up a nomads game and see what the faction numbers for random and nomads are. It could be that in nomads 5 is nomads and 6 is random.

src/test/java/com/faforever/client/game/PlayerCardTooltipControllerTest.java
```diff
@@ -66,8 +73,11 @@ public void testSetOther() {
         .defaultValues()
         .socialStatus(SocialStatus.OTHER);
     Player player = playerBuilder.get();
-    instance.setPlayer(player, 1000);
+    instance.setPlayer(player, 1000, Faction.NOMAD);
```
Askaholic: Yea, not sure about this test.

src/main/java/com/faforever/client/game/PlayerCardTooltipController.java
1-alex98: Well i think what he does making every unknow display as random seems ok to me

src/test/java/com/faforever/client/game/PlayerCardTooltipControllerTest.java
1-alex98: Well yes but let's not blame him I think it is ok. this whole faction number thing is fucked

Askaholic: I'm just saying maybe don't make a test that asserts wrong behavior.

FemtoZetta: I changed the enum name to RANDOM instead of NOMAD. It's not used anywhere else, so shouldn't be a problem.

1-alex98 (approved):

src/main/resources/theme/style.css
```diff
@@ -1002,6 +1002,28 @@
   -fx-opacity: 1;
 }
 
+/***************** Faction label *****************/
+
+.icon.faction-label .text, .icon.faction-label {
+  -fx-font-size: 2em;
+}
+
+.faction-label.aeon .text {
```
Askaholic, line 1011: So are these even needed then?

1-alex98: No u are right be need to remove those

PR #1650, opened 2020-03-29
Remove race condition that results in newbie channel being joined

### LINKED ISSUE #1617: Rarely get connected to #newbie chat without having done so manually
Since 1.1.3 I sometime get connected to newbie chat randomly after starting the client even though I have over 800 global games (and over 60 ladder) and never joined that chat manually. Happened two times so far. Logs are pointless as far as I can see because they don't log which channels you join, so no logs for you...
Axel asked me to open an issue for it. 
In-game name is FemtoZetta (for testing purposes).

1-alex98:
Fixes #1617

### DIFF
--- a/src/main/java/com/faforever/client/chat/PircBotXChatService.java
+++ b/src/main/java/com/faforever/client/chat/PircBotXChatService.java
@@ -12,6 +12,7 @@ import com.faforever.client.player.PlayerOnlineEvent;
 import com.faforever.client.player.PlayerService;
 import com.faforever.client.player.SocialStatus;
 import com.faforever.client.player.UserOfflineEvent;
+import com.faforever.client.player.event.CurrentPlayerInfo;
 import com.faforever.client.preferences.ChatPrefs;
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/client/chat/PircBotXChatService.java
```diff
@@ -103,6 +104,7 @@
   private static final List<UserLevel> MODERATOR_USER_LEVELS = Arrays.asList(UserLevel.OP, UserLevel.HALFOP, UserLevel.SUPEROP, UserLevel.OWNER);
   private static final int SOCKET_TIMEOUT = 10000;
   private static final String NEWBIE_CHANNEL_NAME = "#newbie";
+  public static final int GAMES_FOR_NEWBIE_CHANNEL = 50;
```
Geosearchef: MAX_GAMES

Geosearchef (changes requested):

Geosearchef (approved):

PR #1652, opened 2020-03-30
#1651-improve-chat-readability

### LINKED ISSUE #1651: Improve chat readability
Make the chat easier to read. Using the Python client, a few easy suggestions that would already help a lot:
-make the names/clan/flag right aligned
-add ":" after the name, before the message

Example to show that it does make it easier to visually match names with messages:
Python:
grafik
Java:
grafik

FemtoZetta:
fixes #1651:
-adds ":" after the name in chat
-makes the flag/clan/name right aligned in chat

### DIFF
--- a/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
+++ b/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
@@ -556,7 +556,7 @@ public abstract class AbstractChatTabController implements Controller<Tab> {
     String timeString = timeService.asShortTime(chatMessage.getTime());
     html = html.replace("{time}", timeString)
         .replace("{avatar}", StringUtils.defaultString(avatarUrl))
-        .replace("{username}", login)
+        .replace("{username}", login + ":")
         .replace("{clan-tag}", clanTag)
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -556,7 +556,7 @@ private String renderHtml(ChatMessage chatMessage, URL themeFileUrl, @Nullable I
     String timeString = timeService.asShortTime(chatMessage.getTime());
     html = html.replace("{time}", timeString)
         .replace("{avatar}", StringUtils.defaultString(avatarUrl))
-        .replace("{username}", login)
+        .replace("{username}", login + ":")
```
1-alex98, line 559: needs to be done why I18N Sevice class and the String added to message.properties file. You may say this does not need to be translated but there are languages that work right to left in that case the : would be on the other end

1-alex98 (changes requested):

FemtoZetta:
Gotta fix more stuff (not just translation) and also add the same for the extended version, so I'm closing this for now.

Issue #1663, opened 2020-04-02
Users get stuck in the login screen waiting forever


1-alex98:
Users do not seem to become a "welcome" message and get stuck loging in forever.

https://github.com/FAForever/server/issues/562

Possible ways to improve the situation:

- [ ] Make a timeout after which we cancle log in
- [x] Improve loging and catch and log potentially swallowed exceptions

User posting this on forum:
https://forums.faforever.com/viewtopic.php?f=3&t=18996&p=183021#p183021

**Wanna have the bug fixed quickly?**
Visit Issue hunt...
[![Issue hunt](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg?sanitize=true)](https://issuehunt.io/r/FAForever/downlords-faf-client)

Askaholic:
So in the server log I do see quite a few
```
Client <username> dropped. You have been signed out because you signed in elsewhere.
```
which might mean that old connections aren't being closed properly. Or maybe somehow people trigger two connections at the same time which kick each other out.

PR #1664, opened 2020-04-02
Consider failing login task

1-alex98:
make sure the login future is completed exceptionally if anything unsuspected happens.
Prepares #1663

### DIFF
--- a/src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
+++ b/src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
@@ -95,9 +95,9 @@ import org.springframework.context.annotation.Profile;
 import org.springframework.scheduling.TaskScheduler;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
-import org.springframework.util.ClassUtils;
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -458,10 +476,7 @@ private void parseServerObject(String jsonString) {
             .forEach(consumer -> consumer.accept(serverMessage));
         messageClass = messageClass.getSuperclass();
       }
-      for (Class<?> type : ClassUtils.getAllInterfacesForClassAsSet(messageClass)) {
-        messageListeners.getOrDefault(messageClass, Collections.emptyList())
-            .forEach(consumer -> consumer.accept(serverMessage));
-      }
```
1-alex98, line 464: this loop made no sense at all

1-alex98: it seems to be a left over

1-alex98: look how the variable is not used in the loop and the same call is made in the loop above

PR #1674, opened 2020-04-19
Fix for situations where replays weren't saved locally

### LINKED ISSUE #1666: Local replays list not updated after game ended
**Describe the bug**
Local replays menu not shows new games that I just finished. Also there are no refresh button or something similar.
Also I noticed about default setup for Available replays menu which is set up by default for my games. But it bugged or something because it shows just finished game but replay has 0 B size and shows exception when I try to launch it. I think that it happens because replay is not ready at server side yet. But local replay should be ready when I quit game and should be watchable!

**To Reproduce**
Open Local replays menu, then go play some game, then return to Local replays menu and see that new game not appears in the list.

**Expected behavior**
New game appears right after you quit FA client. Like it was in old good FAF client.

**OS**
Windows

Visit Issue hunt...

andreasley:
Fixes https://github.com/FAForever/downlords-faf-client/issues/1666

Previously, the client would fail to save a replay if the `GameStatus` was set to `CLOSED` before the replay was saved. When FA exited and the client tried to create the replay file, the respective `Game` couldn't be found anymore in the local cache and the task to create the replay file failed.

With this PR, the `Game` is fetched when recording of the replay starts.

### DIFF
--- a/src/main/java/com/faforever/client/replay/ReplayServerImpl.java
+++ b/src/main/java/com/faforever/client/replay/ReplayServerImpl.java
@@ -80,7 +80,7 @@ public class ReplayServerImpl implements ReplayServer {
   }
 
   @Override
-  public CompletableFuture<Integer> start(int gameId, Supplier<Game> onGameInfoFinished) {
+  public CompletableFuture<Integer> start(int gameId, Supplier<Game> gameSupplier) {
     stoppedGracefully = false;
     CompletableFuture<Integer> future = new CompletableFuture<>();
     new Thread(() -> {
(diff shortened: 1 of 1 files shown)


Katharsas:
That might be a solution.

The only possible problem i see here is the object getting replaced in the `uidToGameInfoBean` map while the lobby is running (instead of changed via setters) but i think that never happens (lets hope). But maybe check that once a game lobby has started, the game object is not replaced, only its properties changed.

1-alex98:
Looks reasonable... Maybe u have a creative idea who to keep this from happening again by writing a test...

1-alex98:
Yeah just make it final maybe to address @Katharsas fears

1-alex98:
Actually forget it no idea what @Katharsas means :D

1-alex98:
Because this affects many ppl I guess I am gonna make a new release with this. After u addressed @Katharsas concerns whatever they might be :D

andreasley:
I don't think it's likely that the `Game` bean is ever changed in a way that is problematic here. Especially since we're now fetching it when the game is actually started, so there won't be any changes to the list of players or the map etc.
If the `Game` object is replaced in `uidToGameInfoBean`, that's not a problem since `ReplayServerImpl` just keeps its (now separate) instance which has enough information for the replay file.

As for a test, I don't see how we can test this in a useful way, since it involves interactions with the server and FA.

1-alex98:
We make a RC and see if ppl complain I also dont see reasons why the instance might change. the instances are also not replaces in the uidToGameInfoBean but the existing references are updated there is a method updateFromDto or something like that which updates the instances...

1-alex98:
Building a RC

Katharsas:
Ok then it looks good.

PR #1684, opened 2020-04-27
Only fetch http headers on checking for download file size

### LINKED ISSUE #1683: CheckForUpdateTasks downloads file for file size check
The current implementation of the update tasks downloads chunks or even the whole update file on checking the file size.
This also explains the absurd high download counts on Github.

Brutus5000:
(fixes #1683)


1-alex98 (approved):

Brutus5000:
I have tested this with some analytics tools.
Before the change the client downloads roughly 220kb to catch the file size.
Wiht the change it downloads 0 bytes, just the header information.

PR #1693, opened 2020-05-09
Map generator button #1321

### LINKED ISSUE #1321: Emphasize generate map button
The generate map button is nearly invisible, maybe we should somehow highlight it, e.g. add a "new" tag to it, a generated map logo or change the color.

Katharsas:
- rearranged buttons in create game dialog
- put create_game.fxml grid elements into logical order
- removed default button, instead is shown in list
- map generator button "new" label and color

Fixes #1321


Katharsas:
Currently looks like this:
map_generator_example_2

If anybody knows how to improve this, tell me. Still not sure about position of map generator button (ill see how it looks when i put it above map preview image), and colors. Also not sure how big to make Cancel/Create buttons. If i make them full column width they look like they belog to the corresponding colum which i dont want, and it looks weird:
map_generator_example_1

BlackYps:
I made a quick mockup of a new concept. I think before the players were likely to miss some of the settings on the top. Now the eye is guided better, because there is a clear direction from left to right.

create game concept

I also changed some colors even though that technically affects more than just this window.
I do believe that we should get a nicer color than this ugly green one and the important buttons should not be white. (Search button in the vault and buttons that lead the player to a game. You can easily miss them, especially when they are grey.) So I chose orange for the button and blue because it complements the orange nicely.
What do you think?

Katharsas:
Well first im only interested in making the "Map generator" button more visible (that the point of this PR) and maybe cleaning up the clutter here and there, because we wanted to do that before promoting the map generator in general (like making generator tourney). Im not gonna change the whole color theme or make a major changes to the layout in this PR (however i agree that the current color theme is not optimal). 

I don't see the value from moving general input elements that have nothing to do with the map into the map details area. Also the space that you have put eveything is were the map description is currently (just some maps dont have one, see my second screenshot), so there is no space there. Also its way too little space anyway like a lot of titles wont even fit in there. And you alwasy need to think that in other language the rating text description could be longer, so this is way too small.

Btw, you can pretty easily make your own theme by simply changing colors in style.css:
https://github.com/FAForever/downlords-faf-client/blob/develop/src/main/resources/theme/style.css

BlackYps:
I agree that this PR is not the right place to work on all the changes. I just wanted to set up a new coherent concept for this window.
I think it is important to not have the create game button and important game settings in opposite corners. Thus I ditched the concept of a "map details area" it is now more like a "final settings area".
The generate new map button should probably be under the map column, because this is the most related area. I think it also helps to have the longer text on the buttons because I always think "random map" is the map generator button.
Title and rating line could maybe be two lines to have more space. Most titles are shorter than 20 letters and the title field allows to write longer text inside it, if you have an extra long title.
Edit:
The map description just clutters the window and gives no relevant information when I want to decide what map to play. Except 1% of maps that write something gameplay related there, but the description is accessible in the vault or in the lobby. But that is not worth it to display every map descripion in a place that can be used better.

Katharsas:
>I think it is important to not have the create game button and important game settings in opposite corners

Why? Most forms i have seen have general stuff on top and continue/ok buttons at the bottom. It makes sense because you start at the top, fill everything out until you arrive at the bottom for continue button. In your example i first go to the bottom for map title, than back to the top/left area to select other things, then back to bottom to Contiue. Doesnt make anys sense to me.

The first thing you do should be in a differernt area than the last thing you do, so you can smoothly go through the form.

Katharsas:
Result:
map_generator_example_3

shalkya:
UI wise it seems mostly fine by me, but i'm not too sure about having the "Generate new map" with turquoise background. This colour is used as an indicator of selection in the others column. I get that you want to put it forward, but maybe use another color, or consider that the "NEW" tag is enough to bring attention to it.

At least, it will make people notice about it. Question is when will we remove the "NEW" tag? in 1 year ? Would also be nice to have improvement on the generator, so i can host tourney out of it. But that's not the topic of this PR.

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -140,8 +148,19 @@ public void initialize() {
...
+    Function<FeaturedMod, String> isDefaultModString = mod ->
+      Objects.equals(mod.getTechnicalName(), KnownFeaturedMod.DEFAULT.getTechnicalName()) ?
+      " " + i18n.get("game.create.defaultGameTypeMarker") : null;
+    
+    featuredModListView.setCellFactory(param ->
+        new DualStringListCell<FeaturedMod>(FeaturedMod::getDisplayName, isDefaultModString, false) {
+      @Override
+      protected void init() {
+        // use bind to overwrite updates from css
+        right.textFillProperty().bind(new SimpleObjectProperty<Paint>(Color.GREY));
```
1-alex98: Try avoid UI style set in java code at all costs, use Pseudoclass

1-alex98 (changes requested):

src/main/java/com/faforever/client/game/CreateGameController.java
Katharsas: I can try but it might not be possible due to how `DualStringListCell` works.

1-alex98: it is just give it a style class(possibly in code) and set this style class in css

src/main/java/com/faforever/client/fx/DualStringListCell.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Alternative to {@link StringListCell} that allows to display to differently styled strings in each cell.
+ */
+public class DualStringListCell<T> extends ListCell<T> {
+  protected Label left = new Label();
+  protected Label right = new Label();
+  protected Region spacer = new Region();
+  
+  private HBox container = new HBox(left, spacer, right);
```
1-alex98: This should lso be done in fxm

1-alex98 (changes requested):

Katharsas:
Thanks for completing @axel1200 
Didnt have the time unfortunatly.

src/main/java/com/faforever/client/game/CreateGameController.java
Katharsas: Ok problem was that (default) ended up with other font/size than the other cell. Need to check when i have time.

src/main/java/com/faforever/client/fx/DualStringListCell.java
```diff
@@ -48,12 +26,15 @@ protected void updateItem(T item, boolean empty) {
         setText(null);
         setGraphic(null);
       } else {
-        left.setText(functionLeft.apply(item));
-        right.setText(functionRight.apply(item));
+        if (dualStringListCellController == null) {
+          dualStringListCellController = uiService.loadFxml("theme/dual_string_list_cell.fxml");
```
Katharsas, line 30: ~~This is not very performance friendly.
The JavaFX documentation (in some parent class) explicitely states that updateItem should ONLY reuse ui elements not create or recreate them.~~

Edit: Ok maybe i got this wrong, this usually only happens once per object right?

1-alex98: yes, this is essentially the same as u did...

PR #1701, opened 2020-05-17
rearranged and reworded menu dropdown list

### LINKED ISSUE #1700: Finetune main menu
The preferences option is at the bottom of the dropdown menu in a lot of other software. The preferences should go down to the second to last entry, so it is where the user expects it.
The links to the directories should go up, so the chunk of similar entries is not in the middle but on one of the ends of the list.
"Reveal" should be reworded to the more common word "show".

BlackYps:
Fixes #1700


1-alex98 (approved):

1-alex98:
tested works

PR #1703, opened 2020-05-17
#1702 Create game button gets pushed up

### LINKED ISSUE #1702: Create game button gets pushed up behind other UI elements
**Describe the bug**
when the window is narrow the flowpane with the create game button in the custom games tabe gets pushed up when in tiles view.

**Expected behavior**
Flowpane should expand to bottom

**Screenshots**
grafik

BlackYps:
Fixes #1702 

also increases padding so the tiles line up with the button


1-alex98 (approved):

1-alex98:
tested works

PR #1706, opened 2020-05-18
New news page

BlackYps:
Anihilnine requested me to make the news tab load the soon-to-be-deployed new news page.

The logic for the listview with the news items got kicked out. I hope I didn't forget to delete any leftover code.

Looks like this now:
grafik


src/main/java/com/faforever/client/news/NewsController.java
```diff
@@ -108,38 +87,14 @@ public void onUnreadNewsEvent(UnreadNewsEvent unreadNewsEvent) {
...
-        onLoadingStop();
-        if (!newsItems.isEmpty()) {
-          NewsItem mostRecentItem = newsItems.get(0);
-          preferencesService.getPreferences().getNews().setLastReadNewsUrl(mostRecentItem.getLink());
-          preferencesService.storeInBackground();
-        }
-        newsListView.getSelectionModel().selectFirst();
-      });
+    Platform.runLater(() -> {
+      newsWebView.getEngine().load("https://www.faforever.com/news");
```
Brutus5000: Never ever do hardcoded urls. 
We either load it from the FafApiProperties class and declare it in the application.yml
Or we add it to the remote preferences which would give us a bit more flexibility until things have settled.

src/main/java/com/faforever/client/news/NewsController.java
```diff
@@ -31,25 +31,14 @@
 @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
 public class NewsController extends AbstractViewController<Node> {
 
-  private static final ClassPathResource NEWS_DETAIL_HTML_RESOURCE = new ClassPathResource("/theme/news_detail.html");
-  private final PreferencesService preferencesService;
-  private final I18n i18n;
-  private final NewsService newsService;
```
Brutus5000, line 37: I am missing changes to remove this service and related tests.

Brutus5000 (changes requested):

Katharsas:
Can you hide website menu and headline (is it possible to add CSS to webview?) ?

src/main/java/com/faforever/client/news/NewsController.java
BlackYps: I can do it like the Unitdb, that uses the application.yml and Clientproperties.java
Is that a good way?

BlackYps:
No need to hide the website header. Anihilnine said that the new site won't have a header.
It is possible to add css. In fact there is a style-webview.css already.

anihilnine:
> No need to hide the website header. Anihilnine said that the new site won't have a header.
> It is possible to add css. In fact there is a style-webview.css already.

yeah it will be removed in the website itself

1-alex98:
Wow I don't like that. It removes other features we have while not bringing any new ones:
- Notifying and displaying newly news items
- Displaying a button that redirects to the ladder maps on posts with the corresponding topic
- Refreshing the news in intervals and displaying new 

Including the website seems not the best choice. The user can click on other parts of the website. The web view is not the best. Changes on the website might break the client. What does including faforever.com give as an advantage over the thing there was before. 

I don't wanna maintain this in that way.

1-alex98:
This will end in css making assumptions about the structure of the website etc. or injecting js code that does the same to see what news was looked at...

anihilnine:
Hey axel, sorry mate I tried and talk to you about all of this on Zulip before but I think you are away or something, I did talk to Brutus for an hour on voice. This is my first major project as Promotions Councilor. There are some changes in progress to change the news website as well. It will end up looking like this:
 
newshub

We didnt consider the points you raised about new items, ladder link. I am not sure what do to about that, what do you think?

I did ask for the frame to be refreshed whenever you change tabs, or otherwise I can get JS to do it
And FYI I have also asked to temporarily disable the "launch client at last visited tab" feature so people see the news for a little while.

As far as the user being able to click anywhere goes, I have full control over the web page so I am going to not have the standard website header displayed. I have not been able to test yet but we need to make sure that links, when clicked, open in an external browser and not in the faf client.

The benefits of having it as a website is
a) consistent with website. The website itself will have the same UI. I will probably try and change the website homepage to be news too but thats a different conversation
b) I can actually get the project done without large amounts of dev time from other people, so it makes the project more realistic - there are 4 of us and we already have wordpress, server, html, css, java client change done in 7 days. 
c) flexibility. if there was some huge promo we could remove all the news and add in a fullscreen video instead.

The benefits of having the newsfeed are being able to promote faf internally to all of the players. FAF is an interesting place with many interesting projects like balance patches, Nomads, RPG Mod, Zombie Mod, new maps, new ideas. Plus we can showcase existing lesser known features like what is Zone Control? What is Claustrophobia? Then we can add some zing to tournaments, WWPC. The content will be refreshed every day or three creating a live interesting news hub that players would return to. Part of my agenda as an author will be to plug Ladder - I will always be nodding to Ladder in glory, which is why Ladder entries in my mockup are in gold. I plan on running some stats to see the effect the newshub has - like how many players sign up for tournaments, join nomads discord, play ladder, etc.

I have the assumption that there are lots of Supcom players that don't know about FAF. Personally I played on steam for a year before moving to FAF. I knew vaguelly about FAF the whole time but I just didn't care. The benefit of having the newsfeed on the website is potential players will come and see a vibrant, thriving community. Maybe more people will join - I'm going to measure that too.

I'd love to chat and work out a way to develop this in a way that you are happy with?

anihilnine:
It would be easy for me to make an api on the website to return the Last News Update Date, but I guess you would prefer that to be in the real api. 

We could establish a protocol between the frame and the java client. I don't have any experience with java web frames but in other languages they can intercept links before they are followed - we could make convention like anything prefixed with "fafclient:" is an in client request. Then using some CSS I can hide those html buttons to not display on the website. We dont want a "click here to go to the java client mapvault" on the website news.

Brutus5000:
@anihilnine Do we know the URL of the news endpoint already?

We have two options basically:
Either we put the news url into the static configuration, then we can only change it with a client release, or we put it in the remote preferences (loaded from the server on startup), then I can change it server side if needed. The latter approach offers instant changes for all java client users, but has the drawback that it doesn't integrate well for tests with the test server. (The static config can point to a dedicated news page on the test server if you connect to test.)

anihilnine:
The URL is https://www.faforever.com/newshub

I have no opinion on how to store the url, if I was making the change I would just follow the precedents of the other pages

1-alex98:
We have a way of saving cookies in the settings file for the unit db you can do the same with the news page if u ever intend to store what news the user has already clicked

anihilnine:
Thanks axel but did you read my big post ^^ and are we going to be able to proceed with this web approach of getting newshub into the java client?

1-alex98:
Well I now read it. Still don't think it is a good idea. It seems like yet another legacy(in the future) service. And embedding websites makes stuff harder. Btw just including a big video is not that easy. Youtube embeeds don't work. And embeeding videos was possible before (I men not really as yt embeeds did not work), has it was partly a webview as well.

But well if Brutus is fine with it go ahead

src/main/java/com/faforever/client/news/NewsController.java
```diff
@@ -1,57 +1,37 @@
...
 import org.springframework.context.annotation.Scope;
-import org.springframework.core.io.ClassPathResource;
 import org.springframework.stereotype.Component;
 
-import java.io.InputStreamReader;
-import java.io.Reader;
-
 @Component
 @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
 public class NewsController extends AbstractViewController<Node> {
```
1-alex98, line 23: Check for related tests to remove

BlackYps:
I couldn't find any related tests and I think there aren't any, because I removed a lot of code and still all tests pass.

src/main/java/com/faforever/client/news/NewsController.java
```diff
@@ -108,38 +78,15 @@ public void onUnreadNewsEvent(UnreadNewsEvent unreadNewsEvent) {
...
-          NewsItem mostRecentItem = newsItems.get(0);
-          preferencesService.getPreferences().getNews().setLastReadNewsUrl(mostRecentItem.getLink());
-          preferencesService.storeInBackground();
-        }
-        newsListView.getSelectionModel().selectFirst();
-      });
+    Platform.runLater(() -> {
+      Website website = clientProperties.getWebsite();
+      newsWebView.getEngine().load(website.getNewsHubUrl());
+      onLoadingStop();
```
1-alex98: The loading should be removed the way it is now it is technical impossible to see the loading spinner

1-alex98 (changes requested):

src/main/java/com/faforever/client/news/NewsController.java
1-alex98: Well or listen on the engine finishing the loading

1-alex98 (changes requested):
Missing removal of tests

norraxx:
When tabs are switched, or news tab is not visible, is the webview unloaded?

Well, it'll be main question, memory consumption from top players and slow pc players.

BlackYps:
There are no tests to be removed, because all the news stuff is not covered by tests

1-alex98 (approved):

1-alex98:
@BlackYps every issue needs an issue

BlackYps:
Fixes #1738

1-alex98 (changes requested):
Website texts are cut off

1-alex98:
Screenshot from 2020-05-30 23-11-25

1-alex98 (approved):
just Linux so who cares

Katharsas:
Try defining "letter-spacing" in css @anihilnine

PR #1707, opened 2020-05-19
update java and java fx versions

1-alex98:
Not sure if we should merge it. There are a lot of crashes for me lately and they might be caused by this


build.gradle
```diff
@@ -28,19 +27,19 @@ plugins {
...
 apply plugin: 'propdeps'
-apply plugin: 'jacoco'
-apply plugin: 'com.github.kt3k.coveralls'
 apply plugin: 'distribution'
 
 
-sourceCompatibility = JavaVersion.VERSION_1_10
-targetCompatibility = JavaVersion.VERSION_1_10
+sourceCompatibility = 1.11
+targetCompatibility = 1.11
```
Brutus5000: Just `11` as far as I know.

build.gradle
```diff
@@ -231,10 +230,6 @@ task downloadNativeDependencies(dependsOn: [downloadUnixUid, downloadWindowsUid,
 compileJava.dependsOn 'downloadNativeDependencies'
 compileJava.dependsOn ':webview-patch:build'
 
-task wrapper(type: Wrapper) {
-  gradleVersion = '4.5.1'
-}
```
Brutus5000, line 236: I don't think this should be removed.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Mon Jan 30 18:27:36 CET 2017
+#Sat May 02 19:00:33 CEST 2020
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-all.zip
```
Brutus5000: I can hardly believe that we jump more than 6 (?) version without major issues? From my experience there is always a lot of stuff broken. Did you really test this with the new wrapper? (Especially since you removed the wrapper task.)

Brutus5000 (changes requested):
I support moving to JDK11 even if it adds crashed. Staying on 11 is no option. If there are new bugs we need to find and fix them.

gradle/wrapper/gradle-wrapper.properties
1-alex98: Yes and removing the wrapper task was one thing that I needed to do to get it to work...

Brutus5000: We have the latest Gradle running with wrapper task at work. This is suspicious. Makes me think that then Travis just uses an older Gradle version.

1-alex98: ???

1-alex98: Well instead of overwritting you can edit it yes I can do that instead

1-alex98: Look at @micheljung PR he does the same :D

build.gradle
1-alex98: @micheljung does the same :D

1-alex98: https://stackoverflow.com/questions/25769536/how-when-to-generate-gradle-wrapper-files I custom wrapper task in not needed anymore

gradle/wrapper/gradle-wrapper.properties
micheljung: I was surprised, too, but I executed some Gradle tasks and there were no problems :)

1-alex98:
Sadly the RC does not build it has somthing to do with certs at Install4j download server

PR #1715, opened 2020-05-21
Rework of sub-tabs

### LINKED ISSUE #1712: relocate tabs on the left to be under the menu buttons
Having the tabs vertically on the left wastes a lot of space, because there can't be any other elements under them.
It is also more intuitive to have them under the menu buttons and in a similar style, so you can't miss that they exist. (Can especially be an issue in the leaderboards and play tab).

### LINKED ISSUE #1713: Sub-tabs always default to first entry
The tabs on the left don't remember the state in which you left them.
Example to reproduce: Open 1v1 under the ply tab. Go to vault. Go back to play tab. You are now at custom games again and not at 1v1.
This also happens in the vault and leaderboards
Solution: Remember the last state in which you left a tab.

### LINKED ISSUE #1714: It's hard to see which tab is selected
When navigating in the client it is hard to see at a glance which tab is the selected one, because there is only one very small green line as indicator.

BlackYps:
Fixes #1712 
Fixes #1713 
Fixes #1714 

grafik

### DIFF
--- a/src/main/java/com/faforever/client/play/PlayController.java
+++ b/src/main/java/com/faforever/client/play/PlayController.java
@@ -31,6 +31,7 @@ public class PlayController extends AbstractViewController<Node> {
   public Ladder1v1Controller ladderController;
   public CoopController coopController;
   private boolean isHandlingEvent;
+  private AbstractViewController<?> lastTab;
 
 
   public PlayController(EventBus eventBus) {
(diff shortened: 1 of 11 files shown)


src/main/java/com/faforever/client/play/PlayController.java
```diff
@@ -77,13 +77,6 @@ protected void onDisplay(NavigateEvent navigateEvent) {
     }
   }
 
-  @Override
-  protected void onHide() {
-    customGamesController.hide();
-    ladderController.hide();
-    coopController.hide();
-  }
-
```
1-alex98: Why delete that? That should not be done

1-alex98 (changes requested):

src/main/java/com/faforever/client/play/PlayController.java
```diff
@@ -59,18 +61,22 @@ protected void onDisplay(NavigateEvent navigateEvent) {
...
         ladderController.display(navigateEvent);
+        lastTab = ladderController;
       }
-      if (navigateEvent instanceof OpenCoopEvent) {
+      else if (navigateEvent instanceof OpenCoopEvent) {
         playRootTabPane.getSelectionModel().select(coopTab);
         coopController.display(navigateEvent);
+        lastTab = coopController;
+      } else if (Objects.equals(navigateEvent.getClass(), NavigateEvent.class)) {
+        lastTab.display(navigateEvent);
```
1-alex98: null check

1-alex98 (approved):

PR #1716, opened 2020-05-21
Make popup windows more consistent

BlackYps:
Brings the style of the popup windows (map, mod, and replay detail and create game) more in line with each other and fixes minor UI Issues.

All windows keep sensible min and max sizes and hover in the middle of the screen.

The Create Game window now has a proper cancel button and no superfluous title anymore
"Min" is now in front of the text field instead of behind
grafik

Map detail buttons are lower so they don't crash with the map name
"Description" removed
grafik

Mod detail install button is now a highlighted button
grafik

### DIFF
--- a/src/main/java/com/faforever/client/game/CreateGameController.java
+++ b/src/main/java/com/faforever/client/game/CreateGameController.java
@@ -179,7 +179,7 @@ public class CreateGameController implements Controller<Pane> {
   }
 
   public void onCloseButtonClicked() {
-    onCloseButtonClickedListener.run();
+    getRoot().setVisible(false);
   }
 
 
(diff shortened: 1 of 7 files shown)


1-alex98:
Failing CI

BlackYps:
Fixes #1733

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -179,7 +179,7 @@ public void initialize() {
   }
 
   public void onCloseButtonClicked() {
-    onCloseButtonClickedListener.run();
+    getRoot().setVisible(false);
```
1-alex98, line 182: Don't think that is sufficent to effectively close it

1-alex98 (changes requested):

src/main/java/com/faforever/client/game/CreateGameController.java
BlackYps: Previously this listener was created when the JFXDialog was created. I guess it was needed to make the dialog disappear properly / to trigger the close animation.
Now it just hides and reveals the create game window. It works in the same way in the mod and map details.
If you want to have it "more" closed I need you to go into more detail how you want to achieve that.
Don't know if it's worth the effort just to unload a few elements and one map preview picture from memory, though.

1-alex98 (changes requested):
Please use JfxDialog

1-alex98:
We wanna use jfx on vault not drop jfx in create game dialog

1-alex98:
outdated and no longer needed

PR #1717, opened 2020-05-22
convert show/hide details button in custom games to icon

### LINKED ISSUE #1732: Turn show/hide details button in custom games into an icon
This button would take less space and be pretty self-explanatory if it were an arrow icon.
grafik

BlackYps:
The button now takes less space and is more on the right

### DIFF
--- a/src/main/java/com/faforever/client/game/CustomGamesController.java
+++ b/src/main/java/com/faforever/client/game/CustomGamesController.java
@@ -280,10 +280,10 @@ public class CustomGamesController extends AbstractViewController<Node> {
 
   private void setSidePane(boolean displayed) {
     if (displayed) {
-      toggleSidePaneButton.setText(i18n.get("view.hideSidePane"));
+      toggleSidePaneButton.setText("");
       sidePaneColumn.setMinWidth(sidePaneColumn.getPrefWidth());
     } else {
-      toggleSidePaneButton.setText(i18n.get("view.showSidePane"));
+      toggleSidePaneButton.setText("");
       sidePaneColumn.setMinWidth(0);
     }
   }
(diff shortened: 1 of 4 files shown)


1-alex98:
Screenshot would be nice :D... guess have to look at it for myself

BlackYps:
grafik

1-alex98 (dismissed):

src/main/java/com/faforever/client/game/CustomGamesController.java
```diff
@@ -280,10 +280,10 @@ public void toggleSidePane() {
 
   private void setSidePane(boolean displayed) {
     if (displayed) {
-      toggleSidePaneButton.setText(i18n.get("view.hideSidePane"));
+      toggleSidePaneButton.setText("");
```
axel1200, line 283: Set in fxml

BlackYps:
Changed the icon:

grafik

Fixes #1732

1-alex98:
I really liked the first one better. The windows arrows are also that way on the other site the vault now has the other arrows so I guess we at least need to make it consistent.

Katharsas:
I think its important to make it completely clear which parts of the GUI are interactive and which are not. If buttons are added, they should imo look like buttons, so i would add some background color change to this. The Table/Tiles view switch could also use that.

BlackYps:
I don't really get what windows arrows you mean, but I agree that the first one was better.
I made it look more like a normal button now, that kind of sits on the edge that it is supposed to collapse:
I also fixed the Table/Tiles toggle buttons, so they look consistent to other buttons.
grafik

1-alex98:
merging as soon as tests run through... tell me if forget

PR #1719, opened 2020-05-22
Switch color theme to blue and orange

### LINKED ISSUE #1718: Switch colour theme and make buttons more visible
Important buttons like
 - search button in vault
 - create game button in custom games
 - create button in create game dialogue
are easy to miss / hard to find, especially when they are greyed out, because the white color doesn't really stand out in the various greys that the client uses.

Solution: Color the buttons in a color that attracts attention even when the button is greyed out.

BlackYps:
Fixes #1718 
To color the buttons in a nice way, the two colors have to have some harmony. I couldn't find a good color that fits with the green, so I changed the green to a blue shade.

Looks like this:
grafik
grafik

Quick mockup how it will look with the new tab buttons:
grafik


Brutus5000:
I personally don't like the new colors. And before we switch the color scheme for all users out there, we should rather fix theming itself.

BlackYps:
What needs to be fixed to make themes work?

1-alex98:
Good questions we don't know. @micheljung started it. It is still disabled in the settings but not sure how much is already done in the backgound

1-alex98:
But I agree I don't like it too much. I mean the idea of making buttons more visible is good. Just Oragne- Blue looks not so appealing to me

1-alex98:
That's also why you need to split your PRs. Because your work is great. But if we would refuse the colour schema and with that refuse all your nice changes that would be sad right?

Katharsas:
You know if this is adjustable by players, we could make a "find a good looking new color theme" contest^^
Basically we need at least like 10 different themes so we can decent towards the best-looking one, otherwise there is no chance that we can all agree on a theme.

In this particular color theme i don't like the shade of blue (a less redish blue would be better imo) and we still have grey on grey buttons which im not sure about.

BlackYps:
I don't think a contest is the best idea. Probably more productive to just ask a few people that we think are good at coming up with good color scemes.
I am not particularly invested into the blue and orange, but I think it is important that the "important" buttons are colored in some way. I am open for suggestions for any colors.

BlackYps:
I desaturated the button and changed the hue more to the blue end of the spectrum.

(pic shows the changes together with the changes from #1722 )
grafik

Katharsas:
I like that better. For the orange i think it could use a tiny bit more saturation so it looks less like brown. But now I could live with it. Now that i think about it, blue is the UEF color, we could have a color theme for every faction, wonder if it is possible to switch the theme at runtime.

1-alex98:
Same last one is better

1-alex98:
@Brutus5000 I would say let's use that.

BlackYps:
I increased the saturation and lightness on the yellow button a tiny bit and updated the colors for the charts.
I think it's ready now.
grafik

1-alex98 (approved):

PR #1722, opened 2020-05-25
Rework of sub-tabs II

BlackYps:
We found bugs in the already merged PR #1715 , so we had to revert it.
Here is the fixed version.

### DIFF
--- a/src/main/java/com/faforever/client/chat/ChatController.java
+++ b/src/main/java/com/faforever/client/chat/ChatController.java
@@ -269,7 +269,7 @@ public class ChatController extends AbstractViewController<Node> {
   }
 
   @Override
-  protected void onHide() {
+  public void onHide() {
     super.onHide();
     if (!tabPane.getTabs().isEmpty()) {
       Tab tab = tabPane.getSelectionModel().getSelectedItem();
--- a/src/main/java/com/faforever/client/fx/AbstractViewController.java
+++ b/src/main/java/com/faforever/client/fx/AbstractViewController.java
@@ -26,7 +26,7 @@ public abstract class AbstractViewController<ROOT extends Node> implements Contr
   /**
    * Subclasses may override in order to perform actions when the view is no longer being displayed.
    */
-  protected void onHide() {
+  public void onHide() {
     // To be overridden by subclass
   }
 }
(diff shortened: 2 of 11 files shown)


1-alex98:
naming schema for branches feature/#1715-rework-new-sub-tabs

1-alex98:
Might be that this happens in a different PR that I alredy merged from you, but now the client log is spammed with warning about broken css:
```
2020-05-25 18:28:30.902  WARN 7552 --- [lication Thread] javafx.css                               : Could not resolve '-fx-text-base-color' while resolving lookups for '-fx-text-fill' from rule '*.label' in stylesheet file:/C:/Users/Alexander/IdeaProjects/downlords-faf-client-2/out/production/resources/theme/style.css
``` 
and others as well:
```
2020-05-25 18:28:30.880  WARN 7552 --- [lication Thread] javafx.css                               : Caught 'java.lang.ClassCastException: class java.lang.String cannot be cast to class javafx.scene.paint.Paint (java.lang.String is in module java.base of loader 'bootstrap'; javafx.scene.paint.Paint is in unnamed module of loader 'app')' while converting value for '-fx-background-color' from rule '*.card' in stylesheet file:/C:/Users/Alexander/IdeaProjects/downlords-faf-client-2/out/production/resources/theme/style.css
```

1-alex98 (changes requested):
see above

BlackYps:
Are you sure that you have looked at an up-to-date log? Seems like the client appends the log file everytime and the first time I could indeed find these errors. I then cleared the log file and now I can't trigger the warnings...

1-alex98:
I am sure I ran an up to date version

BlackYps:
I see lots of these
`2020-05-25 22:06:04.313  WARN 4208 --- [JavaFX Application Thread] javafx                                   : Loading FXML document with JavaFX API of version 11.0.1 by JavaFX runtime of version 10.0.2-internal`
but none of the mentioned css errors. Can you tell me what you do to produce the css warnings?
And please don't say "I click random tabs in the client", because I do that ^^

Katharsas:
I got tons of random CSS warnings some months ago which could only be reproduced sometimes.

1-alex98:
Lemme us check together one day

1-alex98:
open the play tab

PR #1729, opened 2020-05-26
Relocate "more" buttons

### LINKED ISSUE #1728: "More" Buttons in the vault are in strange positions
The "more" buttons in the map, mod, and replay vault are bound to the right edge, which looks a bit odd.
Also they are on the same height as the headlines. So especially in the replay vault you get the problem that after you looked through the list, you have to scroll up again to actually press the right button.

Example:
This button leads you to the highest rated replays:
grafik

BlackYps:
Fixes #1728 

Replay Vault:
grafik

Map/Mod Vault:
grafik
The button is part of the flow pane, but it looks a bit odd. If you have suggestions to improve it visually, let me know.


Katharsas:
Map/Mod vault idea: Make button height as big as a tile (100%), allign it left in the tile and rename it to ">>".

1-alex98:
@Katharsas might look good, not sure but does in my imagination

Katharsas:
Would work better with proper icons i guess that look like actual arrow heads.

BlackYps:
something like this? 
grafik

BlackYps:
This is probably better:
grafik

1-alex98 (dismissed):
I still need to test it before merge tho

BlackYps:
Yeah, don't merge it yet. the new button is just a mockup, I have to implement it properly yet.

Katharsas:
Yeah basically. I would prefer the icon to be a little bit smaller (but keep width of button) xD

BlackYps:
grafik

1-alex98 (changes requested):
Needs more button on owned maps

1-alex98 (approved):

PR #1736, opened 2020-05-29
fix ranked 1v1 layout wrapping

### LINKED ISSUE #1735: Layout problem in ranked 1v1 tab
The rearrangement of the tabs revealed the problem that now the rating chart is switching to the right on FullHD.

grafik

Expected behaviour:
The chart should stay under the other cards.

BlackYps:
Fixes #1735 

grafik


1-alex98 (approved):

1-alex98:
need to test that

PR #1752, opened 2020-06-04
Upgrade to Gradle 6.5

micheljung:


1-alex98:
I have a PR that does that as well, but only used Gradle 6.1

1-alex98:
https://github.com/FAForever/downlords-faf-client/pull/1707

PR #1774, opened 2020-06-14
Map gen spawn number

### LINKED ISSUE #1773: Add ability to select the number of players on a generated map
Visit Issue hunt...

1-alex98:
Requires new map gen
Fixes #1773

### DIFF
--- a/src/main/java/com/faforever/client/game/CreateGameController.java
+++ b/src/main/java/com/faforever/client/game/CreateGameController.java
@@ -17,6 +17,7 @@ import com.faforever.client.mod.ModService;
 import com.faforever.client.mod.ModVersion;
 import com.faforever.client.notification.ImmediateErrorNotification;
 import com.faforever.client.notification.NotificationService;
+import com.faforever.client.preferences.GeneratorPrefs;
 import com.faforever.client.preferences.LastGamePrefs;
 import com.faforever.client.preferences.PreferenceUpdateListener;
 import com.faforever.client.preferences.PreferencesService;
@@ -38,6 +39,8 @@ import javafx.scene.control.ListCell;
 import javafx.scene.control.ListView;
 import javafx.scene.control.MultipleSelectionModel;
 import javafx.scene.control.SelectionMode;
+import javafx.scene.control.Spinner;
+import javafx.scene.control.SpinnerValueFactory.IntegerSpinnerValueFactory;
 import javafx.scene.control.TextField;
 import javafx.scene.image.Image;
 import javafx.scene.input.KeyCode;
@@ -105,6 +108,7 @@ public class CreateGameController implements Controller<Pane> {
   public Label versionLabel;
   public CheckBox onlyForFriendsCheckBox;
   public JFXButton generateMapButton;
+  public Spinner<Integer> spawnCountSpinner;
   @VisibleForTesting
   FilteredList<MapBean> filteredMapBeans;
   private Runnable onCloseButtonClickedListener;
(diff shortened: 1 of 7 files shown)


1-alex98:
https://github.com/FAForever/Neroxis-Map-Generator/pull/30

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -377,7 +393,8 @@ public void onRandomMapButtonClicked() {
 
   public void onGenerateMapButtonClicked() {
     try {
-      mapGeneratorService.generateMap().thenAccept(mapName -> {
+      byte spawnCount = (byte) preferencesService.getPreferences().getGeneratorPrefs().getSpawnCountProperty();
```
Geosearchef, line 396: In theory this always equals the currently selected value, but it feels a bit wrong. Maybe read this value from the spawnCountSpinner instead?

src/main/java/com/faforever/client/map/generator/GenerateMapTask.java
```diff
@@ -64,7 +64,7 @@ protected Void call() throws Exception {
     ProcessBuilder processBuilder = new ProcessBuilder();
     processBuilder.inheritIO();
     processBuilder.directory(workingDirectory.toFile());
-    processBuilder.command("java", "-jar", generatorExecutableFile.toAbsolutePath().toString(), ".", String.valueOf(seed), version, mapFilename);
+    processBuilder.command("java", "-jar", generatorExecutableFile.toAbsolutePath().toString(), ".", String.valueOf(seed), "0.1.6", mapFilename, "8");
```
Geosearchef, line 67: Is this intentional? You are hardcoding and ignoring the version field. What is the 8 supposed to do?

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,9 +108,16 @@ private void deleteGeneratedMaps() {
...
 
-
-  public CompletableFuture<String> generateMap() {
-    return generateMap(queryNewestVersion(), seedGenerator.nextLong());
+  /**
+   * We concatenate the spawnCount on the seed, that's is the only way to keep compatibility. Some day in the future
+   * though I proper way of passing generic options would be appreciated. In the future concatenating(making the seed
+   * number long/bigger) options at the end of the seed is preferred over the bit shift.
+   */
+  public CompletableFuture<String> generateMap(Byte spawnCount) {
```
Geosearchef, line 116: use the primitive type

Geosearchef (changes requested):

src/main/java/com/faforever/client/map/generator/GenerateMapTask.java
1-alex98: left overs that I did not delete

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
1-alex98: yes sir

src/main/java/com/faforever/client/game/CreateGameController.java
1-alex98: Well I could. Makes no difference...

1-alex98:
Concerning the funny way of including it in the seed. Atm this is the only way old clients donot break immediately :D I already spok with @Katharsas @shalkya (not sure if thats his github name) @norraxx  about how we fix these compatibility issues in the long term.

1-alex98:
We want to pin the major version (Semantik Versioning) of the map gen and put warnings if you use a higher major version. Also we want to add a variable options part after the seed so that the client does not really care if new options are in the map name. Atm the client makes expectations about the struture of the map name which are pretty strict.

src/main/java/com/faforever/client/map/generator/GenerateMapTask.java
1-alex98: Used to give the gen the number of players via a new arg

PR #1778, opened 2020-06-17
Feature/#1750 MatchFound and MatchCancelled messages

### LINKED ISSUE #1750: Ladder1v1: Add support for `match_found` and `game_launch_cancelled` commands
When a player queueing for ladder has found a match the server will send a message to both players like:
```
{
    "command": "match_found"
}
```
which signals that the client can stop the searching animation.

(on the TMM branch the message will look like this)

```
{
    "command": "match_found"
    "queue": <queue_name>
}
```

If the host of the ladder game fails to connect after a timeout, the server will send a message to both players like:

```
{
    "command": "game_launch_cancelled"
}
```
which signals that the game was cancelled on the server side, and the client should stop waiting for the game to start. If this happens, the client could automatically kill the player's FA instance if it is running. In order to support future enhancements this should also happen for any joining players who's game instances have been started by the matchmaking code.

This should probably also generate an immediate notification (the ones that pop up on the corner of your screen) and could even enter the player into the queue again.

Visit Issue hunt...

cleborys:
Upon receiving `match_found` will stop the ladder searching animation (but *not* call `stopSearchLadder1v1` on the game service) and spawn a notification that you have been matched.

Upon receiving `match_cancelled` will spawn a notification telling you that match setup failed.

ToDo:
 - [x] Write some tests
 - [x] Decide whether the server should always send a `game_launch` an we stop the game on `match_cancelled`, or whether the server will either send `game_launch` or `match_cancelled`, but not both. Currently assuming the latter.

Closes #1750


src/main/java/com/faforever/client/rankedmatch/MatchFoundMessage.java
```diff
@@ -0,0 +1,19 @@
...
+
+  @SerializedName("queue")
+  private QueueName queueName;
+
+  public MatchFoundMessage() {
+    super(FafServerMessageType.MATCH_FOUND);
+  }
+
+  public QueueName getQueueName(){ return queueName; }
+  public void setQueueName(QueueName queueName){ this.queueName = queueName; }
```
Askaholic: You can use Lombok to generate these for you. I think just annotate the class with `@Data`

src/main/java/com/faforever/client/rankedmatch/Ladder1v1Controller.java
```diff
@@ -49,6 +52,7 @@
 import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Component;
 
+import javax.management.Notification;
```
1-alex98: whats that import. Turn on the checkbox in intelij that deletes unused imports maybe

src/main/java/com/faforever/client/rankedmatch/MatchFoundMessage.java
1-alex98: yes please do @Value will do

src/main/resources/i18n/messages_cs.properties
```diff
@@ -74,8 +74,8 @@ ranked1v1.rankingOutOf = ze {0,number,\#} hráčů
 ranked1v1.searching = Hledám soupeře...
 ranked1v1.winLossRation = Poměr výher/proher
 ranked1v1.rankingFormat = \#{0,number,\#}
-ranked1v1.notification.title = Dostupná 1v1 Ranked hra
-ranked1v1.notification.message = Klikni zde pro otevření 1v1 záložky
+ranked1v1.availableOpponentNotification.title = Dostupná 1v1 Ranked hra
+ranked1v1.availableOpponentNotification.message = Klikni zde pro otevření 1v1 záložky
```
1-alex98: never change those files... u need to chnage it on poeeditor

1-alex98 (changes requested):

src/main/resources/i18n/messages_cs.properties
cleborys: Ah, I was renaming an existing property since `ranked1v1.notification` became a bit ambiguous with the new notifications.
Modifying `src/main/resources/i18n/messages.properties` for the defaults is fine, is that right?

I guess I will create the renamed property in `messages.properties` and see if I can get poeeditor to copy over all translations? From what I've heard one might the right to do that on a per-language basis?

1-alex98 (dismissed):
Will merge after I or anybody else than cleborys can confirm it works properly on production. Rebase merge devlop first ;)

1-alex98:
But I have no time atm

Askaholic:
I can test it with the test server tonight. If I remember...

Askaholic:
A few things I noticed:

1. (as guest) When the game is cancelled, you get a cancelled notification, but ForgedAlliance.exe is still launched. IMO it should not launch since we know that connection will probably not succeed (and even if it does, the game will not be registered).
2. After receiving "match found" you can still press the play button while your match is being set up, however it will not show the searching animation so you can press the button as many times as you like. Each time it sends another "matchmaking start" command to the server. (There is also a bit of a server bug here because the server allows you to start searching again while you are setting up your previous match).
3. (as host) After the game launches, you can press the play button to queue again. This is probably an existing bug unrelated to this PR.
4. If the host launches correctly, but the guest never joins, there is no timeout. This still needs to be implemented on the server.

cleborys:
Re 1. Should that go in this PR? I tend towards yes.

Re 2. I thought about adding a third UI state where both buttons are disabled for in between the MatchFound and MatchCancelled / Launch messages. But then I wasn't sure if that's safe because you don't have a way to manually exit that state if you don't get the appropriate messages for some reason.
Let me know what you think is best.

Askaholic:
I just realized number 1 happens because the server still sends the game launch message for backwards compatibility. At some point I’d like to remove this, but technically that will be a breaking change in the server protocol so IMO it should go in v2. I think it makes sense for the client to ignore the game launch message if it just received a match cancelled (maybe within 2 seconds).

For the match found state, I’d say there could be a timeout on the client. If it doesn’t get a match cancelled or game launch message within 1:30, it can trigger the cancelling logic.

1-alex98:
This needs a reabse

cleborys:
Sorry, I so far haven't figured out how to spawn and manage timers for the timed state changes that Askaholic suggested, I'll un-draft this PR when I have.

Geosearchef:
I'm removing this from the TMM project as it's ladder, not TMM related and TMM handles those messages already.

Sheikah45:
I believe this was integrated with tmm

PR #1786, opened 2020-06-19
Feature/#1777 Starting ladder matchmaking methods refactoring

MaxNF:
Now It is working with 1v1 and 2v2 modes and other modes can be added easily. Changed names of corresponding variables and methods to reflect the changes. Changed some tests to correspond to the changed code.

To add a new ladder mode later (for example 3v3 or 4v4) we just need to modify enums: RatingMode enum and KnownFeatureMod enum. Also, we need to add a switch case to the startGame() method in ForgedAllianceService class.

### DIFF
--- a/src/main/java/com/faforever/client/fa/RatingMode.java
+++ b/src/main/java/com/faforever/client/fa/RatingMode.java
@@ -1,7 +1,20 @@
 package com.faforever.client.fa;
 
+import com.faforever.client.game.KnownFeaturedMod;
+
 public enum RatingMode {
   GLOBAL,
   LADDER_1V1,
-  NONE
+  LADDER_2V2,
+  NONE;
+
+  public KnownFeaturedMod getCorrespondingFeatureMod() {
+    switch (this) {
+      case LADDER_1V1: return KnownFeaturedMod.LADDER_1V1;
+      case LADDER_2V2: return KnownFeaturedMod.LADDER_2V2;
+      default: throw new UnsupportedOperationException("FeatureMod for this rating mod is unsupported or not created yet");
+    }
+  }
 }
+
+
--- a/src/main/java/com/faforever/client/game/GameService.java
+++ b/src/main/java/com/faforever/client/game/GameService.java
@@ -89,7 +89,6 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import static com.faforever.client.fa.RatingMode.NONE;
-import static com.faforever.client.game.KnownFeaturedMod.LADDER_1V1;
 import static com.faforever.client.game.KnownFeaturedMod.TUTORIALS;
 import static com.github.nocatch.NoCatch.noCatch;
 import static java.util.Collections.emptyMap;
@@ -147,7 +146,7 @@ public class GameService implements InitializingBean {
 
   private final ObservableList<Game> games;
   private final String faWindowTitle;
-  private final BooleanProperty searching1v1;
+  private final BooleanProperty searchingLadder;
 
   private Process process;
   private boolean rehostRequested;
@@ -192,7 +191,7 @@ public class GameService implements InitializingBean {
 
     faWindowTitle = clientProperties.getForgedAlliance().getWindowTitle();
     uidToGameInfoBean = FXCollections.observableMap(new ConcurrentHashMap<>());
-    searching1v1 = new SimpleBooleanProperty();
+    searchingLadder = new SimpleBooleanProperty();
     gameRunning = new SimpleBooleanProperty();
     currentGame = new SimpleObjectProperty<>();
     games = FXCollections.observableList(new ArrayList<>(),
@@ -304,7 +303,7 @@ public class GameService implements InitializingBean {
       return gameDirectoryFuture.thenCompose(path -> hostGame(newGameInfo));
     }
 
-    stopSearchLadder1v1();
+    stopSearchLadder();
 
     return updateGameIfNecessary(newGameInfo.getFeaturedMod(), null, emptyMap(), newGameInfo.getSimMods())
         .thenCompose(aVoid -> downloadMapIfNecessary(newGameInfo.getMap()))
@@ -325,7 +324,7 @@ public class GameService implements InitializingBean {
 
     log.info("Joining game: '{}' ({})", game.getTitle(), game.getId());
 
-    stopSearchLadder1v1();
+    stopSearchLadder();
 
     Map<String, Integer> featuredModVersions = game.getFeaturedModVersions();
     Set<String> simModUIds = game.getSimMods().keySet();
@@ -484,22 +483,22 @@ public class GameService implements InitializingBean {
     fafService.addOnMessageListener(MatchmakerInfoMessage.class, listener);
   }
 
-  public CompletableFuture<Void> startSearchLadder1v1(Faction faction) {
+  public CompletableFuture<Void> startSearchLadder(RatingMode ratingMode, Faction faction) {
     if (isRunning()) {
-      log.debug("Game is running, ignoring 1v1 search request");
+      log.debug("Game is running, ignoring search request");
       return completedFuture(null);
     }
 
     if (!preferencesService.isGamePathValid()) {
       CompletableFuture<Path> gameDirectoryFuture = postGameDirectoryChooseEvent();
-      return gameDirectoryFuture.thenCompose(path -> startSearchLadder1v1(faction));
+      return gameDirectoryFuture.thenCompose(path -> startSearchLadder(ratingMode, faction));
     }
 
-    searching1v1.set(true);
+    searchingLadder.set(true);
 
-    return modService.getFeaturedMod(LADDER_1V1.getTechnicalName())
+    return modService.getFeaturedMod(ratingMode.getCorrespondingFeatureMod().getTechnicalName())
         .thenAccept(featuredModBean -> updateGameIfNecessary(featuredModBean, null, emptyMap(), emptySet()))
-        .thenCompose(aVoid -> fafService.startSearchLadder1v1(faction))
+        .thenCompose(aVoid -> fafService.startSearchLadder(ratingMode, faction))
         .thenAccept((gameLaunchMessage) -> downloadMapIfNecessary(gameLaunchMessage.getMapname())
             .thenRun(() -> {
               gameLaunchMessage.setArgs(new ArrayList<>(gameLaunchMessage.getArgs()));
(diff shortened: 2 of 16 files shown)


src/main/java/com/faforever/client/fa/RatingMode.java
```diff
@@ -1,7 +1,20 @@
...
+  LADDER_2V2,
+  NONE;
+
+  public KnownFeaturedMod getCorrespondingFeatureMod() {
+    switch (this) {
+      case LADDER_1V1: return KnownFeaturedMod.LADDER_1V1;
+      case LADDER_2V2: return KnownFeaturedMod.LADDER_2V2;
+      default: throw new UnsupportedOperationException("FeatureMod for this rating mod is unsupported or not created yet");
+    }
+  }
```
Askaholic, line 17: We should try to avoid hardcoding references to anything besides global and ladder1v1. There is also no one-to-one relationship between rating mode and game mode like this. We are trying to move away from having a special `ladder1v1` gamemode and having ladder games just be played on the `faf` gamemode, and have the rating information stored separately.

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -484,22 +483,22 @@ public void addOnRankedMatchNotificationListener(Consumer<MatchmakerInfoMessage>
     fafService.addOnMessageListener(MatchmakerInfoMessage.class, listener);
   }
 
-  public CompletableFuture<Void> startSearchLadder1v1(Faction faction) {
+  public CompletableFuture<Void> startSearchLadder(RatingMode ratingMode, Faction faction) {
```
Askaholic, line 486: Maybe it's better to take both rating mode and game mode as arguments here. On the TMM branch both of these parameters will come from the matchmaker queue object.

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -484,22 +483,22 @@ public void addOnRankedMatchNotificationListener(Consumer<MatchmakerInfoMessage>
...
-      return gameDirectoryFuture.thenCompose(path -> startSearchLadder1v1(faction));
+      return gameDirectoryFuture.thenCompose(path -> startSearchLadder(ratingMode, faction));
     }
 
-    searching1v1.set(true);
+    searchingLadder.set(true);
 
-    return modService.getFeaturedMod(LADDER_1V1.getTechnicalName())
+    return modService.getFeaturedMod(ratingMode.getCorrespondingFeatureMod().getTechnicalName())
         .thenAccept(featuredModBean -> updateGameIfNecessary(featuredModBean, null, emptyMap(), emptySet()))
```
Askaholic, line 500: Side note, wouldn't it be better to make sure the game files actually finished updating before we enter the queue? What if the queue pops and we are still downloading files?

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -508,27 +507,27 @@ public void addOnRankedMatchNotificationListener(Consumer<MatchmakerInfoMessage>
...
           }
           return null;
         });
   }
 
-  public void stopSearchLadder1v1() {
-    if (searching1v1.get()) {
-      fafService.stopSearchingRanked();
-      searching1v1.set(false);
+  public void stopSearchLadder(RatingMode ratingMode) {
```
Askaholic, line 522: I think I'd take a String argument `queueName` here. (I know the message field is called `mod`, it's legacy stuff that we're stuck with :P)

src/main/java/com/faforever/client/game/KnownFeaturedMod.java
```diff
@@ -13,6 +13,7 @@
   FAF_DEVELOP("fafdevelop"),
   BALANCE_TESTING("balancetesting"),
   LADDER_1V1("ladder1v1"),
+  LADDER_2V2("ladder2v2"),
```
Askaholic, line 16: Not a thing :). TMM games are played on the `faf` gamemode.

src/main/java/com/faforever/client/rankedmatch/SearchLadderClientMessage.java
```diff
@@ -1,16 +1,17 @@
...
 import com.faforever.client.game.KnownFeaturedMod;
 import com.faforever.client.remote.domain.ClientMessageType;
 
-public class SearchLadder1v1ClientMessage extends MatchMakerClientMessage {
+public class SearchLadderClientMessage extends MatchMakerClientMessage {
 
   private Faction faction;
 
-  public SearchLadder1v1ClientMessage(Faction faction) {
+  public SearchLadderClientMessage(RatingMode ratingMode, Faction faction) {
```
Askaholic, line 12: It's called `mod` but it's really a matchmaker queue technical name.

Askaholic (changes requested):
It's a good start. For your reference the server messages look like this:

```
{
    "command": "game_matchmaking",
    "state": <"start" | "stop">
    "mod": <queue technical_name>
    "faction": <"uef" | "seraphim" | "cybran" | "aeon" | 1-4>
}
```

The queue technical name is sent in the `matchmaker_info` message which looks like this:
```
{
    "command": "matchmaker_info",
    "queues": [
        {
            "queue_name": <queue technical_name>,
            "queue_pop_time": <datetime in ISO format>,
            "boundary_80s": [[<low>, <high>]],
            "boundary_75s": [[<low>, <high>]],
        }
    ]
}
```

MaxNF:
Well. These thoughts make sense. But it becomes more than just refactoring of some methods connected with starting a queue then :). I've tried to solve the issue without changing the existing code much. It would be great to separate game modes from rating queue types. Also, we actually do not need those static definitions of known feature mods. It is better to receive them from a server and deserialize into domain entities.  But all of these changes will require changing other classes responsible for other (not ladder) game modes. Do we really need all that refactoring? If we do, it is better to do it first before adding any new features. The open-closed principle is not respected much as I see and it is the root of the problem. Firstly we need to refactor the existing app-structure to respect this principle. If we do that we would be able to add any new game mode or rating queue by just creating a new class and extending it from the base class.
It's just my thoughts after working with the code couple of hours.

Askaholic:
I will have to ask @axel1200 and @Brutus5000 what they think.

src/main/java/com/faforever/client/fa/RatingMode.java
1-alex98: Then how do u know that a featured mod is a auto match mod :D

1-alex98:
If we wanna go away from hard coding featured mods the client needs to be told what featured mod is a matchmaker mod. Also important for notifications showing that your friend is in a game.

src/main/java/com/faforever/client/game/GameService.java
MaxNF: Why do we need game mode here? This method is dedicated to launching ladder games only.

src/main/java/com/faforever/client/game/GameService.java
MaxNF: Huh? I have not touched the previous behavior that was done for a 1v1 queue. As I see we update the game first and only then (if an update is successful) we send a message to a server to start a search. The only thing which is done after the search is map downloading.

src/main/java/com/faforever/client/fa/RatingMode.java
MaxNF: If we want to avoid it, we have to receive them from a server dynamically. To do that we have to create a new rest API endpoint which will provide the client with all available ladder queue types. Each queue type will encapsulate some information like name, description, required party size, and (the most important!) technical name for the game mode. This technical name will be used in composing a message which will be sent to a server when a player hits the search (play) button.

src/main/java/com/faforever/client/game/GameService.java
MaxNF: We don't need to pass a queue name here. I passed RatingMode object because it has information about the technical name of the mod (which is required to send a stop-message to a server). If we replace RatingMode with a QueueType object (which should be dynamically received from the server), I will pass QueueType instead of RatingMode. Look at this answer: https://github.com/FAForever/downlords-faf-client/pull/1786#discussion_r443147988

src/main/java/com/faforever/client/game/KnownFeaturedMod.java
MaxNF: If we create a QueueType object as I describe here https://github.com/FAForever/downlords-faf-client/pull/1786#discussion_r443147988
we will be able to remove all Ladder mentions in KnownFeaturedMod.

src/main/java/com/faforever/client/rankedmatch/SearchLadderClientMessage.java
MaxNF: Yes. If you implement QueueType endpoint in rest api I will change it to some QueueType object.
https://github.com/FAForever/downlords-faf-client/pull/1786#discussion_r443147988

src/main/java/com/faforever/client/fa/RatingMode.java
MaxNF: Oh, I'm sorry, maybe I just didn't get it or missed smth? You mentioned "matchmaker_info" message. What is this message? Are queue types already implemented on a server side or rest api side?  I do not fully understand the client-server communication process yet (what these messages are and how I can get a response after writing a message), but I'm working on it.
_____
update: I guess I finally understood how to receive responses from a server. Now I just need to know exactly are queue types already implemented on a server-side or not?

src/main/java/com/faforever/client/game/GameService.java
Askaholic: When I say game mode I mean “featured mod”, and matchmaker games can be played on any featured mod.

src/main/java/com/faforever/client/fa/RatingMode.java
Askaholic: @axel1200 there is mod such thing as an “automatch mod”. Automatch is independent of the featured mod and happens on a per game basis (the server sends the corresponding `InitMode.AUTO_MATCH` in the `game_launch` message. In theory it could even be a coop or nomads game, the client should be able to handle that.

Queues have been implemented but not merged/deployed yet. They are sitting on my team matchmaking branch. I will try to merge it soon.

Also be aware of the distinction between API and server. The server has a connection to the client at all times and can send events at any time. But we are trying to keep more generic data queries which don’t depend on what other players are doing to the API.

src/main/java/com/faforever/client/game/GameService.java
Askaholic: Hmm I need to try it out again, but I thought the last time I did, it would enter me I to the queue before it finished updating.

Askaholic:
As I said above, we are not coupling featured mods to init modes. Matchmaker games should be playable on any featured mod. Brutus is working on some `GameType` overhaul that would save this information in the database on a per game basis, but that isn’t even needed in the client because the server sends an `init_mode` field in the `game_launch` message which will signal either autolobby or normal lobby.

src/main/java/com/faforever/client/fa/RatingMode.java
MaxNF: > Also be aware of the distinction between API and server. The server has a connection to the client at all times and can send events at any time. But we are trying to keep more generic data queries which don’t depend on what other players are doing to the API.

Yea I've already realized that. It was not very clear from the beginning. It would be great to have some sort of basic documentation for newcomers with some real code examples.

MaxNF: Please leave complete information about queue type content here when it's deployed.

MaxNF:
What do you mean by?
>  Matchmaker games should be playable on any featured mod. 

As I understood matchmaker games will only be played on original FaF game mode. But we will set different string values to a "mod" field inside a server message object depending on a Queue type. 
I guess I need to implement refactoring first because there is a misunderstanding in some topics between us. So I'm waiting for a queue type deployment and will complete this issue then.

Brutus5000:
There is at least one matchmaker planned that will require a different featured mod: galactic war

Askaholic:
You don’t need to wait for anything. I already described the queue info you get. It’s sent in the `matchmaker_info` command. Featured mod is sent in the `game_launch` message.

MaxNF:
It's too chaotic and unclear what to do, no sense to work on it for me. Sry.

1-alex98:
Yes this is the case and I have repeatbly said to the promo councilor to please not promted such tasks for ppl that wanna start with FAF XD but nobody listens to me

MaxNF:
The issue itself was quite clear. I've done what was described initially with extending and not changing existing code much. But I'm not a magician, I can't read others' thoughts and ideas. If a previous dev comes afterward and throws a bunch of additional information and says that he has an another idea for implementing this stuff then I let him do it himself. I do not want to waste my time on guessing what some older devs actually want.

Askaholic:
> Yes this is the case and I have repeatbly said to the promo councilor to please not promted such tasks for ppl that wanna start with FAF XD but nobody listens to me

BTW nobody claimed that this was a good first issue. 

Also this PR was actually useful, because in the process of working on this, we have discovered some additional work that will need to be done. I am updating the issue accordingly.

PR #1788, opened 2020-06-20
Fix "None" map name retrieved for generated maps

### LINKED ISSUE #1772: MapGen: can't watch replay with generated map
image
you get this error. From what geo said, this isn't a bug related to map generator, but to the client.

Geosearchef: The map generator is deterministic, given a seed
Geosearchef: It detects and auto downloads the required version based on the map name

Katharsas:
Fixes #1772


src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -231,6 +233,7 @@ static Integer parseSupComVersion(byte[] rawReplayBytes) {
   static String parseMapName(byte[] rawReplayBytes) {
     int mapDelimiterIndex = Bytes.indexOf(rawReplayBytes, new byte[]{0x00, 0x0D, 0x0A, 0x1A});
     String mapPath = new String(rawReplayBytes, MAP_NAME_OFFSET, mapDelimiterIndex - MAP_NAME_OFFSET, US_ASCII);
+    // TODO make sure that these are valid characters for a file name? this is potentially unsafe user input
```
1-alex98: Can u specify what you mean, maybe link a GitHub issue. Donot think that a user could use this string to attack another client

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -503,6 +507,14 @@ private void runFafReplayFile(Path path) throws IOException {
       mapName = parseMapFolderName(rawReplayBytes);
     }
 
+    // For map generator games the map name is "None" because replay server gets map name by from DB based on filename
+    // from replay data, and DB does not contain generated maps.
+    if (StringUtils.equalsIgnoreCase(mapName, "None")) {
```
1-alex98, line 521: This is also the case for other featured mods like coop.  How would that work out there?

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -503,6 +507,14 @@ private void runFafReplayFile(Path path) throws IOException {
...
+      String maybeMapGen = parseMapFolderName(rawReplayBytes);
+      if (mapGeneratorService.isGeneratedMap(maybeMapGen)) {
+        mapName = maybeMapGen;
+      }
+    }
```
1-alex98: Could we write a unit test for that?

1-alex98 (changes requested):

src/main/java/com/faforever/client/replay/ReplayService.java
Katharsas: Coop maps seem to be handled in the lines directly above this code. Maybe it was null or empty in the past and due to a replay server change is "None" now for coop maps as well?

src/main/java/com/faforever/client/replay/ReplayService.java
Katharsas: The usual maps dont require calling this beause mapName is in metadata. That mapName is retrived from some server column by matching via filename. So for normal maps we are guarantueed to get server entries. When directly parsing, we are reading a name that comes directly from one players client.

I guess it doesnt matter because downloading the mapo will simply fail if we have invalid characters right?

src/main/java/com/faforever/client/replay/ReplayService.java
Katharsas: Do we have some actual specification that determines when mapName is supposed to be null, empty or "None"? Otherwise the test will just enforce arbitrary things.

1-alex98:
CI failing

src/main/java/com/faforever/client/replay/ReplayService.java
Sheikah45: I added in a check for invalid filename characters

Katharsas:
@Sheikah45 thanks for improving this, looking good imo.

Sheikah45 (dismissed):

Sheikah45 (dismissed):

Sheikah45 (approved):

src/main/java/com/faforever/client/replay/ReplayService.java
Katharsas: Sheikah added a test

1-alex98 (approved):

PR #1790, opened 2020-06-20
Leaderboard now selects the player when searching by rank

dlin2028:
Searching by name selects the player, but searching by rank doesn't. This makes it so that searching by rank also selects the player.


dlin2028:
Why does this decrease coverage on ReplayService.java?

1-alex98:
Well this might be obsolete soon but ok... Some guys are working on pagination for leaderboard

Sheikah45:
stale

PR #1793, opened 2020-06-20
Feature/#1792 theme

1-alex98:
Also wrote a wiki article

### DIFF
--- a/src/main/java/com/faforever/client/preferences/ui/SettingsController.java
+++ b/src/main/java/com/faforever/client/preferences/ui/SettingsController.java
@@ -239,7 +239,14 @@ public class SettingsController implements Controller<Node> {
     });
 
     currentThemeChangeListener = (observable, oldValue, newValue) -> themeComboBox.getSelectionModel().select(newValue);
-    selectedThemeChangeListener = (observable, oldValue, newValue) -> uiService.setTheme(newValue);
+    selectedThemeChangeListener = (observable, oldValue, newValue) -> {
+      uiService.setTheme(newValue);
+      if (oldValue != null && uiService.doesThemeNeedRestart(newValue)) {
+        notificationService.addNotification(new PersistentNotification(i18n.get("theme.needsRestart.message", newValue.getDisplayName()), Severity.WARN,
+            Collections.singletonList(new Action(i18n.get("theme.needsRestart.quit"), event -> Platform.exit()))));
+        // FIXME reload application (stage & application context) https://github.com/FAForever/downlords-faf-client/issues/1794
+      }
+    };
 
     JavaFxUtil.addListener(preferences.getNotification().toastPositionProperty(), (observable, oldValue, newValue) -> setSelectedToastPosition(newValue));
     setSelectedToastPosition(preferences.getNotification().getToastPosition());
@@ -260,7 +267,7 @@ public class SettingsController implements Controller<Node> {
     configureTimeSetting(preferences);
     configureChatSetting(preferences);
     configureLanguageSelection();
-    configureThemeSelection(preferences);
+    configureThemeSelection();
     configureToastScreen(preferences);
     configureStartTab(preferences);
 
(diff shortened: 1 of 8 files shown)


1-alex98:
https://github.com/FAForever/downlords-faf-client/wiki/Themeing

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -239,7 +239,14 @@ public void initialize() {
     });
 
     currentThemeChangeListener = (observable, oldValue, newValue) -> themeComboBox.getSelectionModel().select(newValue);
-    selectedThemeChangeListener = (observable, oldValue, newValue) -> uiService.setTheme(newValue);
+    selectedThemeChangeListener = (observable, oldValue, newValue) -> {
+      uiService.setTheme(newValue);
+      if (newValue.isNeedsRestart()) {
+        notificationService.addNotification(new PersistentNotification(i18n.get("theme.needsRestart.message", newValue.getDisplayName()), Severity.WARN,
```
BlackYps, line 245: when the theme.properties specifies needsRestart=true I get an exception here. I think these are the relevant ones:
`at org.springframework.context.support.AbstractMessageSource.getMessage(AbstractMessageSource.java:153)
	at com.faforever.client.i18n.I18n.get(I18n.java:94)
	at com.faforever.client.i18n.I18n.get(I18n.java:90)
	at com.faforever.client.preferences.ui.SettingsController.lambda$initialize$8(SettingsController.java:245)
[...]
Caused by: java.lang.NumberFormatException: For input string: ""`

BlackYps (changes requested):

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
micheljung: I propose that we don't leave it up to the theme guy to decide whether or not a restart is neede; he might not even know and it might change in future client releases.

Rather the client should figure out itself whether it's needed or not. In its simplest version: require restart if there's more than just a CSS file.
A future client might be able to reload FXML as well.

1-alex98: If he changes fxml files of pop ups or the gif in the ladder1v1 search it might not be needed even tho it is more than just a css. So I say a maintainer should better judge himself. Also concider context menus, changes there also don't require a restart. You are right that in general none CSS changes more often require a restart but this is not 100%of cases the case.

1-alex98: No answer means u agree with me

1-alex98: Fixed

BlackYps (dismissed):

src/main/java/com/faforever/client/theme/Theme.java
```diff
@@ -13,21 +15,20 @@
...
-    this(null, null, null, null);
-  }
-
-  public Theme(String displayName, String author, Integer compatibilityVersion, String themeVersion) {
+  public Theme(String displayName, String author, Integer compatibilityVersion, String themeVersion, boolean needsRestart) {
     this.displayName = new SimpleStringProperty(displayName);
     this.author = new SimpleStringProperty(author);
     this.compatibilityVersion = new SimpleObjectProperty<>(compatibilityVersion);
     this.themeVersion = new SimpleStringProperty(themeVersion);
+    this.needsRestart = new SimpleBooleanProperty(needsRestart);
```
micheljung: I'm still against having such a property.

micheljung (changes requested):

src/main/java/com/faforever/client/theme/Theme.java
1-alex98: So what are u for? No restart pop up, always a pop up? Looking if it contains fxml files?(that's inaccurate) Looking if it contains none css files?

src/main/java/com/faforever/client/theme/UiService.java
```diff
@@ -472,4 +472,11 @@ public JFXDialog showInDialog(StackPane parent, Node content, String title) {
     dialog.show(parent);
     return dialog;
   }
+
+  @SneakyThrows
+  public boolean doesThemeNeedRestart(Theme theme) {
+    try (Stream<Path> stream = Files.list(getThemeDirectory(theme))) {
+      return stream.anyMatch(path -> Files.isRegularFile(path) && !path.endsWith(".css"));
```
BlackYps: This will always return the theme.properties file

1-alex98: What are u talking about? This is a check on wether a theme conatins a file that is not an css file, because if so we need a restart to fully load the theme.

1-alex98: oh yeah i get u now. Hmm yes so I guess I got to include this there as well

Issue #1612, opened 2020-03-07
Add description for game type


FemtoZetta:
The old client showed a description of the different game types (phantom-x, murder party, etc.) when mousing over them.
Can also add it in the client on mouseover when hosting games or maybe in the details part on the right. Info should be somewhere easily accessible in the client at least, because it's not obvious what most of the game types do.

1-alex98:
The data is sent with the server it is only a matter of displaying it

PR #1795, opened 2020-06-21
Add game type description

QTimort:
Pull request that address https://github.com/FAForever/downlords-faf-client/issues/1612
image

### DIFF
--- a/src/main/java/com/faforever/client/fx/DualStringListCellController.java
+++ b/src/main/java/com/faforever/client/fx/DualStringListCellController.java
@@ -27,7 +27,7 @@ public class DualStringListCellController implements Controller<Node> {
     left.setText(apply);
   }
 
-  public void setLeftTextOver(String apply) {
+  public void setLeftTextTooltip(String apply) {
     final WebView web = new WebView();
     uiService.registerWebView(web);
     web.setPrefSize(web.getPrefWidth() / 2, web.getPrefHeight() / 2);
--- a/src/main/java/com/faforever/client/fx/DualStringListCellController.java
+++ b/src/main/java/com/faforever/client/fx/DualStringListCellController.java
@@ -1,14 +1,12 @@
 package com.faforever.client.fx;
 
 import com.faforever.client.theme.UiService;
+import com.google.common.base.Strings;
 import javafx.scene.Node;
-import javafx.scene.control.ContentDisplay;
 import javafx.scene.control.Label;
-import javafx.scene.control.Tooltip;
 import javafx.scene.layout.HBox;
 import javafx.scene.text.Font;
 import javafx.scene.web.WebView;
-import javafx.util.Duration;
 import lombok.RequiredArgsConstructor;
 import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.context.annotation.Scope;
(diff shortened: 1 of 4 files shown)


src/main/java/com/faforever/client/fx/DualStringListCellController.java
```diff
@@ -21,6 +27,19 @@ public void setLeftText(String apply) {
     left.setText(apply);
   }
 
+  public void setLeftTextTooltip(String apply) {
+    final WebView web = new WebView();
+    uiService.registerWebView(web);
```
1-alex98: u need to do a null check on apply... also u need to remove the tooltip on null because the controller might be reused

src/main/java/com/faforever/client/fx/DualStringListCellController.java
```diff
@@ -21,6 +27,19 @@ public void setLeftText(String apply) {
...
+    web.setPrefSize(web.getPrefWidth() / 2, web.getPrefHeight() / 2);
+    web.getEngine().loadContent(apply);
+    final Tooltip  tip = new Tooltip();
+    tip.setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
+    tip.setGraphic(web);
+    tip.setShowDelay(Duration.ZERO);
+    tip.setHideDelay(Duration.seconds(3));
+    left.setTooltip(tip);
```
1-alex98: we try to stick to mvc . So it should be done in fxml file. Either in this one for a serpate one. And further DualStringListCellController is a Class that is potentially used else where as well adding this pretty sepecific behaviour here seems strange, especially when the method is called setLeftTextTooltip, where it is rather addTooltipWithWebview. So yes u might add this webview capabilities here but make sue your naming reflects that it is not a normal tooltip. Also try define Webview and maybe even tooltip in fxml. On the other hand if it is hard to define in FXML I guess we can leave it tho. Further 3 seconds seem to be short. And dividing the pref size by half is also strange. Why do u do it? Maybe just hard code it :D Could even make use units dpending on screen size.

1-alex98 (changes requested):

1-alex98:
And unit test 🤔

1-alex98:
But nice job. Are u on Zulip? Do you want to be invited?

QTimort:
> But nice job. Are u on Zulip? Do you want to be invited?

Yes I'm on Zulip (Timort)

micheljung:
Thanks for contributing! Please make sure to follow [The seven rules of a great Git commit message](http://chris.beams.io/posts/git-commit/#seven-rules).

Askaholic:
Any progress on this?

QTimort:
Hello, no progress so far, I'm quite busy those days, so if someone wants to take this over he can, otherwise I'll get back to it in like 1 - 2 months max

Sheikah45:
@axel1200 I am going to make the changes for this but wanted to know why the featuredmod list uses the dual string cell list in the first place as it seems it only uses the left

Nevermind I see now it is for displaying the default

Sheikah45:
Refactored to use the fxml files currently uses the dual list cell. However if this is an issue we can change it.

1-alex98:
Also needs to be rebased on new design right?

Sheikah45:
I believe I rebased it when I committed although some css might need to be updated to look nice

1-alex98:
Remember me to rebase and test it :D

Sheikah45:
rebased

1-alex98 (approved):
Nomands text is black, do you know why

src/main/resources/theme/dual_string_list_cell.fxml
```diff
@@ -1,12 +1,22 @@
...
 <?import javafx.scene.layout.HBox?>
+<?import javafx.scene.web.WebView?>
 <HBox xmlns:fx="http://javafx.com/fxml/1" fx:id="root" maxWidth="1.7976931348623157E308" spacing="10.0"
       styleClass="dual-string-list-cell" xmlns="http://javafx.com/javafx/11.0.1"
       fx:controller="com.faforever.client.fx.DualStringListCellController">
    <children>
-      <Label fx:id="left" styleClass="left-label"/>
+      <Label fx:id="left" styleClass="left-label">
+         <tooltip>
+            <Tooltip contentDisplay="GRAPHIC_ONLY" showDelay=".5s" hideDelay="2s" showDuration="30s">
```
1-alex98, line 13: Made it show 30 seconds so u can actually read the text

Sheikah45:
The text color is defined in the html that we receive for the description and for some reason in nomads they set the color to black

1-alex98:
Well then so be it

Sheikah45:
Yeah so wherever the html is retrieved from needs to be changed

1-alex98:
database @Brutus5000 @betrok need to do it

PR #1803, opened 2020-06-26
Feature/#1781 map gen options ui

### LINKED ISSUE #1694: MapGen: Proper "create game" UI
The UI for selecting map should be support 2 use-cases:

- User wants so select or random a handmade map from list
- User wants to generate random map with specific settings

Only in the first case it is necessary to show a map list.
Only in the second case it is necesary to show generator options/settings.

The generator-specific UI elements are / could be:
- Select number of player spawns (only 3v3 currently)
- Button to generate map / regenerate with current settings
- Checkbox and input: Generate from seed / name (input for seed/previous map name), greys out other options.
Optional:
- Show a list of recently played seeds/map names.
- Change biome (optional)

So we need some kind of tabs, kinda like this (but of course withg different look)
image

### LINKED ISSUE #1781: MapGen: Support major version changes
This is the first task that needs to be implemented in the client to complete proper argument handling for the map generator as described here: https://github.com/FAForever/Neroxis-Map-Generator/issues/35

Basically, when the client
- needs to know which map gen version it fully supports (example: client supports mapgen 1.8)
- needs to check the major version number of the map name when doing anything with a give map name (like when joining existing mapgen game or starting replay of a mapgen game), for example lets say user tries to replay a game of mapgen version 2.0
- if the major version differs (which is the case in this example), the client refuses to do the operation and tells the user that the client is not compatible and the user should probably update his client.

When this is in completed, new format for mapgen options can be introduced in mapgen 2.x like described in the mapgen issue.

Edit: Maybe a project should be created to include all related issues.

### LINKED ISSUE #1782: MapGen: Reduce validation strictness of map name
Needed for https://github.com/FAForever/Neroxis-Map-Generator/issues/35

Basically right know when the client reads a name of a mapgen map, it will check to make sure the map name contains the seed with correct format/length, even if it just passes that part of the string to the mapgen. This is bad because we want to replace the seed by a base64 string that contains seed and mapgen options, and pass that to the mapgen instead.

If its possible to just remove this (should be) the client can use the same codepath for reading and passing both 1.x and 2.x maps names to the generator, which would be nice. If not, we need to implement branch for different handling of major versions.

Sheikah45:
Connected to FAForever/Neroxis-Map-Generator#37. Fixes #1781, #1782, #1694. This pull request creates an interface and backend for map generation with options. A new dialog window is created when the generate map button is pressed for inputing either a map name to recreate or options to generate a new map with.

On the back end generated map names are generated as neroxis_map_generator_(version)(seed)(options) and passed to the generator jar file. seed and options are base64 encoded strings and encoding is done in the client prior to sending. It should be noted that these map names will not be parsable by older java clients.

The map generator service map name validation was changed to recognize both the old generated names as well as new and hopefully has enough flexibility to absorb future changes.


src/main/java/com/faforever/client/config/ClientProperties.java
```diff
@@ -150,6 +150,9 @@
     private String downloadUrlFormat;
     private String repoAndOwnerName;
     private String queryLatestVersionUrl;
+    private String queryVersionsUrl;
+    private int maxSupportedMajorVersion;
+    private int minSupportedMajorVersion;
```
Katharsas, line 155: I would remove "minSupportedMajorVersion" (we have to have basic support all old versions forever to keep replays playable and even if we did we could build that in later) and all the corresponding checks.

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -375,20 +379,37 @@ public void onRandomMapButtonClicked() {
...
-          initMapSelection();
-          mapListView.getItems().stream()
-              .filter(mapBean -> mapBean.getFolderName().equalsIgnoreCase(mapName))
-              .findAny().ifPresent(mapBean -> {
-            mapListView.getSelectionModel().select(mapBean);
-            mapListView.scrollTo(mapBean);
-            setSelectedMap(mapBean);
+      String generatorVersion = mapGeneratorService.queryMaxSupportedVersion();
+      String majorVersion = generatorVersion.split("\\.")[0];
+      if (majorVersion.equals("1")) {
```
Katharsas: Doesnt this imply that this code needs to be adjusted whenever the major version changes? Why not check for 0 and then do the old way, so we dont need to adjust this on version 2,3 etc. ?

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,241 @@
...
+
+  /**
+   * Remembers if the controller's init method was called, to avoid memory leaks by adding several listeners
+   */
+  private boolean initialized;
+
+  public void initialize() {
+    if (preferencesService.getPreferences().getForgedAlliance().getInstallationPath() == null) {
+      preferenceUpdateListener = preferences -> {
+        if (!initialized && preferencesService.getPreferences().getForgedAlliance().getInstallationPath() != null) {
```
Katharsas: Are you sure this check is not already made by the game and replay services? It should be imo. Have you tried removing this and removing the installation setting and the trying to play?

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,241 @@
...
+          Platform.runLater(this::init);
+        }
+      };
+      preferencesService.addUpdateListener(new WeakReference<>(preferenceUpdateListener));
+    } else {
+      init();
+    }
+  }
+
+  public void init() {
```
Katharsas: private

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -127,17 +144,53 @@ protected String queryNewestVersion() {
     return mainObject.get("tag_name").getAsString();
   }
 
+  @VisibleForTesting
+  public String queryMaxSupportedVersion() {
+    String version = queryNewestVersion();
+    String majorVersion = version.split("\\.")[0];
+    if (Integer.parseInt(majorVersion)>clientProperties.getMapGenerator().getMaxSupportedMajorVersion() ||
```
Katharsas: space around ">", same below

Katharsas (changes requested):

src/main/java/com/faforever/client/config/ClientProperties.java
1-alex98: Well I think some time in the future old replays will break. I personally thnk that is ok even if it should be avoided

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -375,20 +379,37 @@ public void onRandomMapButtonClicked() {
...
-        Platform.runLater(() -> {
-          initMapSelection();
-          mapListView.getItems().stream()
-              .filter(mapBean -> mapBean.getFolderName().equalsIgnoreCase(mapName))
-              .findAny().ifPresent(mapBean -> {
-            mapListView.getSelectionModel().select(mapBean);
-            mapListView.scrollTo(mapBean);
-            setSelectedMap(mapBean);
+      String generatorVersion = mapGeneratorService.queryMaxSupportedVersion();
+      String majorVersion = generatorVersion.split("\\.")[0];
```
1-alex98: I think we have a class that can parse semantic versions... If we do, which I think use that

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -375,20 +379,37 @@ public void onRandomMapButtonClicked() {
...
+      if (majorVersion.equals("1")) {
+        GenerateMapController generateMapController = uiService.loadFxml("theme/play/generate_map.fxml");
+
+        Pane root = generateMapController.getRoot();
+        generateMapController.setCreateGameController(this);
+        JFXDialog dialog = uiService.showInDialog(gamesRoot, root, i18n.get("game.generate.dialog"));
```
1-alex98, line 418: We are trying to get away from JFX since a few weeks. Can u use the notification and a Immediate notification with custom UI. @micheljung what should he do?

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,241 @@
...
+import java.util.Objects;
+import java.util.Random;
+import java.util.function.Function;
+
+@Component
+@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+@RequiredArgsConstructor
+public class GenerateMapController implements Controller<Pane> {
+
+  private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
1-alex98: Use @Slf4j annotation instead

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,241 @@
...
+  private final PreferencesService preferencesService;
+  private final NotificationService notificationService;
+  private final MapGeneratorService mapGeneratorService;
+  public CreateGameController createGameController;
+  public Pane generateMapRoot;
+  public JFXButton generateMapButton;
+  public JFXTextField previousMapName;
```
1-alex98: Use Button and TextField. At least in the Controller

src/main/resources/i18n/messages_cs.properties
```diff
@@ -632,7 +632,7 @@ view.showSidePane=Více detailů
 view.hideSidePane=Méně detailů
 menu.revealReplayFolder=Otevřít složku se záznamy
 menu.revealGamePrefsFile=Otevřít složku s game.prefs souborem
-game.create.generatedMap=Vytvořit mapu
+game.generate=Vytvořit mapu
```
1-alex98: Never change other language files

1-alex98 (changes requested):

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Will do

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/resources/i18n/messages_cs.properties
Sheikah45: Reverted all other language files

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: I am unclear what using @slf4j should replace if I remove the logger then what should be used on line 227?

src/main/java/com/faforever/client/config/ClientProperties.java
Sheikah45: minSupportedMajorVersion is there as a just in case but never gets triggered since it is set to 0 it is there in case it ever would need to be used..

src/main/java/com/faforever/client/game/CreateGameController.java
Sheikah45: Makes sense done

src/main/java/com/faforever/client/game/GenerateMapController.java
1-alex98: https://projectlombok.org/api/lombok/extern/slf4j/Slf4j.html

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Done

src/main/java/com/faforever/client/game/CreateGameController.java
Sheikah45: Used ComparableVersion

src/main/java/com/faforever/client/game/CreateGameController.java
Sheikah45: Decided to keep JFXDialog for now.

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Removed

1-alex98:
Tests are failing and this dramatically reduces test converge. So this needs unit tests...

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -26,6 +26,7 @@
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.jfoenix.controls.JFXButton;
+import com.jfoenix.controls.JFXDialog;
```
1-alex98, line 29: unused, try turning on the option when commiting that removes unused imports

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,233 @@
...
+import com.faforever.client.map.MapBean;
+import com.faforever.client.map.generator.MapGeneratorService;
+import com.faforever.client.mod.FeaturedMod;
+import com.faforever.client.notification.NotificationService;
+import com.faforever.client.preferences.GeneratorPrefs;
+import com.faforever.client.preferences.PreferenceUpdateListener;
+import com.faforever.client.preferences.PreferencesService;
+import com.google.common.annotations.VisibleForTesting;
+import com.jfoenix.controls.JFXButton;
+import com.jfoenix.controls.JFXTextField;
```
1-alex98: remove

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,233 @@
...
+        mapGeneratorService.generateMap(previousMapName.getText()).thenAccept(mapName -> {
+          Platform.runLater(() -> {
+            createGameController.initMapSelection();
+            createGameController.mapListView.getItems().stream()
+                .filter(mapBean -> mapBean.getFolderName().equalsIgnoreCase(mapName))
+                .findAny().ifPresent(mapBean -> {
+              createGameController.mapListView.getSelectionModel().select(mapBean);
+              createGameController.mapListView.scrollTo(mapBean);
+              createGameController.setSelectedMap(mapBean);
+            });
```
1-alex98: ident

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,233 @@
...
+          });
+        });
+      } else {
+        byte spawnCount = spawnCountSpinner.getValue().byteValue();
+        byte landDensity = (byte) (127 - getSliderValue(waterSlider, waterRandom));
```
1-alex98: Byte.MAX_VALUE that exists right?

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,233 @@
...
+        mapGeneratorService.generateMap(optionArray).thenAccept(mapName -> {
+          Platform.runLater(() -> {
+            createGameController.initMapSelection();
+            createGameController.mapListView.getItems().stream()
+                .filter(mapBean -> mapBean.getFolderName().equalsIgnoreCase(mapName))
+                .findAny().ifPresent(mapBean -> {
+              createGameController.mapListView.getSelectionModel().select(mapBean);
+              createGameController.mapListView.scrollTo(mapBean);
+              createGameController.setSelectedMap(mapBean);
+            });
```
1-alex98: indent

1-alex98: ctrl+alt+L

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,13 +119,40 @@ private void deleteGeneratedMaps() {
...
+    return generateMap(version,seedString+'_'+optionString);
+  }
+
+  public CompletableFuture<String> generateMap(ComparableVersion version, byte[] optionArray) {
+    ByteBuffer seedBuffer = ByteBuffer.allocate(8);
+    seedBuffer.putLong(seedGenerator.nextLong());
+    String seedString = Base64.getEncoder().encodeToString(seedBuffer.array());
+    String optionString = Base64.getEncoder().encodeToString(optionArray);
+    return generateMap(version,seedString+'_'+optionString);
   }
```
1-alex98: Make them call each other and avoid code duplication

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,13 +119,40 @@ private void deleteGeneratedMaps() {
...
 
   @VisibleForTesting
-  protected String queryNewestVersion() {
+  @Cacheable(CacheNames.MAP_GENERATOR)
```
1-alex98, line 143: Make sue this is only ever called via a Spring Proxy

1-alex98 (changes requested):

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Should only be called by the bean

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Changed

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/CreateGameController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -375,23 +376,39 @@ public void onRandomMapButtonClicked() {
...
-        Platform.runLater(() -> {
-          initMapSelection();
-          mapListView.getItems().stream()
-              .filter(mapBean -> mapBean.getFolderName().equalsIgnoreCase(mapName))
-              .findAny().ifPresent(mapBean -> {
-            mapListView.getSelectionModel().select(mapBean);
-            mapListView.scrollTo(mapBean);
-            setSelectedMap(mapBean);
+      mapGeneratorService.setGeneratorVersion(mapGeneratorService.queryMaxSupportedVersion());
+      if (mapGeneratorService.getGeneratorVersion().compareTo(new ComparableVersion("1")) < 0) {
```
Brutus5000, line 399: The hardcoded "1" needs at least a comment what is happening here.

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -375,23 +376,39 @@ public void onRandomMapButtonClicked() {
...
+        mapGeneratorService.generateMap().thenAccept(mapName -> {
+          Platform.runLater(() -> {
+            initMapSelection();
+            mapListView.getItems().stream()
+                .filter(mapBean -> mapBean.getFolderName().equalsIgnoreCase(mapName))
+                .findAny().ifPresent(mapBean -> {
```
Brutus5000: linebreak after .findAny()

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,257 @@
...
+  public CheckBox rampRandom;
+  @VisibleForTesting
+  FilteredList<MapBean> filteredMapBeans;
+  private Runnable onCloseButtonClickedListener;
+  private PreferenceUpdateListener preferenceUpdateListener;
+
+  /**
+   * Remembers if the controller's init method was called, to avoid memory leaks by adding several listeners
+   */
+  private boolean initialized;
```
Brutus5000: This is never used?

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -124,26 +152,62 @@ protected String queryNewestVersion() {
...
+
+    ResponseEntity<String> response = restTemplate.exchange(clientProperties.getMapGenerator().getQueryVersionsUrl(), HttpMethod.GET, entity, String.class);
+    JsonElement jsonElement = new JsonParser().parse(response.getBody());
+    for (JsonElement element : jsonElement.getAsJsonArray()) {
+      JsonObject mainObject = element.getAsJsonObject();
+      version.parseVersion(mainObject.get("tag_name").getAsString());
+      if (version.compareTo(maxVersion) < 0 && minVersion.compareTo(version) < 0) {
+        return version;
+      }
+    }
```
Brutus5000: Is there a reason not to use jackson? You can configure a pojo to ignore unknown fields.

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -124,26 +152,62 @@ protected String queryNewestVersion() {
     JsonElement jsonElement = new JsonParser().parse(response.getBody());
     JsonObject mainObject = jsonElement.getAsJsonObject();
 
-    return mainObject.get("tag_name").getAsString();
+    return new ComparableVersion(mainObject.get("tag_name").getAsString());
+  }
+
+  @VisibleForTesting
+  @Cacheable(CacheNames.MAP_GENERATOR)
```
Brutus5000: I'm not entirely sure what happens if you use the cache twice for the same type as the method has no arguments?!

Brutus5000 (changes requested):

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Yeah I am going to say the reason is that I have no idea what a jackson is.

src/main/java/com/faforever/client/game/CreateGameController.java
Sheikah45: Done

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Alright well I had left in the query newest version in case it could be used elsewhere but it is not used anywhere else so I removed it.

src/main/java/com/faforever/client/game/CreateGameController.java
Sheikah45: Added

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Removed

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Brutus5000: Oh. Jackson is **the** json de/serializer in Spring (https://github.com/FasterXML/jackson). Spring uses it by default e.g. if you call a RestTemplate, you tell Spring it needs to return a type T and Jackson will convert it to type T.
(The client also uses Gson which is another json de/serializer by Google, but in the long run, everything should use Jackson).

You can get it via Dependency Injection. The class you need to ask for is an `ObjectMapper`.

Brutus5000: Alternatively you can ask the RestTemplate directly to return you the pojo you need. If it fails during runtime because it get more values than expected than put this as annotation on the pojo: `@JsonIgnoreProperties(ignoreUnknown = true)`

Sheikah45: Have it return from RestTemplate directly

src/main/java/com/faforever/client/game/CreateGameController.java
1-alex98: Not done

Sheikah45: The JFXDialog is used by uiService.showInDialog on line 406 which I thought we were going to leave in if not I can try to find a way to remove it.

Sheikah45:
Screenshot (28)
Current UI

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,246 @@
...
+  private void initPlateauSlider() {
+    GeneratorPrefs generatorPrefs = preferencesService.getPreferences().getGeneratorPrefs();
+    double plateauDensityProperty = generatorPrefs.getPlateauDensityProperty();
+    boolean plateauRandomProperty = generatorPrefs.getPlateauRandomProperty();
+    plateauSlider.setLabelFormatter(getLabelConverter());
+    plateauRandom.setSelected(plateauRandomProperty);
+    plateauRandom.selectedProperty().addListener(((observable, oldValue, newValue) -> {
+      generatorPrefs.setPlateauRandomProperty(newValue);
+      plateauBox.setVisible(!newValue);
+    }));
```
1-alex98: Could have used Bindings here :D

1-alex98: https://www.youtube.com/watch?v=6Zi2L0kHSx4

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,246 @@
...
+        log.error("Map generation failed", e);
+      }
+      if (msg.startsWith("Old Version")) {
+        notificationService.addImmediateErrorNotification(e, "mapGenerator.tooOldVersion");
+        log.error("Map generation failed", e);
+      }
+    } catch (Exception e) {
+      notificationService.addImmediateErrorNotification(e, "mapGenerator.generationFailed");
+      log.error("Map generation failed", e);
+    }
```
1-alex98: I doubt that is is ever called. You probably need to do this in .excpetionally() on the Completable Future. https://www.logicbig.com/tutorials/core-java-tutorial/java-multi-threading/completion-stages-exception-handling.html

1-alex98 (changes requested):

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: I tested it and .exceptionally doesnt work because it is not a completable future that throws the exception rather it is the generatemap function which when it throws the exception never hands back a completable future which is the reason for the try catch but if there is another way it could be structured to use the .exceptionally let me know

Sheikah45: Nevermind I got it and changed the map generator service to return a completable future that throws an exception

Sheikah45: Done and let me know if you meant something different

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
1-alex98: Well ok maybe I was wrong then. I just expected it to be thrown inside the CompletableFuture

1-alex98: It also should be because otherwise u are doing stuff on the ui thread that u should not do. Lemme see what u did...

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,237 @@
...
+  public void onGenerateMapButtonClicked() {
+    onGenerateMap();
+  }
+
+  public void onGenerateMap() {
+      if (!previousMapName.getText().isEmpty()) {
+        mapGeneratorService.generateMap(previousMapName.getText())
+            .exceptionally(throwable -> {
+              handleGenerationException(throwable);
+              return null;
```
1-alex98: If this return null then there is gonna be a nullpointer Exception following in the thenAccept part right XD Also createGameController.initMapSelection(); is called

1-alex98: Maybe not just maybe see how it behaves.

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,237 @@
...
+      } else if (msg.startsWith("Old Version")) {
+        notificationService.addImmediateErrorNotification(cause, "mapGenerator.tooOldVersion");
+        log.error("Map generation failed", cause);
+      } else {
+        notificationService.addImmediateErrorNotification(e, "mapGenerator.generationFailed");
+        log.error("Map generation failed", e);
+      }
+    } else {
+      notificationService.addImmediateErrorNotification(e, "mapGenerator.generationFailed");
+      log.error("Map generation failed", e);
```
1-alex98: Should they not all be warn

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,43 +115,87 @@ private void deleteGeneratedMaps() {
...
+      }
+    }
+    throw new RuntimeException("No Valid Generator Version Found");
   }
 
   public CompletableFuture<String> generateMap(String mapName) {
     Matcher matcher = GENERATED_MAP_PATTERN.matcher(mapName);
     if (!matcher.find()) {
-      throw new IllegalArgumentException("Map name is not a generated map");
+      return CompletableFuture.supplyAsync(() -> {
```
1-alex98: I think there is a static Method on the CompletableFuture class that does exactly that

1-alex98 (changes requested):

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: I tested this and it just errored with the illegalArgument Exception but could also just move the .exceptionally to the end.

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: I am not sure what you me by all of them being warn? is there another function to the notification service?

1-alex98: log.warn(....)

Sheikah45: Ah got it

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Used

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Moved it to the end of the futures

1-alex98 (dismissed):
Still wanna test it for my self

1-alex98:
Remember me if I forget

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -375,23 +376,41 @@ public void onRandomMapButtonClicked() {
...
+        mapGeneratorService.generateMap().thenAccept(mapName -> {
+          Platform.runLater(() -> {
+            initMapSelection();
+            mapListView.getItems().stream()
+                .filter(mapBean -> mapBean.getFolderName().equalsIgnoreCase(mapName))
+                .findAny()
+                .ifPresent(mapBean -> {
+              mapListView.getSelectionModel().select(mapBean);
+              mapListView.scrollTo(mapBean);
+              setSelectedMap(mapBean);
```
micheljung: This code doesn't seem properly formatted; please use/enable IntelliJ's auto formatting

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,237 @@
...
+  public Slider plateauSlider;
+  public CheckBox plateauRandom;
+  public HBox plateauBox;
+  public Slider rampSlider;
+  public CheckBox rampRandom;
+  public HBox rampBox;
+  private Runnable onCloseButtonClickedListener;
+
+  public void initialize() {
+    init();
```
micheljung: This is redundant; inline `init()`

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  private StringConverter<Double> getLabelConverter() {
+    return new StringConverter<>() {
+      @Override
+      public String toString(Double n) {
+        if (n < 127) {
+          return "None";
+        }
+        return "Lots";
```
micheljung: Use `I18n` and message keys.

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,237 @@
...
+      }
+
+      @Override
+      public Double fromString(String s) {
+        if (s.equals("None")) {
+          return 0d;
+        }
+        return 127d;
+      }
```
micheljung, line 108: Conversion from String to a number is most likely never used; just throw an UnsupportedOperationException

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,237 @@
...
+            .thenAccept(mapName -> Platform.runLater(() -> {
+              createGameController.initMapSelection();
+              createGameController.mapListView.getItems().stream()
+                  .filter(mapBean -> mapBean.getFolderName().equalsIgnoreCase(mapName))
+                  .findAny().ifPresent(mapBean -> {
+                createGameController.mapListView.getSelectionModel().select(mapBean);
+                createGameController.mapListView.scrollTo(mapBean);
+                createGameController.setSelectedMap(mapBean);
+              });
+            }))
```
micheljung: Reduce code duplication

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -0,0 +1,237 @@
...
+      if (msg.startsWith("New Version")) {
+        notificationService.addImmediateErrorNotification(cause, "mapGenerator.tooNewVersion");
+        log.warn("Map generation failed", cause);
+      } else if (msg.startsWith("Old Version")) {
+        notificationService.addImmediateErrorNotification(cause, "mapGenerator.tooOldVersion");
+        log.warn("Map generation failed", cause);
+      } else {
+        notificationService.addImmediateErrorNotification(e, "mapGenerator.generationFailed");
+        log.warn("Map generation failed", e);
+      }
```
micheljung: Relying on the exception message is quite fragile; it's not part of an API and may change at any time, staying unnoticed. Instead, the map generator should either throw concrete exceptions or its exception should contain some sort of reason code.

src/main/java/com/faforever/client/map/generator/GithubGeneratorRelease.java
```diff
@@ -0,0 +1,13 @@
+package com.faforever.client.map.generator;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Data;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@Data
+public class GithubGeneratorRelease {
+  private String tag_name;
```
micheljung: Don't break the Java naming convention

```suggestion
  @JsonProperty("tag_name")
  private String tagName;
```

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,62 +115,128 @@ private void deleteGeneratedMaps() {
...
+    ResponseEntity<List<GithubGeneratorRelease>> response = restTemplate.exchange(clientProperties.getMapGenerator().getQueryVersionsUrl(), HttpMethod.GET, entity, new ParameterizedTypeReference<>() {
+    });
+    List<GithubGeneratorRelease> releases = response.getBody();
+    for (GithubGeneratorRelease release : releases) {
+      version.parseVersion(release.getTag_name());
+      if (version.compareTo(maxVersion) < 0 && minVersion.compareTo(version) < 0) {
+        return version;
+      }
+    }
+    throw new RuntimeException("No Valid Generator Version Found");
```
micheljung:
```suggestion
    throw new RuntimeException("No valid generator version found");
```

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,62 +115,128 @@ private void deleteGeneratedMaps() {
...
   }
 
   public CompletableFuture<String> generateMap(String mapName) {
     Matcher matcher = GENERATED_MAP_PATTERN.matcher(mapName);
     if (!matcher.find()) {
-      throw new IllegalArgumentException("Map name is not a generated map");
+      CompletableFuture<String> errorFuture = new CompletableFuture<>();
+      errorFuture.completeExceptionally(new IllegalArgumentException("Map name is not a generated map"));
+      return errorFuture;
```
micheljung:
```suggestion
      return CompletableFuture.failedFuture(new IllegalArgumentException("Map name is not a generated map"));
```

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,62 +115,128 @@ private void deleteGeneratedMaps() {
...
 
-  public CompletableFuture<String> generateMap(String version, long seed) {
+  public CompletableFuture<String> generateMap(ComparableVersion version, String seedAndOptions) {
 
+    ComparableVersion minVersion = new ComparableVersion(String.valueOf(clientProperties.getMapGenerator().getMinSupportedMajorVersion()));
+    ComparableVersion maxVersion = new ComparableVersion(String.valueOf(clientProperties.getMapGenerator().getMaxSupportedMajorVersion() + 1));
+    if (version.compareTo(maxVersion) >= 0) {
+      CompletableFuture<String> errorFuture = new CompletableFuture<>();
+      errorFuture.completeExceptionally(new IllegalArgumentException("New Version not supported"));
+      return errorFuture;
```
micheljung:
```suggestion
      return CompletableFuture.failedFuture(new IllegalArgumentException("New version not supported"));
```

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,62 +115,128 @@ private void deleteGeneratedMaps() {
...
+    }
+    if (version.compareTo(minVersion) < 0) {
+       CompletableFuture<String> errorFuture = new CompletableFuture<>();
+       errorFuture.completeExceptionally(new IllegalArgumentException("Old Version not supported"));
+       return errorFuture;
```
micheljung:
```suggestion
      return CompletableFuture.failedFuture(new IllegalArgumentException("Old Version not supported"));
```

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,62 +115,128 @@ private void deleteGeneratedMaps() {
...
-      if (!VERSION_PATTERN.matcher(version).matches()) {
-        log.error("Unsupported generator version: {}", version);
-        return CompletableFuture.supplyAsync(() -> {
-          throw new RuntimeException("Unsupported generator version: " + version);
-        });
+      if (!VERSION_PATTERN.matcher(version.toString()).matches()) {
+        log.warn("Unsupported generator version: {}", version);
+        CompletableFuture<String> errorFuture = new CompletableFuture<>();
+        errorFuture.completeExceptionally(new RuntimeException("Unsupported generator version: " + version));
+        return errorFuture;
```
micheljung:
```suggestion
        return CompletableFuture.failedFuture(new RuntimeException("Unsupported generator version: " + version));
```

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,62 +115,128 @@ private void deleteGeneratedMaps() {
...
+    String[] seedParts = seedAndOptions.split("_");
+    String seedString = seedParts[0];
+
+    String mapFilename;
+    String seed;
+
+    try {
+      seed = Long.toString(Long.parseLong(seedString));
+    }
+    catch (NumberFormatException nfe) {
```
micheljung:
```suggestion
    } catch (NumberFormatException nfe) {
```

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -107,62 +115,128 @@ private void deleteGeneratedMaps() {
...
+      byte[] seedBytes = Base64.getDecoder().decode(seedString);
+      ByteBuffer seedWrapper = ByteBuffer.wrap(seedBytes);
+      seed = Long.toString(seedWrapper.getLong());
+    }
+
+    // Check if major version 0 which requires numeric seed
+    if (version.compareTo(new ComparableVersion("1")) < 0) {
+      mapFilename = String.format(GENERATED_MAP_NAME, version, seed).replace('/','^');
+    }
+    else {
```
micheljung:
```suggestion
    } else {
```

Please make sure that auto-formatting is enabled

src/main/resources/i18n/messages.properties
```diff
@@ -733,6 +744,8 @@ replay.couldNotOpen=Replay with id ''{0}'' , could not be opened.
 error.game.notTerminatedCorrectly=The game did not terminate correctly and has still certain files open. Some files from the installed game could not be copied over to FAF but unless you recently modified game files in the installed version you probably do not care. We are still starting the game for you.
 map.id=#{0}
 mapGenerator.generationFailed=There was an error during map generation
+mapGenerator.tooNewVersion=This client does not support this version of map generator. Please update client to use this version.
```
micheljung:
```suggestion
mapGenerator.tooNewVersion=This client does not support this version of the map generator. Please update the client to use this version.
```

micheljung (changes requested):
Few minor adjustments but overall, it looks pretty good!
Before we merge this, please squash the commits :-)

src/main/resources/i18n/messages.properties
Sheikah45: Done

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Done

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Done

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Done

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done. Used concrete exceptions.

src/main/java/com/faforever/client/map/generator/GithubGeneratorRelease.java
Sheikah45: Done

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Done

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Done

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/CreateGameController.java
Sheikah45: Done

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -259,7 +260,7 @@ private void initMapSelection() {
     mapListView.getSelectionModel().selectedItemProperty().addListener((observable, oldValue, newValue) -> Platform.runLater(() -> setSelectedMap(newValue)));
   }
 
-  private void setSelectedMap(MapBean newValue) {
+  protected void setSelectedMap(MapBean newValue) {
```
1-alex98, line 275: If you do this to be able to test this. Make it package private instead and annotate with @VisibleForTesting

1-alex98 (dismissed):
Good for me but let's have @micheljung a second go

src/main/java/com/faforever/client/game/CreateGameController.java
Sheikah45: It was changed to protected so that GenerateMapController could set the map selection after generating the map in its futures.

Sheikah45:
UI with added map size
Screenshot (31)

1-alex98:
@Sheikah45 only thing that I dont like is "Map name". Users are gonna think they can name their map. I think it should be made clear that is means "Recreate map using its name" otherwise users will think they are supposed to give the map a name. Will then wonder about the strange error thrown afterwards. So the other Controls should be disabled when the user enters a name I guess, cause they are then use less.

1-alex98:
grafik

1-alex98:
Pushed on your branch tell me if u like it....

1-alex98:
grafik

1-alex98 (dismissed):

Sheikah45:
The more intuitive design is good. I also disabled the options when a map name is given to make it more obvious as well.
Screenshot (32)
Disabled options when map name given

PR #1805, opened 2020-06-29
Feature/#1231 basic search fields for the vault

### LINKED ISSUE #1231: Rework vault, user friendlyness

BlackYps:
I created UI elements to provide basic search fields for the vault. The only things left to do are:
- Make the text fields construct the appropriate api query string.
- Give the user the ability to leave the fields blank while triggering the sort order with the search button.

However, I do not know how to do that properly. The code for the vault search is incredibly abstract to accomodate for changes in the api, which is good, but I have no idea where I would inject the more or less hardcoded string for the basic search fields.

This will eventually fix #1231 

grafik
grafik

### DIFF
--- a/src/main/java/com/faforever/client/map/MapVaultController.java
+++ b/src/main/java/com/faforever/client/map/MapVaultController.java
@@ -102,6 +102,7 @@ public class MapVaultController extends AbstractViewController<Node> {
   public JFXButton nextButton;
   public Label currentPageLabel;
   public HBox paginationHBox;
+  public ScrollPane searchScrollPane;
   private MapDetailController mapDetailController;
   private int currentPage;
   private Supplier<CompletableFuture<List<MapBean>>> currentSupplier;
@@ -125,6 +126,7 @@ public class MapVaultController extends AbstractViewController<Node> {
   public void initialize() {
     super.initialize();
     JavaFxUtil.fixScrollSpeed(scrollPane);
+    JavaFxUtil.fixScrollSpeed(searchScrollPane);
 
     loadingLabel.managedProperty().bind(loadingLabel.visibleProperty());
     showroomGroup.managedProperty().bind(showroomGroup.visibleProperty());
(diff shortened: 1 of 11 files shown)


Brutus5000:
I know people asked for it because the Python client used to do this. For me it's a bad user experience for new people.
What happens if I enter one value? Is the second one ignored?
Can I delete one value afterwards? Is it ignored again?

The selection box solution solved all of these problems...

If you want to keep it like this: Name and Author need at least a checkbox to show if it is active or not. And it should get active automatically if you enter something in the textfield.

BlackYps:
I don't get how this is bad user experience?
All fields should be connected by AND. If it is empty it contributes nothing to the search. If you delete your value again, that field doesn't contribute to the search obviously. Correct me if I'm wrong, but I think that is how most search interfaces work, so people should be familiar with that.

With the current implementation you need to clicks to select the property you want and another two for the operation if you are unlucky. For the basic searches that are probably the vast majority it is way faster if you can just start typing in one of the fields.
I think that is the main reason people want that feature. (Not to mention you need to figure out how some of these properties should be used. Looking at you "map width in px"...)

Brutus5000:
The input box doesn't tell you anything about equals or contains. it also doesn't tell if its mandatory or not.

Following that logic you could also just pre-populate the "advanced" search with Name and Author and mod type (and skip empty values). Then you would know that it's contains and don't need to break the user workflow with switching to advanced (and loosing your search parameters back and forth).

Brutus5000:
For multi-select types like mod-type I'd suggest a multi input select like this:
image

I don't know if there is a similar control in JavaFX but that is what I'll be using in the web vault.

BlackYps:
We could add a small (i) or (?) icon that opens a tooltip on mouseover that explains the functionality.
At the moment the client remembers your input in both panels, so you don't lose anything if you switch between basic and advanced search.
There seems to be no standard element to get this multiselect behavior in JavaFX. I found some examples on how to code that online, but I don't know yet if it is worth the extra effort. I will have to find out how much effort it really is and how bad we need a multiselect list.

BlackYps:
I should probably also include screenshots of the other two tabs...

grafik
grafik

(game type dropdown menu is empty because no controller logic to fill it yet)

Sheikah45:
Covered by Filter search

PR #1820, opened 2020-07-10
Feature/#1763 local replays pagination

### LINKED ISSUE #1078: Don't cap local replay list after 300 files

### LINKED ISSUE #1763: Pagination of local replays
Some users have more than 10.000 local replays, that will be loaded on invocation 
This can cause an error.
Load only the 100 newest and add a pagination with 100 items per page.

jofaul:
Done after unittesting


micheljung:
It looks like there are many commits in this PR that shouldn't be. Please sort this out first.

1-alex98:
@micheljung on the other hand I can also squash this on merge

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -139,19 +142,27 @@
   private final ExecutorService executorService;
   private Thread directoryWatcherThread;
   private WatchService watchService;
-  protected List<Replay> localReplays = new ArrayList<Replay>();
+  protected List<Replay> localReplays = new ArrayList<>();
 
-  public void startLoadingAndWatchingLocalReplays() {
+  public void startLoadingAndWatchingLocalReplays(Pagination pagination) {
```
1-alex98: Passing an UI Elements to a service should never happen!

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -139,19 +142,27 @@
...
+      logger.error("Failed loading total amount of replays", e);
+    }
+
     LoadLocalReplaysTask loadLocalReplaysTask = applicationContext.getBean(LoadLocalReplaysTask.class);
-    taskService.submitTask(loadLocalReplaysTask).getFuture().thenAccept( replays -> {
+    taskService.submitTask(loadLocalReplaysTask).getFuture().thenAccept(replays -> {
       localReplays.clear();
       localReplays.addAll(replays);
-      publisher.publishEvent(new LocalReplaysChangedEvent(this, replays, new ArrayList<Replay>()));
+      publisher.publishEvent(new LocalReplaysChangedEvent(this, replays, new ArrayList<>()));
```
1-alex98: Publish page count here maybe...

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -139,19 +142,27 @@
...
     if (Files.notExists(replaysDirectory)) {
       noCatch(() -> createDirectories(replaysDirectory));
     }
 
+    String replayFileGlob = clientProperties.getReplay().getReplayFileGlob();
+    try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(replaysDirectory, replayFileGlob)) {
+      final int pageCount = (int) Math.ceil((StreamSupport.stream(directoryStream.spliterator(), false).count() / (double) REPLAYS_PER_PAGE));
+      Platform.runLater(() -> pagination.setPageCount(pageCount));
+    } catch (IOException e) {
+      logger.error("Failed loading total amount of replays", e);
```
1-alex98: This is not an error, rather warning I say

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -186,8 +207,8 @@ protected Thread startDirectoryWatcher(Path replaysDirectory) throws IOException
 
   @VisibleForTesting
   protected void onLocalReplaysWatchEvent(WatchKey key) {
```
1-alex98: Have u thought about what happens if a user drops a new replay into his replay folder... that is what happens here.

1-alex98: Maybe u can simplify this method by just going back to the first page :thinking:

1-alex98 (changes requested):
Also misses unit tests

src/main/java/com/faforever/client/replay/ReplayService.java
1-alex98: I am surprised this works with an event publisher. Probably the Controller a singleton and that's why it works. Whatever let's leave it like this.

micheljung:
> @micheljung on the other hand I can also squash this on merge

If you write a proper commit message, sure. But I won't accept the default generated commit message ;-)

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -139,19 +140,29 @@
   private final ExecutorService executorService;
   private Thread directoryWatcherThread;
   private WatchService watchService;
-  protected List<Replay> localReplays = new ArrayList<Replay>();
+  protected List<Replay> localReplays = new ArrayList<>();
+  private double localReplaysItemsSize;
```
1-alex98: whys that a double

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -134,36 +134,47 @@
...
   private final ModService modService;
   private final MapService mapService;
-  private final ApplicationEventPublisher publisher;
+  private final EventBus eventBus;
   private final MapGeneratorService mapGeneratorService;
-  private final ExecutorService executorService;
   private Thread directoryWatcherThread;
-  private WatchService watchService;
-  protected List<Replay> localReplays = new ArrayList<Replay>();
+  protected List<Replay> localReplays = new ArrayList<>();
```
1-alex98, line 129: Written to but never read, right? If so remove

1-alex98 (changes requested):

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -134,36 +134,47 @@
...
+    reloadLocalReplays();
   }
 
-  public Collection<Replay> getLocalReplays() {
-    return localReplays;
+  private void reloadLocalReplays() {
+    Path replaysDirectory = preferencesService.getReplaysDirectory();
+    pageCountLocalReplays = 1;
+    String replayFileGlob = clientProperties.getReplay().getReplayFileGlob();
+    try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(replaysDirectory, replayFileGlob)) {
```
1-alex98: Files.list

1-alex98 (changes requested):

Sheikah45:
Did we want to do the local replay pagination this way? I can rebase and bring it up to current.

1-alex98:
We need to definetely rebase it. How would u do it otherwise?

Sheikah45:
I would probably want to refactor it a bit to make it more similar to the other paginations as much as possible

1-alex98:
ok but we can definitely not use the same code base

Sheikah45:
Maybe not. Also do you think that we can't just make a card for local replays?

Sheikah45:
I also meant more so in the style and naming conventions

1-alex98:
Have fun do what u like :D

Sheikah45:
Fixes #1763

1-alex98 (commented):
Looks neat

PR #1825, opened 2020-07-11
Fix bug in handling of navigation events

### LINKED ISSUE #1807: Clicking view replays doesn't show the replays of a player without having the maps loaded
**Describe the bug**
Clicking view replays doesn't show the replays of a player without having the maps loaded

**To Reproduce**
once in chat, upon clicking show replays on a random player for the first time, the program switches to vault/maps and stays there. Once this action is executed again, the program behaves properly 

**Expected behavior**
Once on view replays is clicked, the program should jump to vault/replays, without having to load maps first

BlackYps:
Fixes #1807 
Posting a navigation event when initializing a tab had side effects when a navigate event (e.g. clicking on the show ladder maps button) took you to a previously uninitialized tab, because that would also trigger the event in initialize().
If we want that, the new behaviour could be improved even further by writing the lastTab information to the preferences, so the client can remember which tabs you had open even over a restart.

### DIFF
--- a/src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
+++ b/src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
@@ -68,8 +68,7 @@ public class LeaderboardsController extends AbstractViewController<Node> {
       if (navigateEvent instanceof OpenLadder1v1LeaderboardEvent) {
         lastTab = ladder1v1LeaderboardTab;
         lastTabController = ladder1v1LeaderboardController;
-      }
-      else if (navigateEvent instanceof OpenGlobalLeaderboardEvent) {
+      } else if (navigateEvent instanceof OpenGlobalLeaderboardEvent) {
         lastTab = globalLeaderboardTab;
         lastTabController = globalLeaderboardController;
       }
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
```diff
@@ -63,13 +66,15 @@ protected void onDisplay(NavigateEvent navigateEvent) {
...
       if (navigateEvent instanceof OpenLadder1v1LeaderboardEvent) {
-        leaderboardRoot.getSelectionModel().select(ladder1v1LeaderboardTab);
-        ladder1v1LeaderboardController.display(navigateEvent);
+        lastTab = ladder1v1LeaderboardTab;
+        lastTabController = ladder1v1LeaderboardController;
       }
-      if (navigateEvent instanceof OpenGlobalLeaderboardEvent) {
-        leaderboardRoot.getSelectionModel().select(globalLeaderboardTab);
-        globalLeaderboardController.display(navigateEvent);
+      else if (navigateEvent instanceof OpenGlobalLeaderboardEvent) {
```
1-alex98: same line as }

1-alex98 (dismissed):

1-alex98 (commented):
Since we had a bug there. How about u try to write a unit test that makes sure this never happens again.
Test if after a X Event the x tab is opened.

BlackYps:
Testing if the correct tab is opened is not enough. We would need to test that it actually shows the ladder maps and not the front page of the map vault. you can actually see that it did that for a brief moment and then the second navigate event resetted the view to the start view. I don't know how to test for this properly. We could try to test if mapService.getLadderMaps() got called, but that would have passed even with this bug...

PR #1826, opened 2020-07-12
Change default database from rackover to spooky

Exotic-Retard:
First off, i have no idea what im doing, and so if this doesnt work i wouldnt be surprised in the slightest. (forgive me)

Also, this is kinda an issue / feature request but i thought it would go along easier if i tried to do it myself first

Anyway, the brick i want to sell is this: the spooky db is good and beautiful while the rackover db is bad and ugly.
the rack one is also without a maintainer while the spooky one does have a maintainer.

I also think that the spooky one should be integrated into faf completely while the rackover one retired, but thats a separate thing. for now i would like for the spooky one to be made default, Check out this comparison:

also, later there will be a version 2 of the database which is even better:

This one however isnt done yet so half its features are missing at the moment.

### DIFF
--- a/src/main/java/com/faforever/client/preferences/Preferences.java
+++ b/src/main/java/com/faforever/client/preferences/Preferences.java
@@ -71,7 +71,7 @@ public class Preferences {
     developer = new DeveloperPrefs();
     gameListSorting = new SimpleListProperty<>(observableArrayList());
     vaultPrefs = new VaultPrefs();
-    unitDataBaseType = new SimpleObjectProperty<>(UnitDataBaseType.RACKOVER);
+    unitDataBaseType = new SimpleObjectProperty<>(UnitDataBaseType.SPOOKY);
     storedCookies = new SimpleMapProperty<>(FXCollections.observableHashMap());
     showPasswordProtectedGames = new SimpleBooleanProperty(true);
     showModdedGames = new SimpleBooleanProperty(true);


1-alex98:
In the past the argument always was:
1. We don't really know spooky. But we did know Rackover.
2. His site seems to not autoupdate and I was told it was outdated quite often.
3. Rackover's repo is in the faf organization and we host it. So we have control over it.

Would be cool if spooky moved the db to FAF organization and we make him admin, if we wanna make it offical unit db. 

But since Rackover left and nobody is working on her stuff I think we can make the change.

1-alex98:
And your change should be ok. I am not sure tho if it changes the default for those that have already installed the client.

Exotic-Retard:
spooky responds fairly quickly, has been keeping his db up to date for the past two years or so and is also working on a newer-er version of the databse. Im sure we can integrate his db into faf and make his update script run automagically on the server, which is currently a powershell script.

And yeah i dont think this will change things for existing users. Some later pr could put in a deeper integration once we decide more things but for now this is the first step.  i can make a bigger pr that removes the rackover DB from the client entirely later :>

The scope of this PR though is just to change the default value for new users, so if that works could you please merge it? thanks <3

1-alex98:
Rackover db also have some hook that makes it auto update I think.

1-alex98:
I mean the faf files and versions are available via api.faforever.com so in theory one could also pool that every day and see there is a new version.

1-alex98 (approved):

spooky:
I'd be more than happy to hand the current db over. I spoke to sheeo at some point to move it to FAF, but if I recall, I need to be an admin of the organization to be able to move the repo there.

As for outdated - the current version does indeed use a lot of lot stuff (hence the work on v2) but it's still usable if the required dependencies are there. There's a dockerfile in the repo that should aid in that process, although, I'm not the author of that, and I haven't really used it myself.

Anyway, if anyone is still interested in taking over the repo, I'd be more than happy to get it off my hands :P (just leave me write rights on in it just in case)

1-alex98:
You can probably talk to @Brutus5000 if you would like to have the repo moved I think he has sufficent rights.

PR #1828, opened 2020-07-12
Feature/#1753 auto update

micheljung:


Sheikah45:
Where did this leave off @micheljung?

micheljung:
@Sheikah45 IIRC, I got it almost finished; it checked for an update, downloaded the update while the client was running and maybe added "Download complete, restart?" notification. When closing the client, the update was finalized and when it started next time, the new version was available.

Another problem was providing the "installed" files on the server; this was also solved by building a ZIP file which was then unpacked and uploaded to a server via SCP; including a de-duplicating script so that we don't store files that never change 100x on the server. This worked perfectly except for java's "module" file which somehow changes every time even though it wasn't changed (maybe there is some timestamp involved?).

IIRC, the remaining issue was that when you now installed the client via MSI, it got e.g. version 1.0.0 but when it auto-updated itself, this version never changed (because for the auto-update, Install4j wasn't involved).

Speaking of Install4J: its auto-update feature doesn't fit well because all it does is updating and executing the newest MSI but the goal of the auto-update was also to reduce the bandwidth and time required to update clients.

I would love to see this finished. I hope the code is still mergeable

Sheikah45:
I have been working on rebasing it some and also trying to find better ways to fix some of the library issues that you had run in to. My main concern is just that the fix for the moment seemed to be that you applied fixes to repos that you own which I am not sure is sustainable but thanks for the info as well I will hoepfully be able to get back to this and start figuring out the issues you mentioned

build.gradle
```diff
@@ -1,48 +1,47 @@
...
-import groovyx.net.http.Method
-import groovyx.net.http.ParserRegistry
 import org.gradle.internal.os.OperatingSystem
 
+import java.nio.file.Paths
+
 buildscript {
   repositories {
-    mavenCentral()
+    jcenter()
```
micheljung, line 10: We shouldn't jcenter() as it's going to shut down

build.gradle
```diff
@@ -271,7 +294,8 @@ dependencies {
   compile("org.jsoup:jsoup:1.11.3")
   compile("com.nativelibs4java:bridj:0.7.0")
   compile("com.github.nocatch:nocatch:1.1")
-  compile("org.luaj:luaj-jse:3.0.1")
+  // Custom version because of https://github.com/luaj/luaj/issues/70
+  compile("com.github.micheljung:luaj:da919e2")
```
micheljung, line 298: This issue was relevant when I used update4j's bootstrapping mechanism (start the updater, then load the application as a Module) but since this doesn't work anyway with JavaFX' WebView (can't be loaded as a module because of classpath issues) and I therefore switched to an update-on-exit approach, this can be reverted.

buildSrc/build.gradle
```diff
@@ -0,0 +1,9 @@
+repositories {
+  jcenter()
```
micheljung, line 2: mavenCentral()

buildSrc/src/main/java/com/faforever/client/Update4jConfigTask.java
```diff
@@ -0,0 +1,60 @@
...
+import org.update4j.Configuration.Builder;
+import org.update4j.FileMetadata;
+import org.update4j.OS;
+
+import java.io.BufferedWriter;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class Update4jConfigTask extends DefaultTask {
```
micheljung, line 16:
```suggestion
/** Generates an update4j configuration file that can be used to update the client. */
public class Update4jConfigTask extends DefaultTask {
```

ci/publish-update4j.sh
```diff
@@ -0,0 +1,38 @@
...
+
+ZIP_FILE_NAME="faf_windows-x64_${APP_VERSION_UNDERSCORES}.zip"
+TAR_FILE_NAME="faf_unix_${APP_VERSION_UNDERSCORES}.tar.gz"
+
+sshNoCheck() {
+  ssh -o "UserKnownHostsFile=/dev/null" -o "StrictHostKeyChecking=no" "$@"
+}
+scpNoCheck() {
+  scp -o "UserKnownHostsFile=/dev/null" -o "StrictHostKeyChecking=no" "$@"
+}
```
micheljung, line 22: There's an easier way: those settings can be written to .ssh/config or something

micheljung:
Which library issues are you referring to, other than lua4j? For that one, see my comment

Sheikah45:
Good to know on the luaj the other was the the webview patch itself

Sheikah45:
Closing as this has gotten stale

micheljung:
@Sheikah45  I still want this to happen and there's a lot of code involved that can still be used. But I assume you mean the branch is obsolete because you created a new one? I saw that in that branch the ci/ folder was deleted, including new files that are required for the auto-updating. Those should be added to your branch before we delete this one.

Sheikah45:
I closed this branch presently because no one is working on it right now and it is rather old. I have another branch that was rebased a while ago but would still need to be majorly rebased again. This branch also hasn't been deleted.

micheljung:
Yes, and w/o a PR the risk of the branch being deleted is high which is why we need to add this file to you branch before the branch gets deleted :)

PR #1845, opened 2020-07-26
Create Abstract Vault Controller

### LINKED ISSUE #1660: Improve map vault pagination
Currently this is what we have: 

And something like this is the standard:  

Would be nice to have this added for improved vault usability.

### LINKED ISSUE #1765: Pagination of the mods-register
By clicking "more" on the mods page the mods-register loads more and more Items at once. That occupies too much local space. (The loaded items aren't fully loaded, some pictures are missing):
-Add Pagination to enable an user friendly interaction with the mods-register .
-Freeing up space for new items, to enhance loading speed. (And enable the loading of all pictures.)
A special sever request needs to be fired when loading new pages.

### LINKED ISSUE #1799: Use proper pagination in Map vault
JavaFX provides a [Pagination Control](https://docs.oracle.com/javafx/2/ui_controls/pagination.htm).
It's very simple to use and should be used instead of our custom implementation in the map vault.

Other paginations should not be updated as part of this issue.

### LINKED ISSUE #1814: Pagination of Mods
While in Vault/Mods, upon clicking the more button, some pictures do not load at all. In Addition to that, the more a user requests for mods, the smaller is the number of actually shown mods, until it throws an Exception at some point.

Sheikah45:
Create an abstract class for Map, Mod, and Online Replay Vaults to use in display for more uniformity


src/main/java/com/faforever/client/theme/UiService.java
```diff
@@ -413,6 +413,16 @@ public void loadThemes() {
     return loader.getController();
   }
 
+  public <T extends Controller<?>> T loadFxml(String relativePath, Class<?> controllerClass) {
+    FXMLLoader loader = new FXMLLoader();
+    loader.setControllerFactory(applicationContext::getBean);
```
1-alex98, line 418: Dont think u need thatand should not do that. Since u already use setController()

Sheikah45:
Will need to solve issue between replay vault using vbox and map and mods using flow pane

Sheikah45:
The API lies about the number of pages when querying mods

1-alex98 (commented):
I would try put far more code in the abstract class. Like almost all of it apart from the code where u load the Entity Detail and the Entity cards. Also would not use fx:include but load with uiService in VaultController and the method that takes a special Controller class.

src/main/java/com/faforever/client/theme/UiService.java
Sheikah45: This was only because the fx:include of the searchBox in the entity needed an application bean provided by the factory but removing that will remove the need for the controller factory as well.

Sheikah45:
I believe that Map, Mod and Online do not use fx:include did you want to have the live and local replays to use this as well?
And working on better abstraction.

src/main/java/com/faforever/client/theme/UiService.java
Sheikah45: Just for context to remove the need to set the controller factory would just need to load and place the fxml for the search, logical node and specification. 
What do you think the adverse effect of setting the controller factory vs manually loading and setting the controller for those?

src/main/java/com/faforever/client/map/MapVaultController.java
```diff
@@ -46,95 +32,43 @@
...
-  private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  @VisibleForTesting
   static final int TOP_ELEMENT_COUNT = 7;
-  @VisibleForTesting
-  static final int LOAD_PER_PAGE = 50;
-  /**
-   * How many mod cards should be badged into one UI thread runnable.
-   */
-  private static final int BATCH_SIZE = 10;
+  private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
1-alex98: Replaye with @Slf4j

1-alex98:
I believe that live and local replay are different so no. But what about the SHOW_ROOM can be not also put that into the abstract part is that not always the same? 7 items and a more button. I would go as far as to say all the abstract controller needs to know is :
What the SHOW_ROOM categories are called and how to query those SHOW_ROOM categories :thinking:

src/main/java/com/faforever/client/theme/UiService.java
1-alex98: ControllerFactory uses the fx:controlle=? in the fxml file and makes a new instance of whatever class there is mentioned. U now just set your own instance that u want.

1-alex98:
Maybe we look at it together some day

src/main/java/com/faforever/client/theme/UiService.java
Sheikah45: Yes the loader sets the controller for the top level file however the search box has its own set of controllers that also need to be set which are gotten by the controller factory unless we change them to all be handed directly in each of those as well.

Sheikah45:
Yeah might be good. I had tried to leave the showroom in fxml because it seemed like so far we had tried to avoid defining the elements inline and thought it might be so that themes or it could be changed easier but we could also just do it inline if there is no reason not to. Also the SHOW_ROOM is set up beforehand rather than being created dynamically and each showroom has a different number of panels so that is defined in a separate fxml. The actual populating is still done by the populate function.

1-alex98:
Lemme see what i would change

src/main/java/com/faforever/client/mod/ModVaultController.java
```diff
@@ -42,167 +31,136 @@
...
     AnchorPane.setTopAnchor(modDetailRoot, 0d);
     AnchorPane.setRightAnchor(modDetailRoot, 0d);
     AnchorPane.setBottomAnchor(modDetailRoot, 0d);
     AnchorPane.setLeftAnchor(modDetailRoot, 0d);
     modDetailRoot.setVisible(false);
 
+    manageModButton = new Button(i18n.get("modVault.manageMods"));
+    manageModButton.getStyleClass().add("button-raised");
+    manageModButton.setOnAction((event -> manageMods()));
+    searchBox.getChildren().add(manageModButton);
```
1-alex98: Avoid defining UI Elements in java code

Sheikah45: I can add this to the to the vault entity and just hide it for the other classes.

Although with regards to buttons in general we do have to set the onAction in java since the fxml doesn't know about the functions without having a controller set there.

1-alex98 (commented):
We need to kill all Show Room Controllers not sure if i did that

1-alex98 (dismissed):

src/test/java/com/faforever/client/vault/VaultEntityControllerTest.java
```diff
@@ -0,0 +1,424 @@
...
+  @Before
+  public void setUp() throws Exception {
+    showRoomRoot = new VBox();
+    showRoomLabel = new Label();
+    showRoomPane = new FlowPane();
+    moreButton = new Button();
+    showRoomPane.setUserData(moreButton);
+
+    when(i18n.get("mock")).thenReturn("mock");
+//    when(preferencesService.getPreferences()).thenReturn(new Preferences());
```
1-alex98: no commented out code

src/test/java/com/faforever/client/vault/VaultEntityControllerTest.java
```diff
@@ -0,0 +1,424 @@
...
+//  public void testOwnedMapsShownWhenUserOwnsMaps() {
+//    when(mapService.getOwnedMaps(anyInt(), anyInt(), anyInt()))
+//        .thenReturn(CompletableFuture.completedFuture(createMockElements(5)));
+//    instance.display(new OpenMapVaultEvent());
+//    WaitForAsyncUtils.waitForFxEvents();
+//    assertTrue(instance.ownedPane.isVisible()
+//        && instance.moreOwnedButton.isVisible()
+//        && instance.ownedMoreLabel.isVisible());
+//  }
+}
```
1-alex98: No commented out code

src/test/java/com/faforever/client/test/AbstractPlainJavaFxTest.java
```diff
@@ -106,6 +107,37 @@ protected void loadFxml(String fileName, Callback<Class<?>, Object> controllerFa
     }
   }
 
+  protected void loadFxml(String fileName, Callback<Class<?>, Object> controllerFactory, VaultEntityController<?> controller) throws IOException {
```
1-alex98, line 84: Could we not unify this with the other loadFXMl there is not realy too much of a difference

1-alex98 (changes requested):

src/test/java/com/faforever/client/vault/VaultEntityControllerTest.java
Sheikah45: whoops removed

src/test/java/com/faforever/client/vault/VaultEntityControllerTest.java
Sheikah45: removed

src/test/java/com/faforever/client/test/AbstractPlainJavaFxTest.java
Sheikah45: combined

1-alex98 (dismissed):

1-alex98:
we need to rebase and test it again

Sheikah45:
Can do

Sheikah45:
Actually my version says that it is up to date with develop and no commits were added

1-alex98:
Github says it is out of date no idea

1-alex98:
oh because probably u need to see develop on this remote cause your are looking at develop from your fork

Sheikah45:
Ah yeah it just hadnt fetched

1-alex98 (dismissed):

1-alex98:
The teams in the online replay vault are wrong. Acctually not seems to be only old replays...

Sheikah45:
Yeah the old replays had teams as 0 and 1 where as the current ones have 0 as no team and use 1 and 2 for the first two teams

PR #1865, opened 2020-08-21
Show reason for not rated games

### LINKED ISSUE #1864: Display the reason for invalid games in the vault
Instead of saying a game is not rated tell the user why.

Visit Issue hunt...

1-alex98:
Fixes #1864


src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -318,6 +315,22 @@ private void populateTeamsContainer() {
     }));
   }
 
+  private void configureRatingControls() {
+    if (!replay.getValidity().equals(Validity.VALID)) {
+      showRatingChangeButton.setVisible(false);
+      notRatedReasonLabel.setVisible(true);
+      notRatedReasonLabel.setText(i18n.get("game.reasonNotValid", replay.getValidity()));
```
Brutus5000: This will not translate the validity reasons, so it's not a proper i18n solution.

1-alex98: ok so I need an I18N key for all of them? ok

1-alex98: was just lazy

Sheikah45:
Would it be better for this to be a tooltip rather than the string itself depending on how long the reasons are?
Or maybe a separate line?

Sheikah45:
Updated it to use i18n for the reasons
@axel1200 for some reason cant request you for review so am pinging

1-alex98:
I can not review my own PR that's why

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -321,6 +315,22 @@ private void populateTeamsContainer() {
     }));
   }
 
+  private void configureRatingControls() {
+    if (!replay.getValidity().equals(Validity.VALID)) {
+      showRatingChangeButton.setVisible(false);
+      notRatedReasonLabel.setVisible(true);
+      notRatedReasonLabel.setText(i18n.get("game.reasonNotValid", i18n.get(i18n.get("game.reasonNotValid.keyFormat", replay.getValidity().ordinal()))));
```
1-alex98: I am pretty sure game.reasonNotValid.keyFormat should not be in I18N, imagine somebody thinks he most translate it then this breaks for the other languages.

1-alex98: Also you got to have a fallback. Maybe somebody adds a new validity reason to the server one day. I suggest if the i18n key does not exist just take the getValidity().name() and say it is the reason.

src/main/resources/i18n/messages.properties
```diff
@@ -655,7 +655,32 @@ replay.playerCountTooltip=Players
 replay.ratingTooltip=Average rating
 replay.qualityTooltip=Quality
 game.showRatingChange=Show rating change
-game.notValid=The game was not rated
+game.reasonNotValid=Game was not rated. Reason: {0}
+game.reasonNotValid.keyFormat=game.reasonNotValid.{0}
```
1-alex98: That should be a constant

src/main/java/com/faforever/client/replay/ReplayDetailController.java
1-alex98: On the other hand then parsing the json would not work so maybe that is not needed. But safety messure would not hurt...

1-alex98:
Also CI fails

src/main/java/com/faforever/client/replay/ReplayDetailController.java
Sheikah45: Hmm yeah good point will just change it back to string formatting unless you think there is a better way

Sheikah45:
How do I run the push tests the tests I run locally all pass

1-alex98:
I am gonna rerun the tests some times the fail for strange reasons

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -66,6 +67,8 @@
 @RequiredArgsConstructor
 public class ReplayDetailController implements Controller<Node> {
 
+  public final String REASON_KEY_FORMAT = "game.reasonNotValid.%d";
```
1-alex98:
```suggestion
  private static final String REASON_KEY_FORMAT = "game.reasonNotValid.%d";
```

1-alex98 (commented):
Approve, after making the one String a real constant

1-alex98 (commented):
Approved

1-alex98:
CI fails:
```
> Task :compileTestJava
/home/travis/build/FAForever/downlords-faf-client/src/test/java/com/faforever/client/replay/ReplayDetailControllerTest.java:126: error: REASON_KEY_FORMAT has private access in ReplayDetailController
    when(i18n.get("game.reasonNotValid", i18n.get(String.format(instance.REASON_KEY_FORMAT, replay.getValidity().ordinal())))).thenReturn("Reason: HAS_AI");
                                                                        ^
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error
```

make it package private ;)

micheljung:
WAIT 😄 Please don't use dynamic message keys.

With static keys, the IDE (or any tool) can always check which key is used where and **if** it's used. With dynamic keys, we lose this feature and things get more difficult to maintain. Moreover, this:

    game.reasonNotValid.keyFormat=game.reasonNotValid.{0}

abuses the i18n mechanism; game.reasonNotValid.{0} is not a localized string but a technical indirection.

Please implement it like so:

1. Create a reason-enum
1. Create one constant for each known reason
1. Assign a `code` to each enum constant (don't use the enum's ordinal) by which the enum constant can be looked up
1. Assign a localization key to each enum constant
1. Implement the i18n lookups accordingly

micheljung (changes requested):
WAIT 😄 Please don't use dynamic message keys.

With static keys, the IDE (or any tool) can always check which key is used where and **if** it's used. With dynamic keys, we lose this feature and things get more difficult to maintain. Moreover, this:

    game.reasonNotValid.keyFormat=game.reasonNotValid.{0}

abuses the i18n mechanism; game.reasonNotValid.{0} is not a localized string but a technical indirection.

Please implement it like so:

1. Create a reason-enum
1. Create one constant for each known reason
1. Assign a `code` to each enum constant (don't use the enum's ordinal) by which the enum constant can be looked up
1. Assign a localization key to each enum constant
1. Implement the i18n lookups accordingly

Sheikah45:
I will change it. Although game.reasonNotValid.keyFormat should no longer exist but I will make sure it is removed.

1-alex98:
@micheljung i think your complains should be addressed

1-alex98 (commented):
if @micheljung does not leave a review I guess we just do it

Sheikah45:
anything to do here?

1-alex98 (commented):
Approve

Issue #1771, opened 2020-06-14
Nomads Replays fail to open, normal replays are fine


Exotic-Retard:
**Describe the bug**
Open any replay containing the nomads mod, the client works fine, but then the game says that the replay is corrupted on launch. Non-nomads replays load fine.

Broken replay (nomads): https://content.faforever.com/faf/vault/replay_vault/replay.php?id=11463794

working replay (not-nomads): https://content.faforever.com/faf/vault/replay_vault/replay.php?id=11762033

99% sure that this is a regression or some weird bug that set itself; the nomads replays worked perfectly fine in the recent past, less than 2-3 weeks ago, considering i was watching nomads replays back then.

**Screenshots**
https://i.imgur.com/IyelwAO.png

Brutus5000:
@Exotic-Retard Can you correlate the problem to a fa-patch? I'm 99,9% sure that replays of featured mods break, when the fa basemod gets an update. (That is an issue of the replay format however)

Exotic-Retard:
looks like you are right, the old replays break when the base patch is updated. i expected them to desync instead of getting corrupted completely though. i guess that sort of explains the issue, but its still a shame i suppose that old replays totally break like that.

Regardless, thanks for your help <3

Sheikah45:
Should have been fixed with #1866 but will still desync due to issues with featuredmod versioning

PR #1866, opened 2020-08-22
Fix coop replay parsing

### LINKED ISSUE #1676: Clientside workaround for coop replays
[Issuehunt badges](https://issuehunt.io/r/FAForever/downlords-faf-client/issues/1676)

## Intro
This is a a sort of mini-guide on how to launch coop replays manually. I'm creating this as a github issue because I think it should be pretty simple to get the client to do this automatically. In the mean time people will need to do the set up themselves.

### Why don't coop replays work?
Coop maps are not stored in the vault like regular maps. This means that the replay server doesn't recognize coop map names, and instead just write `none` into the `fafreplay` header. When the client goes to download the map we see an exception like this:
```
java.io.FileNotFoundException: https://content.faforever.com/faf/vault/maps/none.zip
```

### How to launch coop replays manually
First off, you will want to have the legacy client installed as it will handle updating to the correct version of FA for the replay: https://github.com/FAForever/client/releases

1. Download the replay from the vault. An easy way to do this is to hit the "play" button on the coop tab, and cancel when you see the error, just be aware that it is downloaded into `C:\ProgramData\FAForever\cache\temp.fafreplay`.
2. Change the replay extension from `.fafreplay` to `.coop.fafreplay`. This will tell the legacy client that it should update for the `coop` game mode.
3. Right click the replay and open it with the legacy client. It is probably called `Forged Alliance Forever - Lobby Client`. The legacy client should now update the necessary files, but it will most likely fail with an error message like "Map not found" or a timeout error.
image

4. Start the replay from the command line. Open `cmd` or `powershell` and paste in the following command:
```
C:\ProgramData\FAForever\bin\ForgedAlliance.exe /init init_coop.lua /nobugreport /log C:\ProgramData\FAForever\logs\game.log /replay C:\ProgramData\FAForever\cache\temp.scfareplay
```

## Client workaround
I think we should modify the replay launching code so that it skips the map download for coop maps. That way the user can set up their maps directory manually before launching the replay. We should also make sure that the client downloads the correct version of the FA for the replay, as it stands I think it tries to always download the most recent version (the legacy client handles this correctly).

## Criteria for closing this issue
I want to be able to click the "play" button on the coop leaderboard and have the client download and successfully launch the replay. The client should update the `faf` featured mod to the correct version, but may assume that:
1. The map is already present and at the right version
2. The `coop` featured mod is at the right version

Visit Issue hunt...

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $10.00)

- [askaholic askaholic](https://issuehunt.io/u/askaholic) ($10.00)

### Submitted pull Requests
- [#1866 Fix coop replay parsing](https://issuehunt.io/r/FAForever/downlords-faf-client/pull/1866)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/FAForever/downlords-faf-client/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

Sheikah45:
Reads in map name from replay bytes if game type is coop.
Updates FAF to correct version for replay.

Fixes #1676

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#1676: Clientside workaround for coop replays](https://issuehunt.io/repos/32919066/issues/1676)
---
</details>

### DIFF
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -42,7 +42,6 @@ import com.google.common.primitives.Bytes;
 import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
-import org.eclipse.jgit.util.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationContext;
@@ -509,22 +508,7 @@ public class ReplayService {
     String gameType = replayInfo.getFeaturedMod();
     Integer replayId = replayInfo.getUid();
     Map<String, Integer> modVersions = replayInfo.getFeaturedModVersions();
-    String mapName = replayInfo.getMapname();
-
-    // For some reason in the coop replay the map name is null in the metadata
-    // So we just take it directly from the replay data.
-    if (gameType.equals("coop")) {
-      mapName = parseMapFolderName(rawReplayBytes);
-    }
-
-    // For map generator games the map name is "None" because replay server gets map name by from DB based on filename
-    // from replay data, and DB does not contain generated maps.
-    if (StringUtils.equalsIgnoreCase(mapName, "None")) {
-      String maybeMapGen = parseMapFolderName(rawReplayBytes);
-      if (mapGeneratorService.isGeneratedMap(maybeMapGen)) {
-        mapName = maybeMapGen;
-      }
-    }
+    String mapName = parseMapFolderName(rawReplayBytes);
 
     Set<String> simMods = replayInfo.getSimMods() != null ? replayInfo.getSimMods().keySet() : emptySet();
 
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -513,14 +513,14 @@ public class ReplayService {
 
     // For some reason in the coop replay the map name is null in the metadata
     // So we just take it directly from the replay data.
-    if (StringUtils.isEmptyOrNull(mapName)) {
+    if (gameType.equals("coop")) {
       mapName = parseMapFolderName(rawReplayBytes);
     }
 
     // For map generator games the map name is "None" because replay server gets map name by from DB based on filename
     // from replay data, and DB does not contain generated maps.
     if (StringUtils.equalsIgnoreCase(mapName, "None")) {
-      String maybeMapGen = parseMapName(rawReplayBytes).replaceAll(".scmap", "");
+      String maybeMapGen = parseMapFolderName(rawReplayBytes);
       if (mapGeneratorService.isGeneratedMap(maybeMapGen)) {
         mapName = maybeMapGen;
       }
--- a/src/main/java/com/faforever/client/patch/GameUpdaterImpl.java
+++ b/src/main/java/com/faforever/client/patch/GameUpdaterImpl.java
@@ -73,7 +73,7 @@ public class GameUpdaterImpl implements GameUpdater {
 
     if (!NAMES_OF_FEATURED_BASE_MODS.contains(featuredMod.getTechnicalName())) {
       future = future.thenCompose(aVoid -> modService.getFeaturedMod(FAF.getTechnicalName()))
-          .thenCompose(baseMod -> updateFeaturedMod(baseMod, null))
+          .thenCompose(baseMod -> updateFeaturedMod(baseMod, version))
           .thenAccept(patchResults::add);
     }
 


Sheikah45:
somewhat fixes #1771 as now the replays will play but do desync

Askaholic:
Cool, seems to work!. Which replay did you try that desynced? I would guess that's because it was recorded on an older version of the coop mod. I tried `10674765` which was on game patch 3708 and whatever version of coop I have installed was the right one because it didn't desync.

Sheikah45:
The desync replay I was referencing was 11463794 which is exotics nomads replay.

For the coop replay I get a desync on 10929347

Askaholic:
Yea it desynced for me using the Java client, but I was able to watch it by opening it with the python client first and then launching the replay from the command line. So there might be some update code we could copy from the python client to get the right version of the coop files.

Sheikah45:
Alright I will take a look. I was having some issues with the python client before but will hopefully be able to get it to work.

1-alex98:
or maybe the faf files. I mean coop is based on the faf featured mod if you load the old coop files that's not enough you also need to load the old faf files. However I am not sure if there is even a record that remembers which faf files are needed. @Brutus5000 thought about ways to refactor that whole what featured mod depends  on which others in what version, but we never pushed any changes to the db as far as I remember.

Sheikah45:
Hmm yeah when I run it with the python client I still get the desync issues. 

And @axel1200 the faf files are currently being updated which allows to run the replays at all. The issues seems to be the coop files perhaps the init however in the replay data there doesnt seem to be an entry for the version of the coop used as the id for coop is 25 but the featuredmod versions saved with the replay only go up to 22.

Sheikah45:
It seems to be specifically replays where faf is 3709 that have the desync issues the others above that work

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -513,14 +513,14 @@ private void runFafReplayFile(Path path) throws IOException {
 
     // For some reason in the coop replay the map name is null in the metadata
     // So we just take it directly from the replay data.
-    if (StringUtils.isEmptyOrNull(mapName)) {
+    if (gameType.equals("coop")) {
```
1-alex98: Use the KnowFeaturedMod Enum

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -513,14 +513,14 @@ private void runFafReplayFile(Path path) throws IOException {
 
     // For some reason in the coop replay the map name is null in the metadata
```
1-alex98: Could we maybe test this method with a real replay :D Then we would solve this issue forever

1-alex98 (changes requested):

src/main/java/com/faforever/client/replay/ReplayService.java
Sheikah45: Do you mean just get the map name from the replay data for all maps?

Sheikah45: I made a change assuming that is what you meant

Sheikah45: seems to work

1-alex98: Was absolutely not what I meant XD I meant write a unit test

1-alex98: Concerning what u did I don't know if that's a good idea. No idea what pros and cons are too that

1-alex98: Maybe @micheljung @Wesmania know what pros and cons are about reading the map from the replay file or the replayInfo stored in the databse.

Wesmania: It's the right thing to do for randomized maps, since these will never be stored in the DB. The coop map issue is here: https://github.com/FAForever/server/issues/372
I don't know which is right for all the rest tbh.

Sheikah45: I tested it and it worked for maps in the vault as well. I think an issue would arise if for some reason the map in the lua header was different than the vault or if the map table structure is intended to be changed. Also I believe the replayinfo that is being read here is stored in the replay file as well just it is the part in the faf header vs the lua header.

1-alex98: Hmm well I mean if it works better let's do it this way right.

Sheikah45:
anything to do here?

Askaholic:
Seems fine to me. It might not fix all the issues, but it helps. I haven't gotten around to investigating why the replay was desyncing when I could get it to work manually, and I don't see myself going back to it any time soon.

1-alex98 (approved):

PR #1870, opened 2020-08-29
Store multiple game logs with time stamp

### LINKED ISSUE #777: Seperate log file per game
image

Sheikah45:
Fixes #777

Stores up to 10 game logs in the log folder with the time stamp

### DIFF
--- a/src/main/java/com/faforever/client/preferences/PreferencesService.java
+++ b/src/main/java/com/faforever/client/preferences/PreferencesService.java
@@ -9,6 +9,7 @@ import com.faforever.client.preferences.gson.PropertyTypeAdapter;
 import com.faforever.client.remote.gson.FactionTypeAdapter;
 import com.faforever.client.update.ClientConfiguration;
 import com.github.nocatch.NoCatch.NoCatchRunnable;
+import com.github.nocatch.NoCatchException;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.sun.jna.platform.win32.Shell32Util;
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/client/preferences/PreferencesService.java
```diff
@@ -346,6 +355,22 @@ public Path getThemesDirectory() {
...
+          .sorted(Comparator.comparingLong(p -> p.toFile().lastModified()))
+          .sorted(Collections.reverseOrder())
+          .skip(NUMBER_GAME_LOGS_STORED - 1)
+          .forEach(p -> noCatch(() -> Files.delete(p)));
+    } catch (IOException e) {
+      logger.error("Could not list log directory.", e);
+    } catch (NoCatchException e) {
+      logger.error("Could not delete game log file");
+    }
+    return getFafLogDirectory().resolve(new SimpleDateFormat("'game_'yyyy-MM-dd_HHmmss'.log'").format(new Date()));
```
1-alex98: How is that gonna look are there no seperation between Hours and minutes?

1-alex98 (dismissed):
I like, need to test this some day

src/main/java/com/faforever/client/preferences/PreferencesService.java
Sheikah45: Screenshot (50)

Thought this was a standard time format but can be changed if you have one you like better that is file valid

1-alex98: idk I find it hard to read. Btw the old client used the game uuid as a name. Not sure of that would be an advantage. I think that was so you could match replay and game log.

Sheikah45: That might be better I will just change to that if it works for you

1-alex98: Yeah sure. This is not what we gotta do, I also like the time. I am indecisive.

Sheikah45: Yeah personally I think the time might be more intuitive for the user than game uid and I can add a dash separator for the time

bukajsytlos:
sry maybe nitpicking, but could java.time API be used? I would say game id is better. date can be from FS

1-alex98 (dismissed):

1-alex98:
CI fails and i still gotta test it

1-alex98 (approved):
works

PR #1873, opened 2020-08-29
Bump ice adapter to latency checking version

Geosearchef:


1-alex98:
Hi dear geo,
how important is this? Is there a second ICE Server or are we about to purchase one? I want to release now. If I release with that I need to first make another RC... So if this is something for the long run let's have it in the next release if it is needed somehow I am gonna make a new RC.

Askaholic:
Yes one of the Australian guys has set up a coturn server and we need this so we can test it out and then activate on prod too.

Geosearchef:
It's pretty important, we're waiting for it to enter the second relay server into the server config, but it's untested, so it would be nice if we could do another RC before this release.

Geosearchef:
(the turn server is already configured and running)

1-alex98:
@Geosearchef Sure in that case let's do another RC

PaulWay:
Yay!

PR #1878, opened 2020-09-02
copy local replay info to clipboard

archaeron:
something like this would be cool

This PR adds a context menu to the local replay entries where you can copy the replay id or the replay path

copy-replay-info

(note that I don't write Java for a living, I'm happy to change things for the PR)


src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
```diff
@@ -120,6 +124,25 @@ protected void loadLocalReplaysInBackground() {
   @NotNull
   private TableRow<Replay> replayRowFactory() {
     TableRow<Replay> row = new TableRow<>();
+    ContextMenu contextMenu = new ContextMenu();
+    MenuItem copyReplayId = new MenuItem("Copy Replay Id");
```
1-alex98: we use internationalization so this needs to be translatable. Use the i18n service. Just make a constructor parameter like all the other services. You then enter the english translation in the message.properties file in the resources/i18n folder

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
```diff
@@ -120,6 +124,25 @@ protected void loadLocalReplaysInBackground() {
...
+      clipboard.setContent(content);
+    });
+    MenuItem copyReplayPath = new MenuItem("Copy Replay Path");
+    copyReplayPath.setOnAction(event -> {
+      var replayPath = row.getItem().getReplayFile();
+      final Clipboard clipboard = Clipboard.getSystemClipboard();
+      final ClipboardContent content = new ClipboardContent();
+      content.putString(String.valueOf(replayPath));
+      clipboard.setContent(content);
+    });
```
1-alex98: Maybe extract into method(ctrl+ alt+ m in Intelij)

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
```diff
@@ -120,6 +124,25 @@ protected void loadLocalReplaysInBackground() {
...
+    copyReplayId.setOnAction(event -> {
+      var replayId = row.getItem().getId();
+      final Clipboard clipboard = Clipboard.getSystemClipboard();
+      final ClipboardContent content = new ClipboardContent();
+      content.putString(String.valueOf(replayId));
+      clipboard.setContent(content);
+    });
+    MenuItem copyReplayPath = new MenuItem("Copy Replay Path");
+    copyReplayPath.setOnAction(event -> {
+      var replayPath = row.getItem().getReplayFile();
```
1-alex98: We avoid the var keyword for code style reasons

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
```diff
@@ -120,6 +124,25 @@ protected void loadLocalReplaysInBackground() {
...
+      final Clipboard clipboard = Clipboard.getSystemClipboard();
+      final ClipboardContent content = new ClipboardContent();
+      content.putString(String.valueOf(replayPath));
```
1-alex98: Not sure how String.valueOf divers from that but maybe .getAbsolutePath() would be safer. Not sure if that method exists cause I am not sure what the type of that variable is(since you use var).

1-alex98 (changes requested):

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
archaeron: that came from copy pasting the upper entry 🙈  
thanks

archaeron:
@axel1200 thanks a lot for the review!

what do you think of my changes? I'm happy to change more. Thanks

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
archaeron: what do I do with the other languages?

1-alex98 (dismissed):

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
1-alex98: Nothing they are translated on poeditor.com

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
```diff
@@ -129,6 +137,30 @@ protected void loadLocalReplaysInBackground() {
...
+      content.putString(String.valueOf(replayId));
+      clipboard.setContent(content);
+    });
+    MenuItem copyReplayPath = new MenuItem(i18n.get("vault.replays.copyReplayPath"));
+    copyReplayPath.setOnAction(event -> {
+      Path replayPath = replay.getReplayFile();
+      final Clipboard clipboard = Clipboard.getSystemClipboard();
+      final ClipboardContent content = new ClipboardContent();
+      content.putString(replayPath.toFile().getAbsolutePath());
+      clipboard.setContent(content);
```
Brutus5000: Duplicate code. I don't know if we use clipboard features elsewhere, but if yes a clipboard util class probably makes sense, otherwise at least this should be extracted into a method here.

Brutus5000 (changes requested):

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
archaeron: I have created a `ClipboardUtil` class and used it also in `ChatUserContextMenuController`  and `ReplayDetailController`

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
archaeron: changed this, thanks

src/main/java/com/faforever/client/vault/replay/ReplayVaultController.java
archaeron: I extracted this into the method `replayRowContextMenuFactory`

1-alex98 (approved):

Brutus5000 (approved):

1-alex98:
Please always open an issue for everything you do and link it in your PR by writing "Fixes #number" in your commit message. Please refer on how to format commit messages in the wiki and contribution guidelines

archaeron:
@axel1200 I will do next time. Thanks for the review and the merge! :)

1-alex98:
@archaeron we can invite you to Zulip(chat service if you like). Email me your email :D alexander.von.trostorff@gmail.com

PR #1880, opened 2020-09-03
Add in check for maps added to directory to map directory watcher

### LINKED ISSUE #1879: Maps not added to map list when added to directory manually
Currently maps that are manually added to the maps directory are not added to the maps list in the client forcing a restart for the map to be recognized

**To Reproduce**
Add a map to the maps directory while the client is open

**Expected behavior**
The client should add the map to the installed maps list

**OS**
Windows

Sheikah45:
Fixes #1879


1-alex98 (approved):
Wanna test this still. Maybe we should write a test that validates actions on Create so nobody forgets about it again. Apart from that its nice

1-alex98 (changes requested):
So I tested it on linux and it seems that installing a map now triggers this as well but it triggers it before it is even fully copied (this would also happen on manually copying if stuff is copied file by file).

The Exception in the log is:
```
2020-09-05 20:55:31.511  WARN 32900 --- [     Thread-101] com.faforever.client.map.MapService      : Map could not be read: adaptive_siberia.v0001

com.faforever.client.map.MapLoadException: Map folder does not contain a *_scenario.lua: /home/alex/.faforever/user/My Games/Gas Powered Games/Supreme Commander Forged Alliance/maps/adaptive_siberia.v0001
	at com.faforever.client.map.MapService.lambda$readMap$10(MapService.java:284)
	at java.base/java.util.Optional.orElseThrow(Optional.java:401)
	at com.faforever.client.map.MapService.readMap(MapService.java:284)
	at com.faforever.client.map.MapService.addInstalledMap(MapService.java:258)
	at com.faforever.client.map.MapService.lambda$startDirectoryWatcher$6(MapService.java:208)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
	at com.faforever.client.map.MapService.lambda$startDirectoryWatcher$7(MapService.java:204)
	at com.github.nocatch.NoCatch.noCatch(NoCatch.java:49)
	at com.github.nocatch.NoCatch.noCatch(NoCatch.java:34)
	at com.faforever.client.map.MapService.lambda$startDirectoryWatcher$8(MapService.java:197)
	at java.base/java.lang.Thread.run(Thread.java:832)
```
Such a *_scenario.lua file exists but is not yet copied. We need to do something to avoid this. Either we have a look that needs to be acquired or we just delay the watcher by a few seconds so it does not trigger in between copying files.

Sheikah45:
Well if we delay should it be delayed by longer for various download speeds?

Sheikah45:
I added in a watcher for when the scenario file is written and gave it a timeout of a minute so this should cover the situations of downloading the map or manually copying them.

1-alex98:
@Sheikah45 The map is first downloaded and then unpacked. So a delay of a second would be enough

1-alex98:
OK the whole delay thing is probably shit. I looked at the methods that add a map and it only reads from the scenario.lua. So let's drop all the delay thing and just react to scenario.lua being added and drop all other create actions

1-alex98:
ok we go for sub directory watcher right?

Sheikah45:
Yes and I am going to move it into a function

Sheikah45:
Alright moved it to a sub function. Let me know if you want anything changed.

Sheikah45:
Hmm CI is failing but all tests pass when I run them

src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -266,6 +224,52 @@ private Thread startDirectoryWatcher(Path mapsDirectory) {
     return thread;
   }
 
+  private boolean checkScenarioLua(Path mapPath) {
```
1-alex98: Normally we don't make comments but as this is not a common thing to do maybe explain it in a java doc

Sheikah45:
are there examples of how you think it should be documented

src/main/java/com/faforever/client/map/MapService.java
Sheikah45: are there examples of how you think it should be documented

Sheikah45: Added in the javadoc

Brutus5000 (commented):
Never have I seen such a horrible combination of blocking and threading.

I see a huge load of complexity just for the case that the user could have create a directory that doesn't really contain a map. Wouldn't it be easier to say: Hey, we waited 5 seconds still no scenaria.lua we throw an error?

Alternatively I would think about using a proper async library like reactor which gives you retry *and* error handling. (But this is a general thing to be discussed I guess, we shouldn'T just introduce it for this case)

Sheikah45:
Yeah we can just set it to a timer I just wasn't sure if there would be edge cases that should work but fail with just a timer

1-alex98:
Yes we know that it is suboptimal. But the standard java lib offers to better it seems(at least not if it should work on linux and windows). Tell us what to do @Brutus5000 XD

Sheikah45:
I just used a second delay

1-alex98:
@Brutus5000 u ok with this?

Sheikah45:
anything to do here?

1-alex98:
@Brutus5000 2 days for your opinion other wise gonna merge

Brutus5000 (dismissed):

1-alex98:
Seems like I fucked up

Sheikah45:
why close? Do I need to change something?

Sheikah45:
oh wait was it merged?

PR #1882, opened 2020-09-03
Add in check for local preview for generated maps 

### LINKED ISSUE #1881: Use local preview for generated maps
I would like to display the generated map preview for the host when generating maps. This will allow host to easily regenerate if the map is not to their liking rather than having to go all the way into the lobby.

This should be implemented as checking for the locally created preview and if it is not there then using the image stand in.

Sheikah45:
Fixes #1881

### DIFF
--- a/src/main/java/com/faforever/client/map/MapService.java
+++ b/src/main/java/com/faforever/client/map/MapService.java
@@ -32,6 +32,7 @@ import javafx.collections.FXCollections;
 import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
 import javafx.scene.image.Image;
+import lombok.SneakyThrows;
 import org.apache.maven.artifact.versioning.ComparableVersion;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -49,7 +50,6 @@ import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Service;
 
 import javax.inject.Inject;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -303,7 +306,12 @@ public MapBean readMap(Path mapFolder) throws MapLoadException {
   @Cacheable(value = CacheNames.MAP_PREVIEW, unless = "#result == null")
   public Image loadPreview(String mapName, PreviewSize previewSize) {
```
1-alex98, line 306: Just that you know I think having the random icon was a feature and was not done because guys that wrote it did not know how to fetch the preview...

But since you are the maintainer of map gen I leave it to u, just think about it that maybe having the mystery about how the map looks until you join might be interesting.

src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -303,7 +306,12 @@ public MapBean readMap(Path mapFolder) throws MapLoadException {
...
     if (mapGeneratorService.isGeneratedMap(mapName)) {
-      return mapGeneratorService.getGeneratedMapPreviewImage();
+      try {
+        BufferedImage image = ImageIO.read(forgedAlliancePreferences.getCustomMapsDirectory().resolve(mapName).resolve(mapName + "_preview.png").toFile());
```
1-alex98: Why do we have an AWT image only to convert it to an java fx image later, makes little sense to me pls try do this the java fx way...

1-alex98 (changes requested):

src/main/java/com/faforever/client/map/MapService.java
Sheikah45: yeah i was just lazy and didnt change the import settings but will do so

src/main/java/com/faforever/client/map/MapService.java
Sheikah45: yeah I was just considering whether it should show the preview since we also added the options but am honestly not sure about this yet just implemented to see. This also will only get it for the host before they create the game

src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -303,7 +304,11 @@ public MapBean readMap(Path mapFolder) throws MapLoadException {
   @Cacheable(value = CacheNames.MAP_PREVIEW, unless = "#result == null")
   public Image loadPreview(String mapName, PreviewSize previewSize) {
     if (mapGeneratorService.isGeneratedMap(mapName)) {
-      return mapGeneratorService.getGeneratedMapPreviewImage();
+      try {
+        return new Image(new FileInputStream(forgedAlliancePreferences.getCustomMapsDirectory().resolve(mapName).resolve(mapName + "_preview.png").toFile()));
```
1-alex98: Can u not do new Image(....getURL())

Sheikah45: There is no url for this image as this is getting the image generated on the local filesystem

Brutus5000: Hard to read. My proposal:

```
Path previewPath = forgedAlliancePreferences.getCustomMapsDirectory().resolve(mapName).resolve(mapName + "_preview.png");
return new Image(Files.newInputStream(previewPath));
```

I'm also not totally sure when the exception would be thrown? If the file does not exist yet? That shouldn't be catched via try-catch but with `Files.exists(previewPath)` instead (-> never use Exceptions for control flow).
If it's used for anything else, please add a comment.

Brutus5000 (changes requested):

Brutus5000 (approved):

1-alex98 (approved):

PR #1900, opened 2020-09-12
Update download urls

### LINKED ISSUE #1899: Maps not auto downloading
**Describe the bug**
Maps do not auto download when joining a game due to the server change returning 404 errors for the current urls

**To Reproduce**
join a game with a map you dont have

**Expected behavior**
Should download maps before joining lobby. Vault works fine so just need to change the url

Sheikah45:
Fixes #1899 
fixes download urls

### DIFF
--- a/src/main/java/com/faforever/client/config/ClientProperties.java
+++ b/src/main/java/com/faforever/client/config/ClientProperties.java
@@ -85,7 +85,6 @@ public class ClientProperties {
     private String mapDownloadUrlFormat;
     private String mapPreviewUrlFormat;
     private String replayDownloadUrlFormat;
-    private String modDownloadUrlFormat;
   }
 
   @Data
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -21,7 +21,6 @@ faf-client:
     map_validation_url: ${faf-client.api.base-url}/maps/validate
     mapDownloadUrlFormat: ${faf-client.vault.base-url}/maps/%s.zip
     mapPreviewUrlFormat: ${faf-client.vault.base-url}/map_previews/%s/%s.png
-    modDownloadUrlFormat: ${faf-client.vault.base-url}/maps/%s.zip
 
   website:
       forgot-password-url: ${faf-client.website.base-url}/account/password/reset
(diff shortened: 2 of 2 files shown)


Brutus5000:
Previews, mods and replays also affected, basically everything that contains the path /faf/vault

Brutus5000:
I rechecked it an modDownloadUrlFormat is obsolete and can be removed from the yaml and the properties. I noticed that because the yaml points to the map directory instead.

Sheikah45:
Is it reasonable for links to be fixed by just removing /faf/vault from the url?
Also the modDownloadUrlFormat might be named for clarity in what it is used for but am not sure if it is needed.

src/main/resources/application-prod.yml
```diff
@@ -26,6 +26,6 @@ faf-client:
   status-page-url: https://statuspage.faforever.com
 
   forged-alliance:
-    exe-url: https://content.faforever.com/faf/updaterNew/updates_faf_files/ForgedAlliance.exe
+    exe-url: https://content.faforever.com/updaterNew/updates_faf_files/ForgedAlliance.exe
```
1-alex98: like that @Brutus5000 ?

1-alex98: does not work so i guess not

Sheikah45: think it was only /faf/vault

1-alex98 (dismissed):
works

src/main/resources/application-prod.yml
Sheikah45: just /faf is fine

Brutus5000: Indeed /faf is fine. It was just that the /faf/vault/\<folder> was actually a symlink on the server to ../../\<folder> so I removed it

Brutus5000 (changes requested):
modDownloadUrlFormat can removed completely as I said. (Also having it point to maps is super confusing)

Sheikah45:
So then should the mod vault just call getMapDownloadURL or is it better to rename the link?

Brutus5000:
According to my IntelliJ analysis the variable is not used anywhere. What do you want to rename where?

Sheikah45:
Hmm I thought it was used by the mod but looks like I was wrong

Brutus5000 (approved):

PR #1902, opened 2020-09-13
Use country name for tooltip

### LINKED ISSUE #849: Flag tooltips show ISO country code instead country name, flags don't have tooltips at all places
ISO country codes are not user-friendly. Country names are - they are also shown in the in-game lobby:
image

Example for missing flag tooltips: Tooltip shown for user in chat user list, but not on Play tab on game info tile. Also, when chatting, the conversation window showing the messages has no tooltip for the flags, but everything else (except badge) in the user name.

Sheikah45:
Fixes #849


Sheikah45:
I see there is also CountryCode did we want to use that or the built in locale method that is currently implemented?

If we want to use the Locale method I can also use it in all places where CountryCode is used for consistency otherwise will change this to use CountryCode

1-alex98 (approved):

PR #1904, opened 2020-09-13
Feature/#849 use i18n for all country names

Sheikah45:


src/main/java/com/faforever/client/i18n/I18n.java
```diff
@@ -98,6 +98,10 @@ public Locale getUserSpecificLocale() {
     return this.userSpecificLocale;
   }
 
+  public String getCountryNameLocalized(String isoCode) {
+    return new Locale("", isoCode).getDisplayCountry(this.userSpecificLocale);
```
1-alex98, line 105: maybe we should do a null check here right? Do you know what happens on null here?

Sheikah45: on null there is an exception is isoCode is null so I added in a check.
If isoCode is empty it just returns as empty string

1-alex98 (approved):

PR #1905, opened 2020-09-14
Add in check if game is running

### LINKED ISSUE #714: Check if forged alliance is still running when closing client
Warn user if fa is still running and ask him wether to continue...

Sheikah45:
Fixes #714 

Attempts to add in a check for if the game is running before closing the client currently an alert pops up but the window closes anyway

### DIFF
--- a/src/main/java/com/faforever/client/FafClientApplication.java
+++ b/src/main/java/com/faforever/client/FafClientApplication.java
@@ -24,7 +24,6 @@ import javafx.scene.control.Alert.AlertType;
 import javafx.scene.control.ButtonType;
 import javafx.scene.text.Font;
 import javafx.stage.Stage;
-import javafx.stage.Window;
 import javafx.stage.WindowEvent;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.boot.Banner.Mode;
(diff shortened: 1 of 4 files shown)


Sheikah45:
@micheljung I am trying to add in the check for if FA is still running before closing the client however a closerequest event never gets sent so that event handler cannot be used. Currently it uses onHiding as the event to pick up on however there is no way to stop the window from being hidden. From looking at the code it might be best to implement something in the windowController or override its onCloseClicked method however the windowController is defined in your fxStage and wanted to know if you had any ideas and if we should implement our own windowFXML as there is an option to specify the fxml used.

Sheikah45:
anything to do here?

src/main/java/com/faforever/client/FafClientApplication.java
```diff
@@ -125,6 +136,25 @@ private void showMainWindow(FxStage fxStage) {
...
+    if (applicationContext.getBean(GameService.class).isGameRunning()) {
+      I18n i18n = applicationContext.getBean(I18n.class);
+      NotificationService notificationService = applicationContext.getBean(NotificationService.class);
+      notificationService.addNotification(new ImmediateNotification(i18n.get("exitWarning.title"),
+          i18n.get("exitWarning.message"),
+          Severity.WARN,
+          asList(
+              new Action(i18n.get("yes"), ev -> {
+                Window window = (Window) event.getSource();
+                window.hide();
```
1-alex98: Platform.exit()

1-alex98 (changes requested):

1-alex98 (approved):

1-alex98:
Remember me if CI ran through

Sheikah45:
CI passed

PR #1909, opened 2020-09-14
Correct map preview path

### LINKED ISSUE #1908: Map Preview images not show
Wrong path causes map previews to be missing for not installed maps...

**OS**
Linux or Windows

Visit Issue hunt...

1-alex98:
Sets the new path in the config
Fixes #1908


Sheikah45 (approved):
Works for me

Brutus5000 (approved):
Looks good.
However, for the longterm, I discussed the topic with Downlord and from an architectural perspective we should get rid of hardcoded content-server urls and offer a way to ask for the right path from the API. The server file structure should not influence the client code.

PR #1919, opened 2020-09-17
Fix Load Ladder Maps

### LINKED ISSUE #1918: Show ladder 1v1 maps only works the second time you click on it
**Describe the bug**
The loading of the starting screen of the map vault covers the 1v1 maps if the vault was not loaded before the button clcik

it used to work but we must have broken it when @Sheikah45 and I refactored the vault

Visit Issue hunt...

Sheikah45:
Fixes #1918 

Removes the change listener to the state as that seemed to not do anything and was overridden by the loadshowroom.
Does not cause the show user replays to break.

### DIFF
--- a/src/main/java/com/faforever/client/vault/VaultEntityController.java
+++ b/src/main/java/com/faforever/client/vault/VaultEntityController.java
@@ -18,8 +18,6 @@ import javafx.beans.binding.Bindings;
 import javafx.beans.binding.BooleanBinding;
 import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.SimpleObjectProperty;
-import javafx.beans.value.ChangeListener;
-import javafx.beans.value.ObservableValue;
 import javafx.collections.ObservableList;
 import javafx.scene.Node;
 import javafx.scene.control.Button;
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/vault/VaultEntityController.java
```diff
@@ -327,20 +325,7 @@ protected void onBackButtonClicked() {
...
-            if (newValue != State.UNINITIALIZED) {
-              state.removeListener(this);
-              handleSpecialNavigateEvent(navigateEvent);
-            }
-          }
-        });
-        loadShowRoom();
-      } else {
-        handleSpecialNavigateEvent(navigateEvent);
-      }
```
1-alex98, line 343: Well I think you fixed it but now if you click the back button after the search it showroom is empty. That is why this code existed. Maybe easier would be check if state is uninitialized on back button click and call loadShowRoom then

1-alex98 (changes requested):

src/main/java/com/faforever/client/vault/VaultEntityController.java
Sheikah45: I don't get this error the show room always populates for me with the back button. How are you getting an empty showroom?

Sheikah45: In the code the back button always loads the showroom

1-alex98: Was just an assumption because in the old code that used to be the case

Sheikah45: yeah if you test it and it happens let me know cause it isnt happening for me

1-alex98 (approved):

PR #1924, opened 2020-09-18
Provide reload improvements

### LINKED ISSUE #1917: Live reload in development
My dream is to not need to restart the client after a made changes. There are ways do do this. One could make some code that reloads the complete tab if clicked on play, chat etc. (only for development). Then it needs hot reload of java classes but I would already pleased if it worked only with the theme files. Since the already have a theming support we should be able to built that easily, but maybe Intelij has some functionality on its own for this maybe using that would be interesting. Also Spring has a thing called Spring Developer tools that are supposed to enable live reload, not sure of those are suitable for us. Anyways researching this topic would be interesting.

Visit Issue hunt...

1-alex98:
https://youtu.be/YRedZr6zzbM

Fixes #1917

### DIFF
--- a/.idea/runConfigurations/Main.xml
+++ b/.idea/runConfigurations/Main.xml
@@ -4,7 +4,7 @@
       <env name="SPRING_PROFILES_ACTIVE" value="prod" />
     </envs>
     <option name="MAIN_CLASS_NAME" value="com.faforever.client.Main" />
-    <module name="downlords-faf-client_main" />
+    <module name="downlords-faf-client.main" />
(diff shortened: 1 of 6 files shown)


1-alex98:
So I guess I gotta fix related tests

Sheikah45:
I am hoping to try this out today

Sheikah45:
going to rebase this

1-alex98:
@Sheikah45 should we merge this? I am not sure if it works flawlessly but it should not effect production

1-alex98:
Maybe I should not file watch the default theme in production

Sheikah45:
It worked okay for me for the fxml and others the css was still a little wonky but think it is worth merging as long as production doesn't break

Sheikah45 (dismissed):

Sheikah45 (approved):

PR #1940, opened 2020-09-24
New league leaderboard

BlackYps:
This is still work in progress, but comments are appreciated. At the moment I hacked in some values, because the api for the new data doesn't exist yet. This will obviously change later.
grafik


micheljung:
I like where this is heading 👍

1-alex98:
Please make a draft instead of a pull request

PR #1949, opened 2020-09-25
Issue/#1770 use gametype for determining which games to show

### LINKED ISSUE #1770: TMM: Distinguish between automatch and custom lobby games
Right now there is special handling for ladder games based on which featured mod the game is using. This needs to be extended to all matchmaker games in general. Since the plan is for matchmaker games to be played on an arbitrary featured mod, we need an additional property that determines if and where the lobby is displayed in the client.

~We'll need to add `init_mode` to the `game_info` message sent by the server, and replace special logic which checks for the ladder featured mod to check for the init mode instead.~

We'll need to add `game_type` to the `game_info` message sent by the server. See the PR https://github.com/FAForever/server/pull/635.

https://github.com/FAForever/server/issues/605

Visit Issue hunt...

Askaholic:
Any games with the `custom` game type will show up in the custom games tab regardless of featured mod, same with game type `coop` and the coop tab. Any games with the `matchmaker` game type will not generate friend notifications and also shouldn't let you join them in the context menu, but I couldn't test that since the context menu is broken on this branch.

Closes #1770


src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
```diff
@@ -201,15 +201,12 @@ public void setChatUser(ChatChannelUser chatUser) {
...
-      // TODO: Make this ignore TMM games too and not just ladder
-      // https://github.com/FAForever/downlords-faf-client/issues/1770
       joinGameItem.visibleProperty().bind(newValue.socialStatusProperty().isNotEqualTo(SELF)
           .and(newValue.statusProperty().isEqualTo(PlayerStatus.LOBBYING)
               .or(newValue.statusProperty().isEqualTo(PlayerStatus.HOSTING)))
           .and(Bindings.createBooleanBinding(() -> {
                 return newValue.getGame() != null
-                    && newValue.getGame().getFeaturedMod() != null
-                    && !newValue.getGame().getFeaturedMod().equals(KnownFeaturedMod.LADDER_1V1.getTechnicalName());
+                    && newValue.getGame().getGameType() != GameType.MATCHMAKER;
```
1-alex98, line 209: Just ladder not be excluded as well

src/main/java/com/faforever/client/remote/domain/GameType.java
```diff
@@ -0,0 +1,47 @@
...
+
+public enum GameType {
+
+  UNKNOWN("unknown"),
+  CUSTOM("custom"),
+  MATCHMAKER("matchmaker"),
+  COOP("coop"),
+  TUTORIAL("tutorial");
+
+  private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
1-alex98: @SlF4j not working on enums?

src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
Askaholic: Ladder falls into the category of matchmaker games

src/main/java/com/faforever/client/remote/domain/GameType.java
Askaholic: No idea I just copy and pasted from GameStatus

1-alex98: Try use the annotation

1-alex98:
Can I test this with the server on productin?

Askaholic:
Yes

src/test/java/com/faforever/client/game/GameBuilder.java
```diff
@@ -34,6 +35,7 @@ public GameBuilder defaultValues() {
     game.setMaxRating(800);
     game.setMaxRating(1300);
```
dev-nulled, line 36: This dupe a typo? The 800 will do nothing.  That a 'min' rating perhaps?

1-alex98: This code is only used for testing. But yes this is probably a typo

1-alex98: Make a PR if you like XD @dev-nulled heard u get t-shirts for those

Askaholic: Sadly this one was opened too early so it doesn’t count :P

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
```diff
@@ -297,6 +297,7 @@ private void updateLobbyModeFromGameInfo(GameLaunchMessage gameLaunchMessage) {
       return;
     }
 
+    // TODO: Remove this?
```
1-alex98: I think it should be if(Matchmaker) right? This is needed though I think. Can u check this?

1-alex98 (changes requested):
Could u research that? I think all matchmaker games don't have a lobby hence they should all me autolobby.

1-alex98:
Otherwise tested and works

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
Askaholic: Yea that's what I'm thinking too. We could just get rid of the init mode entirely, and always use autolobby for matchmaker games

1-alex98: Yes can u do so @Askaholic and then I merge this...

Askaholic: Ok, will do

1-alex98:
Also needs rebase

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
Askaholic: So I looked into this, the problem is we didn't add game_type to the game_launch message, so it's not available in this function. I think it's ok to have the extra check for ladder game mode since it would only apply if for some reason the server didn't send the init mode (shouldn't happen, but it doesn't really hurt to have the code in there).

1-alex98 (approved):
I can merge that right?

Askaholic:
Yes

Sheikah45:
Is this all done?

Askaholic:
Yes

Sheikah45 (approved):
Looks good just waiting on checks

PR #1950, opened 2020-09-25
Add back context menu to chat

1-alex98:


1-alex98:
Guess @micheljung already pushed to develop

Issue #1952, opened 2020-09-26
Rename from "Downlord's FAF Client" to "FAF Client"


micheljung:
Remove "Downlord" from the installation directory and application name in preparation for https://github.com/FAForever/downlords-faf-client/issues/1233

tatsujb:
finally!! :D

Sheikah45:
Closed with #2739

PR #1953, opened 2020-09-26
Change installation directory to LocalAppData

### LINKED ISSUE #1951: Install client into LOCALAPPDATA
In preparation to https://github.com/FAForever/downlords-faf-client/issues/1753, change the installation directory to local AppData.

micheljung:
Fixes #1951, #1952

### DIFF
--- a/downlords-faf-client.install4j
+++ b/downlords-faf-client.install4j
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<install4j version="8.0.8" transformSequenceNumber="8">
+<install4j version="8.0.7" transformSequenceNumber="8">
   <directoryPresets config="./src/media/appicon" />
-  <application name="Downlord's FAF Client" applicationId="2848-7798-9769-5013" mediaDir="./build/install4j" lzmaCompression="true" pack200Compression="true" shortName="dfc" publisher="Downlord" publisherWeb="https://github.com/FAForever/downlords-faf-client" allPathsRelative="true" autoSave="true" macVolumeId="e230e730a6fc064c" javaMinVersion="14" javaMaxVersion="14" allowBetaVM="true">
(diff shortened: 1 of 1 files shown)


1-alex98:
How this behave then? Does it still uninstall old version does it read old installer path?

1-alex98:
Did you try what Ingo said https://stackoverflow.com/questions/64077801/how-to-change-installation-directory-in-install4j

1-alex98:
@micheljung should I have a look instead

KionX:
Why LocalAppData?
Why not %USERPROFILE% or %PROGRAMFILES% ?

micheljung:
> Why LocalAppData? Why not %USERPROFILE% or %PROGRAMFILES% ?

`%USERPROFILE%` is a parent directory of `%LOCALAPPDATA%`. Moreover, it's not meant to directly contain installed applications.
`%PROGRAMFILES` is where the client is currently installed into and evaluated permissions are required in order to do so. This continues to cause issues and prevents auto-updating.

`%LOCALAPPDATA%` is the right place for user- and computer-specific applications. Chances are you are using Google Chrome, which is installed there, too.

KionX:
Applications are not installed in LOCALAPPDATA. Only their settings.
USERPROFILE is easier for the user to find.
Google Chrome is installed in PROGRAM FILES.
I think LOCALAPPDATA is an unfortunate choice.

Sheikah45:
One thing to note on this is that the installer no longer requires installation into program files. if a user doesnt have or want to elevate permissions it will install into local app data or they can select the installation location.

Not sure if this changes the status of this pr

micheljung:
> Applications are not installed in LOCALAPPDATA. Only their settings.

Wrong, it's even the [default folder of the Windows Installer](https://docs.microsoft.com/en-us/windows/win32/msi/installation-context) for per-user installations:
image

> Google Chrome is installed in PROGRAM FILES.

If you install it "per user" it's certainly in `%LOCALAPPDATA%`:
image

Just like Discord:
image

Even Microsoft VS Code:
image

Probably `%LOCALAPPDATA%\Programs` is the best choice, but that's still `%LOCALAPPDATA%`

> USERPROFILE is easier for the user to find.

*That* would be an unfortunate choice because it's, as the name suggests, for the user's Profile and not for applications. If "easier for the user to find" was the criteria to select the best directory, then we would install it into `Desktop`. But there are more important things to consider, like access permissions. And that's the main reason why we should no longer install to `%ProgramFiles%` and there is no better location than `%LOCALAPPDATA%`(`/Programs`)

Sheikah45:
stale

PR #1964, opened 2020-09-29
Code signing

1-alex98:
if no CERT_PASSWORD is set or it is empty the executable is not signed...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -175,6 +175,12 @@ task buildInstall4jMediaFiles(type: Install4jTask) {
   group "distribution"
   description 'Creates install4j media files'
   dependsOn 'installDist'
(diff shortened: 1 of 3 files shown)


1-alex98:
Can somebody see what happens if u download the self signed release? If it works out ok we might as well self sign the exe for the time being

Sheikah45:
When I run it the windows defender still warns me but I do see the publisher but still have to click the advanced settings to install

1-alex98:
but it did not get worse right?

Sheikah45:
correct it did not get worse

1-alex98:
So we can also self sign for the time being

1-alex98:
Let's keep it for the time being

Sheikah45:
stale

PR #1971, opened 2020-10-03
Randomize maps in top picks show room on load

### LINKED ISSUE #1970: Randomize maps shown in M&M top Picks show room
Currently on the main page of the maps vault showroom you see the same 7 maps all the time for the more static pages like M&M top picks. It would be nice if these maps rotated so that users are exposed to more maps without needing to click through.

Requested by biass

On every page reload a different set of 7 maps is shown for M&M top picks

Visit Issue hunt...

Sheikah45:
Fixes #1970

### DIFF
--- a/src/main/java/com/faforever/client/map/MapVaultController.java
+++ b/src/main/java/com/faforever/client/map/MapVaultController.java
@@ -42,7 +42,7 @@ public class MapVaultController extends VaultEntityController<MapBean> {
   private final EventBus eventBus;
 
   private MapDetailController mapDetailController;
-  private int recommendedShowRoomPageCount = 1;
+  private Integer recommendedShowRoomPageCount;
 
   public MapVaultController(MapService mapService, I18n i18n, EventBus eventBus, PreferencesService preferencesService,
                             UiService uiService, NotificationService notificationService, ReportingService reportingService) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/map/MapVaultController.java
```diff
@@ -41,6 +42,7 @@
   private final EventBus eventBus;
 
   private MapDetailController mapDetailController;
+  private int recommendedShowRoomPageCount = 1;
```
1-alex98: init with 0

1-alex98: no better make it an integer, so it can be null. Let's make it so the vault does not break if something is misconfigured in the remote preferences

src/main/java/com/faforever/client/map/MapVaultController.java
```diff
@@ -104,8 +95,16 @@ protected Node getEntityCard(MapBean map) {
 
   @Override
   protected List<ShowRoomCategory> getShowRoomCategories() {
+    Random random = new Random();
+    int recommendedPage;
+    if (recommendedShowRoomPageCount > 0) {
```
1-alex98: null check if null just not set the showRoomCategory I would say

1-alex98 (changes requested):

src/main/java/com/faforever/client/map/MapVaultController.java
```diff
@@ -104,8 +95,16 @@ protected Node getEntityCard(MapBean map) {
 
   @Override
   protected List<ShowRoomCategory> getShowRoomCategories() {
```
1-alex98, line 102: can u try test that? just have a look getRecommendedMapsWithPageCount is called with a valid page is called

src/main/java/com/faforever/client/map/MapVaultController.java
Sheikah45: Yeah I can make it null and add a null check but will still have to check for zero too since random.nextInt needs positive

src/main/java/com/faforever/client/map/MapVaultController.java
Sheikah45: It doesn't actually call the function unless the showroom is loaded but I will see if I can mock it and test

1-alex98 (approved):

PR #1973, opened 2020-10-04
Add league leaderboards

BlackYps:
This is still work in progress, but comments are appreciated. At the moment I hacked in some values, because the api for the new data doesn't exist yet. This will obviously change later.
grafik


src/main/java/com/faforever/client/api/dto/DivisionLeaderboardEntry.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Setter;
+
+@Getter
+@Setter
+@EqualsAndHashCode(of = "id")
+@Type("divisionLeaderboardEntry")
+public class DivisionLeaderboardEntry {
+  @Id
+  private String id;
+  private int rank;
```
1-alex98: is that not an enum?

src/main/java/com/faforever/client/leaderboard/Division.java
```diff
@@ -0,0 +1,14 @@
+package com.faforever.client.leaderboard;
+
+import com.faforever.client.api.dto.DivisionName;
+import lombok.Data;
+
+@Data
```
1-alex98: @Value instead and remove privat and final

src/main/java/com/faforever/client/leaderboard/DivisionStat.java
```diff
@@ -0,0 +1,8 @@
+package com.faforever.client.leaderboard;
+
+import lombok.Data;
+
+@Data
```
1-alex98: @Value

1-alex98 (changes requested):

1-alex98:
And a lot more

Sheikah45:
@BlackYps are you still working on this?

BlackYps:
Yes, it is mostly ready, but it's waiting for api support and that took longer than expected.
Also need some tests still.

BlackYps:
All the functionality is now there so I only expect the api queries to maybe change to adapt to the api endpoints.
I am aware that there are still tests missing and there is still some commented out code that I will remove later when I can test the client with real data from the api.
I still want to give you an opportunity to review this already, because as it looks now the api will be ready very soon, and this pull request is rather large.

Brutus5000:
FYI: The server side implementation is broken and I'm still waiting for a fix.

BlackYps:
If you mean that the league service can't parse the game result messages, then we are working on it.

src/main/java/com/faforever/client/domain/LeagueEntryBean.java
```diff
@@ -0,0 +1,96 @@
...
+import lombok.ToString;
+import lombok.Value;
+
+@EqualsAndHashCode(callSuper = true, onlyExplicitlyIncluded = true)
+@ToString(onlyExplicitlyIncluded = true, callSuper = true)
+@Value
+public class LeagueEntryBean extends AbstractEntityBean<LeagueEntryBean> {
+  @ToString.Include
+  @EqualsAndHashCode.Include
+  ObjectProperty<PlayerBean> player = new SimpleObjectProperty<>();
```
Sheikah45: Why include the player in the equals? Shouldn't the id be enough?

src/main/java/com/faforever/client/domain/SubdivisionBean.java
```diff
@@ -0,0 +1,141 @@
...
+
+  public String getMediumImageKey() {
+    return String.format("https://content.faforever.com/divisions/icons/medium/%s%s_medium.png",
+        getDivision().getNameKey(), getNameKey());
+  }
+
+  public String getSmallImageKey() {
+    return String.format("https://content.faforever.com/divisions/icons/small/%s%s_small.png",
+        getDivision().getNameKey(), getNameKey());
+  }
```
Sheikah45: These should be client properties or better yet the urls should be served by the api so if they ever change a client release isnt needed to fix it

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+  @VisibleForTesting
+  protected AutoCompletionBinding<String> usernamesAutoCompletion;
+
+  @Override
+  public void initialize() {
+    scoreLabel.setText(i18n.get("leaderboard.score").toUpperCase());
+    searchTextField.setPromptText(i18n.get("leaderboard.searchPrompt").toUpperCase());
+
+    contentPane.managedProperty().bind(contentPane.visibleProperty());
+    connectionProgressPane.managedProperty().bind(connectionProgressPane.visibleProperty());
```
Sheikah45: use JavaFXUtil.bindManagedToVisible

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+        }
+      }
+      if (foundPlayer == null) {
+        for (LeagueEntryBean leagueEntry : ratingTable.getItems()) {
+          if (leagueEntry.getPlayer().getUsername().toLowerCase().contains(searchText.toLowerCase())) {
+            foundPlayer = leagueEntry;
+            break;
+          }
+        }
+      }
```
Sheikah45: Probably better to just bind an autocompletion to the search box and then work on exact matches

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+    }
+  }
+
+  private void searchInAllDivisions(String searchText) {
+    playerService.getPlayerByName(searchText).thenAccept(playerBeanOptional -> playerBeanOptional.ifPresent(player ->
+      leaderboardService.getLeagueEntryForPlayer(player, season.getId()).thenAccept(leagueEntry -> JavaFxUtil.runLater(() -> {
+        if (leagueEntry == null) {
+          return;
+        }
```
Sheikah45, line 129: This should be outside the JavaFXUtil so we don't queue unnecessarily

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+            .filter(item -> item.getDivision().getIndex() == leagueEntry.getSubdivision().getDivision().getIndex())
+            .findFirst().ifPresent(item -> majorDivisionPicker.getSelectionModel().select(item));
+        subDivisionTabPane.getTabs().stream()
+            .filter(tab -> tab.getUserData().equals(leagueEntry.getSubdivision().getIndex()))
+            .findFirst().ifPresent(tab -> {
+              subDivisionTabPane.getSelectionModel().select(tab);
+              TableView<LeagueEntryBean> newTable = (TableView<LeagueEntryBean>) tab.getContent();
+              newTable.scrollTo(leagueEntry);
+              newTable.getSelectionModel().select(leagueEntry);
+            });
```
Sheikah45: I would be more specific with the JavaFxUtil.runLater wrapping. Wrap it just around the lambdas of the ifPresent

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+      }))));
+  }
+
+  public void setSeason(LeagueSeasonBean season) {
+    this.season = season;
+
+    seasonLabel.setText(i18n.getOrDefault(season.getNameKey(), String.format("leaderboard.season.%s", season.getNameKey())).toUpperCase());
```
Sheikah45: I don't get the default value here. Why not just use the name key directly? It looks like you are formatting it into another key. But that will be the text that is just displayed directly

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+        CompletableFuture<List<LeagueEntryBean>> future = leaderboardService.getEntries(subdivision);
+        futures.add(future.thenAccept(leagueEntries::addAll));
+      });
+      CompletableFuture.allOf(futures.toArray(new CompletableFuture<?>[0])).thenRun(() -> {
+        usernamesAutoCompletion = TextFields.bindAutoCompletion(searchTextField,
+            leagueEntries.stream().map(leagueEntryBean -> leagueEntryBean.getPlayer().getUsername()).collect(Collectors.toList()));
+        usernamesAutoCompletion.setDelay(0);
+      });
+      contentPane.setVisible(true);
+    });
```
Sheikah45: You are making calls to the api and creating futures on the application thread. These need to be taken out of the runLater.

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+  }
+
+  @Override
+  public Tab getRoot() {
+    return leaderboardRoot;
+  }
+
+  private void setCurrentPlayer(PlayerBean player) {
+    playerRatingListener = observable -> updateDisplayedPlayerStats(player);
+    JavaFxUtil.addAndTriggerListener(player.getLeaderboardRatings(), new WeakInvalidationListener(playerRatingListener));
```
Sheikah45: playerRatingListener will be garbage collected right away if defined like this

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+              placementLabel.setVisible(true);
+              placementLabel.setText(i18n.get("leaderboard.placement", leagueEntry.getGamesPlayed(), season.getPlacementGames()));
+            });
+            selectHighestDivision();
+          } else {
+            JavaFxUtil.runLater(() -> {
+              playerDivisionNameLabel.setVisible(true);
+              placementLabel.setVisible(false);
+              playerDivisionImageView.setImage(assetService.loadAndCacheImage(
+                  noCatch(() -> new URL(leagueEntry.getSubdivision().getImageKey())), Paths.get("divisions"), null));
```
Sheikah45: I am looking to remove the usage of noCatch as an FYI

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+          }
+        }
+        plotDivisionDistributions(divisions, leagueEntry);
+      }).exceptionally(throwable -> {
+        log.warn("Error while fetching leagueEntry", throwable);
+        return null;
+      })
+    ).exceptionally(throwable -> {
+      log.warn("Error while loading division list", throwable);
+      return null;
```
Sheikah45: Wouldn't we want to notify the user?

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
Sheikah45: You are loading an image on the application thread should be moved out of the runLater

Sheikah45: Also use Path.of

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+          .findFirst().ifPresent(item -> majorDivisionPicker.getSelectionModel().select(item));
+      onMajorDivisionPicked();
+      subDivisionTabPane.getTabs().stream()
+          .filter(tab -> tab.getUserData().equals(leagueEntry.getSubdivision().getIndex()))
+          .findFirst().ifPresent(tab -> {
+            subDivisionTabPane.getSelectionModel().select(tab);
+            TableView<LeagueEntryBean> newTable = (TableView<LeagueEntryBean>) tab.getContent();
+            newTable.scrollTo(leagueEntry);
+            newTable.getSelectionModel().select(leagueEntry);
+      });
```
Sheikah45: Same thing about being minimal with what we put into JavaFXUtil.runlater

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+    });
+  }
+
+  private void selectHighestDivision() {
+    JavaFxUtil.runLater(() -> {
+      majorDivisionPicker.getSelectionModel().selectLast();
+      onMajorDivisionPicked();
+      subDivisionTabPane.getSelectionModel().selectLast();
```
Sheikah45: onMajorDivision picked calls the api so should not be in a runLater

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+          subDivisionTabPane.getTabs().clear();
+          subdivisions.stream()
+              .filter(subdivision -> subdivision.getDivision().getIndex() == majorDivisionPicker.getValue().getDivision().getIndex())
+              .forEach(subdivision -> {
+                SubDivisionTabController controller = uiService.loadFxml("theme/leaderboard/subDivisionTab.fxml");
+                controller.getRoot().setUserData(subdivision.getIndex());
+                controller.populate(subdivision);
+                subDivisionTabPane.getTabs().add(controller.getRoot());
+                subDivisionTabPane.getSelectionModel().selectLast();
+              });
```
Sheikah45: Same be minimal about what is put into runLater

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -30,6 +46,20 @@
...
+      return 1;
+    }
+    if (o1.getDivision().getIndex() == o2.getDivision().getIndex() && o1.getIndex() > o2.getIndex()) {
+      return 1;
+    }
+    if (o1.getDivision().getIndex() == o2.getDivision().getIndex() && o1.getIndex() == o2.getIndex()) {
+      return 0;
+    }
+    return -1;
+  };
```
Sheikah45:
```suggestion
  private static final Comparator<SubdivisionBean> SUBDIVISION_COMPARATOR = Comparator.<SubdivisionBean, Integer>comparing(subdivision -> subdivision.getDivision().getIndex()).thenComparing(SubdivisionBean::getIndex);
```

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -61,4 +91,135 @@
...
+        .toFuture();
+  }
+
+  @Cacheable(value = CacheNames.LEAGUE_ENTRIES, sync = true)
+  public CompletableFuture<Optional<LeagueEntryBean>> getHighestLeagueEntryForPlayer(PlayerBean player) {
+    ElideNavigatorOnCollection<LeagueSeasonScore> navigator = ElideNavigator.of(LeagueSeasonScore.class).collection()
+        .setFilter(qBuilder().intNum("loginId").eq(player.getId()));
+    return fafApiAccessor.getMany(navigator)
+        // This goes all the way back. Maybe we don't want that
+        .filter(leagueSeasonScore -> leagueSeasonScore.getLeagueSeason().getStartDate().isBefore(OffsetDateTime.now()))
```
Sheikah45: Couldn't this filter be included in the navigator?

Sheikah45 (changes requested):
Got part way through so far but didn't want to chance these messages disappearing.

I noticed there are quite a few instances where the api is called or files are loaded on the application thread. Those should be moved off the application thread to prevent ui lag. Also generally we want to put only the operations that actually change the current UI inside of the runLater calls

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: When I do that the second statement seems to take effect before the first one, thus breaking the functionality. Is there a way to enforce execution order when using multiple runLater calls?

Sheikah45: The runLater's are executed in the order that they are called so if they are not executed in the right order something else is going on

Sheikah45: Also I am not sure why the execution order even matters here as it seems like you are selecting the tabs independent of each other

BlackYps: when I select an item in the majorDivisionPicker it automatically runs onMajorDivisionpicked. That one clears the tabs and adds new ones. Only after that has happened I can pick the correect tab. If that executes too early it might run with the old tabs and then it obviously can't find the league entry. I am not sure how I can best wait for the application thread.

Sheikah45: Well that creates an additional issue somewhat as well as now we are executing the api call on the application thread as well in onMajorDivisonPicked

BlackYps: I refactored onMajorDivisionPicked already so that the api call is outside of the runLater wrappers.

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,400 @@
...
+        }));
+    // The tabs always appear 40px wider than they should for whatever reason. We add a little more to prevent horizontal scrolling
+    JavaFxUtil.runLater(() -> subDivisionTabPane.setTabMinWidth(subDivisionTabPane.getWidth() / subDivisionTabPane.getTabs().size() - 45.0));
+    JavaFxUtil.runLater(() -> subDivisionTabPane.setTabMaxWidth(subDivisionTabPane.getWidth() / subDivisionTabPane.getTabs().size() - 45.0));
```
Sheikah45: These should be wrapped into a single runLater it is better to batch the calls

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
Sheikah45: Alright then yeah the issue is that since the calls there depend on the api call being returned it will likely be executed later. I believe this would also be an issue even when everything is wrapped in the runLater unless that forces the api call to also be run in the api thread.

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -61,4 +91,135 @@
...
+  @Cacheable(value = CacheNames.DIVISIONS, sync = true)
+  public CompletableFuture<List<LeagueBean>> getLeagues() {
+    ElideNavigatorOnCollection<League> navigator = ElideNavigator.of(League.class).collection();
+    return fafApiAccessor.getMany(navigator)
+        .map(dto -> leaderboardMapper.map(dto, new CycleAvoidingMappingContext()))
+        .collectList()
+        .toFuture();
+  }
+
+  @Cacheable(value = CacheNames.DIVISIONS, sync = true)
```
Sheikah45: In the leaderboardService you have an issue with your caches. The Divisions cache is used on two methods, getLatestSeason and getAllSubdivisions, that both have an int as the only parameter these will then get confused so the cache will return the wrong object
The caches only use the parameters as the cache key by default. It doesn't actually cache on a per method basis

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: Maybe this is why it is working now. To make it work properly I would need some sort of callback when onMajorDivisionpicked finishes so I can then select the correct tab. Maybe it is possible with smart usage of event handlers or listeners, but the java documentation is confusing me.

src/main/java/com/faforever/client/domain/SubdivisionBean.java
BlackYps: How likely is that? As long as the folder structure remains we can point to new files by providing a different nameKey

Sheikah45: Well the point would be to enable changes to the folder structure easily. Just like how we do with replays and other things. Paths like this should be dynamically configured since we don't control the source. Just makes our lives easier down the line.

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: Got it to work with the help of a global variable

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: the autocompletion is already there. You mean this is too much effort and we should just delete the logic for partial matches?

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: Why? It is not a local variable

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: The default value is actually the first one. Without a default value we always need to update the season names in time or the whole leaderboard will error out once an unknown season has begun.

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
Sheikah45: Yes I think so because with autocompletion there isn't much need for partial matching

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
Sheikah45: Oh yeah then I was just confused that the the default was the first one but I remember now due to the bar args my bad

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
Sheikah45: You are right I read it wrong

BlackYps:
lol, on my machine all tests pass

src/main/java/com/faforever/client/domain/SubdivisionBean.java
BlackYps: I'll make the league service provide the path, but it is going to take a while to merge and deploy the changes in all needed repositories.

BlackYps:
The 6 remaining code issues look like false positives to me

src/main/java/com/faforever/client/domain/DivisionBean.java
```diff
@@ -0,0 +1,71 @@
...
+
+@EqualsAndHashCode(callSuper = true, onlyExplicitlyIncluded = true)
+@ToString(callSuper = true, onlyExplicitlyIncluded = true)
+@Value
+public class DivisionBean extends AbstractEntityBean<DivisionBean> {
+  StringProperty descriptionKey = new SimpleStringProperty();
+  IntegerProperty index = new SimpleIntegerProperty();
+  @ToString.Include
+  StringProperty nameKey = new SimpleStringProperty();
+  ObjectProperty<LeagueSeasonBean> leagueSeason = new SimpleObjectProperty<>();
```
Sheikah45, line 21: Does this not have a technicalName?

BlackYps: yes, it actually doesn't have a technical name

BlackYps:
Closes #2323

BlackYps:
The url concatenation feels very hacky, but I haven't really found a very straightforward way

Sheikah45:
Why do you need to do url concatenation? If the url is coming from the url it should just be the full url. There should be no reason to concatenate.

BlackYps:
I need to add the name of the png file and that is different for every subdivision. The league just provides the folder where the files are located.

Sheikah45:
Why not just make the urls tied to the subdivision objects so this doesn't have to be done?

BlackYps:
It leads to a lot of duplication. That also makes it harder to change the location of all images at once

Sheikah45:
You can make it a generated value in the API that does the url concatenation for you. Will reduce duplicates in the db while allowing the client to a full url

BlackYps:
Neat, similar to how mapversion does it, I suppose. I just had a look, I'll probably have some questions on the way, but I will begin setting it up.

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
```diff
@@ -1,158 +1,104 @@
...
-        leaderboardComboBox.getItems().clear();
-        leaderboardComboBox.setConverter(leaderboardStringConverter());
-        leaderboardComboBox.getItems().addAll(leaderboards);
-        leaderboardComboBox.getSelectionModel().selectFirst();
-      });
+    leaderboardService.getLeagues().thenAccept(leagues -> leagues.forEach(league ->
+        leaderboardService.getLatestSeason(league.getId()).thenAccept(season -> {
+          LeaderboardController controller = uiService.loadFxml("theme/leaderboard/leaderboard.fxml");
+          controller.setSeason(season);
+          controller.getRoot().setText(i18n.get(String.format("leaderboard.%s", league.getTechnicalName())));
```
Sheikah45: This needs a default value provided otherwise the tab won't open as the key may not be defined

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
```diff
@@ -1,158 +1,104 @@
...
+            notificationService.addImmediateErrorNotification(null, "leaderboard.noLeaderboards");
+          }
+          JavaFxUtil.runLater(() -> {
+            leaderboardRoot.getTabs().sort(Comparator.comparing(tab -> (int) tab.getUserData()));
+            leaderboardRoot.getSelectionModel().select(0);
+            lastTab = leaderboardRoot.getTabs().get(0);
+          });
+        }).exceptionally(throwable -> {
+          log.warn("Error while loading seasons", throwable);
+          notificationService.addImmediateErrorNotification(throwable, "leaderboard.noLeaderboards");
```
Sheikah45: This is a misleading error message as it says no leaderboards returned when in fact there was just an error loading them

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
```diff
@@ -1,158 +1,104 @@
...
+          LeaderboardController controller = uiService.loadFxml("theme/leaderboard/leaderboard.fxml");
+          controller.setSeason(season);
+          controller.getRoot().setText(i18n.get(String.format("leaderboard.%s", league.getTechnicalName())));
+          controller.getRoot().setUserData(league.getId());
+          JavaFxUtil.runLater(() -> leaderboardRoot.getTabs().add(controller.getRoot()));
+          controllers.add(controller);
+        }).thenRun(() -> {
+          if (controllers.isEmpty()) {
+            log.info("Api returned no leagues");
+            notificationService.addImmediateErrorNotification(null, "leaderboard.noLeaderboards");
```
Sheikah45: If this has no throwable attached to it use addImmediateWarnNotification as nothing went wrong there are just no leaderboards defined.

Although I might even prefer having a different view displayed if there are no leaderboards defined rather than pooping up a warning message.

Sheikah45: Should this if statement return from the function since no controllers or tabs are defined? Rather than executing the runLater below

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
```diff
@@ -1,158 +1,104 @@
...
+            lastTab = leaderboardRoot.getTabs().get(0);
+          });
+        }).exceptionally(throwable -> {
+          log.warn("Error while loading seasons", throwable);
+          notificationService.addImmediateErrorNotification(throwable, "leaderboard.noLeaderboards");
+          return null;
+        })
+    )).exceptionally(throwable -> {
+      log.warn("Error while loading leagues", throwable);
+      notificationService.addImmediateErrorNotification(throwable, "leaderboard.noLeaderboards");
```
Sheikah45: Same misleading message as above

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
```diff
@@ -1,158 +1,104 @@
...
-
-  public void onLeaderboardSelected() {
-    contentPane.setVisible(false);
-    searchTextField.clear();
-    if (usernamesAutoCompletion != null) {
-      usernamesAutoCompletion.dispose();
+      leaderboardRoot.getSelectionModel().select(lastTab);
+    } finally {
+      isHandlingEvent = false;
     }
```
Sheikah45, line 104: Why is this all wrapped in a try? Also not sure what the isHandlingEvent is needed for as this just sets the lastTab and does not affect the UI directly.

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,369 @@
...
+
+  private void setTabWidth(double tabPaneWidth, int tabNumber) {
+    // The tabs always appear 40px wider, maybe because of some obscure padding rule.
+    // We also need to subtract the tab area padding to prevent horizontal scrolling.
+    double tabWidth = Math.floor((tabPaneWidth - 20.0) / tabNumber) - 40.0;
+    JavaFxUtil.runLater(() -> {
+      subDivisionTabPane.setTabMinWidth(tabWidth);
+      subDivisionTabPane.setTabMaxWidth(tabWidth);
+    });
+  }
```
Sheikah45, line 117: I feel like this is something that should be set in the fxml. Why does the width vary with the number of tabs?

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,369 @@
...
+
+  private void processSearchInput(String searchText) {
+    playerService.getPlayerByName(searchText).thenAccept(playerBeanOptional -> playerBeanOptional.ifPresent(player ->
+      leaderboardService.getLeagueEntryForPlayer(player, season.getId()).thenAccept(leagueEntry -> {
+        if (leagueEntry == null) {
+          return;
+        }
+        selectLeagueEntry(leagueEntry);
+      })));
+  }
```
Sheikah45: Right now this is running for every letter change correct? I think it would be better to make this an onAction rather than a change listener so we are not pinging the api for partial names

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,369 @@
...
+
+  public void setSeason(LeagueSeasonBean season) {
+    this.season = season;
+
+    seasonLabel.setText(i18n.getOrDefault(season.getNameKey(), String.format("leaderboard.season.%s", season.getNameKey())).toUpperCase());
+    String startDate = season.getStartDate().format(DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM));
+    String endDate = season.getEndDate().format(DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM));
+    seasonDateLabel.setText(i18n.get("leaderboard.seasonDate", startDate, endDate));
```
Sheikah45: The TimeService has a function for converting the date according to the user selected date format so we should use that here as well

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,369 @@
...
+          return null;
+        });
+    setListener(playerService.getCurrentPlayer());
+  }
+
+  private void setUsernamesAutoCompletion(List<SubdivisionBean> subdivisions) {
+    JavaFxUtil.runLater(() -> {
+      majorDivisionPicker.getItems().clear();
+      majorDivisionPicker.getItems().addAll(
+          subdivisions.stream().filter(subdivision -> subdivision.getIndex() == 1).collect(Collectors.toList()));
```
Sheikah45, line 163: Not sure why we are only checking for index 1 here

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,369 @@
...
+    });
+    List<LeagueEntryBean> leagueEntries = new ArrayList<>();
+    List<CompletableFuture<?>> futures = new ArrayList<>();
+    subdivisions.forEach(subdivision -> {
+      CompletableFuture<List<LeagueEntryBean>> future = leaderboardService.getEntries(subdivision);
+      futures.add(future.thenAccept(leagueEntries::addAll));
```
Sheikah45: I would chain the thenAccept when you declare future

Sheikah45: Also why not just extract the usernames in the future since that is all we are using them for. Then we can also make it a set to reduce potential list duplication

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,369 @@
...
+
+  private void setListener(PlayerBean player) {
+    playerRatingListener = observable -> updateDisplayedPlayerStats(player);
+    JavaFxUtil.addAndTriggerListener(player.getLeaderboardRatings(), new WeakInvalidationListener(playerRatingListener));
+  }
+
+  @VisibleForTesting
+  protected void updateDisplayedPlayerStats(PlayerBean player) {
+    leaderboardService.getAllSubdivisions(season.getId()).thenAccept(divisions ->
+      leaderboardService.getLeagueEntryForPlayer(player, season.getId()).thenAccept(leagueEntry -> {
```
Sheikah45: From what I can tell we only ever update the displayed stats for the currently logged in user. If this is the case then there is no need to hand the player into this method. We can instead use the playerService.getCurrentPlayer call here directly or set the currentPlayer as a controller property. This will also allow playerRatingListener to be set only once and remove the need for the setListener method

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,369 @@
...
+    } else {
+      majorDivisionPicker.getItems().stream()
+          .filter(item -> item.getDivision().getIndex() == entryToSelect.getSubdivision().getDivision().getIndex())
+          .findFirst().ifPresent(item -> JavaFxUtil.runLater(() -> majorDivisionPicker.getSelectionModel().select(item)));
+    }
+  }
+
+  private boolean correctDivisionSelected() {
+    return entryToSelect.getSubdivision().getDivision().getIndex() == majorDivisionPicker.getValue().getDivision().getIndex();
+  }
```
Sheikah45: Is there some race condition where we are not confident that we picked the correct division?

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,369 @@
...
+    };
+  }
+
+  @NotNull
+  private StringConverter<Number> integerStringConverter() {
+    return new StringConverter<>() {
+      @Override
+      public String toString(Number object) {
+        if(object.intValue()!=object.doubleValue())
+          return "";
```
Sheikah45, line 362: Is this just checking that it is an integer? Also spacing around the equals

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -41,21 +62,134 @@
...
+  public CompletableFuture<List<LeagueBean>> getLeagues() {
+    ElideNavigatorOnCollection<League> navigator = ElideNavigator.of(League.class).collection();
+    return fafApiAccessor.getMany(navigator)
+        .map(dto -> leaderboardMapper.map(dto, new CycleAvoidingMappingContext()))
+        .collectList()
+        .toFuture();
+  }
+
+  @Cacheable(value = CacheNames.LEAGUE, sync = true)
+  public CompletableFuture<LeagueSeasonBean> getLatestSeason(int leagueId) {
```
Sheikah45: The service should just take in the client domain objects so just take in a League here and extract the id in this method

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -41,21 +62,134 @@
...
+    ElideNavigatorOnCollection<LeagueSeason> navigator = ElideNavigator.of(LeagueSeason.class).collection()
+        .setFilter(qBuilder().intNum("league.id").eq(leagueId))
+        .addSortingRule("startDate", false);
+    return fafApiAccessor.getMany(navigator)
+        .filter(leagueSeason -> leagueSeason.getStartDate().isBefore(OffsetDateTime.now()))
```
Sheikah45: Why not add this filter to the query instead?

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -41,21 +62,134 @@
...
+
+  public CompletableFuture<Integer> getPlayerNumberInHigherDivisions(SubdivisionBean subdivisionBean) {
+    AtomicInteger rank = new AtomicInteger();
+    List<CompletableFuture<?>> futures = new ArrayList<>();
+    return getAllSubdivisions(subdivisionBean.getLeagueSeasonId()).thenCompose(divisions -> {
+      divisions.stream()
+          .filter(division -> SUBDIVISION_COMPARATOR.compare(division, subdivisionBean) > 0)
+          .forEach(division -> {
+            CompletableFuture<Integer> future = getSizeOfDivision(division);
+            futures.add(future.thenAccept(rank::addAndGet));
```
Sheikah45: add the thenAccept to the declaration

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -41,21 +62,134 @@
...
+          });
+      return CompletableFuture.allOf(futures.toArray(CompletableFuture[]::new)).thenApply(unused -> rank.get());
```
Sheikah45: unused -> aVoid just to follow convention

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -41,21 +62,134 @@
...
+    });
+  }
+
+  @Cacheable(value = CacheNames.LEAGUE_ENTRIES, sync = true)
+  public CompletableFuture<Integer> getTotalPlayers(int leagueSeasonId) {
```
Sheikah45: Use the season object as input rather than the id

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -41,21 +62,134 @@
...
+            .intNum("leagueSeason.id").eq(leagueSeasonId)
+            .and()
+            .intNum("score").gt(-1))
+        .pageSize(1);
+    return fafApiAccessor.getManyWithPageCount(navigator)
+        .toFuture()
+        .thenApply(Tuple2::getT2);
+  }
+
+  public CompletableFuture<Integer> getSizeOfDivision(SubdivisionBean subdivision) {
```
Sheikah45, line 131: If we are only interested int he count here then we should use the same approach as the getTotalPlayers method and use the pageCount to get the number of entries so that we don't make additional unnecessary calls to enrich the leagueEntrybeans that we aren't going to use

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -41,21 +62,134 @@
...
+    return getEntries(subdivision).thenApply(List::size);
+  }
+
+  @Cacheable(value = CacheNames.LEAGUE_ENTRIES, sync = true)
+  public CompletableFuture<LeagueEntryBean> getLeagueEntryForPlayer(PlayerBean player, int leagueSeasonId) {
```
Sheikah45: use the season object

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -41,21 +62,134 @@
...
+      if (leagueSeasonScores.isEmpty()) {
+        return CompletableFuture.completedFuture(Collections.emptyList());
+      }
+      List<Integer> playerIds = leagueSeasonScores
+          .parallelStream()
+          .map(LeagueSeasonScore::getLoginId)
+          .collect(Collectors.toList());
+      return playerService.getPlayersByIds(playerIds).thenApply(playerBeans ->
+          leagueSeasonScores
+              .parallelStream()
```
Sheikah45: not really a need that I see for the parallel stream here

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -41,21 +62,134 @@
...
+                  playerBeans
+                      .stream()
+                      .filter(playerBean -> playerBean.getId().equals(leagueSeasonScore.getLoginId()))
+                      .map(playerBean -> leaderboardMapper.map(leagueSeasonScore, playerBean, new CycleAvoidingMappingContext())))
+              .collect(Collectors.toList()));
+    });
+  }
+
+  @Cacheable(value = CacheNames.DIVISIONS, sync = true)
+  public CompletableFuture<List<SubdivisionBean>> getAllSubdivisions(int leagueSeasonId) {
```
Sheikah45: Use the league object

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -41,21 +62,134 @@
...
+        .setFilter(qBuilder().intNum("league.id").eq(leagueId))
+        .addSortingRule("startDate", false);
+    return fafApiAccessor.getMany(navigator)
+        .filter(leagueSeason -> leagueSeason.getStartDate().isBefore(OffsetDateTime.now()))
+        .next()
+        .map(dto -> leaderboardMapper.map(dto, new CycleAvoidingMappingContext()))
+        .toFuture();
+  }
+
+  public CompletableFuture<Integer> getPlayerNumberInHigherDivisions(SubdivisionBean subdivisionBean) {
```
Sheikah45: Probably want to cache this

src/main/java/com/faforever/client/mapstruct/LeaderboardMapper.java
```diff
@@ -41,4 +48,41 @@
...
+  @Mapping(target = "leagueLeaderboard", source = "leaderboard")
+  LeagueSeason map(LeagueSeasonBean bean, @Context CycleAvoidingMappingContext context);
+
+  @Mapping(target = "index", source = "divisionIndex")
+  DivisionBean map(LeagueSeasonDivision dto, @Context CycleAvoidingMappingContext context);
+
+  @Mapping(target = "divisionIndex", source = "index")
+  LeagueSeasonDivision map (DivisionBean bean, @Context CycleAvoidingMappingContext context);
+
+  @Mapping(target = "leagueSeasonId", expression = "java(Integer.parseInt(dto.getLeagueSeasonDivision().getLeagueSeason().getId()))")
```
Sheikah45:
```suggestion
  @Mapping(target = "leagueSeasonId", source = "leagueSeasonDivision.leagueSeason.id")
```

Will automatically include null safety in the generated method and convert the types for you

src/main/java/com/faforever/client/player/PlayerInfoWindowController.java
```diff
@@ -401,7 +425,9 @@ public void onRatingTypeChange() {
     if (ratingTypeComboBox.getValue() != null) {
       ratingHistoryChart.setVisible(false);
       loadingHistoryPane.setVisible(true);
-      loadStatistics(ratingTypeComboBox.getValue()).thenRun(() -> JavaFxUtil.runLater(this::plotPlayerRatingGraph));
+      if (player != null) {
+        loadStatistics(ratingTypeComboBox.getValue()).thenRun(() -> JavaFxUtil.runLater(this::plotPlayerRatingGraph));
+      }
```
Sheikah45, line 429: Why would player be null?

src/main/java/com/faforever/client/player/PlayerInfoWindowController.java
```diff
@@ -198,6 +198,15 @@ private void setAvailableAchievements(List<AchievementDefinition> achievementDef
   }
 
   public void setPlayer(PlayerBean player) {
+    if (player.getLeaderboardRatings().isEmpty()) {
+      log.info("no ratings found");
+      updateRatings(player).thenAccept(player1 -> JavaFxUtil.runLater(() -> setOnlinePlayer(player1)));
```
Sheikah45: I would use something like updatedPlayer rather than just player1

src/main/java/com/faforever/client/teammatchmaking/PartyMemberItemController.java
```diff
@@ -137,6 +141,24 @@ private void setPlayerProperties() {
...
+  @VisibleForTesting
+  protected void setLeagueInfo() {
+    leaderboardService.getHighestLeagueEntryForPlayer(player).thenAccept(leagueEntry -> JavaFxUtil.runLater(() -> {
+      if (leagueEntry.isEmpty() || leagueEntry.get().getSubdivision() == null) {
+        leagueLabel.setText(i18n.get("teammatchmaking.inPlacement").toUpperCase());
+        leagueImageView.setVisible(false);
+      } else {
+        leagueLabel.setText(i18n.get("leaderboard.divisionName",
+            i18n.get(leagueEntry.get().getSubdivision().getDivisionI18nKey()),
+            leagueEntry.get().getSubdivision().getNameKey()).toUpperCase());
```
Sheikah45: Need a default here in case the key doesn't exist yet

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingController.java
```diff
@@ -181,13 +181,31 @@ private void setQueueHeadingLabel() {
...
 
+  private void setLeagueInfo() {
+    leaderboardService.getHighestLeagueEntryForPlayer(player).thenAccept(leagueEntry -> JavaFxUtil.runLater(() -> {
+      if (leagueEntry.isEmpty() || leagueEntry.get().getSubdivision() == null) {
+        leagueLabel.setText(i18n.get("teammatchmaking.inPlacement").toUpperCase());
+        leagueImageView.setVisible(false);
+      } else {
+        leagueLabel.setText(i18n.get("leaderboard.divisionName",
+            i18n.get(leagueEntry.get().getSubdivision().getDivisionI18nKey()),
+            leagueEntry.get().getSubdivision().getNameKey()).toUpperCase());
```
Sheikah45: need default here

Sheikah45 (changes requested):
Getting there. Also would like to see full test coverage at least in the leaderboardService

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: Because I want the tabs to always be in total the width of the tabpane. Sometimes it is five, sometimes only one. And generally the number could be arbitrary.

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: Because we want one entry for each major division. This assumes that every subdivision starts with index 1 so we will catch all major divisions

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
Sheikah45: Ah gotcha I thought this was setting the width of the tabpane

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: Yes, with default behaviour it would frequenty put values like 5.5 on the axis which makes no sense when we have integers

Sheikah45: Setting tickUnit in the fxml didn't work?

src/main/java/com/faforever/client/mapstruct/LeaderboardMapper.java
BlackYps: When refactoring the leaderboardService I noticed that I don't really need the leagueSeasonId in the subdivision. I can get it by using the division that is attached to the subdivision

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
BlackYps: This could maybe lead to inconsistencies when the entries in a subdivision change, because the getEntries method needs to use a different cache and we could get a problem with cache timeouts. I think it is still a good idea to cache it and maybe nobody really notices anyway, because you would have to switch division tabs and actively look for inconsistencies.

src/main/java/com/faforever/client/mapstruct/LeaderboardMapper.java
Sheikah45: So this property will just be removed?

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: We also use `getLeagueEntryForPlayer` in `processSearchInput` to then select the right tab and select the entry for the player the user searched for

Sheikah45: Yes there is no need to remove that method just in this case the listener and handing updateDisplayedPlayerStats accepting a player are not needed in the leaderboardController since the value never changes

BlackYps: Oh wait, I missunderstood. Yeah, I guess that makes sense

src/main/java/com/faforever/client/mapstruct/LeaderboardMapper.java
BlackYps: yes, unless you say that you want a reference to the season directly in the subdivisionBean

Sheikah45: Nah just making sure I read it right

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
BlackYps: Getting no leagues from the api means that something is seriously wrong, so I find a warning appropriate. Do you have ideas what we could display on the page if we can't load it normally?

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
BlackYps: I actually have no idea. I mimicked the structure of the onDisplay method of other controllers with subtabs (like the play or replay tab), because I assumed there is a specific reason to do it that way.

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: Where can the user select a date format?

src/main/java/com/faforever/client/player/PlayerInfoWindowController.java
BlackYps: I can't explain it, but when I leave this check out I regularly get NPEs there when opening the info window. It's very weird because the player should get set in line 210.

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
Sheikah45: Yes I just think that the current message could imply that there just are no leagues rather than that something went wrong while loading the leagues.

As far as display maybe something like a simple uh-oh page message perhaps.

This would allow us to refresh the page again at least as well.

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: I didn't know about tickUnit. However it doesn't seem to respect that setting...

BlackYps: I see, I have to disable autoranging. This also means that I manually, but still automatically have to set the upper bound and the tickUnit to sensible levels otherwise this happens:  
grafik
Seems like we need some code for the logic anyway. Which way do you prefer? I think I prefer the string formatter more, because the intention is to get only integers on the axis and the string formatter makes that more clear.

Sheikah45: Yeah then just the string formatter works. Is annoying it is needed like that though

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
BlackYps: The addImmediateWarnNotification displays an uh-oh message. So that is what you had in mind?

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: It's not the first thing that is annoying about javafx :P

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: No, this is a helper method to compare the division index of the entry with the selected one of the division picker. I did this, because the evaluation is very long and we need it to adjust the behaviour accordingly. For example if you searched for someone that is in silver III it will select that tab again if you switched over to gold and back to silver. However we don't want to have gold III selected. It should be the highest division when there is no specific reason to select a different tab.
I'll rename the method to make its use clearer.

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
Sheikah45: Yeah that works as well

Sheikah45:
As an FYI this changes the fonts of all tables which is probably unwanted.
image

Sheikah45:
Also why are we using a new font in the leaderboard tab we should keep it consistent and use one already in the client

BlackYps:
It's not a new font, it has already been used in the matchmaker tab

Sheikah45:
Ah okay I thought it had looked different but it is probably just that it is bigger

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,377 @@
...
+    double tabWidth = Math.floor((tabPaneWidth - 20.0) / tabNumber) - 40.0;
+    JavaFxUtil.runLater(() -> {
+      subDivisionTabPane.setTabMinWidth(tabWidth);
+      subDivisionTabPane.setTabMaxWidth(tabWidth);
+    });
+  }
+
+  public void processSearchInput() {
+    String searchText = searchTextField.getText();
+    if (StringUtils.isEmpty(searchText)) {
```
Sheikah45: can we use isblank here

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
Sheikah45: use TimeService.asDate and it is set in the SettingsController

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,377 @@
...
+    leaderboardService.getAllSubdivisions(season).thenAccept(this::setUsernamesAutoCompletion)
+        .exceptionally(throwable -> {
+          contentPane.setVisible(false);
+          log.warn("Error while loading division list", throwable);
+          notificationService.addImmediateErrorNotification(throwable, "leaderboard.failedToLoadDivisions");
+          return null;
+        });
+
+    playerRatingListener = observable -> updateDisplayedPlayerStats();
+    JavaFxUtil.addAndTriggerListener(playerService.getCurrentPlayer().getLeaderboardRatings(), new WeakInvalidationListener(playerRatingListener));
```
Sheikah45, line 156: This can be move to the initializer

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
Sheikah45: This can be removed

Sheikah45 (changes requested):

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: There are two issues I see with this: The setting is explicitely in the chat settings, so I would expect it to only affect the chat. Secondly, I want the dateformat to be medium (Jan 5, 2021) instead of short (01/05/2021), so I am not convinced that using the timeservice functionality here is a good idea

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: `updateDisplayedPlayerStats` requires the season to be set, otherwise the user will see an error message

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
Sheikah45: That setting is no longer chat only and just needs to be moved in the settings ui. As far as the dateformat length then I would just modify the asDate function to take in the formatStyle as a parameter as well. That way the date formats can be consistent as the user set them according to the locale they want across all times

src/main/java/com/faforever/client/player/PlayerInfoWindowController.java
Sheikah45: Ah it is probably because the rating type gets set before the player is populated in the initialization

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingController.java
```diff
@@ -181,13 +182,32 @@ private void setQueueHeadingLabel() {
...
+        leagueLabel.setText(i18n.get("teammatchmaking.inPlacement").toUpperCase());
+        leagueImageView.setVisible(false);
+      } else {
+        SubdivisionBean subdivision = leagueEntry.get().getSubdivision();
+        leagueLabel.setText(i18n.get("leaderboard.divisionName",
+            i18n.getOrDefault(subdivision.getDivision().getNameKey(), subdivision.getDivisionI18nKey()),
+            subdivision.getNameKey()).toUpperCase());
+        leagueImageView.setImage(assetService.loadAndCacheImage(
+            subdivision.getMediumImageUrl(), Paths.get("divisions"), null
+        ));
```
Sheikah45: Don't want to load and cache images on the application thread

src/main/java/com/faforever/client/teammatchmaking/PartyMemberItemController.java
```diff
@@ -137,6 +142,25 @@ private void setPlayerProperties() {
...
+        leagueLabel.setText(i18n.get("teammatchmaking.inPlacement").toUpperCase());
+        leagueImageView.setVisible(false);
+      } else {
+        SubdivisionBean subdivision = leagueEntry.get().getSubdivision();
+        leagueLabel.setText(i18n.get("leaderboard.divisionName",
+            i18n.getOrDefault(subdivision.getDivision().getNameKey(), subdivision.getDivisionI18nKey()),
+            subdivision.getNameKey()).toUpperCase());
+        leagueImageView.setImage(assetService.loadAndCacheImage(
+            subdivision.getMediumImageUrl(), Paths.get("divisions"), null
+        ));
```
Sheikah45: Don't load and cache on the application thread

src/main/java/com/faforever/client/preferences/ChatPrefs.java
```diff
@@ -97,14 +95,6 @@ public void setTimeFormat(TimeInfo time) {
...
   }
 
-  public DateInfo getDateFormat() {
-    return dateFormat.get();
-  }
-
-  public void setDateFormat(DateInfo date) {
-    this.dateFormat.set(date);
-  }
-
```
Sheikah45: Would be nice if we kept these for one or two release and use them in the migratePreferences method of the preferences service so people don't lose the settings

src/main/java/com/faforever/client/player/PlayerInfoWindowController.java
```diff
@@ -198,6 +198,15 @@ private void setAvailableAchievements(List<AchievementDefinition> achievementDef
   }
 
   public void setPlayer(PlayerBean player) {
+    if (player.getLeaderboardRatings().isEmpty()) {
+      log.info("no ratings found");
```
Sheikah45: This seems more like a debug than info. It also isn't very informative.  I would at least add the name of the player or something to tell us what is causing it or from where. But if this is expected then I am not sure if it is worth logging.

src/main/java/com/faforever/client/leaderboard/SubDivisionTabController.java
```diff
@@ -0,0 +1,94 @@
...
+    subDivisionTab.setText(subdivision.getNameKey());
+
+    leaderboardService.getEntries(subdivision).thenAccept(leagueEntryBeans -> {
+      leaderboardService.getPlayerNumberInHigherDivisions(subdivision).thenAccept(count ->
+          leagueEntryBeans.forEach(entry -> entry.setRank(count + 1 + leagueEntryBeans.indexOf(entry))));
+      ratingTable.setItems(observableList(leagueEntryBeans));
+    }).exceptionally(throwable -> {
+      log.warn("Error while loading leaderboard entries for division " + subdivision, throwable);
+      notificationService.addImmediateErrorNotification(throwable, "leaderboard.failedToLoad");
+      return null;
```
Sheikah45: Is this all on the application thread? If so  it shouldn't be. If not then the subDivisionTab.setText should be wrapped in a runLater

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,380 @@
...
+  @VisibleForTesting
+  protected AutoCompletionBinding<String> usernamesAutoCompletion;
+
+  @Override
+  public void initialize() {
+    scoreLabel.setText(i18n.get("leaderboard.score").toUpperCase());
+    searchTextField.setPromptText(i18n.get("leaderboard.searchPrompt").toUpperCase());
+
+    JavaFxUtil.bindManagedToVisible(contentPane);
+    JavaFxUtil.bindManagedToVisible(connectionProgressPane);
```
Sheikah45:
```suggestion
    JavaFxUtil.bindManagedToVisible(contentPane, connectionProgressPane);
```

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,380 @@
...
+      if (!children.contains(nodeToAdd)) {
+        children.add(nodeToAdd);
+        nodeToAdd.setViewOrder(-0.5);
+      }
+    });
+
+    JavaFxUtil.addListener(node.boundsInParentProperty(), (ov, oldBounds, bounds) -> {
+      nodeToAdd.setLayoutX(Math.round(bounds.getMinX() + bounds.getWidth() / 2 - nodeToAdd.prefWidth(-1) / 2));
+      nodeToAdd.setLayoutY(Math.round(bounds.getMaxY() - nodeToAdd.prefHeight(-1) * 0.5));
+    });
```
Sheikah45, line 338: I would almost make this its own class but maybe not necessary

src/main/java/com/faforever/client/config/CacheConfig.java
```diff
@@ -73,6 +76,9 @@ public CacheManager cacheManager() {
         new CaffeineCache(REPLAYS_RECENT, newBuilder().expireAfterWrite(10, MINUTES).build()),
         new CaffeineCache(MAP_GENERATOR, newBuilder().expireAfterWrite(10, MINUTES).build()),
         new CaffeineCache(LEADERBOARD, newBuilder().expireAfterWrite(5, MINUTES).build()),
+        new CaffeineCache(LEAGUE, newBuilder().expireAfterWrite(10, MINUTES).build()),
+        new CaffeineCache(LEAGUE_ENTRIES, newBuilder().expireAfterWrite(1, MINUTES).build()),
```
Sheikah45, line 80: Why so short for the entries?

Sheikah45 (changes requested):

src/main/java/com/faforever/client/domain/SubdivisionBean.java
```diff
@@ -0,0 +1,156 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+
+import java.net.URL;
+
+@Slf4j
+@EqualsAndHashCode(callSuper = true, onlyExplicitlyIncluded = true)
+@ToString(onlyExplicitlyIncluded = true, callSuper = true)
+@Data
```
Sheikah45: @ Value

Sheikah45 (changes requested):

Sheikah45:
Also in the leaderboard table make sure the columns are wide enough by default so that the text doesn't get shortened in the headers
image

src/main/java/com/faforever/client/preferences/ChatPrefs.java
BlackYps: Do you use some sort of convention to mark things as deprecated in the client?

src/main/java/com/faforever/client/config/CacheConfig.java
BlackYps: The idea was that league entries will change all the time as people play games. The divisions for example are way more static info. If you find 1 minute excessive, I can increase it.

BlackYps:
Huh, on my linux machine it doesn't do that. Seems like there is a tiny difference how the client gets displayed. I'll make sure to check it under windows.

Edit: It must be something else, even on windows it looks normal on my machine. I'll increase the width by 10px, let me know if that fixes it.

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: If we at some point want to have more bar charts like this, then sure. Right now I don't see the use case, so it doesn't seem to be worth the effort

src/main/java/com/faforever/client/preferences/ChatPrefs.java
Sheikah45: @ Deprecated

Sheikah45:
> Edit: It must be something else, even on windows it looks normal on my machine. I'll increase the width by 10px, let me know if that fixes it.

This was taken on my 1920x1080 with the client set to full screen

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
Sheikah45: That makes sense

BlackYps:
I noticed a warning in the log, but I don't understand how to fix it
`
/home/runner/work/downlords-faf-client/downlords-faf-client/src/main/java/com/faforever/client/mapstruct/CycleAvoidingMappingContext.java:42: warning: Lifecycle method has multiple matching parameters (e. g. same type), in this case please ensure to name the parameters in the lifecycle and mapping method identical. This lifecycle method will not be used for the mapping method 'com.faforever.client.domain.LeagueEntryBean map(com.faforever.commons.api.dto.LeagueSeasonScore dto, com.faforever.client.domain.PlayerBean player, @Context com.faforever.client.mapstruct.CycleAvoidingMappingContext context)'
`

src/main/java/com/faforever/client/teammatchmaking/PartyMemberItemController.java
```diff
@@ -137,6 +139,23 @@ private void setPlayerProperties() {
...
+        leagueLabel.setText(i18n.get("teammatchmaking.inPlacement").toUpperCase());
+        leagueImageView.setVisible(false);
+      } else {
+        SubdivisionBean subdivision = leagueEntry.get().getSubdivision();
+        leagueLabel.setText(i18n.get("leaderboard.divisionName",
+            i18n.getOrDefault(subdivision.getDivision().getNameKey(), subdivision.getDivisionI18nKey()),
+            subdivision.getNameKey()).toUpperCase());
+        leagueImageView.setImage(leaderboardService.loadDivisionImage(subdivision.getMediumImageUrl()));
+        leagueImageView.setVisible(true);
+      }
```
Sheikah45, line 155: This is still loading the image inside of a JavaFxutil.runLater

Sheikah45:
> I noticed a warning in the log, but I don't understand how to fix it 

This is because the cycle avoiding context doesn't know where to use the leagueSeasonScore or the PlayerBean as the target. So one would need to be named the same as the parameter to the CycleAvoidingContext

Sheikah45 (changes requested):
See my comment above about the loading the image in the application thread. After that I think we should be pretty close

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -0,0 +1,374 @@
...
+    leaderboardService.getAllSubdivisions(season).thenAccept(this::setUsernamesAutoCompletion)
+        .exceptionally(throwable -> {
+          contentPane.setVisible(false);
+          log.warn("Error while loading division list", throwable);
+          notificationService.addImmediateErrorNotification(throwable, "leaderboard.failedToLoadDivisions");
+          return null;
+        });
+
+    playerRatingListener = observable -> updateDisplayedPlayerStats();
+    JavaFxUtil.addAndTriggerListener(playerService.getCurrentPlayer().getLeaderboardRatings(), new WeakInvalidationListener(playerRatingListener));
```
Sheikah45: Actually one last thing since the playerRatingListener is never changed and takes no input it can be made final and moved to the initializer, and only when the season is set do we add the listener to the leaderboardRatings

Sheikah45:
The last thing I have some issue with is how the league season names are handled. Right now it looks like each season will need a different name key even though the only thing that is changing is the number at the end. Can we just parameterize that so we don't have so many redundant messages in the i18n.

Also the name keys as currently returned from the service are not present

src/main/java/com/faforever/client/teammatchmaking/PartyMemberItemController.java
BlackYps: the other usages of `loadAndCacheImage` that I checked don't bother with loading the image outside of the application thread, so I thought the call to the service might switch the thread back.
So I load the image into a local variable and then only set that in the application thread?

Sheikah45: hmmm yeah I just realized that we force the loadAndCacheImage to load in the background for fetching fresh images but not for loading from cache. If you add the true flag to load from background when loading from cache as well then this should actually be fine

BlackYps:
Ah yes, that is a remnant from my idea to maybe have seasons like "2021 season", but now that we just number them it is not needed.
The keys don't match because the latest version of the league service hasn't been deployed yet. Same with the api and the image urls

Sheikah45:
Even in the case of 2021 season that would just be the word season and some identifier which could be generalized some

BlackYps:
Hm, every season has a unique name key. If we only want the season number, we would have to extract just the last part of the name key, but that seems more like an ugly hack. I guess the proper solution is to add a new column in the league db that explicitly has the number of the season. This would need a new league service release and an api change to cut down the number of messages from 8 to 4. Not sure if it's worth it

Sheikah45:
But that number goes up every single time there is a new season right so right now we have 8 but that will grow with every season

BlackYps:
Hm, that's right.
testShowModDetail() keeps randomly failing. When I rerun the job it's fine. Have you seen that before?

Sheikah45:
There were some issues with this on develop I thought I had fixed

Sheikah45 (approved):
Are we all good here @BlackYps?

Issue #1978, opened 2020-10-09
Filter aeolus by country 


tatsujb:
Someone on reddit was looking for duch people in a certain rating range to play with on faf.

I answered that he could filter aeolus by rating : 
![2jjHe1o_d](https://user-images.githubusercontent.com/5132359/95566416-cded1f80-0a21-11eb-910f-eaf7ea0f7af5.jpg)

(Which I discovered on the spot and thought was really cool!)

And i said for the dutch nationality he could just run through the list he gets and pm those with a dutch flag.

That seemed to me like an underwhelming answer. 

I feel like being able to filter by country would be a nice addition!  :)

chrsin:
Created a pull request for this:
https://github.com/FAForever/downlords-faf-client/pull/1979

PR #1979, opened 2020-10-10
#1978 Implemented a filter for country in the chat window.

chrsin:
Pull request adds country filtering for the chat.

First time contributor here.
I come from a .NET core web development background, so I expect something to be missing or poorly implemented.
Would love feedback! Especially regarding tests. (I'm not sure they ran properly on my machine. Saw a lot of exception messages but the tests seemed to pass anyway.)


tatsujb (dismissed):

src/main/java/com/faforever/client/chat/CountryFlagService.java
```diff
@@ -35,6 +39,24 @@
...
+    startsWith = startsWith.toLowerCase();
+
+    ArrayList<String> result = new ArrayList(countries.length);
+    for(String country : countries)  {
+      if(country.toLowerCase().startsWith(startsWith))
+        result.add(country);
+    }
+
+    return result;
+  }
```
1-alex98: A lot of code for a simple thing:
```suggestion
  return Arrays.stream(Locale.getISOCountries())
        .filter(country -> startsWith == null || country.startsWith(startsWith))
        .collect(Collectors.toList());
```

src/main/java/com/faforever/client/chat/CountryFlagService.java
```diff
@@ -35,6 +39,24 @@
...
+    if(startsWith == null) {
+      return Arrays.asList(countries);
+    }
+
+    startsWith = startsWith.toLowerCase();
+
+    ArrayList<String> result = new ArrayList(countries.length);
+    for(String country : countries)  {
+      if(country.toLowerCase().startsWith(startsWith))
+        result.add(country);
```
1-alex98: for code style reasons we use {} always

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -30,20 +33,24 @@
 public class UserFilterController implements Controller<Node> {
 
   private final I18n i18n;
+  private CountryFlagService flagService;
```
1-alex98: final
```suggestion
  private final CountryFlagService flagService;
```

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -30,20 +33,24 @@
...
   public MenuButton gameStatusMenu;
   public GridPane filterUserRoot;
   public TextField clanFilterField;
   public TextField minRatingFilterField;
   public TextField maxRatingFilterField;
   public ToggleGroup gameStatusToggleGroup;
+  public ComboBox countryFilterField;
```
1-alex98: A ComboBox of what. See how Intelij warns you. In order to achieve type safety please use the Type parameter:
```suggestion
  public ComboBox<String> countryFilterField;
```

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -159,6 +172,27 @@ boolean isGameStatusMatch(ChatChannelUser chatUser) {
     }
   }
 
+  boolean isCountryMatch(ChatChannelUser chatUser) {
+    var fieldValue = countryFilterField.getEditor().getText();
```
1-alex98: We avoid var for code style reasons

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -159,6 +172,27 @@ boolean isGameStatusMatch(ChatChannelUser chatUser) {
...
+    if(fieldValue.isEmpty())
+      return true;
```
1-alex98:
```suggestion
    if(fieldValue.isEmpty()){
      return true;
    }
```

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -159,6 +172,27 @@ boolean isGameStatusMatch(ChatChannelUser chatUser) {
...
+
+    Optional<Player> playerOptional = chatUser.getPlayer();
+
+    if(!playerOptional.isPresent())
+      return false;
```
1-alex98: {}

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -159,6 +172,27 @@ boolean isGameStatusMatch(ChatChannelUser chatUser) {
...
+
+    Player player = playerOptional.get();
+
+    var country = player.getCountry();
```
1-alex98: ...

src/main/resources/i18n/messages_de.properties
```diff
@@ -793,4 +793,5 @@ settings.general.startTab = Starter "Tab"
 startTab.selectTab = Wählen Sie den "Tab", mit dem Sie beginnen möchten\:
 startTab.message = Hier können Sie den "Tab" auswählen, mit dem der Client startet. Sie können dies später in den Einstellungen ändern, wenn Sie möchten.
 startTab.save = Speichern
-vault.replays.onlyShowLastYear = Nur ein Jahr alte
\ No newline at end of file
+vault.replays.onlyShowLastYear = Nur ein Jahr alte
+chat.filter.countryPrompt=Country
```
1-alex98: Translations are made on poeditor.com don't ever commit changes to message properties file other than the English one

1-alex98 (changes requested):
- [ ] Needs unit tests
- [ ] Use localized language names

src/main/java/com/faforever/client/chat/CountryFlagService.java
1-alex98: Java stream api

1-alex98:
> * [ ]  Needs unit tests
> * [ ]  Use localized language names

I18n::getCountryNameLocalized

1-alex98:
@chrsin Exceptions do not make tests fail depending on what thread they are on. So usually there are a ton of Exceptions when running the tests. But as long as they are shown as passed that is not that bad

src/main/java/com/faforever/client/chat/CountryFlagService.java
chrsin: Didn't know about the stream API. Thanks really helpful! Ended out using a method since there was a bit much logic for a one liner.

src/main/java/com/faforever/client/chat/UserFilterController.java
chrsin: Ended up going with a TextField instead. The ComboBox was confusing as you could end up seeing values that didn't match what you were typing.

src/main/java/com/faforever/client/chat/UserFilterController.java
chrsin: Being explicit is better. I agree

src/main/resources/i18n/messages_de.properties
chrsin: Roger that. I have removed the changes for the localized files.

tatsujb (dismissed):

chrsin:
I'm not sure if the fact coveralls failed because of a 0.07 increase should block for this merge. But I won't be able to work on this until at least saturday.

1-alex98:
Next time you can request my review again if you are finished

src/main/java/com/faforever/client/chat/CountryFlagService.java
```diff
@@ -24,6 +30,12 @@
   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
   private static final Collection<String> NON_COUNTRY_CODES = Arrays.asList("A1", "A2", "");
+  private final I18n i18n;
+
+  public CountryFlagService(I18n i18n) {
+    this.i18n = i18n;
+  }
```
1-alex98: Use @RequiredArgsConstructor above the class instead

src/main/java/com/faforever/client/chat/UserFilterController.java
1-alex98: still

1-alex98 (changes requested):

tatsujb (approved):

1-alex98 (approved):
I would have expected a drop-down with suggestions but ok

1-alex98:
Still gotta test it locally

chrsin:
> I would have expected a drop-down with suggestions but ok

I found the text field to be more intuitive. Since I could get DK when typing "de" for an example. (It was maching "Denmark").
Since it filters on every keystroke you get your user list as the result set which was what I used intuitively anyways every time I tested it.

chrsin:
@axel1200 do I need to do more here for this to be merged?

1-alex98:
Yes annoy me to merge it

PR #1987, opened 2020-10-15
Feature/#955 setting advanced logging

Atrocyte:


Atrocyte:
Found a bug, I'll try again

PR #1992, opened 2020-10-18
Catch runtime exception in generated map folder deletion

### LINKED ISSUE #1991: Client crashes if unable to delete map generator maps
**Describe the bug**
Catch the exception issue a warning. Sometimes AV keeps program from deleting maps. Should not keep us from starting.

**OS**
Windows

Visit Issue hunt...

Sheikah45:
Fixes #1991

### DIFF
--- a/src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
+++ b/src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
@@ -113,6 +113,8 @@ public class MapGeneratorService implements InitializingBean {
             .forEach(p -> noCatch(() -> FileUtils.deleteRecursively(p)));
       } catch (IOException e) {
         log.error("Could not list custom maps directory for deleting leftover generated maps.", e);
+      } catch (RuntimeException e) {
+        log.error("Could not delete generated map folder");
       }
     }
   }


src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -113,6 +113,8 @@ private void deleteGeneratedMaps() {
             .forEach(p -> noCatch(() -> FileUtils.deleteRecursively(p)));
       } catch (IOException e) {
         log.error("Could not list custom maps directory for deleting leftover generated maps.", e);
+      } catch (RuntimeException e) {
+        log.error("Could not delete generated map folder");
```
1-alex98, line 117: Should we inform the user 🤔

1-alex98 (approved):

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45: We could I just followed how it handled not being able to list the custom maps directory. 
The only real reason to inform I think would be so that they could delete it themselves if they wanted to but not sure if necessary or depending on how often it happens might just be annoying.

PR #1998, opened 2020-10-18
Add Catalan

1-alex98:


src/test/java/com/faforever/client/i18n/I18nTest.java
```diff
@@ -103,7 +108,11 @@ public void rounded() {
   }
 
   @Test
-  public void getAvailableLanguages() {
-    assertThat(instance.getAvailableLanguages(), hasSize(13));
+  public void testLoadedLanguagesAreComplete() throws IOException {
+    final Path path = Paths.get("src", "main", "resources", "i18n");
```
micheljung, line 112: Reading files and folders like this is bad practice since it will produce different results depending on the working directory of the test runner, which can be different in the IDE, Gradle, CI and even run configurations.

The safest way is always to operate on the class path (even though that's not as easy)

1-alex98: I read gradle automatically sets working directory for tests

1-alex98: So i do the getResource thing?

PR #2000, opened 2020-10-19
Adding Advanced Logging

### LINKED ISSUE #955: Setting for advanced logging
@micheljung is that possible. Pretty sure it is, not sure how tho, but I wanna keep the idea. So we should make a setting to turn on advanced logging, so that everything is logged including debug, I would say so people can turn that on if they have an issue and aditionally information is needed.

Atrocyte:
Allowing the user to increase the log level in Settings > General when clicking the Start button.
The increased log level will remain for the rest of the session.
All loggers inside the com.faforever package will be set to DEBUG.

Fixes #955

### DIFF
--- a/src/main/java/com/faforever/client/preferences/ui/SettingsController.java
+++ b/src/main/java/com/faforever/client/preferences/ui/SettingsController.java
@@ -153,6 +153,7 @@ public class SettingsController implements Controller<Node> {
   public Spinner<Integer> gameDataCacheTimeSpinner;
   public CheckBox allowReplayWhileInGameCheckBox;
   public Button allowReplayWhileInGameButton;
+  public CheckBox debugLogToggle;
 
   private final InvalidationListener availableLanguagesListener;
 
@@ -335,6 +336,8 @@ public class SettingsController implements Controller<Node> {
         clientUpdateService.checkForUpdateInBackground();
       }
     });
+
+    debugLogToggle.selectedProperty().bindBidirectional(preferences.debugLogEnabledProperty());
 
     initUnitDatabaseSelection(preferences);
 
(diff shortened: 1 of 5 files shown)


1-alex98 (commented):
This is for example interesting to debug issues while logging in. Maybe remembering this and setting it on startup would be appropriate.

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -558,5 +562,16 @@ public void onAddAutoChannel() {
...
+  public void onDebugLogToggle() {
+    Level logLevel = Level.DEBUG;
+    final LoggerContext loggerContext = ((Logger) LoggerFactory.getLogger(MethodHandles.lookup().lookupClass())).getLoggerContext();
+    loggerContext.getLoggerList()
+        .stream()
+        .filter(logger -> logger.getName().startsWith("com.faforever"))
+        .forEach(logger -> ((Logger) LoggerFactory.getLogger(logger.getName())).setLevel(logLevel));
+
+    log.info("Switching FA Forever logging configuration to {}", logLevel.levelStr);
+  }
```
1-alex98: Business logic should be in a service

Atrocyte: OK, I've moved this logic to the PreferencesService, which I think is the most suitable Service for logging configurations.

Sheikah45:
Where did this leave off @axel1200?

Atrocyte:
If required I can change the behaviour so that the application will remember enabled DEBUG logging once pressed, so issues on logging in can be checked.
I set the toggle per session initially, so the application won't write too much DEBUG logging in case the user forgets to disable it.

Sheikah45:
Yes it would be nice if the option persisted between sessions

PR #2007, opened 2020-10-26
Feature/#2006 chat changes

### LINKED ISSUE #2006: Optimize chat user list updates
Currently the loading of player information for chat via the services is done in the ChatUserItemController.
This causes the services to be re-called for every single user every time a change is made in the list. With the amount of users growing this will turn into a problem with slow downs.

The information called by the services should be stored in the ChatUserItem so that they can only be called when necessary.

Sheikah45:
This pull request is in response to #2006.

It is trying to use the event bus to modify elements of the chatUserItem and remove calls to the various services from chatUserItemController.
This will hopefully reduce the cpu usage as currently all the services are called for all users whenever the chat user list changes and with this they will not be associated with the specific cell but rather the chat user.

### DIFF
--- a/src/main/java/com/faforever/client/chat/ChannelTabController.java
+++ b/src/main/java/com/faforever/client/chat/ChannelTabController.java
@@ -296,8 +296,8 @@ public class ChannelTabController extends AbstractChatTabController {
     JavaFxUtil.addListener(preferencesService.getPreferences().getChat().chatColorModeProperty(), chatColorModeChangeListener);
     addUserFilterPopup();
 
-    chatUserListView.setItems(filteredChatUserList);
     chatUserListView.setCellFactory(param -> new ChatUserListCell(uiService));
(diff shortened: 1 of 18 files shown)


Sheikah45:
Currently I have it so that everything is initialized fine still need to add events/make sure when changes happen they are reflected properly

Sheikah45:
Alright this should be fully functional now however still need to do unit tests

Brutus5000 (commented):
Apart from the ignored tests it looks good.

Brutus5000:
Tested this. Now I'm stuck on an endless "Connecting to chat". Doesn't happen on develop.

Sheikah45:
Recent commit should have fixed the endless connecting. Seemed to be a race condition so added back the user change listener

Brutus5000:
Yep, works again

Brutus5000 (dismissed):

Sheikah45:
Potentially addresses #1892

PR #2010, opened 2020-11-01
Feature/#1400 Information on games

### LINKED ISSUE #1400: information on games
hover over swords in chat should should give information on the hosted/ being played game.

Marc-Spector:
Closes #1400 

<details>
  <summary>Demonstration</summary>

  demo</details>


Sheikah45:
With the optimization of the chat user list this will need to be rebased and redone slightly

Marc-Spector:
Done.

src/main/java/com/faforever/client/chat/ChatUserService.java
```diff
@@ -128,6 +130,15 @@ public void onChatUserGameChange(ChatUserGameChangeEvent event) {
       chatChannelUser.setStatusImage(null);
     }
   }
+
+  private String getLocalizedStatusGame(PlayerStatus status) {
+    return switch (status) {
```
Sheikah45: This would be better if the i18n key was an instance variable of the enum. See RatingMode.java for an example.

Sheikah45 (changes requested):
Tested and it works. One change to make the enum consistent with others.

src/main/java/com/faforever/client/chat/ChatUserService.java
Marc-Spector: Okay, I can do it better, but what about a separate `IDLE` variable that does not have a string? I do not want to put an empty string `''` or `null` for this because a man will have to write extra checks for an empty string or null before getting a localized string else we will get exception `NoSuchMessageException`.

How about adding this check to the `get(String, Object...)` method of `I18n` class? Like:
```java
public String get(String key, Object... args) {
    return key == null || key.isEmpty() ? null : get(userSpecificLocale, key, args);
  }
```

Or to implement a method like:
```java
public String tryGet(String key, Object... args) {
    try {
      return get(key, args);
    } catch (NoSuchMessageException e) {
      return null;
    }
  }
``` 

What do you think about it?

Sheikah45: I think it would be best to just give idle a message that way if it is used in the future it can be as well.

src/main/java/com/faforever/client/chat/UserFilterController.java
```diff
@@ -209,19 +209,13 @@ private void updateGameStatusMenuText(PlayerStatus status) {
...
-      default:
-        throw new ProgrammingError("Uncovered player status: " + status);
-    }
+    String key =  switch (status) {
+      case PLAYING -> PLAYING.getI18nKey();
+      case LOBBYING -> LOBBYING.getI18nKey();
+      case IDLE -> IDLE.getI18nKey();
+      default -> throw new ProgrammingError("Uncovered player status: " + status);
+    };
+    gameStatusMenu.setText(i18n.get(key));
```
Sheikah45: You can just use this no need for the switch statement

`gameStatusMenu.setText(i18n.get(status.getI18nKey()));`

Sheikah45 (changes requested):
Outside of an apparent mistake in a language file it looks good. Needs another rebase just to catch up to develop though as well.

src/main/resources/i18n/messages_nl.properties
```diff
@@ -197,6 +197,7 @@ game.full.action = Klik om spel te openen
 game.mods.twoAndMore = {0} en {1,number,\#} meer
 game.gameStatus = Spel status
 game.gameStatus.playing = Spelen
+game.gameStatus.hosting = Hosting
```
Sheikah45: If this isnt translated it shouldnt be here

src/main/resources/i18n/messages_ru.properties
```diff
@@ -205,6 +205,7 @@ game.full.action = Нажмите, чтобы открыть игру
 game.mods.twoAndMore = {0} и {1,number,\#} подробнее
 game.gameStatus = Статус
 game.gameStatus.playing = Играет
+game.gameStatus.hosting = Хостит игру
```
1-alex98: no translations are ever made on github!!! poeditor.com

src/main/java/com/faforever/client/chat/ChatUserItemController.java
```diff
@@ -70,14 +73,17 @@
   private final WeakInvalidationListener weakFormatInvalidationListener;
   private final WeakInvalidationListener weakColorPerUserInvalidationListener;
 
+  public ImageView playerMapImage;
   public ImageView playerStatusIndicator;
+  protected Tooltip statusGameTooltip;
+  protected Tooltip gameInfoTooltip;
```
1-alex98, line 78: java fx tooltips are dangerous in terms of memory leaks. Especially here where content is frequently changed

1-alex98: but you seem to take care of it

src/main/java/com/faforever/client/chat/ChatUserItemController.java
```diff
@@ -151,8 +151,72 @@ public void initialize() {
...
+    countryTooltip = new Tooltip();
+    countryTooltip.showDelayProperty().set(Duration.millis(250));
+    Tooltip.install(countryImageView, countryTooltip);
+
+    statusGameTooltip = new Tooltip();
+    statusGameTooltip.setShowDuration(Duration.seconds(30));
+    statusGameTooltip.setShowDelay(Duration.ZERO);
+    statusGameTooltip.setHideDelay(Duration.ZERO);
+    Tooltip.install(playerStatusIndicator, statusGameTooltip);
   }
```
1-alex98, line 173: Chat user items are in a recycler view. Meaning tons of them are created and destroyed while scrolling. So all this is safer if it runs on mouse over only

1-alex98 (changes requested):

1-alex98:
If you want to we can invite you to Zulip(chat service) @ivan-gryzunov

src/main/java/com/faforever/client/chat/ChatUserItemController.java
Sheikah45: This was a change in the chat optimization so might not be related to this pr specifically and would probably need to be done in another pr.

src/main/resources/i18n/messages_ru.properties
Marc-Spector: Sorry. Now I understand how it works. I will remove all changes excluding the `messages.properties` file.

src/main/java/com/faforever/client/chat/ChatUserItemController.java
Marc-Spector: I understand that there is no need to redo this code?

Sheikah45: Yeah if it is redone it wont be in this pr

Sheikah45 (approved):
This looks good to me if checks pass

Marc-Spector:
@axel1200 Thanks! What I should do that to get the invitation? Do you need my email?

Sheikah45:
Yeah we would need an email to send an invite to

ivan-gryzunov:
@Sheikah45 v23620@gmail.com

Sheikah45:
Should come through in a day or so

PR #2012, opened 2020-11-01
#1651 Improve Chat readability

### LINKED ISSUE #1651: Improve chat readability

BlackYps:
Fixes #1651 

Makes names right aligned and reduces font weight on clan tag to normal. I didn't implement the colon, because it is controversial and more of a preference thing than objectively better.
Long names disturbing the layout is still an issue, but that seems rather difficult to fix.
grafik

### DIFF
--- a/src/main/resources/theme/chat/compact/chat_section.html
+++ b/src/main/resources/theme/chat/compact/chat_section.html
@@ -1,9 +1,11 @@
 <table class="chat-section compact chat-section-{username}">
   <tr>
-    <td class="author">
-      <span class="author" style="{inline-style}">
+    <td class="decoration">
         <img class="avatar" src="{avatar}"/>
         <img class="country-flag" src="{country-flag}"/>
+    </td>
+    <td class="author">
+      <span class="author" style="{inline-style}">
         <span class="clan-tag {css-classes}"
               onmouseover="showClanInfo('{clan-tag}')"
               onmouseout="hideClanInfo()"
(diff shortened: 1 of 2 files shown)


Sheikah45 (dismissed):
Looks good to me

Brutus5000:
This is a super opinionated topic and to me it drastically reduces the readability. And we won't come to an agreement here.
Suggestion: If you want to clone the python client as close as possible add a 3rd chat format: "Classic Python"

BlackYps:
The idea is not to clone the python client layout. Can you elaborate what causes the drastically reduced readability for you?
Can we roll out themes? The actual changes are in the css so it would make sense to ship them as a theme, if merging this is too controversial.

Brutus5000:
Technically I don't need to argue why I think the current state is better, since **you** are the one who wants it different.

Anyway: In a spaced or tabular layout in the western hemisphere we use left-aligned text for ages. The whole world is used to it. Why? Because our text flow is left-to-right. And the readability is best, if all elements of the same column start at exactly the same position from where our text flow starts.

In the right-aligned version, I need to search for the right place where the username starts, which makes it harder to read.

BlackYps:
Non-technically we all want a better client so it is good to know what makes a new idea bad :P
And I don't want to persuade you. We already settled that this is opinionated.

BlackYps:
I reverted back to left alignment. Now it contains fixes where the alignment of the text was broken 1 or 2 pixels and adds handling of too long names that would overflow into the message area.
Screenshot_20201108_194038

Brutus5000 (approved):

PR #2013, opened 2020-11-04
Add button to add channel

### LINKED ISSUE #532: Allow to join channels not only if all channels are closed
Add an + Tab which allows to add new channels, similar to the no Open channels view

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46103097-allow-to-join-channels-not-only-if-all-channels-are-closed?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1990: Join channel via a plus button as last tab
Allow joining of IRC channels via UI

Visit Issue hunt...

Sheikah45:
Fixes #1990


Brutus5000:
Tested and works, but not everybody is familiar with IRC, so I think we should emphasize that it needs to start with a `#` (or show an error otherwise instead of just blanking the text field)

bukajsytlos:
question is if the user is bother by the underlying chat technology? should we abstract that and simply prefix automatically?

Sheikah45:
Currently it gives the user an error if it doesnt start with #. I am not sure about adding the abstraction since the /join command would require the # still and it might make people think they don't need it there.

1-alex98 (dismissed):
I like maybe Kubko is correct and we should add the # automatically but otherwise quite good. log statement should probably not be warn level

Brutus5000:
No we shouldn't otherwise it's not consistent with the `/join` command also doesn't differentiate between chatrooms and private chat.

PR #2014, opened 2020-11-04
Add option for date format

### LINKED ISSUE #2009: Date Format
There is an option to change the time format in the settings. An option to change the date format would be a nice addition since currnetly it's in US format only.

Sheikah45:
Fixes #2009


1-alex98 (approved):
Looks good, maybe a unit test

Brutus5000:
Why do we not work with locales instead which would cover dates, times and numbers?

Sheikah45:
We do work with locales. The auto option uses the system default locale. This is just an option to override that for users who don't like their locale similar to the time

PR #2021, opened 2020-11-07
Remove clanMenu from ChatUserItemController and use clanLabel instead

### LINKED ISSUE #2020: Remove Clan Menu and Clan object from chat user list
**Describe the bug**
Currently the clan api call takes a very long time and ends up blocking other messages as well as causing high cpu usage.
If this is removed there is no need for the clan api call and could potentially reduce client cpu usage.

Sheikah45:
Fixes #2020

The removal of the clan menu would cut down significantly on a blocking api call which is part of the issue with the cpu usage.
If it is worth the slowdown then we could keep it.


Sheikah45:
Not going to use for now

PR #2022, opened 2020-11-08
Feature/#1890 Fix a bug when an player is not visible in chat list after removing from friend list

### LINKED ISSUE #1890: Adding then removing player from friend list removes them from chat list
Removing a player from the chat list removes their name from the chat list entirely

**Reproduce**
Add then remove player from friend list

**Expected behavior**
They should move back to the others list

**OS**
Windows

Marc-Spector:
Fixes #1890

### DIFF
--- a/src/main/java/com/faforever/client/chat/CategoryOrChatUserListItem.java
+++ b/src/main/java/com/faforever/client/chat/CategoryOrChatUserListItem.java
@@ -13,6 +13,17 @@ import lombok.Value;
  */
 @Value
 class CategoryOrChatUserListItem {
+
   ChatUserCategory category;
   ChatChannelUser user;
+
+  CategoryOrChatUserListItem(ChatUserCategory category) {
+    this.category = category;
+    this.user = null;
+  }
+
+  CategoryOrChatUserListItem(ChatChannelUser user) {
+    this.user = user;
+    this.category = null;
+  }
 }
--- a/src/main/java/com/faforever/client/chat/ChannelTabController.java
+++ b/src/main/java/com/faforever/client/chat/ChannelTabController.java
@@ -64,8 +64,10 @@ import org.springframework.util.Assert;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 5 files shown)


Marc-Spector:
I have additionally added two very important tests (`testWhenUserAddPlayerToFriendlyListByAccidentallyAndCancelsActionImmediately` and `testWhenUserAddPlayerToFoeListByAccidentallyAndCancelsActionImmediately`). Without my fix, these tests will fail.

Marc-Spector:
The PR is ready to review

src/main/java/com/faforever/client/chat/CategoryOrChatUserListItem.java
```diff
@@ -13,6 +13,15 @@
...
   ChatUserCategory category;
   ChatChannelUser user;
+
+  public boolean isUserItem() {
+    return user != null;
+  }
+
+  public boolean isCategoryItem() {
+    return category != null;
+  }
```
Brutus5000: These are only ever used followed by an .equals().
Wouldn't it be easier to write a
`public boolean contains(Object item)` ?

src/main/java/com/faforever/client/chat/CategoryOrChatUserListItem.java
```diff
@@ -13,6 +13,15 @@
  */
 @Value
 class CategoryOrChatUserListItem {
```
Brutus5000, line 15: While refactoring this class, why not offer either individual constructors only taking a category or a user or offering static builder function `of(ChatUserCategory category)` and  `of(ChatChannelUser user)`

src/main/java/com/faforever/client/chat/CategoryOrChatUserListItem.java
Marc-Spector: Ok, it is a good idea.

Marc-Spector: I will not write this method. There is an easier way.

Marc-Spector: See please https://github.com/FAForever/downlords-faf-client/pull/2022/files#diff-55c0277ffe170fd8988215bc63e3d03909588a2d72fb0fa2e9cca785cdd12bb9R500

Brutus5000: Why did you drop `@Value`? You could just configure the default constructor to be private and call the private constructor from your custom . If you drop @Value you loose ToString, Equals and HashCode features from Lombok

Marc-Spector: Reverted.

Sheikah45 (approved):
Looks good to me

PR #2025, opened 2020-11-10
Fix mod more number on tiles

### LINKED ISSUE #1054: Don't display all mods in tile view
image

When there are more than 3 mods, it should say "MOD1, MOD2, and N more"

Sheikah45:
Fixes #1054


src/main/java/com/faforever/client/game/GameTileController.java
```diff
@@ -119,7 +119,7 @@ private String getSimModsLabelContent(ObservableMap<String, String> simMods) {
         .collect(Collectors.toList());
 
     if (simMods.size() > 2) {
-      return i18n.get("game.mods.twoAndMore", modNames.get(0), modNames.size());
+      return i18n.get("game.mods.twoAndMore", modNames.get(0), simMods.size() - 1);
```
Brutus5000: Looking at the ticket I would have expected size-2

Brutus5000 (changes requested):

src/main/java/com/faforever/client/game/GameTileController.java
Sheikah45: Yeah me too but it was implemented to just show the first mod and then the number if there are more than two mods. I would guess because some mods have long names

Brutus5000: But then the if-statement makes no sense? Shouldn't that be > 1 then? And the i18n string is also misleading...

Sheikah45: Yeah I can change the string title. And if there are 2 then it shows both names if there are more than two it shows name and n more

Brutus5000 (approved):

PR #2028, opened 2020-11-10
Add option to color users by group

### LINKED ISSUE #1264: irc user color suggestions
Please allow users to set colors for user groups (eg all friends in blue and all mods in purple)
changing colors for individual users can be tiring and you can only do it for users who are online atm (unless you edit your client prefs i guess)

having group colors would make things a lot easier - bonus points if you make it so that individually set colors overwrite the user group color so i can make all the mods white and color gieb pink as god wanted him to be

Sheikah45:
Fixes #1264 

This will enable users to set the color by group through context menu on the user category item. Users can also over ride this for individual users. The color is chosen from the highest tier group that is set to a non null value.

Screenshot (163)


Sheikah45:
Would be nice if any one has color suggestions or ways to make it themeable

Brutus5000 (commented):
To make it themable you should add CSS classes dynamically to the list items if the grouping option is enabled

Sheikah45:
Yeah I agree but one issue currently is that all the color operations use the color object to assign and color the text however there isnt a good way to go from css class named color to a color object.

Oh wait nevermind I get what you mean and how to do it

Sheikah45:
hmm actually that would only work for the chat user list the actual chat uses the color object so it is still the same problem

Brutus5000:
I don't think I understand your problem. If you try to change the color you need to know which groups the ChatChannelItem belongs already therefore you could add category-based css classes on the listview item.

Sheikah45:
I am just trying to avoid adding another property to the chatUser class as that will require adding an additional way of setting color as currently none of the chat colors use baked in theme colors

Sheikah45:
Hoping to merge this soon

PR #2043, opened 2020-11-22
Remove JFoenix dependency

Brutus5000:
I have no clue if these animations are still required.


Sheikah45 (commented):
The color picker doesn't let me select it sometimes but not sure if this is because of the change or because the list is updating

PR #2073, opened 2020-12-23
Clean up api includes

Sheikah45:


Brutus5000 (commented):
The api query keys should also be static strings (sort, filter include)

PR #2074, opened 2020-12-23
Add PreferencesBuilder and rename prefs variables

### LINKED ISSUE #246: Create PreferencesBuilder
In test classes, we quite often mock a `Preferences` tree.
It would be easier if we created one builder that builds up the whole object tree (either as a Mock object or as a "real" object, I guess the real object would be sufficient).

Then, replace all mocked `Preferences` with that builder.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/35277121-create-preferencesbuilder?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Sheikah45:
Fixes #246

### DIFF
--- a/src/test/java/com/faforever/client/preferences/PreferencesBuilder.java
+++ b/src/test/java/com/faforever/client/preferences/PreferencesBuilder.java
@@ -1,0 +1,660 @@
+package com.faforever.client.preferences;
+
+import com.faforever.client.chat.ChatColorMode;
+import com.faforever.client.chat.ChatFormat;
+import com.faforever.client.chat.ChatUserCategory;
+import com.faforever.client.game.Faction;
+import com.faforever.client.game.GamesTilesContainerController.TilesSortingOrder;
+import com.faforever.client.main.event.NavigationItem;
+import com.faforever.client.map.generator.GenerationType;
+import com.faforever.client.preferences.Preferences.UnitDataBaseType;
+import com.faforever.client.vault.search.SearchController.SortConfig;
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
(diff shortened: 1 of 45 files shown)


Brutus5000 (commented):
The variable renaming in this PR is inconsistent.

In the GeneratorPrefs you remove the `Property` suffix of all variables, but in the Preferences you add the `Prefs` suffix everywhere.

Also the suffix `Prefs` itself does not improve readability. However, it would make sense if you offer these prefs directly under the PreferencesService, because then you can drop shitton of `.getPreferences()` parts.

src/test/java/com/faforever/client/preferences/PreferencesBuilder.java
```diff
@@ -0,0 +1,660 @@
...
+    preferences = new Preferences();
+  }
+
+  public static PreferencesBuilder create() {
+    return new PreferencesBuilder();
+  }
+
+  public PreferencesBuilder defaultValues() {
+    return this;
+  }
```
Brutus5000, line 41: Where are the default values coming from?

Sheikah45: The default values are set the ones set when the preferences object is created although in that case it may be better to just remove the default values

Sheikah45:
I made the names to be consistent with what the other defaults were however you are right in its inconsistency so will see if I can change it to allow these preferences to come from the service directly

PR #2075, opened 2020-12-24
Clean up detail controllers and add unit tests

Sheikah45:


Brutus5000 (commented):
Please don't name a PR add unit tests and then add changes in the main application.
I didn't intend to review unit tests.

src/main/java/com/faforever/client/mod/Mod.java
```diff
@@ -19,7 +18,7 @@
   private final StringProperty author;
   private final ObjectProperty<OffsetDateTime> createTime;
   private final ObjectProperty<OffsetDateTime> updateTime;
-  private final ObjectProperty<Player> uploader;
+  private final StringProperty uploader;
```
Brutus5000, line 21: Why move to String?

src/main/resources/i18n/messages.properties
```diff
@@ -890,5 +890,7 @@ games.create.enforceRating=Enforce rating
 review.currentVersion=Current Version
 review.version=Version {0}
 review.date={0} on {1}
+review.delete.error=This review could not be deleted
+review.save.error=This review could not be deleted
```
Brutus5000: could not be saved

src/main/java/com/faforever/client/mod/Mod.java
Sheikah45: I moved to string for consistency with author and also since only the username is needed

PR #2114, opened 2021-01-14
Automatically update map when hosting a game

### LINKED ISSUE #1986: Automatically update map when hosting a game
I've a few maps which got a few plays, one actually got over 3k plays now and I updated the map and added textures and stuff but people still keep playing the old version...

It would be a nice feature to have to either

+force hosts to download the latest version of the map (?)
+or at least display a notice that there is a newer version and that the host can download the map before he creates the lobby

Marc-Spector:
Closes #1986


Marc-Spector:
Do I need to add unit tests? What do you think?

Sheikah45:
Tests are always good. Helps to make sure that things don't get broken by future changes

src/main/java/com/faforever/client/map/MapBean.java
```diff
@@ -243,6 +243,20 @@ public int compareTo(@NotNull MapBean o) {
...
+
+    ComparableVersion v1 = this.getVersion();
+    ComparableVersion v2 = o.getVersion();
+    if (v1 == null || v2 == null) {
+      throw new CompareMapVersionException(String.format("cannot compare map versions, map '%s' has null version",
+          v1 == null ? getFolderName() : o.getFolderName()));
+    }
+    return v1.compareTo(v2);
+  }
+
```
Sheikah45: I would move this operation to the MapService as the MapBean is really just a data storage object

Sheikah45: Also the latestMapVersion should always be the most recent so you don't have to check the version specifically just check if the latest is the same as the one you are comparing to.

Marc-Spector: Good! I did not think about it.

src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -475,6 +482,60 @@ public void evictCache() {
...
+
+  public CompletableFuture<Optional<MapBean>> getUpdatedMapIfExist(MapBean map) {
+    return getLatestVersionMap(map).thenApply(optional -> {
+      if (optional.isPresent()) {
+        MapBean latestMap = optional.get();
+        try {
+          if (isNewVersionMap(latestMap, map)) {
+            return Optional.of(latestMap);
+          }
+        } catch (CompareMapVersionException ignored) {  }
```
Sheikah45: should at least add a log statement

Sheikah45 (changes requested):
Very thorough tests!
Looks good just a log statement in the ignored exception would be nice

Sheikah45 (approved):

PR #2115, opened 2021-01-14
Refactor/#1309 notification service

### LINKED ISSUE #1309: Re factor the way notifications are called
Make all the logic in the Notification service and define what notifications there are via an enum(at least the standard ones), also call the audioService from there.

Sheikah45:
Refactor notification service to be event based
Closes #1309


Sheikah45:
I could not figure out a good way to use a notificationType Enum

Sheikah45:
dropped for now

PR #2123, opened 2021-01-19
Clean up updateMap futures and make tests deterministic

Sheikah45:
Was running some tests and some of the updateMap tests were failing randomly. This cleans up some of the tests and the future and tasks for the updateMaps in mapService


Brutus5000 (approved):
I'm not that deep into the topic but looks good

PR #2127, opened 2021-01-23
Use Flowless over listview for chat user list

Sheikah45:
This uses Flowless Virtual Flow for the chat user list. Notably this removes the resetting of the chat user list everytime a user joins or leaves the channel. This will reduce the gpu load from constantly removing and creating the images and associated graphics

### DIFF
--- a/src/main/java/com/faforever/client/chat/ChatUserService.java
+++ b/src/main/java/com/faforever/client/chat/ChatUserService.java
@@ -49,9 +49,6 @@ public class ChatUserService implements InitializingBean {
 
   public void populateClan(ChatChannelUser chatChannelUser) {
     if (chatChannelUser.isDisplayed()) {
-      if (chatChannelUser.getClan().isPresent()) {
-        return;
-      }
       chatChannelUser.getPlayer().ifPresent(player -> {
         if (player.getClan() != null) {
           clanService.getClanByTag(player.getClan())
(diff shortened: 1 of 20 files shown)


src/main/java/com/faforever/client/chat/ChatUserService.java
```diff
@@ -174,30 +167,60 @@ public void associatePlayerToChatUser(ChatChannelUser chatChannelUser, Player pl
...
+      chatChannelUser.setSocialStatusChangeListener((observable, oldValue, newValue) -> {
+        if (!oldValue.equals(newValue)) {
+          populateColor(chatChannelUser);
+        }
+      });
+      chatChannelUser.setGameStatusChangeListener((observable, oldValue, newValue) -> {
+        if (!oldValue.equals(newValue)) {
+          populateGameImages(chatChannelUser);
+        }
+      });
```
micheljung: This throws NPE if `oldValue` is null. You can use `Objects.equals(oldValue, newValue)` to avoid this

micheljung (changes requested):

micheljung:
If this works properly, it's awesome

src/main/java/com/faforever/client/chat/ChatUserService.java
Sheikah45: good catch

Brutus5000:
For general information: I did not notice any performance improvements on Linux on this branch. The client still eats 50-100% of one core non-stop.

Sheikah45:
The bottleneck on linux appears to be the transparent webview which this does not change. This is specifically targeted at stopping the redrawing of the chat user list when a new user joins or leaves, this is going to be most notable during peak times with lots of users hopping in and out.
But ultimately if the bottle neck is something else for a certain system then the impact may be negligible

Brutus5000:
Is this somehow adding a dependency to a particular JavaFX version? From my understanding of the project it should not.

Sheikah45:
This should add no dependencies to a specific javafx version

PR #2131, opened 2021-01-26
Add discord invite over matchmaker chat

BlackYps:
Our player councillor requested more visibility for discord so here is a message for the tmm chat.


src/main/java/com/faforever/client/chat/MatchmakingChatController.java
```diff
@@ -68,6 +79,15 @@ public void setChannel(String partyName) {
...
     matchmakingChatTabRoot.setText(partyName);
+    String topic = i18n.get("teammatchmaking.chat.topic");
+    topicText.getChildren().clear();
+    Arrays.stream(topic.split("\\s"))
+        .forEach(word -> {
+          Label label = new Label(word + " ");
+          label.setStyle("-fx-font-weight: bold; -fx-font-size: 1.1em;");
+            topicText.getChildren().add(label);
+        });
+    topicText.getChildren().add(discordLink);
```
Sheikah45, line 90: If this is always the same it doesnt need to be reset with every setChannel. I would either move it out of the setChannel or make it a true topic for the channel.

Sheikah45 (changes requested):
Also if we move the button here we should remove the one at the top as it gives no real information

src/main/java/com/faforever/client/chat/MatchmakingChatController.java
BlackYps: setChannel only runs once per controller, because currently we clear the tab and make a new one whenever the party owner changes.
setChannel is basically the initialize function.

BlackYps:
I agree about the button at the top. Though I will ask FtX what he thinks about it.

BlackYps:
FtX approved the removal of the upper button

Sheikah45 (approved):

PR #2144, opened 2021-02-12
Add back rating search for replay

### LINKED ISSUE #2143: Add ability to search replays by rating
Replay search by rating was removed due to avoiding the hardcoded ratings in the database and the underspecified problem that rating is mean-3*deviation. This could be added back if we just assume a value for deviation and adjust accordingly

Sheikah45:
Fixes #2143 

This estimates the rating so there will be some false positives or missed replays potentially


Sheikah45:
@Brutus5000 I want to know from you if searching like this will kill the server

Sheikah45:
@Brutus5000 is there anyway to do this where it wouldn't kill the server?

Sheikah45:
As I was testing this it didn't seem to take too long of a time

Issue #2132, opened 2021-01-30
The custom game table view is listing duplicates and not refreshing correctly


dev-nulled:
If you leave the 'play > custom games > table' view open for a length of time (hours?) you the game list does not update correctly.  You will see duplicates show up in the list and new games may or may not appear.

Clicking to tiles, then back to table refreshes the game list back to what it should be.  This was happening prior to the change where custom games is no longer the default view for play.

My suspicion is there is a timeout issue with these games that is jamming things up.  

Screenshots attached - the table view, clicking tiles, then immediately back to table view.  A few duplicate games - but also worth noting the list of games is very different.

![screen 0](https://user-images.githubusercontent.com/47699221/106363735-0164fa00-62f0-11eb-925b-d16484630181.png)
![screen 1](https://user-images.githubusercontent.com/47699221/106363740-09249e80-62f0-11eb-8c6b-3037cba935df.png)
![screen 2](https://user-images.githubusercontent.com/47699221/106363745-0d50bc00-62f0-11eb-9c80-72890442545b.png)

dev-nulled:
![screen 4](https://user-images.githubusercontent.com/47699221/106369246-6121cc00-6315-11eb-935e-0a260b235a70.png)

Another shot after a few hours.

Wizek:
I have also an issue to report that may very well be connected to this or have the same root cause: I often create servers, and they used to fill up rather quickly, sometimes in minutes, usually at most in 30 minutes.

Ever since 1.4.0 came out, they take agonizingly long to fill up, 1-2-3 hours sometimes, even when there is only one slot left from 12. I suspect this is because many players just don't see the game at all in their client.

I'm not sure if I should report this as a separate issue or if it is best to keep it here as a comment.

dev-nulled:
@Brutus5000  Think I found the issue.  Will be doing a bit more testing to confirm.

Sheikah45:
This is most likely an issue with the comparator for the rating range column

PR #2149, opened 2021-02-16
table view fix for custom games Issue #2132

dev-nulled:
First time actually sending code... so let me know where things need to be done differently.  I think this resolves the issues I was seeing with custom games, table view - where you would end up seeing the list go catawampus with duplicates over time.  The min values were null

https://github.com/FAForever/downlords-faf-client/issues/2132

Not sure if zero is a correct floor as a default value.  UI will display >=0 rather than blank for these results too, which may or may not be intended.

### DIFF
--- a/src/main/java/com/faforever/client/remote/domain/RatingRange.java
+++ b/src/main/java/com/faforever/client/remote/domain/RatingRange.java
@@ -24,7 +24,7 @@ public class RatingRange implements Comparable<RatingRange> {
   }
 
   public Integer getMin() {
-    return min;
+    return min == null ? 0 : min;
   }
 
   public Integer getMax() {
--- a/src/test/java/com/faforever/client/remote/domain/RatingRangeTest.java
+++ b/src/test/java/com/faforever/client/remote/domain/RatingRangeTest.java
@@ -1,0 +1,25 @@
+package com.faforever.client.remote.domain;
+
+import org.junit.Test;
+import org.junit.jupiter.api.DisplayName;
+
+import static org.junit.Assert.assertEquals;
+
+public class RatingRangeTest {
+
+  private RatingRange ratingRange;
+
+  @Test
+  @DisplayName("Confirm one as minimum")
+  public void testGetMinWithIntegers(){
+      ratingRange = new RatingRange(1, 2);
+      assertEquals(1, ratingRange.getMin().intValue());
+  }
+
+  @Test
+  @DisplayName("Confirm null as minimum, treating as zero")
+  public void testGetMinWithNulls(){
+    ratingRange = new RatingRange(null, 2);
+    assertEquals(0, ratingRange.getMin().intValue());
+  }
+}


src/main/java/com/faforever/client/remote/domain/RatingRange.java
```diff
@@ -24,7 +24,7 @@ public int compareTo(@NotNull RatingRange o) {
   }
 
   public Integer getMin() {
-    return min;
+    return min == null ? 0 : min;
```
Brutus5000, line 27: 0 is not the minimum rating. It can be negative as well.

Sheikah45:
What lead you to believe that this is the issue? If it is because of a null value it would be better to do the null handling where needed rather than set a minimum value because there is no minimum value since ratings can be negative as well and this would have the side effect of causing those low rated players to get the warning on every game.

I am unable to reproduce the issue on my end so it is difficult for me to find where it goes wrong.

dev-nulled:
@Sheikah45 I'm also digging into that code as well, but for something where the min is empty, zero seems like a reasonable default.  There is still the possibility of a game where the bottom is less than zero (thus the waiting for a -300 to spin up a game for a few hours) and that still works - as that was not null.  This might have something to do with players who are still in a grey state or never played a ladder game.  Still trouble shooting that.  The root cause was the getMin() function is called on the table render.  The compare checks for null, but the getter does not.  There are a few other places where getMin was goofing up and not having a check - listing on the replay vault also gets horked up over time.  My account (*heelix") is one of those grey accounts, if this is a user issue where it renders fine for some, not for the others.

I linked a few screen shots in the issues.  After an hour or so of the refresh of the table layout, all the games will look like one of the early games.  It works like it use to with that min tweak.  This seemed to have broken just before the default was changed from Custom games to Matchmaking.

I'd actually default the min value to zero on the creation of the RankingRange object in the constructor, and allow it to be set lower... but I wanted to actually talk to someone about this first.  Using an Optional is tempting if we want both a value and a null state.

Without the fix, I see this exception spammed in my logs all day long.

Exception in thread "JavaFX Application Thread" java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "com.faforever.client.remote.domain.RatingRange.getMin()" is null
	at com.faforever.client.remote.domain.RatingRange.compareTo(RatingRange.java:23)
	at com.faforever.client.remote.domain.RatingRange.compareTo(RatingRange.java:5)
	at javafx.scene.control.TableColumnBase.lambda$static$0(TableColumnBase.java:138)
	at com.sun.javafx.scene.control.TableColumnComparatorBase$TableColumnComparator.doCompare(TableColumnComparatorBase.java:120)
	at com.sun.javafx.scene.control.TableColumn

Sheikah45:
Ah got it so it seems to be an issue when sorting the table by the rating range. the fix would then be to just define a comparator that also accounts for nulls for that column

Sheikah45:
But we still dont want to have a default min value in the rating range as that would have unintended side effects with regards to the display and showing games to players and determining elgibility.

Sheikah45:
Take a look at #2150 which should fix the issue

dev-nulled:
@Sheikah45 The way this is done, it preserves the negative scores if present/set.  The current table view is very unusable right now, if you leave it open for any length of time.

I see some poor soul who is hosting a -500 to 1000 right now.

ranking

Sheikah45:
Yes but if the range is not specified with this implementation then it will default to 0 and all the negative rated players will get warnings that this game is not intended for them which is not the intended behavior. 

No Rating range means no one should get warnings and that there is no intended range.

heelix:
Good point.  I'll pull your branch and see what I see on the table
refreshing.  The label shows >=0 to zero, but not sure if a negative player
actually sees a block.  Will confirm here in a sec.

dev-nulled:
@Sheikah45 Your update to the compare fixes the table layout issue as well as the weirdness I was seeing with the Vault replays.  Going to close mine.

PR #2159, opened 2021-02-23
Show all search result in a leaderboard (show all possible matches)

### LINKED ISSUE #1775: Show all search results
In the Leaderboards tab, the search is done as the user types something on the keyboard, showing the first matching item and the rest of users beneath it. This is however not practical and not efficient, thus it is aimed to narrow the search results down to users matching the input pattern: 
ex. if "kr" is given, the results should look something like "krristi"; "kraboom" and so on.

Marc-Spector:
Closes #1775 

UI elements
image

### DIFF
--- a/src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
+++ b/src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
@@ -63,7 +63,6 @@ public class LeaderboardsController extends AbstractViewController<Node> {
         leaderboardComboBox.getItems().clear();
         leaderboardComboBox.setConverter(leaderboardStringConverter());
         leaderboardComboBox.getItems().addAll(leaderboards);
-        leaderboardComboBox.getSelectionModel().selectedItemProperty().addListener(onLeaderboardSelected());
         leaderboardComboBox.getSelectionModel().selectFirst();
       });
       return null;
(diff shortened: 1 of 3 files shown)


Marc-Spector:
The PR is ready to review

src/main/resources/theme/leaderboard/leaderboards.fxml
```diff
@@ -16,7 +16,7 @@
       fx:controller="com.faforever.client.leaderboard.LeaderboardsController">
     <Separator minHeight="10.0"/>
     <HBox alignment="CENTER_LEFT" spacing="10.0">
-        <ComboBox fx:id="leaderboardComboBox" onAction="#onLeaderboardSelected"/>
```
Sheikah45: Why the change from the actionEvent to the listener? I would think the actionEvent would be better and less prone to leaks

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
```diff
@@ -125,17 +132,26 @@ public Leaderboard fromString(String string) {
...
+      searchTextField.clear();
+      if (usernamesAutoCompletion != null) {
+        usernamesAutoCompletion.dispose();
+      }
+      leaderboardService.getEntries(newValue).thenAccept(leaderboardEntryBeans -> {
+        ratingTable.setItems(observableList(leaderboardEntryBeans));
+        usernamesAutoCompletion = TextFields.bindAutoCompletion(searchTextField,
+            leaderboardEntryBeans.stream().map(LeaderboardEntry::getUsername).collect(Collectors.toList()));
+        usernamesAutoCompletion.setDelay(0);
+        contentPane.setVisible(true);
```
Sheikah45: I don't see the need for the change to the listener over the onAction method

Sheikah45 (changes requested):

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
```diff
@@ -125,17 +132,26 @@ public Leaderboard fromString(String string) {
...
-    });
+      searchTextField.clear();
+      if (usernamesAutoCompletion != null) {
+        usernamesAutoCompletion.dispose();
+      }
+      leaderboardService.getEntries(newValue).thenAccept(leaderboardEntryBeans -> {
+        ratingTable.setItems(observableList(leaderboardEntryBeans));
+        usernamesAutoCompletion = TextFields.bindAutoCompletion(searchTextField,
+            leaderboardEntryBeans.stream().map(LeaderboardEntry::getUsername).collect(Collectors.toList()));
+        usernamesAutoCompletion.setDelay(0);
```
1-alex98: Don't you think this is a bad thing that the client asks the api with a request on the user table (big table) every letter you type in. Say the username is TrustTheFall. Makes 12 http requests and sql queries XD

1-alex98: I mean there should definitely be a big delay to ensure the server does not get too many requests

1-alex98 (changes requested):

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
Sheikah45: This action is on the comboBox choice not the textField input I believe

Sheikah45: So the api is only called when the user changes the leaderboard they are searching. Then that list is used to populate the autocompletion.

1-alex98: Then this only work with first 10 000 again ... well ok

1-alex98: Please call oldValue and newValue with proper names then I would probably better understand what you are doing. newValue-> newlySelectedLeaderBoard

src/main/resources/theme/leaderboard/leaderboards.fxml
Marc-Spector: To simplify the code. We don't need to manually call the `leaderboardComboBox.getValue()` method. The listener will automatically give us the newly selected value. But if this is done to be less prone to leaks, then okay.

src/main/java/com/faforever/client/leaderboard/LeaderboardsController.java
Marc-Spector: Okay

Sheikah45 (approved):

PR #2166, opened 2021-02-26
Add try catch to i18n.get

### LINKED ISSUE #2165: Difference in translation arguments can break the client
When there are differences in the translation arguments or an argument itself is translated such as number to nummer it can lead to crashing the Application thread of the client which results in a non-operational client.

Ideally this situation should just catch the error and return the english rather than crashing.

Sheikah45:
Fixes #2165

### DIFF
--- a/src/main/java/com/faforever/client/i18n/I18n.java
+++ b/src/main/java/com/faforever/client/i18n/I18n.java
@@ -98,7 +98,7 @@ public class I18n implements InitializingBean {
     try {
       return messageSource.getMessage(key, args, locale);
     } catch (Exception e) {
-      log.error("Could not load message {} with locale {} defaulting to US english", key, locale.toString(), e);
+      log.error("Could not load message {} with locale {} defaulting to US english", key, locale, e);
       return messageSource.getMessage(key, args, Locale.US);
     }
   }
@@ -111,7 +111,7 @@ public class I18n implements InitializingBean {
     try {
       return messageSource.getMessage(key, args, defaultMessage, locale);
     } catch (Exception e) {
-      log.error("Could not load message {} with locale {} defaulting to US english", key, locale.toString(), e);
+      log.error("Could not load message {} with locale {} defaulting to US english", key, locale, e);
       return messageSource.getMessage(key, args, defaultMessage, Locale.US);
     }
   }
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/i18n/I18n.java
```diff
@@ -93,15 +95,25 @@ public String get(String key, Object... args) {
...
     return getWithDefault(userSpecificLocale, defaultMessage, key, args);
   }
 
   public String getWithDefault(Locale locale, String defaultMessage, String key, Object... args) {
-    return messageSource.getMessage(key, args, defaultMessage, locale);
+    try {
+      return messageSource.getMessage(key, args, defaultMessage, locale);
+    } catch (Exception e) {
+      log.error(String.format("Could not load message %s with locale %s defaulting to US english", key, locale.toString()), e);
+      return messageSource.getMessage(key, args, Locale.US);
```
Brutus5000: The default message is missing here

src/main/java/com/faforever/client/i18n/I18n.java
```diff
@@ -93,15 +95,25 @@ public String get(String key, Object... args) {
   }
 
   public String get(Locale locale, String key, Object... args) {
-    return messageSource.getMessage(key, args, locale);
+    try {
+      return messageSource.getMessage(key, args, locale);
+    } catch (Exception e) {
+      log.error(String.format("Could not load message %s with locale %s defaulting to english", key, locale.toString()), e);
```
Brutus5000: Slf4j logger never need String.format or MessageFormat. It works out of the box with `{}` as the placeholder

src/main/java/com/faforever/client/i18n/I18n.java
```diff
@@ -93,15 +95,25 @@ public String get(String key, Object... args) {
...
   public String getWithDefault(String defaultMessage, String key, Object... args) {
     return getWithDefault(userSpecificLocale, defaultMessage, key, args);
   }
 
   public String getWithDefault(Locale locale, String defaultMessage, String key, Object... args) {
-    return messageSource.getMessage(key, args, defaultMessage, locale);
+    try {
+      return messageSource.getMessage(key, args, defaultMessage, locale);
+    } catch (Exception e) {
+      log.error(String.format("Could not load message %s with locale %s defaulting to US english", key, locale.toString()), e);
```
Brutus5000: Same as above

Brutus5000 (changes requested):

src/main/java/com/faforever/client/i18n/I18n.java
Sheikah45: It cannot take the variable argument to the string and log the throwable but I can just do string concatenation

Brutus5000 (approved):

PR #2179, opened 2021-03-04
Sort map pool maps by size and display name

### LINKED ISSUE #2178: Sort queue map list by display name

Sheikah45:
Fixes #2178


Brutus5000 (commented):
I don't it makes sense to sort on a single page? Shouldn't the sorting be on the collected result of all pages?

Sheikah45:
This does sort on all the results because for the map pools we have to get and combine them all then the pagination is done locally on the return line of the function

Brutus5000:
Oh so we are just faking a page result. Hmm. Weird.

Sheikah45:
Yes it is done that way currently just because we dont show the individual rating pools and instead show them all which we can't paginate from the api.

PR #2182, opened 2021-03-07
Rework invite ui

### LINKED ISSUE #2181: Invite player ui is unintuative it seems to me.
I am supposed to click enter and then the currently selected entry gets invited. Super strange. Took me a hile to get behind how it is working.
So here my suggestion.
Have buttons behind names to invite. No enter or double click or whatever needed. Just clcik a button
image

Visit Issue hunt...

1-alex98:
Fixes #2181


1-alex98:
image

1-alex98:
Codacy is wrong I am right I used lombok and codacy does not know that

Sheikah45 (commented):
Can we also add some tests for it?

BlackYps:
Please re-add inviting by pressing enter and keyboard navigation of the invite list. It helps a lot when you want to invite multiple people.

src/main/java/com/faforever/client/teammatchmaking/InvitePlayerController.java
```diff
@@ -83,28 +112,8 @@ public Pane getRoot() {
...
-  private void invite() {
-    playersListView.getSelectionModel().getSelectedItems().forEach(player -> {
-      teamMatchmakingService.invitePlayer(player);
-      invitedPlayersListView.getItems().add(player);
-    });
-    playerTextField.setText("");
-    JavaFxUtil.runLater(() -> playersListView.getSelectionModel().selectFirst());
-  }
-
-  public void onKeyPressed(KeyEvent keyEvent) {
```
BlackYps, line 99: Please re-add inviting by pressing enter and keyboard navigation of the invite list. It helps a lot when you want to invite multiple people.

BlackYps (changes requested):

src/main/java/com/faforever/client/teammatchmaking/InvitePlayerController.java
1-alex98: I find that super not intuitive this is exactly what I want to get rid of. I want to confirm my search so I press enter. But instead of searching it invites the player on top of the list. U can not imagine how many ppl I  invited by accident.

1-alex98: Yes it is definitely slower. But who expects such behavior of a search dialog.

1-alex98: And i did not think that was a time critical thing here... Where it is important if I need 10 or 13 seconds to invite my friends

1-alex98:
> Can we also add some tests for it?

Yes wanna make sure you are ok with the change first

src/main/java/com/faforever/client/teammatchmaking/InvitePlayerController.java
BlackYps: i think I don't understand your workflow. You said you wanted to confirm the search and accidentally invited the wrong player. Right now nothing happens when you press enter. So what did you want to confirm? Also shouldn't it invite the right player, because you searched for him?
You are right that it is not time critical, but evertime you want to play tmm with your friends you have to invite them, so I wanted to make the invitation process with as few clicks as possible, so you don't get annoyed when you do it for the 50th time.

Sheikah45: My perspective on this is that the enter to invite seems a little unintuitive to me as well. And accidentally pressing enter while partially searched and inviting someone else seems easy. I think that the loss of pressing enter is not much since the list first shows your friends and then you can just click the button to invite them right away. If you want to invite someone faster you can just add them as a friend

BlackYps: Right, I forgot about that friend feature. I am fine this then.

Sheikah45:
This should be good but still needs tests

1-alex98:
yeah

Sheikah45:
Are tests going to be added?

1-alex98:
on it

1-alex98:
Codacy sadly does not know about Lombok

Sheikah45:
Yeah I am always surprised they haven't add detection for that yet

1-alex98:
Testwise there is not much else I could test in a meaningfull manner

src/main/java/com/faforever/client/fx/IconButtonListCell.java
```diff
@@ -0,0 +1,41 @@
...
+        configureControllerCallback.accept(new IconButtonListCellControllerAndItem<>(iconButtonListCellController, item));
+        setGraphic(iconButtonListCellController.getRoot());
+      }
+    });
+  }
+
+  @Value
+  public static class IconButtonListCellControllerAndItem<E> {
+    IconButtonListCellController iconButtonListCellController;
+    E item;
```
Sheikah45, line 39: wouldn't it be easier to just add a getter for the controller rather than adding a whole new class as list cell already has getItem

Sheikah45 (changes requested):

src/main/java/com/faforever/client/fx/IconButtonListCell.java
1-alex98: No the Controller does not know about the item and also does not need to

Sheikah45: yes the controller doesn't know about it but the list cell does know about it.

Sheikah45 (approved):
Looks good to me

Issue #1804, opened 2020-06-27
Search field for mods and presets.


zfxstd:
I'm always frustrated when i want to create game with mods. I have a lot of mods downloaded and i need every time to find mods which i want to activate in my big modlist. The main problem - is time. You need spend some minutes just to find it all, sometimes you have dublicate mods and you not sure which mod that was.

One of my solutions is to add search field, like is it now for maps. 

Second solution, which going to be best one, adding presets. For example, i can save settings like gamename\map\mods\etc in game create menu and then load in 1-2 clicks, so i dont need to search and spend much time for this.

Thanks. 
P.s. Sorry for my bad english :)

1-alex98:
Presets are a big change. Needs extra control and UI for that

ghost:
working on this !

zfxstd:
That's awesome! Thanks.

PR #2183, opened 2021-03-09
Implement mod search

ghost:
Implemented mod search bar for searching mods
https://github.com/FAForever/downlords-faf-client/issues/1804

### DIFF
--- a/src/main/java/com/faforever/client/mod/ModManagerController.java
+++ b/src/main/java/com/faforever/client/mod/ModManagerController.java
@@ -102,17 +102,18 @@ public class ModManagerController implements Controller<Parent> {
       @Override
       public void changed(ObservableValue<? extends String> observable, String oldValue, String newValue) {
         if (newValue.isEmpty()) {
-          modVersionFilteredList.setPredicate(null);
+          modVersionFilteredList.setPredicate(viewToggleGroup.getSelectedToggle() == uiModsButton ? UI_FILTER : SIM_FILTER);
         } else {
-          modVersionFilteredList.setPredicate(modVersion -> modVersion.getDisplayName().toLowerCase().contains(newValue.toLowerCase()));
-        }
-        if (!modVersionFilteredList.isEmpty()) {
-          modListView.getSelectionModel().select(0);
+          modVersionFilteredList.setPredicate(getCombinedFilter(newValue));
         }
       }
     });
     EventHandler keyEventHandler = new SearchTextFieldKeyEventHandler(modListView,modVersionFilteredList);
     modSearchTextField.setOnKeyPressed(keyEventHandler);
+  }
+
+  private Predicate<ModVersion> getCombinedFilter(String newValue){
+    return modVersion -> (viewToggleGroup.getSelectedToggle() == uiModsButton ? modVersion.getModType() == ModType.UI : modVersion.getModType() == ModType.SIM) && modVersion.getDisplayName().toLowerCase().contains(newValue.toLowerCase());
   }
 
   private void loadActivatedMods() {
--- a/src/main/java/com/faforever/client/mod/ModManagerController.java
+++ b/src/main/java/com/faforever/client/mod/ModManagerController.java
@@ -2,16 +2,22 @@ package com.faforever.client.mod;
 
 import com.faforever.client.fx.Controller;
 import com.faforever.client.fx.StringListCell;
+import com.faforever.client.map.MapBean;
 import com.faforever.client.mod.ModVersion.ModType;
+import javafx.beans.value.ChangeListener;
+import javafx.beans.value.ObservableValue;
 import javafx.collections.ObservableList;
 import javafx.collections.transformation.FilteredList;
+import javafx.fxml.FXML;
 import javafx.scene.Parent;
 import javafx.scene.control.ListCell;
 import javafx.scene.control.ListView;
 import javafx.scene.control.MultipleSelectionModel;
 import javafx.scene.control.SelectionMode;
+import javafx.scene.control.TextField;
 import javafx.scene.control.ToggleButton;
 import javafx.scene.control.ToggleGroup;
+import javafx.scene.input.KeyCode;
 import javafx.scene.input.MouseEvent;
 import javafx.scene.layout.VBox;
 import javafx.util.Callback;
@@ -50,6 +56,7 @@ public class ModManagerController implements Controller<Parent> {
   public ToggleButton simModsButton;
   public ListView<ModVersion> modListView;
   public VBox root;
+  public TextField modSearchTextField;
 
   private FilteredList<ModVersion> modVersionFilteredList;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/faforever/client/mod/ModManagerController.java
```diff
@@ -86,6 +93,38 @@ public void initialize() {
...
+        }
+        if (!modVersionFilteredList.isEmpty()) {
+          modListView.getSelectionModel().select(0);
+        }
+      }
+    });
+    modSearchTextField.setOnKeyPressed(event -> {
+      MultipleSelectionModel<ModVersion> selectionModel = modListView.getSelectionModel();
+      int currentMapIndex = selectionModel.getSelectedIndex();
+      int newMapIndex = currentMapIndex;
```
micheljung: rather mod index?

ghost: ah right, I think we should refactor this logic out cause this might be repeated elsewhere, but don't know how much work does it require

micheljung: you could extract the event listener into a separate class, passing all arguments into the constructor

src/main/java/com/faforever/client/mod/ModManagerController.java
```diff
@@ -86,6 +97,22 @@ public void initialize() {
     loadActivatedMods();
 
     modListView.scrollTo(modListView.getSelectionModel().getSelectedItem());
+
+    modSearchTextField.textProperty().addListener(new ChangeListener<String>() {
+      @Override
+      public void changed(ObservableValue<? extends String> observable, String oldValue, String newValue) {
+        if (newValue.isEmpty()) {
+          modVersionFilteredList.setPredicate(null);
```
Sheikah45: We would want to still preserve the ui or sim filter when the field is empty

src/main/java/com/faforever/client/mod/ModManagerController.java
```diff
@@ -86,6 +97,22 @@ public void initialize() {
...
+        } else {
+          modVersionFilteredList.setPredicate(modVersion -> modVersion.getDisplayName().toLowerCase().contains(newValue.toLowerCase()));
```
Sheikah45: We would want the filter to search only the ui mods or sim mods based on which button was selected

src/main/java/com/faforever/client/mod/ModManagerController.java
```diff
@@ -86,6 +97,22 @@ public void initialize() {
...
+        }
+        if (!modVersionFilteredList.isEmpty()) {
+          modListView.getSelectionModel().select(0);
```
Sheikah45: Selection is used in a different way for the mod list. It indicates which mods are enabled so we likely do not want to be auto selecting mods for the user as this can get confusing especially since the list is multi select

Sheikah45 (changes requested):
This looks like it is coming along well. 

Two things I noticed were that the UI and SIM filters were not preserved while searching. Additionally the selection of elements in the mod list indicates which mod is enabled so we don't want to auto select for the user. Also it is a multi select list rather than a single select so it is not the same as the map list.

src/main/java/com/faforever/client/mod/ModManagerController.java
ghost: fixed

src/main/java/com/faforever/client/mod/ModManagerController.java
ghost: fixed

src/main/java/com/faforever/client/mod/ModManagerController.java
ghost: removed

src/main/java/com/faforever/client/mod/ModManagerController.java
```diff
@@ -86,6 +97,23 @@ public void initialize() {
...
+      public void changed(ObservableValue<? extends String> observable, String oldValue, String newValue) {
+        if (newValue.isEmpty()) {
+          modVersionFilteredList.setPredicate(viewToggleGroup.getSelectedToggle() == uiModsButton ? UI_FILTER : SIM_FILTER);
+        } else {
+          modVersionFilteredList.setPredicate(getCombinedFilter(newValue));
+        }
+      }
+    });
+    EventHandler keyEventHandler = new SearchTextFieldKeyEventHandler(modListView,modVersionFilteredList);
+    modSearchTextField.setOnKeyPressed(keyEventHandler);
```
Sheikah45: Probably don't want this bound because of the multi select model

src/main/resources/theme/mod_manager.fxml
```diff
@@ -1,46 +1,39 @@
...
         </ToggleButton>
-        <ToggleButton fx:id="simModsButton" maxWidth="1.7976931348623157E308" minWidth="-Infinity"
-                      mnemonicParsing="false" onAction="#onShowSimMods" selected="true" styleClass="game-list-type"
-                      text="%mod.manager.simMods" toggleGroup="$viewToggleGroup" HBox.hgrow="ALWAYS">
+        <ToggleButton fx:id="simModsButton" maxWidth="1.7976931348623157E308" minWidth="-Infinity" mnemonicParsing="false" onAction="#onShowSimMods" prefWidth="152.0" selected="true" styleClass="game-list-type" text="%mod.manager.simMods" toggleGroup="$viewToggleGroup" HBox.hgrow="ALWAYS">
             <tooltip>
-                <Tooltip text="%mod.manager.simModsTooltip"/>
+                <Tooltip text="%mod.manager.simModsTooltip" />
             </tooltip>
         </ToggleButton>
```
Sheikah45: The UI and SIM buttons should be the same size

Sheikah45 (changes requested):

Sheikah45:
Is this still being worked on?

ghost:
> Is this still being worked on?

oh i did not have time , will fix this soon

src/main/resources/theme/mod_manager.fxml
ghost: fixed

src/main/java/com/faforever/client/mod/ModManagerController.java
ghost: removed

src/main/resources/theme/mod_manager.fxml
```diff
@@ -30,7 +31,10 @@
...
     </HBox>
-    <ListView fx:id="modListView" maxHeight="1.7976931348623157E308" VBox.vgrow="ALWAYS">
+    <HBox>
+        <TextField fx:id="modSearchTextField" HBox.hgrow="ALWAYS" minWidth="-Infinity"  maxWidth="1.7976931348623157E308" ></TextField>
+    </HBox>
+    <ListView fx:id="modListView" maxHeight="1.7976931348623157E308">
```
Sheikah45, line 37: It would be nice to have spacing between the buttons, the search box, and the list view vertically.

Sheikah45: Would be good to have the same search... prompt text in the box

Sheikah45 (changes requested):
Almost there. The only thing I would like to see is that now when you search the selections are not preserved. Searching should still show which ones are currently enabled.

Sheikah45:
Gonna close if not being worked on

ghost:
> Gonna close if not being worked on

maybe someone can finish this instead of me?

ghost:
> > Gonna close if not being worked on
> 
> maybe someone can finish this instead of me?

ill try to finish this tomorrow , if not you can close this , sorry about this really short on time  :( and really tired

PR #2195, opened 2021-03-21
Add in detection of replay available flag

### LINKED ISSUE #2194: Use replay available flag in vault

Sheikah45:
Fixes #2194


src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -221,20 +221,27 @@ public void setReplay(Replay replay) {
...
+                watchButton.setText(i18n.get("game.replayFileMissing"));
+                watchButton.setDisable(true);
+              }
+            }));
+      } else {
+        downloadMoreInfoButton.setText(i18n.get("game.replayNotAvailable"));
+        downloadMoreInfoButton.setDisable(true);
+        watchButton.setText(i18n.get("game.replayNotAvailable"));
+        watchButton.setDisable(true);
+      }
```
Brutus5000: The replayFileMissing case will no longer occur due to replaySize. It's not 100% guaranteed but 99,99%.
Also maybe it would be nice to distinguish betweeen "not available" and "not yet available". E.g. if the game finished over an hour ago I would not expect it to show up at all, but if it is younger than (or if no scoreTime set) it is probably still in processing.

Sheikah45: Alright so the scoreTime is only set once the file is written? Or I might just do the hour.
And if the 99.99% is good then I can just remove the other logic just wasnt sure

Brutus5000: No, probably not. The scoreTime is written when the lobby server has receive all game ended messages. That is not necessarily the same moment the replayserver wrote the file. I think this requires the game to be closed, but I'm not sure.

Brutus5000 (changes requested):
Now `game.replayFileMissing` is not used at all. Would like to use it for older replays where no replay will ever come.

PR #2212, opened 2021-04-15
Download all matchmaker maps before queueing

### LINKED ISSUE #1767: Download all ladder maps before being able to join the queue
From discord,
Tagada:
>  Another idea/possible problem
I suppose that Tmm will work like ladder regarding setting up the match which means that if a player doesn't have the map that was selected the client will start downloading it which creates two problems: 
1st: You are able to cancel the download of the map giving you basically the chance to avoid maps you don't like and fuck with the queue + waste time of other people.
2nd. People with really bad internet connection may download the map for some time ( the bigger ones) and if you don't have any loading screen / information that people are connecting then players may think sth broke and before all 4-8 players download and are ready to join someone may quit thinking it's bugged.

>I see 2 potential solutions that should also be implemented into ladder (as this is a problem there aswell, not as big tho since only 2 people are connecting to each other), either make the process of downloading the map invisible to player so that he can't interact with it while displaying info that there is download happening or Force the client to download all the maps before player is accepted to join the queue ( this could lead to people with worse internet connection to wait for quite some time (depending on the pool size too)) which may be a bad thing since they will sit there downloading first patch then maps instead of playing their first game.

Visit Issue hunt...

Sheikah45:
Fixes #1767 

This downloads all your matchmaker maps before queueing. Currently only works for the party leader since it is on the queue button clicked and there is no way to know when the other party members have finished downloading their maps.

### DIFF
--- a/src/main/java/com/faforever/client/map/MapService.java
+++ b/src/main/java/com/faforever/client/map/MapService.java
@@ -542,6 +542,13 @@ public class MapService implements InitializingBean, DisposableBean {
     return fafService.getMatchmakerMapsWithPageCount(matchmakerQueue.getQueueId(), meanRating, count, page);
   }
 
+  public CompletableFuture<List<MapBean>> getAllMatchmakerMaps(MatchmakingQueue matchmakerQueue) {
+    Player player = playerService.getCurrentPlayer().orElseThrow(() -> new IllegalStateException("No user is logged in"));
+    float meanRating = Optional.ofNullable(player.getLeaderboardRatings().get(matchmakerQueue.getLeaderboard().getTechnicalName()))
+        .map(LeaderboardRating::getMean).orElse(0f);
+    return fafService.getAllMatchmakerMaps(matchmakerQueue.getQueueId(), meanRating);
+  }
+
   private CompletableFuture<Void> downloadAndInstallMap(String folderName, URL downloadUrl, @Nullable DoubleProperty progressProperty, @Nullable StringProperty titleProperty) {
     if (mapGeneratorService.isGeneratedMap(folderName)) {
       return mapGeneratorService.generateMap(folderName).thenRun(() -> {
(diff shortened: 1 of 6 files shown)


Askaholic:
For party members you could probably still trigger it when the queue is joined? So long as it won't block other maps from downloading in case the game_launch message comes with an uninstalled map before everything has been downloaded.

I guess you just have to consider how you can detect if the required map is already in the process of being downloaded. For example if the game launch message arrives while the required map is already partially downloaded.

Sheikah45:
Yes that is the main reason why I did not implement it in that way. Right now if the scenario file exists then it thinks the map is downloaded already so could launch with a partial map file which is bad.

It would be easier if we had the server implementation that allowed people to ready up for the matchmaker in a party

Sheikah45:
I am not sure there is a good way to do it without overcomplicating with file watchers or tying the services together

Sheikah45:
I do not know if I can merge this in good faith when there are 200MB maps in the matchmaker pools as it may just block some people from being able to queue. It takes a while to download even on my own system with reliable internet.

Askaholic:
I'm not a fan of the 200mb maps, that just doesn't seem worth it to me. However, isn't that sort of the point of this PR? To try preventing game timeout due to downloading maps.

Sheikah45:
Somewhat but i have heard of users who have issues downloading those maps at all because of their size and so this would prevent them from ever being able to queue for ladder for one map that they may or may not get.

Askaholic:
I think I would tend to agree. On one hand, if the download fails, you could still start the queue, but then it would try to redownload and fail every time you queued, which would just be unnecessary annoyance. Plus some people (like me) have datacaps...

Sheikah45:
Yeah and this iteration wouldn't even let you queue until it knew all the authored maps were downloaded

Sheikah45:
Due to new large map sizes being introduced to the queue and players with limited internet connections this will not be implemented as it may end up blocking the queue for some users over a single map or stressing player's data limits

PR #2221, opened 2021-04-30
#2217 - Make "Upload map" dialog as scrollable

### LINKED ISSUE #2217: "Upload now" and "Cancel" buttons are not visible at the "Upload map" window when using low screen resolution
If you have a low screen resolution (for example, 1366x768), then the "Upload now" and "Cancel" buttons go down outside the window and do not respond to mouse clicks (see screenshot, taken from discord) 
image

Marc-Spector:
Fixes #2217 
<details> 
  <summary>Demonstration</summary>
   
demo

</details>


Marc-Spector:
~~Maybe make all dialogs scrollable by default via `showInDialog(...)` of class `UiService`?~~
Maybe alto better to add new `isScrollable` parameter to `showInDialog(...)` method of `UiService` class? What do you think about this?

Sheikah45:
It is probably best to just make dialogs scrollable by default

Sheikah45 (approved):

PR #2228, opened 2021-05-02
#1217 - Player rating graph - show rating on mouse over

### LINKED ISSUE #1217: Player rating graph - show rating on mouse over
In the player statistics window a rating graph is shown. Display the rating the user is currently hovering over as a number.

http://forums.faforever.com/viewtopic.php?f=2&t=17068&start=10#p170823

Marc-Spector:
Closes #1217


Marc-Spector:
<details>
  <summary>Demo (updated)</summary>  
     
demo


</details>

Marc-Spector:
The PR is ready to review

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,176 @@
...
+  private Tooltip hoverTooltip;
+
+  private boolean valid = false;
+  private final Map<Integer, Double> yValuesMap = new HashMap<>(); // key - pixel in chart background
+
+  public LineChartWithHover(@NamedArg("xAxis") Axis<X> xAxis, @NamedArg("yAxis") Axis<Y> yAxis) {
+    super(xAxis, yAxis);
+    chartBackground = (Region) lookup(".chart-plot-background");
+    if (chartBackground == null) {
+      throw new RuntimeException("'Chart background' object was no found. Using LineChartWithHover class is impossible");
```
Sheikah45:
```suggestion
      throw new RuntimeException("'Chart background' object was not found. Using LineChartWithHover class is impossible");
```

Sheikah45 (changes requested):
It would be nice to add some tests to the LineChartWithHover

src/main/java/com/faforever/client/fx/LineChartWithHover.java
Marc-Spector: You checked the code of old commits. The exception is no longer used and replaced with the `available` flag.

src/main/java/com/faforever/client/chat/PlayerRatingChartTooltipController.java
```diff
@@ -0,0 +1,55 @@
...
+    dateFormat = DateFormat.getDateInstance(DateFormat.MEDIUM, timeService.getCurrentDateLocale());
+  }
+
+  @Override
+  public Node getRoot() {
+    return root;
+  }
+
+  @Override
+  public void currentXY(Double date, Double rating) {
```
Sheikah45: call it setXY

src/main/java/com/faforever/client/chat/PlayerRatingChartTooltipController.java
```diff
@@ -0,0 +1,55 @@
...
+    JavaFxUtil.runLater(() -> {
+      dateLabel.setText(dateFormat.format(TimeUnit.SECONDS.toMillis(date.longValue())));
+      ratingLabel.setText(Long.toString(Math.round(rating)));
+    });
+  }
+
+  @Override
+  public void empty() {
```
Sheikah45: call it clear

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,190 @@
...
+  private void prepareHoverTooltip() {
+    hoverTooltip = JavaFxUtil.createCustomTooltip(tooltipContainer.getRoot());
+    hoverTooltip.setAutoFix(false);
+    hoverTooltip.setHideDelay(Duration.ZERO);
+    hoverTooltip.setShowDelay(Duration.ZERO);
+    hoverTooltip.setShowDuration(Duration.INDEFINITE);
+    Tooltip.install(chartBackground, hoverTooltip);
+  }
+
+  private void showCurrentValues(MouseEvent event) {
```
Sheikah45: I would prefer setValues

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,190 @@
...
+      if (yValue != null) {
+        tooltipContainer.currentXY(xValue, yValue);
+      } else {
+        tooltipContainer.empty();
+      }
+      moveTooltipAndLineToCursor(event);
+    }
+  }
+
+  private void moveTooltipAndLineToCursor(MouseEvent event) {
```
Sheikah45: It appears this is being called twice. Once inside of showCurrentValues and again as a direct mouse event listener. Only call it once

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,190 @@
...
+    if (chartBackground == null) {
+      log.warn("'Chart background' object was no found. The hover tooltip will not be displayed.");
+      available = false;
+    } else {
+      chartBackground.getParent().getChildrenUnmodifiable().stream()
+          .filter((node) -> node != chartBackground)
+          .forEach((node) -> node.setMouseTransparent(true));
+      chartBackground.setOnMouseMoved((event) -> {
+        showCurrentValues(event);
+        moveTooltipAndLineToCursor(event);
```
Sheikah45: This is a potentially redundant call as it is already called in showCurrentValues

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  @VisibleForTesting
+  Region getChartBackground() {
+    return chartBackground;
+  }
+
+  public interface LineChartTooltip extends Controller<Node> {
+
+    void currentXY(Double x, Double y);
```
Sheikah45: setXY

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,190 @@
...
+
+    void empty();
```
Sheikah45: clear

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,190 @@
...
+      valid = true;
+    }
+  }
+
+  @VisibleForTesting
+  Region getChartBackground() {
+    return chartBackground;
+  }
+
+  public interface LineChartTooltip extends Controller<Node> {
```
Sheikah45: This should be moved to its own class

src/main/java/com/faforever/client/util/TimeService.java
```diff
@@ -119,7 +119,7 @@ private Locale getCurrentTimeLocale() {
 
   }
 
-  private Locale getCurrentDateLocale() {
+  public Locale getCurrentDateLocale() {
```
Sheikah45: This should not be made public. Conversion should use asDate instead of making its own method.

src/main/java/com/faforever/client/chat/PlayerRatingChartTooltipController.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public Pane root;
+  public Label dateLabel;
+  public Label ratingLabel;
+
+  private DateFormat dateFormat;
+
+  @Override
+  public void initialize() {
+    dateFormat = DateFormat.getDateInstance(DateFormat.MEDIUM, timeService.getCurrentDateLocale());
```
Sheikah45: There is not need to have its own date format should just use timeService.asDate

src/main/java/com/faforever/client/chat/PlayerRatingChartTooltipController.java
Sheikah45: Also the XValues are always longs and the rating values are always integers and should be typed as such and use the primitive types

src/main/java/com/faforever/client/chat/PlayerRatingChartTooltipController.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @Override
+  public Node getRoot() {
+    return root;
+  }
+
+  @Override
+  public void currentXY(Double date, Double rating) {
+    JavaFxUtil.runLater(() -> {
+      dateLabel.setText(dateFormat.format(TimeUnit.SECONDS.toMillis(date.longValue())));
```
Sheikah45:
```suggestion
     dateLabel.setText(timeService.asDate(Instant.ofEpochSeconds(date)));
```

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,190 @@
...
+    List<LineTo> lines = elements.stream().filter((e) -> e instanceof LineTo).map((e) -> (LineTo) e).collect(Collectors.toList());
+    if (lines.size() < 2) {
+      return;
+    }
+
+    for (int i = 0; i < lines.size() - 1; i++) {
+      LineTo line1 = lines.get(i);
+      LineTo line2 = lines.get(i + 1);
+      int x1 = (int) Math.floor(line1.getX());
+      int x2 = (int) Math.floor(line2.getX());
```
Sheikah45: you can just cast directly to int it has the same effect

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,190 @@
...
+      Number yValue1 = getDisplayedYValue(line1.getY());
+      Number yValue2 = getDisplayedYValue(line2.getY());
```
Sheikah45: getDisplayedValue already returns a double and you use them as doubles no need to cast to Number

Sheikah45: Need a null check here

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,190 @@
...
+
+  private void putYValueMap(int x, Number y) {
+    Double value = yValuesMap.get(x);
+    yValuesMap.put(x, value != null
+        ? (value + y.doubleValue()) / 2
+        : y.doubleValue());
+  }
+
+  private Double getDisplayedXValue(double displayPosition) {
+    return (Double) getXAxis().getValueForDisplay(displayPosition);
```
Sheikah45: This casting will only work when the axis is of double objects so this class may only be valid for when X and Y are doubles

src/main/java/com/faforever/client/fx/LineChartWithHover.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  private Double getDisplayedYValue(double displayPosition) {
+    return (Double) getYAxis().getValueForDisplay(displayPosition);
```
Sheikah45: same as with X

Sheikah45 (changes requested):
Apologies for the delay in the review did not realize this was ready.

I found some confusing usage of the generics in the LineChart class where there is direct casting to a double which is not valid in all cases so should either be extended to be really generic or the types should be narrowed to only being doubles.

Also throughout it is better to use the primitive types where possible rather than the boxed types.

src/main/java/com/faforever/client/fx/LineChartWithHover.java
Marc-Spector: Hmmm... It is impossible.
image

Marc-Spector:
@Sheikah45 Thanks for the review. Updated the PR.

src/test/java/com/faforever/client/fx/MouseEvents.java
```diff
@@ -11,4 +11,12 @@ private MouseEvents() {
   public static MouseEvent generateClick(MouseButton button, int clicks) {
     return new MouseEvent(MouseEvent.MOUSE_CLICKED, 0, 0, 0, 0, button, clicks, false, false, false, false, false, false, false, false, false, false, null);
   }
+
+  /**
+   * @param x the x with respect to the node.
+   * @param y the y with respect to the node.
+   */
+  public static MouseEvent generateMouseMoved(int x, int y) {
```
1-alex98, line 19: FxRobot class does such things I think

src/main/java/com/faforever/client/fx/PlayerRatingChart.java
```diff
@@ -0,0 +1,187 @@
...
+      int x1 = (int) Math.floor(line1.getX());
+      int x2 = (int) Math.floor(line2.getX());
+      int rating1 = getDisplayedRatingValue(line1.getY());
+      int rating2 = getDisplayedRatingValue(line2.getY());
+      if (rating1 == Integer.MIN_VALUE || rating2 == Integer.MIN_VALUE) {
+        continue;
+      }
+      int distance = x2 - x1;
+      if (distance > 1) {
+        double k = (double) (rating2 - rating1) / distance;
```
1-alex98: Bad variable naming. Maybe `gradient` or what is this k?

src/main/java/com/faforever/client/fx/PlayerRatingChart.java
```diff
@@ -0,0 +1,187 @@
...
+    List<LineTo> lines = elements.stream().filter((e) -> e instanceof LineTo).map((e) -> (LineTo) e).collect(Collectors.toList());
+    if (lines.size() < 2) {
+      return;
+    }
+
+    for (int i = 0; i < lines.size() - 1; i++) {
+      LineTo line1 = lines.get(i);
+      LineTo line2 = lines.get(i + 1);
+      int x1 = (int) Math.floor(line1.getX());
+      int x2 = (int) Math.floor(line2.getX());
```
1-alex98: leftXCoordinate rightXCordinate maybe

src/main/java/com/faforever/client/fx/PlayerRatingChart.java
```diff
@@ -0,0 +1,187 @@
...
+      Series<Double, Double> series = getData().get(i);
+      if (series.getNode() instanceof Path) {
+        buildData(((Path) series.getNode()).getElements());
+        break;
+      }
+    }
+  }
+
+  private void buildData(List<PathElement> elements) {
+    List<LineTo> lines = elements.stream().filter((e) -> e instanceof LineTo).map((e) -> (LineTo) e).collect(Collectors.toList());
```
1-alex98: e stands for element? or energy 🤔 XD

src/main/java/com/faforever/client/fx/PlayerRatingChart.java
```diff
@@ -0,0 +1,187 @@
...
+    return value != null ? value.longValue() : Long.MIN_VALUE;
+  }
+
+  private int getDisplayedRatingValue(double displayPosition) {
+    Double value = getYAxis().getValueForDisplay(displayPosition);
+    return value != null ? value.intValue() : Integer.MIN_VALUE;
+  }
+
+  @Override
+  protected void seriesChanged(Change<? extends Series> c) {
```
1-alex98: `change`

src/main/java/com/faforever/client/chat/PlayerRatingChartTooltipController.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  public Node getRoot() {
+    return root;
+  }
+
+  public void setDateAndRating(long dateValueInSec, int rating) {
+    JavaFxUtil.runLater(() -> {
+      dateLabel.setText(timeService.asDate(Instant.ofEpochSecond(dateValueInSec)));
+      ratingLabel.setText(Long.toString(Math.round(rating)));
```
Sheikah45: rating is already an int why round it?

src/main/java/com/faforever/client/fx/PlayerRatingChart.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+      } else {
+        putMapRatingValue(x1, rating1);
+      }
+    }
+  }
+
+  private void putMapRatingValue(int x, double rating) {
+    Double currentRating = ratingMap.get(x);
+    ratingMap.put(x, currentRating != null ? (currentRating + rating) / 2 : rating);
```
Sheikah45: look at the merge function of a map. It would make this easier to read

Sheikah45 (changes requested):

src/main/java/com/faforever/client/chat/PlayerRatingChartTooltipController.java
Marc-Spector: I missed this code when I updated the type of the `rating` parameter from `double` to `int`. Fixed.

src/test/java/com/faforever/client/fx/MouseEvents.java
Marc-Spector: Cannot find the method I need. The `moveTo` method is not suitable because he moves the cursor physically. The `generateMouseMove` simply imitates the movement of the cursor to the node for invoking of `mouseMoved` listener.

1-alex98: ok

src/main/java/com/faforever/client/fx/LineChartWithHover.java
1-alex98: (int)

1-alex98: Your screenshot does not show a cast

src/main/java/com/faforever/client/chat/PlayerRatingChartTooltipController.java
Geosearchef: The client code uses `String.valueOf()` in most other locations, so I'd go for that one instead of `Long.toString()`. It's also null safe in case the object in question is null, which doesn't matter for this case though.

src/main/java/com/faforever/client/fx/PlayerRatingChart.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  private long getDisplayedDateValue(double displayPosition) {
+    Double value = getXAxis().getValueForDisplay(displayPosition);
+    return value != null ? value.longValue() : Long.MIN_VALUE;
```
Geosearchef:
```
return Optional.ofNullable(getXAxis().getValueForDisplay(displayPosition))
             .map(ClassNameGoesHere::longValue)
             .else(Long.MIN_VALUE)
```
Just a suggestion, this temp value variable and tertiary op look very verbose, not sure if it's any better though.

src/main/java/com/faforever/client/chat/PlayerRatingChartTooltipController.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override
+  public Node getRoot() {
+    return root;
+  }
+
+  public void setDateAndRating(long dateValueInSec, int rating) {
+    JavaFxUtil.runLater(() -> {
+      dateLabel.setText(timeService.asDate(Instant.ofEpochSecond(dateValueInSec)));
+      ratingLabel.setText(Integer.toString(rating));
```
Sheikah45: one small change here. i18n has a number method so this would become
`ratingLabel.setText(i18n.number(rating))`

Sheikah45 (changes requested):

src/main/java/com/faforever/client/fx/PlayerRatingChart.java
Marc-Spector: Done. :)

Sheikah45 (dismissed):

src/main/java/com/faforever/client/chat/PlayerRatingChartTooltipController.java
Marc-Spector: Done.

Sheikah45:
Although checks are failing as i18n is not in the test class instantiation

Sheikah45 (approved):

1-alex98 (approved):

PR #2236, opened 2021-05-09
Fix review version not being set

### LINKED ISSUE #2235: Writing a new review freezes the map vault
There is a null for the version

Sheikah45:
Fixes #2235

### DIFF
--- a/src/main/java/com/faforever/client/vault/review/ReviewController.java
+++ b/src/main/java/com/faforever/client/vault/review/ReviewController.java
@@ -49,13 +49,12 @@ public class ReviewController implements Controller<Pane> {
 
   private Consumer<Review> onSendReviewListener;
   private Consumer<Review> onDeleteReviewListener;
-  private Optional<Review> review;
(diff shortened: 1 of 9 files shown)


src/main/java/com/faforever/client/vault/review/ReviewController.java
```diff
@@ -76,7 +77,7 @@ public void initialize() {
   public void setReview(Optional<Review> optionalReview) {
```
Brutus5000: I know you didn't write this code, but I like to have mistakes fixed when I see them.

This method is the perfect example why you only ever return Optionals but never use them as parameters:

- The code checks for the Optional. If it's missing, you return. (The orElse feature is not even used)
- Due to that the code forgets to check if the Optional is null (this is always possible)
- The optional needs to be unwrapped.

Instead of checking for Optional it would be more useful to check for null instead. If the caller has an Optional hat hand he can just do .orElse(null)

PR #2255, opened 2021-05-30
Update Mods before hosting

### LINKED ISSUE #1680: Auto update mods
As the title says, maybe a feature that would automatically update mods would be nice.

If someone for example tries to host a game, with V2 of a certain mod, but V5 is the latest, the client should recognise V2 as hidden in the vault (maybe just a simple name-check, since that's how automatically hiding old versions works already?), and auto-download the latest version before the game launches. 

This would reduce the hassle for mod users, while also preventing the old versions from circulating after they have been updated. (as they are no longer obtainable from the vault anyway)

1-alex98:
Fixes #1680


src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -407,26 +409,45 @@ private void onGenerateMap() {
...
 
-  private void hostGame(MapBean map) {
-    Set<String> mods = modManagerController.apply().stream()
+  private void hostGameAfterModUpdate(MapBean map) {
+    List<ModVersion> selectedModVersions = modManagerController.getSelectedModVersions();
+
+    modService.updateModVersionsAndActivateThose(selectedModVersions)
+        .exceptionally(throwable -> {
+          log.error("Error when updating selected mods", throwable);
+          notificationService.addImmediateErrorNotification(throwable, "game.create.errorUpdatingMods");
```
1-alex98: Can not be recovered because it might lead to inconsistencies

1-alex98:
Tested it by downloading outdated mod worked flawlessly

Marc-Spector:
@1-alex98 I think maybe add a notification that the selected mods have been updated so that players know and can go to the mod vault to read the changelog?

Marc-Spector:
And add an option "Mods updated" in "Notifications" settings?

1-alex98:
You want a persistent notification?

1-alex98:
Do ppl write change logs?

1-alex98:
The feature of notifying players about an updated mod is a separate feature and should not be part of this pull request. Mods are also updated if you join a game that has a newer version and there you also don't get a notification

1-alex98:
But I guess you could open an issue for that

Sheikah45:
I don't think I have ever seen a changelog in the mod vault for any mods

Marc-Spector:
> The feature of notifying players about an updated mod is a separate feature and should not be part of this pull request. Mods are also updated if you join a game that has a newer version and there you also don't get a notification

I did not think about it. You are right. my idea is useless. :)

Sheikah45:
> 
> 
> > The feature of notifying players about an updated mod is a separate feature and should not be part of this pull request. Mods are also updated if you join a game that has a newer version and there you also don't get a notification
> 
> I did not think about it. You are right. my idea is useless. :)

I would not say useless. But yeah I don't know if it would fit in with other similar changes we have made.

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -407,26 +409,45 @@ private void onGenerateMap() {
...
+        })
+        .thenAccept(modVersions -> hostGame(map, getUUIDsFromModVersions(modVersions)));
+  }
+
+  @NotNull
+  private Set<String> getUUIDsFromModVersions(List<ModVersion> modVersions) {
+    return modVersions.stream()
         .map(ModVersion::getUid)
         .collect(Collectors.toSet());
+  }
```
Sheikah45: I think this might be better just as a thenCompose chain on the completableFutures similar to how it is done in the gameService as right now the hostGameAfter naming convention is a little confusing as there are actually multiple steps here

src/main/java/com/faforever/client/mod/ModService.java
```diff
@@ -441,4 +442,30 @@ private void addMod(Path path) {
   public void destroy() {
     Optional.ofNullable(directoryWatcherThread).ifPresent(Thread::interrupt);
   }
+
+  @Async
+  @SneakyThrows
+  public CompletableFuture<List<ModVersion>> updateModVersionsAndActivateThose(final List<ModVersion> selectedModVersions) {
+    final List<ModVersion> newlySelectedMods = new ArrayList<>(selectedModVersions);
```
Sheikah45: I think updateAndActivateModVersions would be better

src/main/java/com/faforever/client/mod/ModService.java
```diff
@@ -441,4 +442,30 @@ private void addMod(Path path) {
...
+    selectedModVersions.forEach(installedModVersion -> {
+      try {
+        ModVersion modVersionFromApi = getModVersionById(installedModVersion.getUid()).get();
+        ModVersion latestVersion = modVersionFromApi.getMod().getLatestVersion();
+        boolean isLatest = latestVersion.equals(installedModVersion);
```
Sheikah45, line 458: Do these equality actually work since the fields of ModVersion are properties and not the direct values

src/test/java/com/faforever/client/mod/ModServiceTest.java
```diff
@@ -405,6 +408,61 @@ public void testLoadThumbnail() throws MalformedURLException {
...
+
+    when(fafService.getModVersion(eq(modVersion.getUid())))
+        .thenReturn(CompletableFuture.completedFuture(modVersion));
+
+    Files.createFile(gamePrefsPath);
+
+
+    List<ModVersion> modVersions = instance.updateModVersionsAndActivateThose(List.of(modVersion)).get();
+
+    verify(taskService, times(1)).submitTask(any());
```
Sheikah45: add verification that the modService never tries to download the mod in this case

Sheikah45 (changes requested):

BlackYps:
This needs an option to turn off this behavior. See Brutus' comment in the linked issue. Especially point 1. and 4.
Maybe just notifying about outdated mods is the better solution?

1-alex98:
I disagree with @Brutus5000 there. It is not like we keep players from playing. The updating only triggers if you are really playing that exact mod. Playing outdated and broken mods is a huge issue in FAF. Players have almost no way of updating their mod. Many tecnical help threads concerning mods end up in "there is a new version of this mod". Have you updated? Then most mods do not have frequent updates. And lastely mods are big and mods are not big on average. Apart from the mods that introduce new units. But they are only updated if you play them(Blackops etc.) and I think Blackops have not been updated in years i'd guess. Brewlan neither. So I really don't know what mod that would apply to (SCTA maybe, not that it had any playerbase I guess)... But I can build such an option even if I am 100% sure it is probably not needed(nobody will turn it off/notice the change)

src/main/java/com/faforever/client/mod/ModService.java
1-alex98: yes it compares uids

src/test/java/com/faforever/client/mod/ModServiceTest.java
1-alex98: This is the verifyication

1-alex98: I can make it more explicit

Sheikah45: Ah yeah I must have missed that as I was expecting something on modService

1-alex98:
Now you can turn it off if you like to, even tho I see little sense in it. Also btw if you are not host then you are forced to map and mod updates as well all the time and you can not turn that off 🤔

Sheikah45 (approved):

BlackYps:
I agree that it would normally not be a problem, but if a mod gets updated with a broken version you would not be able to play with that mod at all.
Anyway, I won't argue further because you added the option already, so it's all gucci.

PR #2257, opened 2021-05-30
Remove cookie service

Sheikah45:
In prefs the cookies cause an illegal reflective action to occur which is preventing us from using java16. Cookies are only used by the unitdb  Controller and from what I can tell no relevant cookies to the db are actually stored making it rather useless.


1-alex98:
It was supposed to store the unit db settings... Don't ask me if it actually still works :D

1-alex98:
Anyways its functionality is not that relevant

1-alex98 (approved):

Sheikah45:
Yeah it seems like the type of db is stored in the preferences itself

1-alex98:
No it was the settings of @Rackover 's unit db

Rackover:
if I remember correctly you can pass these settings as a string in a GET parameter, but I may be wrong :) 
https://github.com/FAForever/UnitDB/blob/master/index.php#L32 Check it out

Have a nice day ♥

1-alex98:
Yeah no idea it is good the cookie service is gone probably

PR #2258, opened 2021-06-01
Mention users with dash

### LINKED ISSUE #2202: mentionPattern does not match names that end in `-`
People who's names end in `-` can't be pinged.
img

After messing around a bit, I think this is due to the `\b` word boundary matching. As `-` is not a word character, a space can never match after it (word boundary requires word before and non-word after, or vice versa). You can still ping these people if you put a letter directly after the `-` in order to trigger the word boundary matching.

```java
Pattern p = Pattern.compile("\\b(" + "Box-" + ")\\b", CASE_INSENSITIVE);
Matcher m = p.matcher("I'm Box-");

System.out.println(m.find());
// Prints "false"

Matcher m2 = p.matcher("I'm Box-x");
System.out.println(m2.find());
// Prints "true"
```

Visit Issue hunt...

1-alex98:
Fixes #2202

### DIFF
--- a/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
+++ b/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
@@ -250,7 +250,7 @@ public abstract class AbstractChatTabController implements Controller<Tab> {
   }
 
   public void initialize() {
-    mentionPattern = Pattern.compile("\\b(" + Pattern.quote(userService.getUsername()) + ")\\b", CASE_INSENSITIVE);
+    mentionPattern = Pattern.compile("\\b(" + Pattern.quote(userService.getUsername()) + ")\\b|[ ]", CASE_INSENSITIVE);
 
     initChatView();
 


1-alex98:
Tried testing it is impossible

Askaholic:
Can’t you write some tests like I did on the issue. Make a Matcher and call find.

What about users who’s name starts with a dash?

1-alex98:
Well the code does a shit tone of other things including reading files etc. and because we do not load a spring context this is incredible hard to mock away. Well I could just refactor to make it more testable I guess

1-alex98:
I stay with can not be tested

Askaholic:
Just test the regex. You don’t need to test the code that uses the regex.

Just do something like
```java
assertTrue(CHAT_PATTERN.matcher("Hello Box-!").find();
```

Whatever the syntax is

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -250,7 +250,7 @@ public void setReceiver(String receiver) {
   }
 
   public void initialize() {
-    mentionPattern = Pattern.compile("\\b(" + Pattern.quote(userService.getUsername()) + ")\\b", CASE_INSENSITIVE);
+    mentionPattern = Pattern.compile("\\b(" + Pattern.quote(userService.getUsername()) + ")\\b|[ ]", CASE_INSENSITIVE);
```
Askaholic, line 253: Btw this doesn’t catch the case where the username is the only thing on the line. Like, this will not match the string "Box-".

PR #2260, opened 2021-06-04
Use new recommended maps and mods attribute for top picks

Sheikah45:


1-alex98:
I get no maps/mods displayed is that because no M&M favourites have been set yet?

Sheikah45:
correct none have been set yet. But there are ones set on the test server

PR #2263, opened 2021-06-05
Bugfix/#1305 Stop multiple sounds

### LINKED ISSUE #1305: Multiple ping sounds in a row
Not a big issue, but it's pretty irritating when you're doing something else on the PC and someone sends multiple PM's and it just keeps spamming that annoying notification sound. One's enough, I know you've sent me messages, don't need a reminder every few seconds. also i heard this was said already, but flashing icon when you get pm'd like with faf would be nice (like it is on python), instead of the small hard-to-notice "2" on faf icon.

Sheikah45:
Fixes #1305

### DIFF
--- a/src/main/java/com/faforever/client/audio/AudioService.java
+++ b/src/main/java/com/faforever/client/audio/AudioService.java
@@ -23,7 +23,7 @@ public class AudioService implements InitializingBean {
   private static final String INFO_SOUND = "theme/sounds/info.mp3";
   private static final String MENTION_SOUND = "theme/sounds/mention.mp3";
   private static final String PRIVATE_MESSAGE_SOUND = "theme/sounds/pm.mp3";
-  private static final long SOUNDS_DELAY = 30000;
+  private static final long SILENCE_PERIOD_AFTER_SOUND = 30000;
 
   private final PreferencesService preferencesService;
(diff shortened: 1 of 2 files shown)


Brutus5000:
How can delaying without a queue solve this? Don't you just move the problem to a later point?
I would expect we'd need an aggregating function only firing every few seconds.

Sheikah45:
The title might be misleading. It doesnt actually delay. It just wont play any new sounds for 30 seconds after a sound is played

src/main/java/com/faforever/client/audio/AudioService.java
```diff
@@ -21,10 +23,12 @@
   private static final String INFO_SOUND = "theme/sounds/info.mp3";
   private static final String MENTION_SOUND = "theme/sounds/mention.mp3";
   private static final String PRIVATE_MESSAGE_SOUND = "theme/sounds/pm.mp3";
+  private static final long SOUNDS_DELAY = 30000;
```
Brutus5000: Maybe SILENCE_PERIOD_AFTER_SOUND ?

Sheikah45: yeah probably better

Brutus5000 (approved):

PR #2271, opened 2021-06-12
Extend mention pattern

### LINKED ISSUE #2202: mentionPattern does not match names that end in `-`

Sheikah45:
Closes #2202

### DIFF
--- a/src/test/java/com/faforever/client/chat/AbstractChatTabControllerTest.java
+++ b/src/test/java/com/faforever/client/chat/AbstractChatTabControllerTest.java
@@ -54,7 +54,9 @@ import static java.util.concurrent.CompletableFuture.completedFuture;
 import static org.hamcrest.Matchers.isEmptyString;
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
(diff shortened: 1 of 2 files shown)


src/test/java/com/faforever/client/chat/AbstractChatTabControllerTest.java
```diff
@@ -331,4 +333,18 @@ public void testChannelNamesTransformedToHyperlinks() {
     String output = instance.replaceChannelNamesWithHyperlinks("Go to #moderation and report a user");
     assertThat(output, is("Go to <a href=\"javascript:void(0);\" onClick=\"java.openChannel('#moderation')\">#moderation</a> and report a user"));
   }
+
+  @Test
+  public void testMentionPattern() {
```
Askaholic, line 338: Some more suggestions to test. I’m not even sure what the desired behavior would be for these tho cuz they’re kinda weird:
- `-Box--`
- `-Box-Box-`

Maybe also:
- `[-Box-]`

Since that would have failed with my lazy A-z character class.

1-alex98 (approved):

PR #2276, opened 2021-06-13
Clean up teammatchmaking service and controller

### LINKED ISSUE #2088: TMM: Invite notification doesn't always show up
An invite triggers two notifications:
A persistent notification on the top and a transient notification (popup toast). The popup notification seems to not show up for some people.

### LINKED ISSUE #2089: TMM: Players in queue is not visible when you are not party leader
Some people reported that they can't see the "In queue" label when they are not the party leader. I can't reproduce this issue, so I need some more information. Preferably with logs, but I am not sure if we will find anything in them.

### LINKED ISSUE #2111: Team Matchmaker Tests not entirely deterministic
Some of the Team Matchmaker Tests will randomly fail in the CI. When it is rerun they pass.

### LINKED ISSUE #2148: Optimize Matchmaker Display Logic
Currently the matchmaker display code is not very optimized as it does things like call the api on every message even when the queue information is stored locally, re-render every matchmaker object every time etc. It would be nice if this could be optimized to have a lower impact on GPU and CPU.

Sheikah45:
Currently there is little separation between the service and ui controller for teammatchmaking. The service is frequently changing the UI directly and the controller is modifying service variables directly.

This pr aims to clean up the code.

Closes #2148

### DIFF
--- a/src/main/java/com/faforever/client/fx/JavaFxUtil.java
+++ b/src/main/java/com/faforever/client/fx/JavaFxUtil.java
@@ -220,7 +220,7 @@ public final class JavaFxUtil {
     }
     if (image.isBackgroundLoading() && image.getProgress() < 1) {
       // Let's hope that loading doesn't finish before the listener is added
-      JavaFxUtil.addListener(image.progressProperty(), new ChangeListener<Number>() {
+      JavaFxUtil.addListener(image.progressProperty(), new ChangeListener<>() {
         @Override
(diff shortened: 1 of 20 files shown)


Sheikah45:
Just need to finish fixing tests but it should be functional now

Sheikah45:
Alright this should be good to go now tests and all

Sheikah45:
Would really appreciate a review on this maybe @1-alex98 ?

PR #2294, opened 2021-06-28
Use jackson over gson

### LINKED ISSUE #492: Use Jackson instead of Gson
<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/41506334-use-jackson-instead-of-gson?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Sheikah45:
Closes #492


Sheikah45:
This pr got a little more involved then I was expecting so apologies for the number of files changed

Sheikah45:
Should be good to go now

Sheikah45:
will probably add tests for all the messages later today

src/main/java/com/faforever/client/fa/relay/ice/event/GpgClientMessageEvent.java
```diff
@@ -0,0 +1,14 @@
...
+
+
+import com.faforever.client.fa.relay.outbound.GpgOutboundMessage;
+import lombok.Value;
+
+@Value
+public class GpgClientMessageEvent {
+  GpgOutboundMessage gpgMessage;
+
+  public GpgClientMessageEvent(GpgOutboundMessage gpgMessage) {
```
Brutus5000: redundant for @Value class

src/main/java/com/faforever/client/remote/domain/inbound/FafInboundMessage.java
```diff
@@ -6,12 +6,11 @@
...
  * field.
- *
  */
 
-public abstract class FafServerMessage extends ServerMessage {
+public abstract class FafInboundMessage extends InboundMessage {
 
-  protected FafServerMessage() {
+  protected FafInboundMessage() {
     super(MessageTarget.CLIENT);
```
Brutus5000: Not sure why we still need this.

Brutus5000 (changes requested):
What's with all these `public static final String COMMAND` variables? Shouldn't this string be moved to the json subtypes so they vanish from the actual class dropping all these annoying manual constructors? Check faf-java-commons lobby example on how to do it.

src/main/java/com/faforever/client/remote/domain/inbound/FafInboundMessage.java
Sheikah45: I use this class to set the target for all the faf subclass messages

src/main/java/com/faforever/client/fa/relay/ice/event/GpgClientMessageEvent.java
Sheikah45: Ah thanks for the catch

Sheikah45:
I was using the `COMMAND` field similarly to the `NAME` field that is seen in the api in some places and for some documentation of how the name is tied to the class in case it needs to be used elsewhere but could be convinced it makes no sense.

src/main/java/com/faforever/client/discord/DiscordEventHandler.java
```diff
@@ -17,24 +17,37 @@
 public class DiscordEventHandler extends DiscordEventHandlers {
   private final ApplicationEventPublisher applicationEventPublisher;
   private final NotificationService notificationService;
+  private final ObjectMapper objectMapper = new ObjectMapper();
```
Brutus5000: Why instantiate your own? This is very costly

src/main/java/com/faforever/client/discord/DiscordEventHandler.java
```diff
@@ -17,24 +17,37 @@
...
 
   public DiscordEventHandler(ApplicationEventPublisher applicationEventPublisher, NotificationService notificationService) {
     this.applicationEventPublisher = applicationEventPublisher;
     this.notificationService = notificationService;
     ready = this::onDiscordReady;
     disconnected = this::onDisconnected;
     errored = this::onError;
-    spectateGame = this::onSpectate;
-    joinGame = this::onJoinGame;
+    spectateGame = spectateSecret -> {
```
Brutus5000: why is the try catch not inside the onSpectate method?

src/main/java/com/faforever/client/discord/DiscordEventHandler.java
```diff
@@ -17,24 +17,37 @@
...
+      try {
+        onSpectate(spectateSecret);
+      } catch (JsonProcessingException e) {
+        log.warn("Could not discord spectate game", e);
+      }
+    };
+    joinGame = joinSecret -> {
```
Brutus5000: same here

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
```diff
@@ -395,76 +364,77 @@ public void stopSearchMatchmaker() {
...
+        .reason(reason)
+        .duration(duration)
+        .period(periodType.name())
+        .build());
   }
 
   @Override
   public void closePlayersGame(int playerId) {
-    writeToServer(new ClosePlayersFAMessage(playerId));
+    writeToServer(ClosePlayersFAMessage.builder().user_id(playerId).build());
```
Brutus5000: user_id ? please fix

src/main/java/com/faforever/client/remote/domain/RatingRange.java
```diff
@@ -1,37 +1,69 @@
...
 
-  private final Integer min;
-  private final Integer max;
+  Integer min;
+  Integer max;
 
   public RatingRange(Integer min, Integer max) {
     this.min = min;
     this.max = max;
   }
```
Brutus5000, line 19: obsolete

src/main/java/com/faforever/client/remote/domain/VictoryCondition.java
```diff
@@ -1,44 +1,16 @@
...
-    return victoryCondition;
-  }
-
-  public Object getValue() {
-    return value;
-  }
+  @JsonProperty("0") DEMORALIZATION,
+  @JsonProperty("1") DOMINATION,
+  @JsonProperty("2") ERADICATION,
+  @JsonProperty("3") SANDBOX,
```
Brutus5000: Can the int-value really be replaced with a String value?

src/main/java/com/faforever/client/remote/domain/inbound/faf/GameLaunchMessage.java
```diff
@@ -0,0 +1,33 @@
...
+  public static final String COMMAND = "game_launch";
+
+  /**
+   * Stores game launch arguments, like "/ratingcolor d8d8d8d8", "/numgames 236".
+   */
+  List<String> args;
+  int uid;
+  String mod;
+  String mapname;
+  @NonNull
```
Brutus5000: Does this @NonNull really make sense here? Is the rest allowed to be null?

micheljung:
I once wanted to do this but it ended up being a ton of work. Kudos to @Sheikah45 for going through with it ❤️

micheljung (dismissed):

micheljung:
I didn't look through it in detail but in general, I approve ;)

Brutus5000 (changes requested):
Some intermediate comments

src/main/java/com/faforever/client/remote/domain/VictoryCondition.java
Sheikah45: I tested and it works but I also have a better way to do it now so will change it

src/main/java/com/faforever/client/remote/FafServerAccessorImpl.java
Sheikah45: I believe this already was fixed but will double check

src/main/java/com/faforever/client/discord/DiscordEventHandler.java
Sheikah45: Because the java compiler complains about the exception not being caught here during the callable declaration

src/main/java/com/faforever/client/discord/DiscordEventHandler.java
Sheikah45: same as above

src/main/java/com/faforever/client/remote/domain/inbound/faf/GameLaunchMessage.java
Sheikah45: Yeah I don't think it makes much sense standing on its own

src/main/java/com/faforever/client/remote/domain/RatingRange.java
Sheikah45: Removed

src/main/java/com/faforever/client/remote/domain/VictoryCondition.java
Sheikah45: oh wait never mind the unknown ruins my other way

src/main/java/com/faforever/client/remote/domain/RatingRange.java
Sheikah45: Actually this has to be present because the normal constructor is not produced by Value and it is needed to instantiate with a null as one of the values

src/main/java/com/faforever/client/discord/DiscordEventHandler.java
Sheikah45: Yeah I will just use the autowired one there

src/main/java/com/faforever/client/remote/domain/VictoryCondition.java
Sheikah45: Bah nevermind I realize it doesn't work because the victory condition is read in java commons and that has a different type. Really the LocalReplayInfo probably should be in the commons library instead of here I think along with the VictoryCondition definition

Sheikah45:
Had to rebase with the oauth login so more changes needed now

src/main/java/com/faforever/client/discord/DiscordEventHandler.java
micheljung: This explanation doesn't make sense to me. If `onSpectate` handles the exception and doesn't declare it, the compiler won't complain here.

Sheikah45: Ah nevermind I misunderstood what Brutus was asking I understand now

Sheikah45:
Alright the oauth stuff was incorporated

PR #2310, opened 2021-07-16
bump ice adapter to 3.1.0 - relay latency check fallback, increased logging

Geosearchef:
Use fallback servers for estimating the latency in case the relay server doesn't respond to ICMP echo requests. This prevents usage of the defaut relay in case its latency would be too high but couldn't be estimated due to server misconfiguration. (which is currently the case for oceanians and prevents quite a lot of them from playing)

Also increases log verbosity by adding info about the available types of candidates and the chosen candidate pair after connectivity establishment to ease identification of issues in absence of JavaFX needed for the debug window.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 version=unspecified
 javafxPlatform=unspecified
-faf_ice_adapter_version=3.0.0
+faf_ice_adapter_version=3.1.0
 faf_uid_version=4.0.4
 springBootVersion=2.4.0
 springSecurityOauth2AutoconfigureVersion=2.3.4.RELEASE
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 version=unspecified
 javafxPlatform=unspecified
-faf_ice_adapter_version=3.1.0
+faf_ice_adapter_version=3.1.1
 faf_uid_version=4.0.4
 springBootVersion=2.4.0
 springSecurityOauth2AutoconfigureVersion=2.3.4.RELEASE


Sheikah45:
I feel like those tray icon messages are just going to be a nuisance to the user every time they start a game. They seem like messages that are better left in the log as they don't really have much info the end user can act on or really understand the consequences of.

Additionally what is the behavior if geosearchef.de was to go down or no longer be online?

Geosearchef:
When I talk about logging I mean logging to the log. The increased verbosity all goes to the file.

There's one tray icon message I added which is the ICMP echo fallback which is there so you actually know something is going wrong. Last time this went unnoticed for the past few months and oceanian players just wondered why they can't play or their games lag without anyone being aware of or fixing the issue.

About geosearchef.de in there, all of this is mainly a work around. Usually faforever.com should respond to ICMP echo messages, but the server admins had no time to fix that during the past months. So if that doesn't work, the adapter will fall back to test.faforever.com (beware,just for latency estimation, relaying still happens via the normal server). This currently works, there are players running a modified version of the adapter using the test server. And only if test also goes down, it will fallback to my server (simply cause it was the only one that I could think of located in the same data center and it's better than nothing, even if it goes down).

The right solution is to fix the server configuration, but I don't have any hope of that happening soon, so we need to find other solutions to allow players to well, play.

Geosearchef:
The tray icon currently tells you when it falls back and when it fails. What we could of course do is to just show the fail notification, not the fall back.

I suspect though that this will lead to the issue never getting fixed and becoming the new norm.

Brutus5000:
Where is the fallback server defined? Does it make sense to have the client pass it as an argument for different environments?

Geosearchef:
Currently it's hardcoded in the adapter. It just replaces faforever.com by test. faforever.com if faforever.com fails.

There's no reason to pass this from the server or client. This is a work around in case the server is misconfigured.

Sheikah45:
When I looked at the ice adapter commits there were two tray icon messages added and I am just wary of adding too many as people already hate the toast messages that pop up. 

It would seem like the log messages are enough.

Geosearchef:
In this case, no, a log message is not enough. We should remove the default "ice adapter has started" notification though.

Geosearchef:
I removed the "ice adapter started" notification which was long overdue, but I'm not happy at all with removing the only indicator that there is an error occuring.

Geosearchef:
As the main server now works again, I removed my server as a fallback.

Sheikah45 (approved):

Issue #1783, opened 2020-06-18
TMM: Add match confirmation (ready check) for matchmaker games


Askaholic:
For proposed protocol specification see: https://github.com/FAForever/server/issues/607 and https://github.com/FAForever/server/pull/608. The latter PR contains a feature complete implementation that is ready to be tested with a supporting client.

A feature which has already been requested a few times for 1v1 and will be especially relevant for bigger matchmaker games is having a "ready" check before the game starts. The server would send a `match_info` message which would trigger a popup in the client that looks something like: 

![MatchFound](https://user-images.githubusercontent.com/17941539/85073488-65481180-b1aa-11ea-8074-5d56f73da261.png)

Whenever someone clicks the `ready` button, a new `match_info` message will be sent so that the client can update the UI. Once everyone accepts the match, the launch sequence will happen as normal. If the timeout is reached before everyone accepts, then the match cancellation sequence will happen and the party will be entered back into the queue.

**Wanna have the bug fixed quickly?**
Visit Issue hunt...
[![Issue hunt](https://github.com/BoostIO/issuehunt-materials/raw/master/v1/issuehunt-button-v1.svg?sanitize=true)](https://issuehunt.io/r/FAForever/downlords-faf-client)

RoxyBananaSmoothie:
Hi there, appreciate you guys probably already have a long list of things to fix. I just want to say that this "ready to play" function is absolutely essential for the 2 v 2 ladder. We can't really expect people to stay and wait in the queue for hours without leaving the computer.  My friend was in a situation that he was in a queue for a few hours, and the game launched while he was away from his desk. His team mate somehow didn't realise he was afk for 4..5 mins into the game (I know it's strange that he didn't notice within the first minute). Now he blocked my friend and threaten to report him...

bukajsytlos:
If you leave your PC, you should leave queue too.

RoxyBananaSmoothie:
Technically, yes. In reality, you can easily be on your PC and still miss the launch if you have been in a queue for hours and forgotten about it. That's easily done in my opinion.

Katharsas:
Current status:
Implemented in server PR https://github.com/FAForever/server/pull/608 as described in https://github.com/FAForever/server/issues/607 .
With that server PR, the server will only start a game with `game_launch` when all players have sent `match_ready`, and the client cannot send `match_ready` without the PR because that would close the connection. Therefore we need a two stage implementation on the server:

[...]

PR #2324, opened 2021-08-11
Send `match_ready` to server immediatley on `match_found` (#1783)

Katharsas:
Since i do not have a full local environment i can only start testing this when https://github.com/FAForever/server/pull/820 is released on test or prod.

Waiting for clarification on https://github.com/FAForever/server/issues/607#issuecomment-897205805


Sheikah45:
the ready party message is already implemented. You should just use that.

Katharsas:
But this is not the ready party message, its an entirely different thing in terms of meaning and potential evolution? I don't think those two should be mixed up.

Sheikah45:
The new message needs to be implemented in the lobby client

Sheikah45:
This is out of date now

PR #2334, opened 2021-08-27
Feature/#2333 status indicator

micheljung:
If a service is offline AND there is a scheduled announcement active, only the announcement is shown:

image

If a service is offline and there are is no active announcement, the offline services are shown:

image

They could surely be improved visually and we could differentiate between an upcoming and an active announcement but for now, that will do.


Sheikah45:
I saw a commit widening the loginWebView but that shouldn't be done. It was kept narrow because for users with high resolution screens it tried to load the background image at too high a resolution and crashed the client.

It is an issue with the graphics library used in the webview.

micheljung:
@Sheikah45 

Thx for the heads up. I reverted it already but know I know why it was there :) The embedded web view will be gone anyways with #2335

build.gradle
```diff
@@ -328,3 +328,5 @@ dependencies {
 
   codacy("com.github.codacy:codacy-coverage-reporter:-SNAPSHOT")
 }
+
+task copyUpdaterToResources{}
```
Sheikah45: Was this meant to be added?

src/main/java/com/faforever/client/status/Incident.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Value;
+
+import java.time.OffsetDateTime;
+
+@Value
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Incident {
```
Sheikah45, line 15: Can the class names be of the status dtos be a little more descriptive of what they are and what they are. Right now they are a little easy to get confused with other classes or share names with "meta" classes

src/main/java/com/faforever/client/status/Message.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Value;
+
+import java.time.OffsetDateTime;
+
+@Value
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Message {
```
Sheikah45, line 15: Maybe a more descriptive name

src/main/java/com/faforever/client/status/Service.java
```diff
@@ -0,0 +1,30 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Value;
+
+import java.time.OffsetDateTime;
+import java.util.List;
+
+@Value
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Service {
```
Sheikah45, line 13: Maybe a more descriptive name

src/main/java/com/faforever/client/status/ServiceStatusService.java
```diff
@@ -0,0 +1,33 @@
...
+import com.faforever.client.config.ClientProperties;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.web.reactive.function.client.WebClient;
+import org.springframework.web.reactive.function.client.WebClient.Builder;
+import reactor.core.publisher.Flux;
+
+/** REST client for statping-ng API. */
+@org.springframework.stereotype.Service
+@Lazy
+public class ServiceStatusService {
```
Sheikah45: Maybe just StatPingService? ServiceStatusService just is a little confusing as to whether that is the server services or the client services

Sheikah45 (changes requested):
Looking good. Although I noticed some of the names are a little generic and could lead to confusion down the line so would be great if those were made a little clearer.

Also one thing to consider is that for players who have auto login enabled none of this will be shown unless services affecting login are down and errors out the login process.

Which, thinking about it are probably the only services players care about....

micheljung:
@1-alex98 are you admin on Codacy? Can you configure the JUnit5 Method name pattern? I'd commit a PMD `ruleset.xml` to configure it but I fear that it won't just configure this one rule, [but make this rule the **only one**](https://pmd.sourceforge.io/pmd-5.1.1/howtomakearuleset.html):

`.codacy/ruleset.xml`:
```
<?xml version="1.0"?>

<ruleset name="Customized Rules"
         xmlns="http://pmd.sourceforge.net/ruleset/2.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://pmd.sourceforge.net/ruleset/2.0.0 https://pmd.sourceforge.io/ruleset_2_0_0.xsd">

    <rule ref="category/java/codestyle.xml/MethodNamingConventions">
        <properties>
            <property name="methodPattern" value="[a-z][a-zA-Z0-9]*" />
            <property name="staticPattern" value="[a-z][a-zA-Z0-9]*" />
            <property name="nativePattern" value="[a-z][a-zA-Z0-9]*" />
            <property name="junit3TestPattern" value="test[A-Z0-9][a-zA-Z0-9_]*" />
            <property name="junit4TestPattern" value="[a-z][a-zA-Z0-9_]*" />
            <property name="junit5TestPattern" value="[a-z][a-zA-Z0-9_]*" />
        </properties>
    </rule>

</ruleset>
```

It's [commonly accepted to break the Java method naming convention for unit tests](https://dzone.com/articles/7-popular-unit-test-naming) and use `_` to make them more readable.

build.gradle
micheljung: Thx; no, you're right, this needs to be removed. My IntelliJ somehow requires this and I have no idea where I can fix it :(

src/main/java/com/faforever/client/status/Incident.java
micheljung: Do you have a suggestion for a class name that represents [the `Incident` API object](https://docs.statping.com/#folder-incidents), other than `Incident`? We could call it `StatusPingIncident` but that's rather redundant and an anti-pattern because the package makes it clear what this class belongs to.

I could add a Javadoc but other than "Represents an Incident" and a link to https://docs.statping.com/#folder-incidents I wouldn't know what to explain.

micheljung: Alright, I gave it a try

micheljung:
> Also one thing to consider is that for players who have auto login enabled none of this will be shown unless services affecting login are down and errors out the login process.
> 
> Which, thinking about it are probably the only services players care about....

That's a good point. We might add such notifications elsewhere in the future but I think the most important case is when users can't log in.
We could also add periodic checks but that would cause quite some useless network traffic. A more efficient option would be to have only the lobby server poll StatsPing and notify all clients as needed.

But for now, I think this is a good step forward :)

src/main/java/com/faforever/client/status/Message.java
micheljung: Wouldn't know what but I added a JavaDoc

src/main/java/com/faforever/client/status/Service.java
micheljung: Wouldn't know what but I added a JavaDoc

src/main/java/com/faforever/client/status/ServiceStatusService.java
micheljung: Renamed it to StatPingService. Wanted to keep the name technology-agnostic but it's StatPing-specific anyway.

src/main/java/com/faforever/client/status/Service.java
Sheikah45: Yeah let me think of what might fit. The javadoc might solve the issue. But just want to avoid people seeing it in the future and thinking it is a meta class for the spring Services when really it is just a dto.

Sheikah45 (dismissed):
Do we need to wait until the stat ping is live?

The names should be good with the included javaDoc

micheljung:
> Do we need to wait until the stat ping is live?

Fortunately not, it throws an exception but reactor silently swallows it,

The client needs to handle StatPing not being available anyways.

Sheikah45:
When I started this up it seems there was a change that messed with the alignment of the webview and port specifers even when invisible.

The webview should be filling the vertical space and the fields should be adjacent

image

image

Sheikah45:
It would probably be good to add error logging to the StatPingService and properly return a Flux.empty()

src/main/java/com/faforever/client/status/StatPingService.java
```diff
@@ -0,0 +1,33 @@
...
+  public Flux<Service> getServices() {
+    return getMany("/services", Service.class);
+  }
+
+  public Flux<Message> getMessages() {
+    return getMany("/messages", Message.class);
+  }
+
+  private <T> Flux<T> getMany(String path, Class<T> type) {
+    return webClient.get().uri(path).retrieve().bodyToFlux(type);
```
Sheikah45: It would probably be good to add error logging and properly return a Flux.empty() using onErrorResume

micheljung: Added it

micheljung:
> The webview should be filling the vertical space and the fields should be adjacent

Fixed:
image

micheljung:
Btw I now use https://status.faf.micheljung.ch/api until @Brutus5000 can provide everything required to officially host StatPing.

micheljung:
> @1-alex98 are you admin on Codacy? Can you configure the JUnit5 Method name pattern? I'd commit a PMD `ruleset.xml` to configure it but I fear that it won't just configure this one rule, [but make this rule the **only one**](https://pmd.sourceforge.io/pmd-5.1.1/howtomakearuleset.html):

@1-alex98 did you see my comment?

Sheikah45:
Was there anything here that needed to change? I know we ran into some latent issues when looking at the update branch but wasn't sure which pr it was

micheljung:
As far as I'm aware, this is ready to be merged.

Sheikah45 (approved):

micheljung:
Why did GitHub create a new commit with the pull request's title as commit message? 😠

Sheikah45:
I am not sure. Normally it uses the first commit as the commit name

Issue #2333, opened 2021-08-26
Add service status indicators


micheljung:
Users should be notified when services are down. #2330 didn't help with that but it made room for something better :)

Therefore, I set up https://github.com/FAForever/faf-status (which isn't officially deployed yet but available under https://status.faf.micheljung.ch) which comes with [an API](https://docs.statping.com/#doc-api-detail), finally allowing us to display status information within the client.

That's what I'm going to do.

micheljung:
Fixed by #2334

PR #2338, opened 2021-09-01
Replace embedded login web view with login via browser

micheljung:
Builds on top of #2333 

image
image
image
image
image


src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -396,6 +297,49 @@ public void onDownloadUpdateButtonClicked() {
...
+        .setRemotePort(Integer.parseInt(replayServerPortField.getText()));
+
+    clientProperties.getIrc()
+        .setHost(ircServerHostField.getText())
+        .setPort(Integer.parseInt(ircServerPortField.getText()));
+
+    clientProperties.getApi().setBaseUrl(apiBaseUrlField.getText());
+    clientProperties.getOauth().setBaseUrl(oauthBaseUrlField.getText());
+
+    OAuthValuesReceiver oAuthValuesReceiver = applicationContext.getBean(OAuthValuesReceiver.class);
```
Sheikah45, line 331: Why retrieve this via the application context rather than the constructor?

src/main/java/com/faforever/client/update/ClientConfiguration.java
```diff
@@ -55,4 +58,9 @@
     private String message;
     private URL releaseNotesUrl;
   }
+
+  @Data
+  public static class OAuth {
+    private Collection<URI> redirectUris;
```
Sheikah45: Shouldn't these be apart of the ServerEndpoints object as the could be different depending on the endpoint used

src/main/resources/i18n/messages_fr.properties
```diff
@@ -1104,7 +1104,6 @@ map.all = Toutes versions
 map.current = Version courante
 login.remember = Se souvenir de moi
 login.failed = Erreur durant la connexion
-login.badState = L'état retourné par le service utilisateur ne correspond pas à l'état initial. Si le problème persiste, veuillez solliciter l'aide technique sur le forum ou sur le canal Discord.
```
Sheikah45: We shouldn't manually change this file

src/main/resources/i18n/messages_zh_TW.properties
```diff
@@ -1061,7 +1061,6 @@ map.all = 歷史版本
 map.current = 目前版本
 login.remember = 記住我
 login.failed = 登入時發生錯誤
-login.badState = State returned by user service does not match initial state if this continues to occur please reach out to technical help on the forum or discord channel
```
Sheikah45: Shouldn't manually change

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  public CompletableFuture<URI> getRedirectUri() {
+    return addressFuture;
+  }
+
+  @Value
+  public static class Values {
+    String code;
+    String sate;
```
Sheikah45: state

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
```diff
@@ -0,0 +1,134 @@
...
+          .append("\r\n")
+          .append(html);
+    }
+  }
+
+  private Values readValues(Socket socket) throws IOException {
+    // Don't use try-with-resources since socket must not be closed yet
+    BufferedReader reader = new BufferedReader(new InputStreamReader(socket.getInputStream()));
+    String query = reader.readLine();
+    String code = extractValue(query, CODE_PATTERN);
```
Sheikah45: If this is just parsing the URL wouldn't it be simpler to just use the URLEncodeUtils to parse the url and query parameters rather than have the custom regex

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
```diff
@@ -0,0 +1,134 @@
...
+    JavaFxUtil.assertBackgroundThread();
+    if (clientProperties.isUseRemotePreferences()) {
+      return preferencesService.getRemotePreferences().getOAuth().getRedirectUris().stream()
+          .filter(uri -> uri.getPort() > 0)
+          .filter(uri -> ALLOWED_HOSTS.contains(uri.getHost()))
+          .collect(Collectors.toList());
+    }
+    return IntStream.range(60000, 65535)
+        .mapToObj(port -> URI.create("http://localhost:%d/".formatted(port)))
+        .collect(Collectors.toList());
```
Sheikah45: Does this just scan all possible ports? What use case is this for?

Sheikah45 (changes requested):

src/main/java/com/faforever/client/login/LoginController.java
micheljung: Because it's not a singleton but a prototype bean:
```java
@Component
@RequiredArgsConstructor
@Slf4j
public class OAuthValuesReceiver {
```
which is needed because it's stateful.

If it was injected via constructor, LoginController would receive only once instance and in order for it to be able to re-use this instance, the value receiver would need to be mutable (resetting `addressFuture`). Immutability, however, makes it easier to write, understand and use.

src/main/java/com/faforever/client/update/ClientConfiguration.java
micheljung: I also thought about this but it's a bit tricky; the redirect URIs are not server endpoints as in "where do I need to connect to". They are always `localhost` with some "random" port. But since the port must not be random but whitelisted by the OAuth server, the client needs to know which ports (or rather URIs), are whitelisted.

I added a JavaDoc.

src/main/resources/i18n/messages_fr.properties
micheljung: Reverted

src/main/resources/i18n/messages_zh_TW.properties
micheljung: Reverted

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
micheljung: fixed, thx

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
micheljung: Agree, however, `URLEncodeUtils` doesn't fit since it's a client-side library to "encode" URLs whereas this is server-side parsing of an HTTP request: `GET /?code=1234&state=abcd HTTP/1.1`. I'm not aware of any library that can easily parse this. I could use Spring classes to start an embedded Tomcat or so but that seems to be total overkill.

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
micheljung: If remote preferences aren't used, the server can't tell us which redirect URIs/ports are whitelisted (as in the comment above). This is the case when running the client in offline mode (where there is no OAuth anyway) or in unit-tests where the port doesn't matter; the list just needs to be long enough so that a free port can be found.

Spring provides `org.springframework.util.SocketUtils#findAvailableTcpPort()` which would seem perfect, however, from experience I know that it is not reliable: it tries to open a port and if that's successful, it will close the port and return its number, assuming that it will still be free when the caller uses it to open a port - for which there is no guarantee.

I added a comment.

src/main/java/com/faforever/client/update/ClientConfiguration.java
Sheikah45: Yes but in terms of configuration these ports could be different between the test and production server so having a single value across both might not work

micheljung:
Thanks for the review.

Other things I changed:
* Localize `login_success.html`
* Unit test `OAuthValuesReceiver`

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
Sheikah45: If this is based around unit tests then wold it not be better to define the port in the test and then mock the client configuration to return the uris with the opened port. And in the case of us being offline then this should never be called anyway.

src/main/java/com/faforever/client/update/ClientConfiguration.java
micheljung: That is true, but how would moving it to `ServerEndpoints` fix this? Either way it's read from the exact same configuration file.

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
micheljung: How can a unit test know which port will be free to open?

Sheikah45: We can just create the same list of ports in the unit test and hand it over by mocking PreferencesService

src/main/resources/i18n/messages.properties
```diff
@@ -225,43 +225,43 @@ game.chat.message = Message
...
-vault.replays.newest = Most Recent
-vault.replays.highestRated = Highest Rated
-vault.replays.displayQuery = Display search query
-vault.replays.queryPrompt = Build a query above or enter one manually
-vault.replays.searchError = An error has occurred while searching for replays.
+vault.replays.newest=Most Recent
+vault.replays.highestRated=Highest Rated
+vault.replays.displayQuery=Display search query
+vault.replays.queryPrompt=Build a query above or enter one manually
+vault.replays.searchError=An error has occurred while searching for replays.
```
micheljung: Such changes need to be reverted

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
micheljung: Fair enough. And what should the `OAuthValuesReceiver` do if `clientProperties.isUseRemotePreferences()` is false? That doesn't necessarily translate to offline mode.

Sheikah45: In that case the values would have to be input into a nominal redirect_uris textbox like the other endpoints are on the login screen. So a method to specify those manually needs to be added.

micheljung: I added it to the login form

1-alex98:
nice pr

src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -396,6 +302,49 @@ public void onDownloadUpdateButtonClicked() {
...
+    });
+    oAuthValuesReceiver.receiveValues().thenAccept(values -> {
+      String actualState = values.getState();
+      String expectedState = userService.getState();
+      if (!expectedState.equals(actualState)) {
+        handleInvalidSate(actualState, expectedState);
+        return;
+      }
+      loginWithCode(values.getCode());
+    });
```
1-alex98: Exception silently swallowed. I imagine port being blocked could happen and would be good if user got a message instead of nothing happening

1-alex98: .expectionally()

src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -396,6 +302,49 @@ public void onDownloadUpdateButtonClicked() {
...
+        .setPort(Integer.parseInt(ircServerPortField.getText()));
+
+    clientProperties.getApi().setBaseUrl(apiBaseUrlField.getText());
+    clientProperties.getOauth().setBaseUrl(oauthBaseUrlField.getText());
+
+    OAuthValuesReceiver oAuthValuesReceiver = applicationContext.getBean(OAuthValuesReceiver.class);
+    oAuthValuesReceiver.getRedirectUri().thenAccept(redirectUri -> {
+      clientProperties.getOauth().setRedirectUri(redirectUri);
+      platformService.showDocument(userService.getHydraUrl());
+    });
```
1-alex98: .expectionally here as well 🤔

1-alex98 (changes requested):
Looks good

Sheikah45:
@micheljung are you still working on this? Conflicts need to be resolved as well as the error notification alex mentioned

micheljung:
@Sheikah45 yes, will push the fixes soon

Sheikah45:
Closing as it is stale reopen if worked on

src/main/java/com/faforever/client/login/LoginController.java
micheljung: Fixed

src/main/java/com/faforever/client/update/readme.md
```diff
@@ -0,0 +1,6 @@
+# How it works
+
+When 
+
+`Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\2848-7798-9769-5013
+
```
Sheikah45: Is this an accidental addition?

src/main/java/com/faforever/client/config/ClientProperties.java
```diff
@@ -218,7 +220,7 @@ public void updateFromEndpoint(ServerEndpoints serverEndpoints) {
   @Data
   public static class Oauth {
     private String baseUrl;
-    private String redirectUrl;
+    private URI fallbackRedirectUri = URI.create("");
```
Sheikah45: What is the purpose of this fallback?

Right now none are defined in the properties. Additionally with the usage of specific ports it is unlikely any hard coded fallback can be used without a controlled browser.

src/main/java/com/faforever/client/update/readme.md
micheljung: Yes, deleted it already

src/main/java/com/faforever/client/config/ClientProperties.java
micheljung: Per your request :) https://github.com/FAForever/downlords-faf-client/pull/2338#discussion_r708171890

src/main/java/com/faforever/client/login/LoginController.java
micheljung: Added

src/main/java/com/faforever/client/config/ClientProperties.java
Sheikah45: Ah gotcha the name threw me off. I understand now. I might not call it fallback though, it is supposed to be for inputting parameters for local setup testing.

micheljung: Yeah it's complicated. I added a JavaDoc. Couldn't think of a better name so far.

micheljung:
@Sheikah45 can you please review and resolve the open conversations? If there's nothing left we can merge this.

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
```diff
@@ -0,0 +1,147 @@
...
+      OAuthEndpoint endpoint = oAuthEndpoint.orElseThrow(() -> new IllegalStateException("No endpoint has been specified"));
+      return endpoint.getRedirectUris().stream()
+          .filter(uri -> uri.getPort() > 0)
+          .filter(uri -> ALLOWED_HOSTS.contains(uri.getHost()))
+          .collect(Collectors.toList());
+    }
+    if (fallbackRedirectUri.toASCIIString().isBlank()) {
+      throw new NoFallbackRedirectUriException();
+    }
+    return Collections.singletonList(fallbackRedirectUri);
```
Sheikah45: We probably want to switch this so that the specified url takes priority over the remote preferences. Otherwise it makes testing on a local setup harder as it requires actively disabling the remote preferences check.

Sheikah45 (changes requested):
I think the current "fallback" which is specified in the textbox should be the priority as it is with the rest of the environment text boxes. 

As it is now it would be impossible to run the client using a localhost for testing unless you change the spring profile to disable the remote preference checks.

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
micheljung: How about now?

Sheikah45: Works

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
```diff
@@ -0,0 +1,155 @@
...
+          return readWithUri(uri);
+        } catch (SocketException e) {
+          log.warn("Port " + uri.getPort() + " is probably already in use", e);
+        }
+      }
+      throw new IllegalStateException("Could not read from any redirect URI: " + redirectUris);
+    });
+  }
+
+  @SneakyThrows
```
Sheikah45: This is not needed as nothing throws a checked exception here that I can see

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
```diff
@@ -0,0 +1,155 @@
...
+
+    if (redirectUris.isEmpty()) {
+      throw new NoRedirectUriException("No valid redirect URI has been provided by endpoint: " + endpoint.getUrl());
+    }
+
+    return redirectUris;
+  }
+
+  @SneakyThrows
+  private Values readWithUri(URI uri) throws SocketException {
```
Sheikah45, line 97: What is throwing the SocketException? All the methods I saw specified just IOExceptions. Why not just throw the IOException and catch that? Then we can remove the SneakyThrows and has basically the same effect.

Sheikah45 (changes requested):
Just saw the sneaky throws and had some questions

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
micheljung: Trying to open a socket on a port that is already occupied throws a SocketException. If I throw IOException, the caller is forced to handle which is undesired because it only wants to handle SocketException.

Sheikah45: I would still think it might be better to also handle the IOexception should that happen. But since it is inside a completableFuture call then I guess it is okay.

I am just not a big fan of sneaky throws.

Sheikah45: Most of my thinking or reasoning is that especially on these more network based things I have had too many experiences of sneaky throws just swallowing connection issues and the like and then the user is just left stranded.

Although hopefully in this case it is caught by the exceptionally.

micheljung: Checked exceptions were arguably a bad idea, which is also why they don't exist e.g. in Kotlin. `@SneakyThrows` just kind of retro-fits this so you might want to rethink your position ;-)

In case of a blocked port, I need to continue looking for another port. In case of any other IOException I don't want to keep looking but let the exception propagate.

I do see your point but the problem is not `@SneakyThrows` but actually, JavaFX not having an unhandled exception handler. The very same thing can happen with NPEs.

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
micheljung: Good catch

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
Sheikah45: Yep understood they are just different bandaids over a Library problem

Sheikah45: If it wasn't called primarily from a completable future I would say wrap it in one of the notifiable exceptions but unfortunately those don't work inside completable futures.

Sheikah45 (approved):

ghost:
How does it work? Can you shortly describe the process? Like:
user clicking -> client redirect to auth -> ... -> how does it catch authorization process?

Brutus5000:
The client starts a webserver on localhost and the oauth flow redirects there .

ghost:
Thanks, i get it

PR #2339, opened 2021-09-01
Create new gameoptions command line switch

### LINKED ISSUE #2345: Enable client to relay game options arguments
The server will send game options for automatches with different options depending on the queue. The client needs to process these and relay them to the game when it is starting.

BlackYps:
Is this actually all that is needed? It seems so easy.
Depends on https://github.com/FAForever/faf-java-commons/pull/69

closes #2345


Sheikah45:
No you would need to look at the relevant pr in the fa repo. It looked like Askaholic was trying to make each argument have its own trigger

Sheikah45:
see https://github.com/FAForever/fa/pull/3385

Sheikah45:
Oh apparently I am not up to date on that pr maybe it is that easy but would need testing

BlackYps:
From my understanding we wanted to have one command line trigger followed by all arguments, because that would be easier.

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -540,6 +540,7 @@ public Game getByUid(int uid) {
               gameLaunchMessage.getArgs().add("/team " + gameLaunchMessage.getTeam());
               gameLaunchMessage.getArgs().add("/players " + gameLaunchMessage.getExpectedPlayers());
               gameLaunchMessage.getArgs().add("/startspot " + gameLaunchMessage.getMapPosition());
+              gameLaunchMessage.getArgs().add("/gameoptions " + gameLaunchMessage.getGameOptions());
```
Askaholic: How does java handle `String + Map`? I have a feeling there needs to be some conversion happening here to construct the string correctly.

Sheikah45: Probably need to use a Collectors.joining

Sheikah45:
Also why is the server sending the options if they are persisted in the db? Wouldn't it make sense to get the params for the queue from the api maybe?

Sheikah45:
Although that might require more changes but just a thought

BlackYps:
I think it's nice to have the possibility that the server can modify the commands. And it makes the most sense to me to pack everything in the same message, instead of having to get one piece of the information from a different location.

BlackYps:
Wait, the client is already gathering the rest of the queue information from the api anyway. Idk, I'll let @Askaholic decide what he thinks is the better approach. He also knows better what he needs for galactic war.

Askaholic:
This way is more flexible, it allows the queue to make different game setups on a per game basis. This isn’t currently implemented but I’m thinking about future expansions like a MOTW queue that changes its settings dynamically.

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -540,7 +541,10 @@ public Game getByUid(int uid) {
               gameLaunchMessage.getArgs().add("/team " + gameLaunchMessage.getTeam());
               gameLaunchMessage.getArgs().add("/players " + gameLaunchMessage.getExpectedPlayers());
               gameLaunchMessage.getArgs().add("/startspot " + gameLaunchMessage.getMapPosition());
-              gameLaunchMessage.getArgs().add("/gameoptions " + gameLaunchMessage.getGameOptions());
+              Map<String, String> gameOptions = gameLaunchMessage.getGameOptions();
+              gameLaunchMessage.getArgs().add("/gameoptions " + gameOptions.keySet().stream()
```
Sheikah45: Use entrySet over keySet. EntrySet provides an object that has both the key and associated value.

BlackYps:
Do I see this correctly that there is no test yet, that tests the command line options for non-empty values?

Sheikah45:
probably

Sheikah45:
There were no generic game options like this before

src/main/java/com/faforever/client/game/GameService.java
BlackYps: Like this?
```java
gameLaunchMessage.getArgs().add("/gameoptions " + gameLaunchMessage.getGameOptions().entrySet().stream()
                  .map(entry -> entry.getKey() + ":" + entry.getValue())
                  .collect(Collectors.joining(" ")));

File: src/test/java/com/faforever/client/game/GameServiceTest.java [hunk 3]
```diff
@@ -669,6 +670,24 @@ public void startSearchMatchmaker() throws IOException {
...
+    Map<String, String> gameOptions = new HashMap<>();
+    gameOptions.put("Share", "ShareUntilDeath");
+    gameOptions.put("UnitCap", "500");
+    GameLaunchResponse gameLaunchMessage = GameLaunchMessageTestBuilder.create().defaultValues().team(1).expectedPlayers(4).mapPosition(3).gameOptions(gameOptions).get();
+    when(gameUpdater.update(any(), any(), any(), any())).thenReturn(completedFuture(null));
+    when(mapService.download(gameLaunchMessage.getMapName())).thenReturn(completedFuture(null));
+    when(fafService.startSearchMatchmaker()).thenReturn(completedFuture(gameLaunchMessage));
+    instance.startSearchMatchmaker();
+    verify(forgedAllianceService).startGame(
+        gameLaunchMessage.getUid(), null, List.of("/team", "1", "/players", "4", "/startspot", "3", "/gameoptions", "Share:ShareUntilDeath UnitCap:500"),
```
BlackYps: Whats a good way to also allow `"UnitCap:500 Share:ShareUntilDeath"`?

Askaholic: Does Java have an ordered map? You could use that instead of a HashMap.

BlackYps: Seems like using LinkedHashMap in the kotlin file is the way to go.

Sheikah45: Does order actually matter? If it doesn't then I would not use a Linked Hash map.

There is a contains in any order matcher as well

Askaholic: oh btw I think the game is expecting the args to look like `["/gameoption", "UnitCap:500", "Share:ShareUntilDeath"]` not `["/gameoption", "UnitCap:500 Share:ShareUntilDeath"]`

Sheikah45: What is the difference there they just get appended onto the command line arg with a space as a joiner. I would prefer the second as that follows the pattern of key and then value rather than having a multivalued key.

Sheikah45: On the verification you can also use an argument captor and do the verification after the verify call

Askaholic: The difference is that one has 2 command line args while the other has 3. As it stands the game doesn’t try to split the args by space, just by colon. I mean maybe we could add that but imo just pass the args correctly.

Sheikah45: I think it would be better to keep it as a single key and single value to maintain the same pattern as all the other argument pairs.

BlackYps: Currently the client splits all arguments at spaces: https://github.com/FAForever/downlords-faf-client/blob/221e3fb2309212167304c3977087bb652337b890/src/main/java/com/faforever/client/game/GameService.java#L612

Sheikah45: Maybe although I don't think that field is actually used but I guess it is okay

BlackYps:
The checks will only pass once https://github.com/FAForever/faf-java-commons/pull/69 is merged. Then it should work.

1-alex98:
@Sheikah45 Probably this thing should be released asap... The thing is we will need to release it and then force all players onto that version if we want to make sure everybody sets the correct options if he is host. Since it does also work if the server does not send the options. Consider merging it even tho other projects might not be ready ;)

1-alex98:
Because the sooner it is out the sooner we can think about starting to use it.

Sheikah45:
Yes I just haven't gotten word if this is the finalized format yet

Sheikah45:
Also the commons pr needs to be merged first and that one assumes everything will be strings when in the server that is not the case so that needs to be resolved first

BlackYps:
We tested that with Brutus yesterday and the parsing to strings works.

BlackYps:
We used this branch to do our tests with the game, so it is well tested and everything works. There are no further changes planned.

Sheikah45 (approved):

PR #2344, opened 2021-09-11
update to javafx17

Sheikah45:
This is the next LTS version of javafx so we will likely not update beyond this for a while unless there is some feature we desperately need


Sheikah45:
Hmm I forgot the transparent webview is getting fixed in javafx 18 guess we wont be on this long

Brutus5000:
JavaFX is not Java. There is no LTS release coming.
Also why close? Version 17 still improves the WebView on Linux.

Sheikah45:
Well they say this on the website. Although it might mean more commercially.
`JavaFX 17 is the latest release of JavaFX. JavaFX 17 is the second long term support release after JavaFX 11`
https://gluonhq.com/products/javafx/

And there were some larger changes with javaFX17 so I wasn't sure if I wanted to go through the effort now to update it or wait for javaFX 18 to do all that since it has the real value with the transparent webView

micheljung:
What Brutus meant is, JavaFX is not included in Java but a dependency that we can update independently of our Java version. Probably.

Brutus5000:
No that's not what I meant, but interesting to know. But it makes no difference. Similar to Oracle LTS, I would assume the Gluon JavaFX 17 LTS patches will also be available to paying customers only.

Sheikah45:
This would need updating for some of the css of some elements as there were changes for some in javaFX 17

Sheikah45:
Need to fix the pagination. Looks like this with javafx17
image

Brutus5000:
@Sheikah45 Any updates on this?

Sheikah45:
For this javafx17 branch it just changes some of the styles of the client so I need to go in and figure out what the css needs to be changed to, to make it look the same again.

Sheikah45:
Superceded by #2623

PR #2353, opened 2021-09-22
Add ability to start games offline and remove mock services

Sheikah45:
This adds a play offline button to the login screen so players can just play offline with their current faf version.

Also removes all the mock services associated with the offline profile as they were never used.


Sheikah45:
Might be a good idea to add a check for a valid FAF Data directory and if it doesn't exist then tell them they need to login at least once or download it or something

Askaholic:
No auto downloads for offline play right?

Sheikah45:
Correct it just uses what ever is already downloaded

Sheikah45:
If the FAF exe doesn't exist I might just launch the steam one that they have. That or throw an error. Not sure which is better

BlackYps:
I don't know if launching the steam version is a good idea. That makes it even less obvious that FAF intents to use its own custom game version. I would prefer to throw an error that explains to the user that FAF is meant to download a separate, differently balanced version.

Sheikah45:
Yeah that probably makes the most sense

src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -440,7 +441,16 @@ private void showLoginProgess() {
   }
 
   public void onPlayOfflineButtonClicked() {
-    gameService.startGameOffline();
+    try {
+      gameService.startGameOffline();
+    } catch (Exception e) {
+      if (e.getCause() instanceof IOException) {
+        notificationService.addImmediateWarnNotification("offline.noExe");
```
BlackYps, line 448: If I remove the exe I don't get the error message, but instead I get prompted to locate the forged alliance directory.

Sheikah45: Can you verify that the game path didn't get removed from your prefs by some test. You will get the game directory pop up if it doesn't know where the supcom install directory is. This logic happens in the game service.

BlackYps: Seems like it happened because I just have the exe on linux to trick the client. I just tested it on windows with a regular faf install and there it works as expected when I rename the exe.

BlackYps (approved):

PR #2373, opened 2021-09-27
Implement improved error logging and user notifications

### LINKED ISSUE #2368: Remove usage of @SneakyThrows and NoCatch
As described in #2367 the error handling in the client is rather poor with just showing the stack trace to the user. At the moment this is mostly due to using noCatch and SneakyThrows which wraps any exception into a Runtime removing our ability to properly handle errors. 

To start we should properly check exceptions and remove usages of SneakyThrows and noCatch in the client

**Describe alternatives you've considered**
There is also a possibility of using a global error handler but would like require more error trace parsing rather than a more local solution

Sheikah45:
Closes #2368


micheljung:
Sorry for your lost time but I strongly oppose this. Catching exceptions, logging a warn message and returning null and such things is an antipattern. Please read https://dzone.com/articles/need-robust-software-make-it-fragile-1

Sheikah45:
Yes I can understand the return and can replace those with a rethrow of a non checked exception. 

Although would still prefer to catch and rethrow the unchecked exception because as it is now those exceptions are never actually logged during run as far as I can tell so this allows us to log the correct exception.

micheljung:
If you replace `@SneakyThrows` or NoCatch with an unchecked exception, you're doing pretty much exactly what those two methods already do but with more code.

Exceptions should always be logged where they are handled, not where they are thrown or wrapped.

Sheikah45:
Yes but the issue is that many of the checked exceptions that were being wrapped were throwing to either the platform application runner which didn't log them or they weren't logged because the higher methods which were run didn't have a catch statement for them due because they were unaware they could be thrown.

Sheikah45:
Also most of these exceptions are io exceptions where it would be better to just notify the user about file access permissions. 

In the current for some of these the issue is just that the exception occurred but was never logged and never notified the user. Like this was the case with the missing avatar icons for a while. There was an issue with the url encoding but since the exception was never logged or caught and users never got a notification then it was actually pretty hard to track down why some avatars just never showed up.

micheljung:
Swallowed exceptions are bad, I agree. However, the solution we need is one that makes sure that uncaught exceptions are somehow always logged, rather than logging exceptions where they happen (which [is a mistake](https://stackify.com/common-mistakes-handling-java-exception#mistake3)).

For instance, in the case of an unchecked exception like `NullPointerException`, having to remember putting try-catch around every operation is worse than having to remember to implement an `exceptionally()`. Handling uncaught exceptions solves this, too.

If there is no way for us to implement an uncaught exception handler for JavaFX and Completable future, then our next best option is to configure static code analysis. This could be to disallow `thenApply` and force `handle` instead or something alike.

micheljung:
I don't know if `Thread.setDefaultUncaughtExceptionHandler` would be enough. Briefly looking at [JDK-8100937](https://bugs.openjdk.java.net/browse/JDK-8100937) looks like it should. As for CompletableFuture, [this answer](https://stackoverflow.com/a/62684489/513451) says there is no way except some ugly hack.

Brutus5000:
My suggestion was to remap the checked exceptions with business-runtime exceptions we can later check on.

Sheikah45:
Yes although even wrapping them in business runtime exceptions doesn't necessarily solve the swallowed exception issue unless I am missing something.

Unless we add catches for them everywhere

Brutus5000:
I'm not exactly sure how the current "Uh oh" error dialog is working, but in best case it's like a global exception handler and you can put some logic in there.

Sheikah45:
Currently that has to be explicitly called by the notificationService wherever the error is handled. But could be adapted to be called by a global handler. And each custom exception having a message and translation key.

Sheikah45:
One question I have is should errors that come from poorly formatted data such as bad urls or bad paths that we get from the server for loading images be displayed to the user?

Or should those just be logged and then the asset not displayed or some stand-in displayed like we currently do for unknown maps.

micheljung:
> One question I have is should errors that come from poorly formatted data such as bad urls or bad paths that we get from the server for loading images be displayed to the user?
> 
> Or should those just be logged and then the asset not displayed or some stand-in displayed like we currently do for unknown maps.

If getting bad paths from the server is something that can be expected, then yes, we can silently log it (fail-safe). If it's not to be expected, then it should be displayed and fail loudly (fail-fast). See my initial link.

Sheikah45:
My main concern would be if there was any way for another player to insert a bad link somewhere then that would give people the ability to effectively spam clients

micheljung:
And if someone hacked the server or cosmic rays would flip a bit, the same could happen. But that's not the right thing to worry about

Sheikah45:
We already have had instances of malicious users exploiting game lobbies to cause crashes and delays so I am not keen to introduce a way for them to spam clients as well. So will probably fail safe on anything that comes from users

micheljung:
> If getting bad paths from the server is something that can be expected, then yes, we can silently log it (fail-safe). If it's not to be expected, then it should be displayed and fail loudly (fail-fast). See my initial link.

Like I said, if it's something that can be expected (bad user input) then yes, otherwise no.
Surely whatever receives the bad user input should validate it, however.

Sheikah45:
Also something else to consider. Currently some checked exceptions are handled locally. 

For our use case when we wrap them into our own unchecked exceptions would it be best to let these all be handled by the global uncaught exception handler or are there cases where it would still be good to handle them locally.

With regards to the fail fast or fail safe discussion above I was mostly thinking in terms of are the exceptions logged only or is a warning/error also shown to the user.

micheljung:
> Currently some checked exceptions are handled locally.

Do you have a concrete example to discuss?

> would it be best to let these all be handled by the global uncaught exception handler or are there cases where it would still be good to handle them locally.

That really depends on the case

> With regards to the fail fast or fail safe discussion above I was mostly thinking in terms of are the exceptions logged only or is a warning/error also shown to the user.

I see. We should never "fail safe", always "fail fast". Whether the exception is then displayed to the user is another question.

Sheikah45:
Yes I agree on the fail fast. That is why I clarified that most of previous comments have been on do we then show the error to the user.

With regards to where we handle something locally an example is a map load exception.

It is currently handled locally in mapService line 297 in the addInstalledMap function. This could instead have the try removed and be handled by the uncaught exception handler.


https://github.com/FAForever/downlords-faf-client/blob/develop/src/main/java/com/faforever/client/map/MapService.java#L297

micheljung:
https://github.com/FAForever/downlords-faf-client/blob/1c798f0eb1a2cf3b54f6977d40798c205b6fb6f8/src/main/java/com/faforever/client/map/MapService.java#L296-L298

This was done deliberately. It is to be expected that some of them are corrupt and can't be loaded. Therefore, one corrupt map should not cause map loading to abort. We could move corrupt maps to another directory [like we do with replays](https://github.com/FAForever/downlords-faf-client/blob/f682b932a7727374d9861b8680f5d67d2329fb3c/src/main/java/com/faforever/client/replay/ReplayService.java#L226) and notify the user about it. But IMO this is a valid case of fail-safe because it is to be expected and the behavior doesn't cause subsequent errors, leave the program in an invalid state or pose any risk.

Sheikah45:
Yes that makes sense although wouldn't it be best to catch that exception in the methods like load all maps and the directory watcher thread. Rather than the single add method.

micheljung:
Yes and no: I agree that `addInstalledMap` itself should not handle the exception. But since it's called from 4 different locations and the error handling should stay the same, I'd either `4 places -> tryAddInstalledMap() -> addInstalledMap()` or rename the method: `4 places -> tryAddInstalledMap()`

Sheikah45:
Alright I added a global exception handler and only catch the expected exceptions which mostly revolve around potential bad or unknown paths from the server

PR #2376, opened 2021-09-28
Add names for 4v4 queues

BlackYps:
This is the final piece for enabling 4v4 matchmaker queues.
Should we also add a name key for a possible event queue?


Askaholic:
Should we rename 1v1 and 2v2 to use the 'Competitive' keyword as well?

src/main/resources/i18n/messages.properties
```diff
@@ -997,6 +997,8 @@ teammatchmaking.queueTitle.inParty = The party leader will join the queues
 teammatchmaking.queueTitle.inQueue = Your party is now in queue
 teammatchmaking.queue.ladder1v1 = 1v1 Queue
 teammatchmaking.queue.tmm2v2 = 2v2 Queue
+teammatchmaking.queue.tmm4v4ShareUntilDeath = 4v4 Casual
+teammatchmaking.queue.tmm4v4FullShare = 4v4 Competitive
```
Askaholic: Is that really what we're going to make the technical names of the queues? Wouldn't it make sense to have the technical names also just be `tmm4v4_casual` and `tmm4v4_comp`?

Sheikah45: api values should be camelCase 😝

BlackYps: Now that I think about it, I guess you are right. I wanted to include what the actual difference is, but in the db where the nameKey is defined the gameoptions are right next to it, so that is probably not needed.

BlackYps:
We decided to use "Queue" because we didn't want to give the impression that this is the tryhard area, but it should be the default way to play FAF. That's why even "Ranked" was rejected.

Askaholic:
Well it is a tryhard area tho, especially ladder 1v1 which has never been anything but a tryhard queue. And now that we're adding a casual queue namely casual 4v4 I think it's even more OK to call the other queues Competitive. I just don't think it makes sense to have them be inconsistent like that just because of historical reasons. We should take this opportunity to adjust the names for all the queues.

src/main/resources/i18n/messages.properties
Askaholic: Well I was just referencing the technical names as they'd appear in the db and in the server protocol (they should be snake case). I assume the API will transform those but idk.

BlackYps:
But you don't *need* to tryhard on ladder. You can just chill and be 100 rating lower than you could be. I should probably try to find where we have previously discussed this.

Askaholic:
Well by that logic you don't *need* to not tryhard in casual either. You don't *need* to do anything.

BlackYps:
That is entirely correct, but not really an argument to name them all Competitive :P

Askaholic:
We're not naming them all 'Competitive' though, because now we have a queue that is dedicated to being casual (which we didn't before). There would be 3 called 'Competitive' and one called 'Casual'. The only thing I don't like, now that I've been looking at those two words side by side for a while is that they look visually pretty similar since they start with the same letter.

BlackYps:
-> zulip

BlackYps:
I believe this is ready to be merged now that everyone is happy with the queue names.

src/main/resources/i18n/messages.properties
```diff
@@ -995,8 +995,10 @@ teammatchmaking.queueTitle.inGame = Can't join queues while you are in a game
 teammatchmaking.queueTitle.memberInGame = Can't join queues while someone is in a game
 teammatchmaking.queueTitle.inParty = The party leader will join the queues
 teammatchmaking.queueTitle.inQueue = Your party is now in queue
-teammatchmaking.queue.ladder1v1 = 1v1 Queue
-teammatchmaking.queue.tmm2v2 = 2v2 Queue
+teammatchmaking.queue.ladder1v1 = 1v1
+teammatchmaking.queue.tmm2v2 = 2v2
```
Sheikah45, line 999: Did you just remove the word queue for consistency?

BlackYps: Yes, seemed a bit weird to me to append "Queue" to some but not to all. Also with the new queues the queue buttons won't be as big anymore, so we don't need a filler word as much.

PR #2377, opened 2021-09-29
Change dns resolver of webclient

### LINKED ISSUE #2374: Some users cannot resolve api.faforever.com
**Describe the bug**
From the forum:
updated to 1.6 today, after having not played for a good while, however when i try to log in i get the following message

java.util.concurrent.CompletionException: org.springframework.web.reactive.function.client.WebClientRequestException: failed to resolve ' api.faforever.com ' after 2 queries ; nested exception is java.net.UnknownHostException: failed to resolve ' api.faforever.com ' after 2 queries 

**To Reproduce**
Some minority of users cannot resolve api.faforever.com using the webclient from spring

**Expected behavior**
Users should be able to easily connect to the api like they have with client 1.5.0

Visit Issue hunt...

Sheikah45:
Closes #2374

### DIFF
--- a/src/main/java/com/faforever/client/config/WebClientConfigurer.java
+++ b/src/main/java/com/faforever/client/config/WebClientConfigurer.java
@@ -1,0 +1,31 @@
+package com.faforever.client.config;
+
+import com.faforever.client.api.JsonApiReader;
+import com.faforever.client.api.JsonApiWriter;
+import io.netty.resolver.DefaultAddressResolverGroup;
+import lombok.RequiredArgsConstructor;
+import org.springframework.boot.web.reactive.function.client.WebClientCustomizer;
+import org.springframework.http.client.reactive.ClientHttpConnector;
+import org.springframework.http.client.reactive.ReactorClientHttpConnector;
+import org.springframework.stereotype.Component;
+import org.springframework.web.reactive.function.client.WebClient;
+import reactor.netty.http.client.HttpClient;
+
+@Component
+@RequiredArgsConstructor
+public class WebClientConfigurer implements WebClientCustomizer {
+  private final JsonApiWriter jsonApiWriter;
+  private final JsonApiReader jsonApiReader;
+
+  @Override
+  public void customize(WebClient.Builder webClientBuilder) {
+    HttpClient httpClient = HttpClient.create().resolver(DefaultAddressResolverGroup.INSTANCE);
+    ClientHttpConnector clientHttpConnector = new ReactorClientHttpConnector(httpClient);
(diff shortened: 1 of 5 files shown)


micheljung:
Your commit message should reference the issue like:

```
Increase timeout and dns queries per resolve

Fixes #2374
```

Assuming this is what this is about.

Sheikah45:
This only seems to fix the issue for a small amount of users.

I will probably try using Apache's http client instead

Sheikah45:
Trying to switch the default resolver to see if this resolves the issue as seen here https://github.com/reactor/reactor-netty/issues/1431

Sheikah45:
It seems to be a known bug of the netty dns resolver and spring 2.4.0

micheljung:
> It seems to be a known bug of the netty dns resolver and spring 2.4.0

Can you link this please

Sheikah45:
I linked it above and here https://github.com/reactor/reactor-netty/issues/1431
I haven't been able to find where it is tracked on netty itself though. But lots of stack overflow saying this was a resolution as well https://stackoverflow.com/questions/67075057/unknowhhostexception-using-spring-cloud-starter-gateway-with-spring-boot-2-4-0-a

Sheikah45:
Received confirmation from some users that this fixes the problem

src/main/java/com/faforever/client/config/WebClientConfigurer.java
```diff
@@ -0,0 +1,31 @@
...
+@RequiredArgsConstructor
+public class WebClientConfigurer implements WebClientCustomizer {
+  private final JsonApiWriter jsonApiWriter;
+  private final JsonApiReader jsonApiReader;
+
+  @Override
+  public void customize(WebClient.Builder webClientBuilder) {
+    HttpClient httpClient = HttpClient.create().resolver(DefaultAddressResolverGroup.INSTANCE);
+    ClientHttpConnector clientHttpConnector = new ReactorClientHttpConnector(httpClient);
+    webClientBuilder.defaultHeader("User-Agent", "downlords-faf-client")
```
micheljung, line 24: We're debranding the client and here we hardcode the user agent? :) nonono

IIRC there's already a property to be read, and I'd include the client version as well.

Sheikah45: I change all the user agents in the debrand pr. Also I did not see a property but can double check and change in the debrand pr.

Sheikah45: Also apparently the user agent for the webviews can't change unless the website is also changed because that is used to stop the header and footer from appearing

Issue #1753, opened 2020-06-05
Use update4j for auto-updates


micheljung:
It has always bothered me that users need to:
- Download the newest version
- Close the client
- Execute the installer

Even if the version just contains a small bugfix.

2y ago, [update4j](https://github.com/update4j/update4j) came into existence and today, it seems quite mature. I'd like to try implementing it into the client.

Best case scenario, the client will auto-update itself in the background (after asking for permission to do so) and the updated version will be loaded upon the next start.

1-alex98:
Like u personal 😱 ? Nice

mordechaim:
Hi, update4j contributed here. 
What's your experience so far? Feel free to ask, I'll try my best to assist.

micheljung:
@Sheikah45 just wanted to ask if you didn't want to continue on this, now I see you committed something :) Nice, I would love to see this implemented. IIRC the remaining issue were:

1. The user installs e.g. 1.0.0 using the wizard. Then, despite auto-updating, Windows will still see 1.0.0 installed. Maybe we can use the Install4J API to update the installation.
2. The generated JRE module file somehow changes every time (adding ~30mb) even though nothing changed. Still better than downloading the whole client, though.
3. Right now, files are uploaded to a private server of mine. We'd probably want a new container on the server that allows uploading files via SCP. Alternatively, we could upload it via API but I think this requires unnecessary work and ends up being less flexible.

Let me know if you have any questions or want to work on it together.

Sheikah45:
Yeah I was working on rebasing this and was going to start again on it once I finished the gson to Jackson and oauth PRS.
I will definitely reach out to you probably for assistance.

With regards to the update files and the server that is definitely something we will need to work with Brutus on to see if he thinks that is feasible.

Although why do you think they could not just be on the git release?

micheljung:
The idea of the auto updater (and the way update4j works) is to download only files that changed. Therefore, all client files need to be hosted relative to some URL. With GitHub releases, this isn't viable nor possible.

Keeping such releases around on our servers is quite space consuming. To reduce this, the deployment script makes use of hardlinks; if a file didn't change it's only stored once. This way, new releases usually only take up a few KiB of additional disk space.

Once this is in place, the question becomes whether we still want to publish offline installers or if we'd want to provide a web-installer that's small and will download the files as needed. Such an installer could even allow selecting the version to install, making it easy to install an older version, too.

PR #2379, opened 2021-10-01
Debrand the client

Sheikah45:


src/icomoon/Downlord's FAF Client.json
```diff
@@ -1,6 +1,6 @@
 {
   "metadata": {
-    "name": "Downlord's FAF Client",
+    "name": "FAF Client",
```
micheljung, line 3: File can be renamed as well

micheljung:
There are additional places like installation path etc. which I also changed in the update branch (see #1753)

Sheikah45:
@micheljung how did you create the splash screen initially that needs to be changed as well. I see the png and psd files

Sheikah45:
Alright all the places that can be changed should be changed now

Sheikah45:
Note the user agent cannot be changed as the website will then include the header and footer

micheljung:
> @micheljung how did you create the splash screen initially that needs to be changed as well. I see the png and psd files

Open the PSD in Photoshop, edit and save the PNG :) let me know if you don't have it

Sheikah45:
Yep already done. Just had to look around for a font match

PR #2381, opened 2021-10-03
Utilize inheritance for context menus

BlackYps:
This is preparation so we can use a context menu for the leaderboard without introducing immense code duplication. I extracted this from the league PR to make review easier and it is not really tied to the new league leaderboard anyway.


src/test/java/com/faforever/client/fx/PlayerContextMenuControllerTest.java
```diff
@@ -0,0 +1,132 @@
...
+  private PlatformService platformService;
+
+  private PlayerContextMenuController instance;
+  private PlayerBean player;
+
+  @BeforeEach
+  public void setUp() throws Exception {
+    instance = new PlayerContextMenuController(clientProperties, playerService,
+        eventBus, uiService, platformService);
+    loadFxml("theme/player_context_menu.fxml", clazz -> instance);
```
BlackYps: Somehow the initialize() method doesn't get called, so I get NullPointerExceptions when calling setPlayer(). I don't understand why it doesn't load the fxml properly though.

BlackYps:
Fixes #2382

BlackYps:
I think I fixed all the problems now. The new structure should make it quite easy to add context menus to the players in the replay and custom games tabs.

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
```diff
@@ -0,0 +1,249 @@
...
+    socialSeparator.visibleProperty().bind(addFriendItem.visibleProperty().or(
+        removeFriendItem.visibleProperty().or(
+            addFoeItem.visibleProperty().or(
+                removeFoeItem.visibleProperty()))));
+    gameSeparator.visibleProperty().bind(joinGameItem.visibleProperty().or(
+        watchGameItem.visibleProperty().or(
+            viewReplaysItem.visibleProperty())));
+    moderatorActionSeparator.visibleProperty().bind(kickGameItem.visibleProperty().or(
+        kickLobbyItem.visibleProperty().or(
+            broadcastMessage.visibleProperty())));
```
Sheikah45: Would be good to have the closing parentheses match the level of the open parentheses. Would make it easier to see

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
```diff
@@ -0,0 +1,249 @@
...
+    PlayerBean player = getPlayer();
+    joinGameHelper.join(player.getGame());
+  }
+
+  public void onWatchGameSelected() {
+    PlayerBean player = getPlayer();
+    try {
+      replayService.runLiveReplay(player.getGame().getId());
+    } catch (Exception e) {
+      log.error("Cannot display live replay", e.getCause());
```
Sheikah45: Are you sure all exceptions thrown here will have a cause

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
```diff
@@ -0,0 +1,249 @@
...
+    moderatorService.closePlayersLobby(getPlayer());
+  }
+
+  protected void setModeratorOptions(Set<String> permissions, PlayerBean player) {
+    boolean notSelf = !player.getSocialStatus().equals(SELF);
+
+    JavaFxUtil.runLater(() -> {
+      kickGameItem.setVisible(notSelf & permissions.contains(GroupPermission.ADMIN_KICK_SERVER));
+      kickLobbyItem.setVisible(notSelf & permissions.contains(GroupPermission.ADMIN_KICK_SERVER));
+      broadcastMessage.setVisible(notSelf & permissions.contains(GroupPermission.ROLE_WRITE_MESSAGE));
```
Sheikah45: Why does broadcast message need to be notSelf?

Sheikah45 (changes requested):

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
BlackYps: Apparently it does not. I'll also transfer it to only the chat context menu, because I think it makes the most sense there.

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
BlackYps: I'm not. This was just copied over.

BlackYps:
Apparently codacy changes its mind every commit about long parameter lists...

Sheikah45:
Yes it changes its mind about what are new issues

src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
```diff
@@ -64,56 +53,67 @@
...
-  private void initializeListeners() {
+  @NotNull
+  private StringListCell<AvatarBean> avatarCell() {
+    return new StringListCell<>(
+        AvatarBean::getDescription,
+        avatarBean -> {
+          URL url = avatarBean.getUrl();
+          if (url == null) {
+            return null;
+          }
```
Sheikah45: The loadAvatar method and subsequent loadAndCacheImage already have null checks so this is not needed here

src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
```diff
@@ -127,12 +127,11 @@ private void initializeListeners() {
...
         sendPrivateMessageItem.setVisible(socialStatus != SELF);
-        addFriendItem.setVisible(socialStatus != FRIEND && socialStatus != SELF);
+        addFriendItem.setVisible(optionalPlayer.isPresent() && socialStatus != FRIEND && socialStatus != SELF);
         removeFriendItem.setVisible(socialStatus == FRIEND);
-        addFoeItem.setVisible(socialStatus != FOE && socialStatus != SELF);
+        addFoeItem.setVisible(optionalPlayer.isPresent() && socialStatus != FOE && socialStatus != SELF);
         removeFoeItem.setVisible(socialStatus == FOE);
-        reportItem.setVisible(socialStatus != SELF);
-
+        reportItem.setVisible(optionalPlayer.isPresent() && socialStatus != SELF);
```
Sheikah45: Why were these optionalPlayer.isPresent checks added? If player isn't present then socialStatus should be null so the only dependency is socialStatus

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
```diff
@@ -0,0 +1,247 @@
...
+
+    socialSeparator.visibleProperty().bind(addFriendItem.visibleProperty().or(
+        removeFriendItem.visibleProperty().or(addFoeItem.visibleProperty().or(removeFoeItem.visibleProperty()))
+    ));
+    gameSeparator.visibleProperty().bind(joinGameItem.visibleProperty().or(
+        watchGameItem.visibleProperty().or(viewReplaysItem.visibleProperty())
+    ));
+    moderatorActionSeparator.visibleProperty().bind(kickGameItem.visibleProperty().or(
+        kickLobbyItem.visibleProperty().or(broadcastMessage.visibleProperty())
+    ));
```
Sheikah45: Probably good to break each .or onto its own line

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
```diff
@@ -0,0 +1,247 @@
...
+    JavaFxUtil.runLater(() -> {
+      kickGameItem.setVisible(notSelf & permissions.contains(GroupPermission.ADMIN_KICK_SERVER));
+      kickLobbyItem.setVisible(notSelf & permissions.contains(GroupPermission.ADMIN_KICK_SERVER));
+    });
+  }
+
+  @NotNull
+  protected PlayerBean getPlayer() {
+    return player;
+  }
```
Sheikah45: both the getter and the player attribute are protected which makes the getter obsolete. You can just use the player property directly.

Or at least make the player property private but since there is no additional logic to the getter it doesn't make much sense

src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
```diff
@@ -64,56 +53,67 @@
...
+          if (url == null) {
+            return null;
+          }
+          return new ImageView(avatarService.loadAvatar(avatarBean));
+        });
+  }
+
+  @Override
+  protected void initializeListeners() {
     chatUserPropertyInvalidationListener = observable -> {
```
Sheikah45: Rather than redefining the listener over again and completely overriding the super classes listener which largely sets the same values. It may be better to just have a separate listener that listens on the player object and subsequently sets the player of the context menu. This should reduce duplication and may allow us to simplify the chatUser object

src/main/java/com/faforever/client/teammatchmaking/PartyMemberContextMenuController.java
```diff
@@ -1,150 +1,52 @@
...
+                                          UiService uiService) {
+    super(eventBus, joinGameHelper, moderatorService, notificationService, playerService, replayService, uiService);
   }
 
-  public ContextMenu getContextMenu() {
-    return partyMemberContextMenuRoot;
+  @Override
+  public void initialize() {
+    super.initialize();
+    sendPrivateMessageItem.setOnAction(event -> onSendPrivateMessageSelected());
```
Sheikah45: Part of me feels like the sendPrivateMessageItem is really needed at all bot eh

src/main/resources/theme/player_context_menu.fxml
```diff
@@ -9,61 +9,60 @@
...
-          <ComboBox fx:id="avatarComboBox" maxWidth="1.7976931348623157E308"/>
-      </content>
-    </CustomMenuItem>
+      <CustomMenuItem fx:id="avatarPickerMenuItem" hideOnClick="false" mnemonicParsing="false" visible="false">
+          <content>
+          </content>
+          <content>
+              <ComboBox fx:id="avatarComboBox" maxWidth="1.7976931348623157E308"/>
+          </content>
+      </CustomMenuItem>
```
Sheikah45: Would be good to have consistency with the initial visibility and just set them all to false and let the initialization turn on the appropriate ones

Sheikah45 (changes requested):

src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
BlackYps: Without these, the items still show up for IRC users. ~~It seems like getting an objectProperty never returns a null value.~~
Edit: It's just that we with the != checks it also passes if the value is null

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
BlackYps: I need the getter to be able to override it in the chatUserContextMenu. I also need access to the player attribute when overriding setPlayer for the leaderboardContextMenu (which is not part of this PR)
```  
public void setPlayer(PlayerBean player) {
    super.player = player;
}
```
although there might be a better way to do this so the attribute can stay private.

src/main/java/com/faforever/client/teammatchmaking/PartyMemberContextMenuController.java
BlackYps: Don't know how many people know that you can also doubleclick a user. (I wasn't.)

src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
Sheikah45: ah yes gotcha forgot about chat only

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
Sheikah45: I would probably go the route of still running the super.setPlayer() and then doing whatever extra you need. Also I am not sure what you would need to override a getter for that is protected if the attribute is protected as well.

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
```diff
@@ -0,0 +1,261 @@
...
+    kickGameItem.setOnAction(event -> onKickGame());
+    kickLobbyItem.setOnAction(event -> onKickLobby());
+
+    initializeListener();
+
+    copyUsernameItem.setVisible(true);
+    socialSeparator.visibleProperty().bind(addFriendItem.visibleProperty()
+        .or(removeFriendItem.visibleProperty()
+            .or(addFoeItem.visibleProperty()
+                .or(removeFoeItem.visibleProperty()))
```
Sheikah45: These ors are nested right now but don't need to be. You can just close the or on each line

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
```diff
@@ -0,0 +1,261 @@
...
+    ));
+    gameSeparator.visibleProperty().bind(joinGameItem.visibleProperty()
+        .or(watchGameItem.visibleProperty()
+            .or(viewReplaysItem.visibleProperty())
```
Sheikah45: nested ors

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
```diff
@@ -0,0 +1,261 @@
...
+    ));
+    moderatorActionSeparator.visibleProperty().bind(kickGameItem.visibleProperty()
+        .or(kickLobbyItem.visibleProperty()
+            .or(broadcastMessage.visibleProperty())
```
Sheikah45: nested ors

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
Sheikah45: Also in the chatUserContextMenu wouldn't it be better to have a listener on the playerproperty that would set the player and also check for a player when the chat user is set. Then you would not even need to override the getter. I think the Illegal State exception that is thrown currently can be removed as it is a perfectly acceptable state due to chat only users.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
BlackYps: Sounds like a good plan

BlackYps:
I don't know how I managed to break the two failing tests. When I try it out in the client everything gets displayed as it should.

Sheikah45:
Probably a race condition between the application thread and the test thread. Are you waiting for FX events?
Or could be that something isn't mocked properly

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
```diff
@@ -0,0 +1,294 @@
...
+  public MenuItem watchGameItem;
+  public MenuItem viewReplaysItem;
+  public SeparatorMenuItem moderatorActionSeparator;
+  public MenuItem kickGameItem;
+  public MenuItem kickLobbyItem;
+  public MenuItem broadcastMessage;
+  public CustomMenuItem avatarPickerMenuItem;
+  public ComboBox<AvatarBean> avatarComboBox;
+
+  protected PlayerBean player;
```
Sheikah45: I would still make this private and create a protected getter. That way child classes can't set the player without attaching the required listeners

src/main/java/com/faforever/client/chat/ChatUserContextMenuController.java
```diff
@@ -2,168 +2,106 @@
...
   }
 
-  private void initializeListeners() {
+  @Override
+  protected void initializeListener() {
+    super.initializeListener();
     chatUserPropertyInvalidationListener = observable -> {
-      Optional<PlayerBean> optionalPlayer = chatUser.getPlayer();
-      optionalPlayer.ifPresent(player -> setModeratorOptions(moderatorService.getPermissions(), player));
+      // TODO: We should only set the player once and not every time the player object changes
```
Sheikah45: Probably good to have a separate playerchangelistener for this then that is only on the chatUser.playerProperty

src/main/java/com/faforever/client/fx/PlayerContextMenuController.java
BlackYps: But I actually don't want that. Have a look at the LeaderboardContextMenuController in the league PR. There I use it for offline players and need to get rid of all the options that only make sense for online players.

Sheikah45: Then it would probably be better to just override setPlayers for that subclass.

Is there something that is preventing that?

BlackYps: I can't do this in the subclass when player is private:
```
  public void setPlayer(PlayerBean player) {
    super.player = player;
  }
```

BlackYps: And I don't really see the issue with leaving the player attribute protected. The setter is public, so the attribute is still lower visibility than the setter.

Sheikah45: Yes if it was just a pure setter I might agree but since this setter has additional logic to it. I think it is probably better to force all subclasses to go through that additional logic and then change what they need for consistency sake

src/main/java/com/faforever/client/fx/OnlinePlayerContextMenuController.java
```diff
@@ -0,0 +1,238 @@
...
+    joinGameItem.setOnAction(event -> onJoinGameSelected());
+    watchGameItem.setOnAction(event -> onWatchGameSelected());
+    kickGameItem.setOnAction(event -> onKickGame());
+    kickLobbyItem.setOnAction(event -> onKickLobby());
+
+    avatarComboBox.setCellFactory(param -> avatarCell());
+    avatarComboBox.setButtonCell(avatarCell());
+    avatarPickerMenuItem.visibleProperty().bind(Bindings.createBooleanBinding(() ->
+            !avatarComboBox.getItems().isEmpty() && getPlayer().getSocialStatus() == SELF,
+        avatarComboBox.getItems()));
```
Sheikah45: I try to avoid binding to objects that are updated by api or server calls as that extends the reach of things running on the application thread into the program.

Sheikah45 (commented):
Looking pretty good just a minor comment

src/main/java/com/faforever/client/fx/OnlinePlayerContextMenuController.java
BlackYps: I don't really understand. Isn't the logic that changes objects always running in the non-application thread anyway? How does the existence of an api call change the situation?
Apart from that general question, do you have a suggestion how we could change the code in this instance?

Sheikah45: When you bind to a property then whenever that observable is changed the binding will executed in the same thread as the change occurred in. In the past when the binding was to an object updated by a server message or something that was directly updated in the server then it resulted in the update failing due to not being executed in the application thread. 

Because of this I have been taking the approach to rather than binding to explicitly non pure ui based values to instead use weak listeners and then inside of the listener ensue the JavaFx.runLater is wrapping the relevant updates.

BlackYps: I had to change the tests a little, because now it won't hide the avatar picker when your social status changes from self to other. In reality this will never happen, because it would require a different player to login in the client without destroying the context menu, so I didn't bother to implement that edge case.

Sheikah45 (dismissed):

src/main/java/com/faforever/client/fx/OnlinePlayerContextMenuController.java
Sheikah45: And yeah I do consider if after this pr we should just make it so that you can only set the player once

Sheikah45 (approved):

PR #2389, opened 2021-10-10
Add ability to close matchmaker games when match is cancelled

### LINKED ISSUE #2188: Close FA on `match_cancelled` message
# Motivation
Sometimes a matchmaker game can fail to start because one of the players involved does not connect fast enough. When this happens the game will not be recorded in any way. No achievements will be processed, no rating change will be calculated, and no entry will show up in the replay vault. However, sometimes players continue to wait for a while and the game does end up launching several minutes later. The players then end up playing a non-existent game and may be dissappointed when it doesn't show up in their history.

# How to fix it
When a matchmaker game times out during setup, the server sends a message that looks like this:
```json
{"command": "match_cancelled"}
```
Upon receiving this message, the client should react by either 
1. Terminating the running FA process if it has been launched already, or
2. Ignoring the "game_launch" command that might follow in the next few seconds.

NOTE: Number 2 is only necessary due to legacy client support. In version 2 of the protocol, the server will not send a `game_launch` command if the match has been cancelled.

It may also be a good idea to indicate that the match was cancelled in the UI somewhere so the user knows what's going on.

Related: #1750

Visit Issue hunt...

Sheikah45:
Fixes #2188


Sheikah45:
Right now there is a small possibility that if a player exits the queue after a match and then request and gets a new game that the new game is the one that was closed.

@Askaholic is this even a situation that can happen? 

Getting a new matchmaker game before the previous one was cancelled?

Askaholic:
I don’t quite understand the situation you’re describing. Like if the player tries to join a custom game before the match is cancelled? The server will prevent that because it won’t let you join customs while you’re in a non-idle state such as when you’re waiting for a ladder game to start.

Sheikah45:
I am saying that some one gets a match and the game launches. Then for whatever reason they close the window. Are they then able to requeue server side before they get the match cancelled message?

Askaholic:
Yea I think that can happen actually. As soon as they close the game, the server will see that and reset the state tracking.

Sheikah45:
Hmm okay. Well as of right now there would be a chance that of they started another matchmaker game before they got the cancelled message the new game would be closed. I wonder how often that would happen

Askaholic:
Yea that would probably be a good thing for me to write an integration test for on the server. I think in practice it would be pretty hard to do because of the queue pop interval of 3 minutes and the game timeout of about 2 minutes. But with bigger games and more players in the queue the interval could go down and the timeout could go up so it might be possible that they overlap. 

I wonder if it would help if we added the game id to the match_cancelled message. It would have to be optional because it’s possible for the match to fail before it creates the game id, but that should only be in rare circumstances where there is a server bug or an improperly setup map pool. If you had the game id then maybe you could associate the process handle with a particular id and only terminate the process if the ids match.

Sheikah45:
Yeah I can currently associate the process with a game id I just wanted to know if I actually needed to add another state variable like that or I could get away with just tracking whether it is a matchmaker game or not.

Sheikah45:
Although actually thinking about it more it actually might be necessary to have that game id in the match cancelled since there is no way for me to associate a match cancelled with the match launch outside of assuming they are serially ordered.

Of course I easy solution may be to just not allow people to get another game or join the queue until the game has officially been cancelled. Would also encourage people not to abort the launch process.

Askaholic:
I think both would be a good idea.

PR #2391, opened 2021-10-11
Remove unused partiesInQueue attribute

BlackYps:
This is never used and I'd like to remove the boundary80 and boundary75 infos from the server anyway because they are pretty useless.

### DIFF
--- a/src/main/java/com/faforever/client/domain/MatchmakerQueueBean.java
+++ b/src/main/java/com/faforever/client/domain/MatchmakerQueueBean.java
@@ -24,7 +24,6 @@ public class MatchmakerQueueBean extends AbstractEntityBean<MatchmakerQueueBean>
   StringProperty technicalName = new SimpleStringProperty();
   ObjectProperty<OffsetDateTime> queuePopTime= new SimpleObjectProperty<OffsetDateTime>();
   IntegerProperty teamSize = new SimpleIntegerProperty(0);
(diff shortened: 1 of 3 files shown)


Sheikah45:
Might not be used now but why remove it?
Also curious what the boundaries are supposed to represent

BlackYps:
The boundaries are supposed to give an estimate to the rating range for other players that would achieve 75 or 80% trueskill game quality with that player. I guess they were used by the python client for the "A person in your rating range is searching for a game. Click here to join them." message. However, the ranges are unreliable, because it just gives a hardcoded range and the trueskill quality also depends on the deviation. Also we alter the acceptable quality for every unsucessful queue pop.
For team queues we use a different system entirely so I would just remove this.

Sheikah45:
Makes sense. Although also why remove the parties in queue instead of just the boundaries? I don't really see a problem with the partiesInQueue being there

BlackYps:
Being able to tell the number of parties in queue is basically a side effect of how the info for the boundaries is structured. To make it still available I would have to create a new field in the server message. That is not a problem, but I didn't really find a good use case for that info, so I went with removing it in the client entirely.
We don't have to do that, if you have an idea what it would be useful for, I'd also be happy to keep it.

Sheikah45:
Ah I didn't realize it was a derived property

Sheikah45 (approved):

Issue #1946, opened 2020-09-24
Custom filter for displaying lobbies in play tab


GrunttiNoob:
a tickbox for applying and saving custom lobby filters 
![kuva](https://user-images.githubusercontent.com/71841656/94181917-d9c8d580-fea8-11ea-958c-0c901bd81d8a.png)

Examples for filters many players would appreciate:

MapNameIncludes: "gap" + "astro"
MinRating: "2000"

ryanwalder:
I would also like to see a custom filter. To expand slightly:

* Filter by Map
* Filter by Sim Mod (eg: Phantom)
* Filter by Mod
* Filter by map player count (min/max players)
* Filter by Title
* Filter by host

Ideally this would all be accessible via a Query like "MAP !=foo,bar,baz AND MODS !=2x, blackops AND SIMMOD = phantom" or similar.

Also slightly unrelated but being able to have a notification based on a query would be awesome if you're looking out for a game on a specific map or simmod etc...

PR #2407, opened 2021-10-24
Proposal: Notify users about interesting lobbies

SKeeneCode:
This is a draft proposal for a feature that is unfinished but has a functional prototype. I'm putting this out here for discussion because while I personally will use this feature it may not have sufficient utility to warrant developing further.

Users in FAF have the ability to be notified when friends join lobbies because those lobbies are obviously likely to be _interesting_ to the user. There are other potential use-cases where a user might like to be notified of a lobby without having to be on the play tab, for example:

- Players who wish to be notified when a lobby is almost full to avoid lobby sim.
- Players who want to be notified about lobbies of their favorite maps around their level. (For example I might only care about Setons lobbies with an average rating of 1500 or higher, or Gap/Astro with an average rating of under 1000).

This PR allows the user to receive these notifications using the following conditions:

image

Example notifications:

image
image

I would like to put this idea out for discussion to establish its utility and what changes might be made.

(Much of the logic to determine whether a lobby is interesting or not is found in [CustomGameController.java](https://github.com/FAForever/downlords-faf-client/pull/2407/commits/4ea76f80e9df130d6f37016f59a07dc9764108ba#diff-9a738b2fbce284d83c29c5878044140696b475eff46dd4695b7d7e68a4eb05de) but that only happened because I had to be in there to listen in on the [filtered list of games ](https://github.com/FAForever/downlords-faf-client/pull/2407/commits/4ea76f80e9df130d6f37016f59a07dc9764108ba#diff-9a738b2fbce284d83c29c5878044140696b475eff46dd4695b7d7e68a4eb05deR134-R145). If needed in the future I would refactor this out to its own class.


Sheikah45:
This in my mind is very similar to #1946 just with notifications rather than a filter.

I think the filter would be better. With regards to the notifications I know plenty of people hate the toast notifications so I don't know if it would be good to introduce more of those.

I would also not put the options into the settings window but would create a separate window that can be opened from the custom games tab

SKeeneCode:
> I think the filter would be better. 

This is fair - do you think if a proper way to filter lobbies was developed it would make sense to have the notifications as a part of that?

>With regards to the notifications I know plenty of people hate the toast notifications so I don't know if it would be good to introduce more of those.

I don't understand this concern because notifications are completely optional? And the implementation is that they are off by default so only players who specifically like and want the notifications can have them.

> I would also not put the options into the settings window but would create a separate window that can be opened from the custom games tab

I placed it in the settings notifications tab because it seemed the most obvious place - it is the location where users decide on what notifications they want to see after all. I also wanted to avoid any more clutter in the play tab, but if a way to filter lobbies was established it may make sense to put it there.

Sheikah45:
Potentially notifications could be a part of it and the option to turn those notifications off and on could be in the settings but the options defining which notifications or games to filter should not be in the settings container. They just introduce too much clutter

SKeeneCode:
One potential issue in wanting to have a lobby filter and wanting to notify the user is the obvious use-case where the user has lobbies they are interested in and want to be notified about but don't want that filter applied to the list of lobbies. Perhaps the solution would be to set up the lobby filter and then have two checkboxes, one to apply the filter to the list of lobbies and one to be notified?

Sheikah45:
Yes that could work

Askaholic:
I think its a cool idea, and different enough from filters to be useful. Having notifications means you can just be hanging out in Aeolus and still know when a good lobby opens up without having to constantly switch back to the play tab. You could also just be doing other stuff on your computer and know when something is hosted without having to look at the client at all.

src/main/java/com/faforever/client/game/CustomGamesController.java
```diff
@@ -152,6 +180,85 @@ public TilesSortingOrder fromString(String string) {
...
+        notifyUserAboutInterestingLobby(game);
+      }
+    }
+  }
+
+  private void notifyUserAboutInterestingLobby(GameBean gameBean) {
+    String mapName = gameBean.getMapFolderName();
+    try {
+      CompletableFuture<Optional<MapVersionBean>> mapVersion = mapService.findByMapFolderName(gameBean.getMapFolderName());
+      mapName = mapVersion.get().get().getMap().getDisplayName();
```
Sheikah45, line 213: We cannot make calls to the api on every single gameInfo message this will easily over load the server

SKeeneCode: Can you clarify about this gameInfo message?

This mapService function is only called when the lobby is both interesting and a notification is needed. `findByFolderName` also checks to see if the user has the map locally before trying to call the API so it very well may not make any API calls. In the worst case scenario the user has no maps downloaded and is also interested in every single lobby and has the re-notify time set to minimum (1 minute) meaning every 60 seconds an API call is made for each lobby (plus 1 for each new lobby when they pop up).

This should easily be solved by caching the map display names? That way a single API call needs to be made for each separate map that is hosted and not local to the user which would be what max... 30 API calls the first hour and then less and less calls as time goes on as the cache grows? At most 10% or 20% of the userbase would use this feature. Surely the overhead would be minimal?

SKeeneCode:
> I think its a cool idea, and different enough from filters to be useful. Having notifications means you can just be hanging out in Aeolus and still know when a good lobby opens up without having to constantly switch back to the play tab. You could also just be doing other stuff on your computer and know when something is hosted without having to look at the client at all.

This is the kind of use-case which makes this useful to me personally. I don't hang out in Aeolus much and will spend hours doing something else and not look at FAF. I've probably missed many lobbies I would have loved to play.

While the opinion of the devs of this feature is invaluable I feel something like this would benefit from asking the opinion of the wider player base. What would be an appropriate way to approach this?

src/main/java/com/faforever/client/game/CustomGamesController.java
Sheikah45: it more about the fact that all many clients would be making the same call to the api at the same time. Even with caching there are many maps with many different names that cannot necessarily be cached

BlackYps:
> Players who wish to be notified when a lobby is almost full to avoid lobby sim.

To me this looks like a feature that players would think they want, but is actually an anti-feature. It will increase the problem of lobbies staying empty for a long time and when it eventueally reaches a certain number of players the rest joins quickly.
This feature basically tries to offload wait times to other players, but it can't reduce the wait time over all players. I would not recommend implementing that.

Sheikah45:
If there is no one working on this then i will close it soon TM

SKeeneCode:
Yes, sorry, I'll close it. I don't have the time (or the desire tbh) to implement lobby filtering so that these notifications can be integrated alongside it. Hopefully somebody in the future who really wants lobby filtering can include this feature in the future.

PR #2408, opened 2021-10-28
Make TokenService thread safe

Sheikah45:


src/main/java/com/faforever/client/api/SessionExpiredEvent.java
```diff
@@ -1,4 +1,7 @@
 package com.faforever.client.api;
 
-public class SessionExpiredEvent {
+public class SessionExpiredEvent extends RuntimeException{
```
Brutus5000: Uh please don't call Exceptions Events

src/main/java/com/faforever/client/api/TokenService.java
```diff
@@ -23,92 +27,106 @@
...
-            preferencesService.getPreferences().getLogin().setRefreshToken(null);
-            preferencesService.storeInBackground();
-            eventBus.post(new SessionExpiredEvent());
-          });
-    }
+  private void resetRetrievalMono() {
+    tokenRetrievalMono = Mono.defer(this::retrieveToken)
+        .cacheInvalidateIf(OAuth2AccessToken::isExpired)
+        .doOnNext(token -> log.debug("Token still valid for {} seconds", token.getExpiresIn()));
+  }
```
Brutus5000: The whole caching looks odd. Did you have a look at [CachedMono](https://projectreactor.io/docs/extra/snapshot/api/reactor/cache/CacheMono.html) ?

Basically I would put lambda doing the refresh operation inside and it should be executed everytime the mono is subscribed and either not initialized yet or expired.

src/main/java/com/faforever/client/api/SessionExpiredEvent.java
Sheikah45: Doh this is what happens when you code too late into the night.

End up thinking things like this are reasonable 🙄

src/main/java/com/faforever/client/api/TokenService.java
Sheikah45: Aha figured I was missing something like this. Thanks

Sheikah45: I wasn't really able to figure out a good way to invalidate the cache mono when the token was expired. So I switched instead to using cacheInvalidateWhen with a mono that is delayed by the expiration time and a mono that is triggered on logout events.

If you know of a good way to invalidate the cacheMono on token expiration and logout let me know

PR #2417, opened 2021-11-06
Add detection of non ascii title

### LINKED ISSUE #2414: No warning that the title of a game should contain only english text and (or) numbers
Trying to create a game with the title with non-english text (for example `+800 ты не пройдешь!`) and click "Create". 
Nothing will happen after that.

The following was printed in the log file:

> 2021-11-02 18:29:02.793  INFO 3940 --- [pool-3-thread-10] c.f.c.patch.GameBinariesUpdateTaskImpl   : Updating binaries to 3724
> 2021-11-02 18:29:02.870  INFO 3940 --- [reactor-tcp-nio-2] com.faforever.commons.lobby.FafLobbyApi  : Outbound channel cancelled
> 2021-11-02 18:29:02.871  INFO 3940 --- [reactor-tcp-nio-2] com.faforever.commons.lobby.FafLobbyApi  : Inbound channel cancelled
> 2021-11-02 18:29:02.872  INFO 3940 --- [reactor-tcp-nio-2] com.faforever.commons.lobby.FafLobbyApi  : Disconnected from server

Client Version:
2021.10.0

Sheikah45:
Closes #2414


src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -211,27 +211,29 @@ private void init() {
...
-      case CONNECTING -> i18n.get("game.create.connecting");
+    String createGameTextKey = switch (lobbyConnectionState) {
+      case DISCONNECTED -> "game.create.disconnected";
+      case CONNECTING -> "game.create.connecting";
       case CONNECTED -> {
-        if (Strings.isNullOrEmpty(titleTextField.getText())) {
-          yield i18n.get("game.create.titleMissing");
+        if (Strings.isNullOrEmpty(title)) {
+          yield "game.create.titleMissing";
+        } else if (!StandardCharsets.US_ASCII.newEncoder().canEncode(title)) {
```
Askaholic, line 223: Need to make a new encoder object everywhere, or can there just be a static final one? Or maybe not static, but just final and init in constructor.

src/main/resources/i18n/messages.properties
```diff
@@ -160,6 +160,7 @@ game.create.searchMap = Search…
 game.create.gameType = Game type
 game.create.mods = Mods
 game.create.titleMissing = Title is missing
+game.create.titleNotAscii = Title must contain only ascii characters
```
Askaholic: ASCII all caps?

src/main/java/com/faforever/client/game/CreateGameController.java
Sheikah45: Charset encoder is not thread safe, so would need to make sure this can only be called by one thread at a time but since it is inside a listener that is not a guarantee

Askaholic: Java never ceases to disappoint me :P

PR #2426, opened 2021-11-21
Check testOnReport method

Marc-Spector:


Sheikah45:
Fixed in https://github.com/FAForever/downlords-faf-client/pull/2427

PR #2428, opened 2021-11-23
#803 - Redesign "Coop" tab

### LINKED ISSUE #803: Play -> Co-Op - Make layout work on smaller window sizes (currently cut and overlapping elements)
Leaderboard table colums to narrow even with maximized window:
coop_leaderboard_4_players_games

Overlapping and cut elements:
coop_overlapping_and_cut_elements

Marc-Spector:
Fixes #803

New view 
image

### DIFF
--- a/src/main/java/com/faforever/client/coop/CoopController.java
+++ b/src/main/java/com/faforever/client/coop/CoopController.java
@@ -13,12 +13,9 @@ import com.faforever.client.game.GameService;
 import com.faforever.client.game.GamesTableController;
 import com.faforever.client.game.NewGameInfo;
 import com.faforever.client.i18n.I18n;
-import com.faforever.client.map.MapService;
-import com.faforever.client.map.MapService.PreviewSize;
 import com.faforever.client.mod.ModService;
 import com.faforever.client.notification.NotificationService;
 import com.faforever.client.replay.ReplayService;
-import com.faforever.client.reporting.ReportingService;
 import com.faforever.client.theme.UiService;
 import com.faforever.client.util.TimeService;
 import com.faforever.commons.lobby.GameStatus;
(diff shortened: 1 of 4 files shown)


Marc-Spector:
I removed the `map preview` view because all image URLs of missions are broken. For example https://content.faforever.com/maps/previews/small/x1ca_coop_001.v0026.png from `Black Day` mission

Marc-Spector:
If you remove the version number, the URL will work correctly. https://content.faforever.com/maps/previews/small/x1ca_coop_001.png

~~Random error? Do I need to manually remove the version number before calling the `mapService.loadPreview()` method? What do you think about this?~~

It seems that only the `Black Day` mission contains small and large map preview images.

Sheikah45:
If we removed the loading of map previews might as well also remove the map column from the table

src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -244,11 +238,7 @@ private CoopMissionBean getSelectedMission() {
   }
 
   private void setSelectedMission(CoopMissionBean mission) {
-    JavaFxUtil.runLater(() -> {
-      descriptionWebView.getEngine().loadContent(mission.getDescription());
-      mapImageView.setImage(mapService.loadPreview(mission.getMapFolderName(), PreviewSize.SMALL));
```
Sheikah45, line 249: Remove the map column as well

Sheikah45 (changes requested):

src/main/java/com/faforever/client/coop/CoopController.java
Marc-Spector: The column cannot be removed because it is being used elsewhere. I just hid it instead. Also hid rating range column. It is always empty and takes up extra space.

Marc-Spector: image

Sheikah45: Yeah hiding is good

Sheikah45 (approved):

PR #2430, opened 2021-11-29
Support emojis

### LINKED ISSUE #436: Support emojis
<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/41864552-support-emojis?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F14200913&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Marc-Spector:
Closes #436

Final view
image


Marc-Spector:
Please note that the PR is still not in its final stages.
I need some feedback. Depending on the feedback, I will continue to work with this or close it.

Sheikah45:
What are you looking for feedback on specifically?

Marc-Spector:
Anything. UI, implementation, code complexity....

Sheikah45:
Gotcha will take a look later today

src/main/java/com/faforever/client/chat/ChatUserService.java
```diff
@@ -42,9 +57,44 @@
...
   @Override
   public void afterPropertiesSet() {
     eventBus.register(this);
+    loadAndVerifyEmoticons();
+  }
+
+  private void loadAndVerifyEmoticons() {
+    try {
+      emoticonsGroups = new ObjectMapper().readValue(EMOTICONS_JSON_FILE_RESOURCE.getFile(),
+          TypeFactory.defaultInstance().constructCollectionType(List.class, EmoticonsGroup.class));
```
Sheikah45: We don't want to create a new ObjectMapper every time as they are expensive to create. We have an ObjectMapper that is created on application start and can be injected by adding it to the constructor through spring

Sheikah45: I also would just make a class that is a list of emoticons rather than using the type factory to deserialize or just read it in as an array

src/main/java/com/faforever/client/chat/ChatUserService.java
```diff
@@ -42,9 +57,44 @@
...
+      emoticonShortcodes = new HashMap<>();
+      List<Emoticon> emoticons = emoticonsGroups.stream().flatMap(emoticonsGroup -> emoticonsGroup.getEmoticons().stream()).collect(Collectors.toList());
+      for (Emoticon emoticon : emoticons) {
+        String shortcode = emoticon.getShortcode();
+        if (emoticonShortcodes.containsKey(shortcode)) {
+          throw new ProgrammingError("Shortcode `" + shortcode + "` is already taken");
+        }
+        emoticonShortcodes.put(shortcode, emoticon.getSvgFilePath());
```
Sheikah45: No need to collect to a list here if we are just going to iterate through it again. Just use forEach

src/main/java/com/faforever/client/chat/ChatUserService.java
```diff
@@ -33,6 +46,8 @@
 @RequiredArgsConstructor
 public class ChatUserService implements InitializingBean {
 
+  private static final ClassPathResource EMOTICONS_JSON_FILE_RESOURCE = new ClassPathResource("images/emoticons/emoticons.json");
```
Sheikah45: It would be better to create an EmoticonService rather than embed it in this one as they don't have much to do with each other

src/main/java/com/faforever/client/chat/ChatUserService.java
```diff
@@ -42,9 +57,44 @@
...
+          throw new ProgrammingError("Shortcode `" + shortcode + "` is already taken");
+        }
+        emoticonShortcodes.put(shortcode, emoticon.getSvgFilePath());
+      }
+    } catch (IOException e) {
+      log.error("cannot read emoticons.json file", e);
+      emoticonsGroups = Collections.emptyList();
+    } catch (ProgrammingError e) {
+      log.error(e.getMessage(), e);
+      emoticonsGroups = Collections.emptyList();
```
Sheikah45: Probably want some user notification here otherwise it will just go unnoticed, also maybe don't want to fail safe since neither of these are really user input errors but more errors in the client code

Sheikah45 (commented):
Looks decent to me so far. I think we would want to create a separate EmoticonService rather than coopt the ChatUserService as that is not really its purpose.

Additionally will have to dig in a little more to the cost benefits of having the svgs as separate files vs having them in css

Marc-Spector:
> Additionally will have to dig in a little more to the cost benefits of having the svgs as separate files vs having them in CSS

I'm not sure, but ... What to do with complex paths that contain different colors? How to combine them and put them in the `-fx-shape` attribute. I didn't find any information on Google.

Sheikah45:
What do you mean by complex paths?

Marc-Spector:
For example, content of `1f601.svg`:

image

Marc-Spector:
Found another way not to use separate SVG files.

Sheikah45:
Ah yeah that is much better

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -170,6 +187,9 @@ public AbstractChatTabController(WebViewConfigurer webViewConfigurer,
     this.eventBus = eventBus;
     this.countryFlagService = countryFlagService;
     this.chatUserService = chatUserService;
+    this.emoticonService = emoticonService;
+
+    EMOTICON_SHORTCODE_DETECTOR_PATTERN = emoticonService.getEmoticonShortcodeDetectorPattern();
```
Sheikah45: This is only used in one place so I would just inline the get call. Right now it's definition as a constant like this doesn't really match. If we keep it as a property I would jsut use normal naming convention but since it is part of the emoticonService I don't see a reason to have it in this class as well.

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -687,4 +725,36 @@ protected void onDisplay() {
   protected void onHide() {
 
   }
+
+  public Button getEmoticonsButton() {
+    return emoticonsButton;
+  }
```
Sheikah45: This getter is unnecessary nothing uses it and I don't think we want to expose this button anyway outside of as necessary for the fxml

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -687,4 +725,36 @@ protected void onDisplay() {
...
+
+  public void openEmoticonsPopupWindow() {
+    Button emoticonsButton = getEmoticonsButton();
+    Bounds screenBounds = emoticonsButton.localToScreen(emoticonsButton.getBoundsInLocal());
```
Sheikah45: The getter here is unnecessary

src/main/java/com/faforever/client/chat/emoticons/EmoticonService.java
```diff
@@ -0,0 +1,64 @@
...
+  private final ObjectMapper objectMapper;
+
+  private List<EmoticonsGroup> emoticonsGroups;
+  private final HashMap<String, String> shortcodeToBase64SvgContent = new HashMap<>();
+  private Pattern emoticonShortcodeDetectorPattern;
+
+  @Override
+  public void afterPropertiesSet() throws IOException {
+    loadAndVerifyEmoticons();
+    initializePattern();
```
Sheikah45: I would move initializePattern into loadAndVerifyEmoticons so that way whenever emoticons are loaded the pattern is up to date

src/main/java/com/faforever/client/chat/emoticons/EmoticonsWindowController.java
```diff
@@ -0,0 +1,60 @@
...
+      controller.setGroup(group, onEmoticonClicked());
+      nodes.add(controller.getRoot());
+    });
+    root.getChildren().addAll(nodes);
+  }
+
+  @VisibleForTesting
+  protected Consumer<String> onEmoticonClicked() {
+    return shortcode -> {
+      textInputControl.appendText(" ".concat(shortcode).concat(" "));
```
Sheikah45:
```suggestion
      textInputControl.appendText(" " + shortcode + " ");
```

src/main/java/com/faforever/client/chat/emoticons/EmoticonsWindowController.java
```diff
@@ -0,0 +1,60 @@
...
+      textInputControl.requestFocus();
+      textInputControl.selectEnd();
+    };
+  }
+
+  public void associateWith(TextInputControl textInputControl) {
```
Sheikah45: Just use @Setter on the TextinputControl field
```suggestion
  public void setTextControl(TextInputControl textInputControl) {
```

Sheikah45 (changes requested):

Sheikah45:
It would also be good to add documentation somewhere on how to add emojis

Marc-Spector:
Ok. I need time to write the documentation. Where to place it? Wiki?

Sheikah45:
Yes that would be best

Sheikah45:
https://github.com/FAForever/downlords-faf-client/wiki/Adding-Emojis

Sheikah45 (approved):

PR #2434, opened 2021-12-06
Download patch files from mirrors

### LINKED ISSUE #2433: Download game patches from configured list of mirrors
Some users experience extremely low download speeds (especially on patch days) due to throttling on our content server. The client should be able to download the files from one or more mirrors added through the settings. It will need to verify the checksums against the expected values returned by the API to ensure that the files can be trusted.

Visit Issue hunt...

Askaholic:
Closes #2433

I also made it so that before trying to download ForgedAlliance.exe, the client will try to use the currently installed version and just patch the version number as the majority of updates don't actually change the exe other than the version number. If the checksum of the patched file doesn't match then it will download the exe as usual.

image


src/main/java/com/faforever/client/io/ChecksumMismatchException.java
```diff
@@ -0,0 +1,22 @@
...
+  @Getter
+  private final String expected;
+  @Getter
+  private final String actual;
+
+  public ChecksumMismatchException(Object item, String expected, String actual) {
+    this.item = item;
+    this.expected = expected;
+    this.actual = actual;
+  }
```
Sheikah45:
```suggestion
@Getter
@RequiredArgsConstructor
public class ChecksumMismatchException extends Exception {
  private final Object item;
  private final String expected;
  private final String actual;
```

Sheikah45: Is object really the most restrictive type? Couldn't it be something like File or Path?

src/main/java/com/faforever/client/io/DownloadService.java
```diff
@@ -1,41 +1,91 @@
...
+   */
+  public void downloadFileWithMirrors(URL url, Path targetFile, ByteCountListener progressListener, String md5sum) throws IOException, NoSuchAlgorithmException, ChecksumMismatchException {
+    for (URL mirrorUrl : getMirrorsFor(url)) {
+      try {
+        downloadFile(mirrorUrl, targetFile, progressListener, md5sum);
+        return;
+      } catch (FileNotFoundException e) {
+        log.warn("Could not download file at " + mirrorUrl);
+      } catch (ChecksumMismatchException e) {
+        log.warn(String.format("Checksum downloaded from %s did not match. Expected %s got %s", mirrorUrl, e.getExpected(), e.getActual()));
```
Sheikah45: You don't need String.format here. Also if you have the toString method the same as this method then use that
```suggestion
        log.warn("Checksum did not match from {}", mirrorUrl, e);
```

src/main/java/com/faforever/client/io/DownloadService.java
```diff
@@ -1,41 +1,91 @@
...
+      } catch (IOException e) {
+        log.warn("Could not download file at " + mirrorUrl, e);
```
Sheikah45:
```suggestion
        log.warn("Could not download file at {}" , mirrorUrl, e);
```

src/main/java/com/faforever/client/io/DownloadService.java
```diff
@@ -1,41 +1,91 @@
...
+    downloadFile(url, targetFile, progressListener, md5sum);
+  }
+
+  /*
+   * Download a file from an URL using a temporary path and copy it to targetFile if it downloaded and the checksum
+   * matched.
+   */
+  public void downloadFile(URL url, Path targetFile, ByteCountListener progressListener, String md5sum) throws IOException, NoSuchAlgorithmException, ChecksumMismatchException {
     Path tempFile = Files.createTempFile(targetFile.getParent(), "download", null);
+    log.debug("Downloading file " + url + " to " + tempFile);
```
Sheikah45:
```suggestion
    log.debug("Downloading {} to {}", url, tempFile);
```

src/main/java/com/faforever/client/io/DownloadService.java
```diff
@@ -46,4 +96,25 @@ public void downloadFile(URL url, Path targetFile, ByteCountListener progressLis
...
   }
+
+  /*
+   * Get URLs to the same file on all available mirrors.
+   */
+  public List<URL> getMirrorsFor(URL url) {
+    return preferencesService.getPreferences().getMirror().getMirrorURLs().stream()
+        .map(mirror -> getMirrorURL(mirror, url))
+        .filter(Optional::isPresent)
+        .map(Optional::get)
```
Sheikah45:
```suggestion
        .flatMap(Optional::stream)
```

src/main/java/com/faforever/client/patch/GameUpdaterImpl.java
```diff
@@ -102,16 +102,20 @@ public GameUpdater addFeaturedModUpdater(FeaturedModUpdater featuredModUpdater)
...
-            throw new UnsupportedOperationException("Unable to patch Forged Alliance to the required version " +
-                "due to conflicting version running", throwable);
-          } else if (!allowReplaysWhileInGame) {
+            if (throwable.getCause() instanceof AccessDeniedException) {
+              throw new UnsupportedOperationException("Unable to patch Forged Alliance to the required version " +
+                  "due to conflicting version running", throwable);
+            } else if (throwable.getCause() instanceof ChecksumMismatchException) {
+              throw new CompletionException(throwable.getCause());
+            } else {
+              log.info("Launching anyway", throwable);
```
Sheikah45:
```suggestion
              log.info("Ignored error while updating featured mod", throwable);
```

src/main/java/com/faforever/client/preferences/MirrorPrefs.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class MirrorPrefs {
+  private final ListProperty<URI> mirrorURLs;
+
+  public MirrorPrefs() {
+    mirrorURLs = new SimpleListProperty<>(FXCollections.observableArrayList());
+  }
+
+  public ObservableList<URI> getMirrorURLs() {
+    return mirrorURLs.get();
```
Sheikah45:
```suggestion
  private final ObservableList<URI> mirrorURLs;

  public MirrorPrefs() {
    mirrorURLs = FXCollections.observableArrayList();
  }

  public ObservableList<URI> getMirrorURLs() {
    return mirrorURLs;
```

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -310,11 +313,19 @@ public void initialize() {
...
     autoChannelListView.getItems().addListener((ListChangeListener<String>) c -> preferencesService.storeInBackground());
     autoChannelListView.managedProperty().bind(autoChannelListView.visibleProperty());
     autoChannelListView.visibleProperty().bind(Bindings.createBooleanBinding(() -> !autoChannelListView.getItems().isEmpty(), autoChannelListView.getItems()));
 
+    mirrorURLsListView.setSelectionModel(new NoSelectionModel<>());
+    mirrorURLsListView.setFocusTraversable(false);
+    mirrorURLsListView.setItems(preferencesService.getPreferences().getMirror().getMirrorURLs());
+    mirrorURLsListView.setCellFactory(param -> uiService.<RemovableListCellController<URI>>loadFxml("theme/settings/removable_cell.fxml"));
+    mirrorURLsListView.getItems().addListener((ListChangeListener<URI>) c -> preferencesService.storeInBackground());
+    mirrorURLsListView.managedProperty().bind(mirrorURLsListView.visibleProperty());
```
Sheikah45:
```suggestion
   JavaFxUtil.bindManagedToVisible(mirrorURLsListView);
```

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -310,11 +313,19 @@ public void initialize() {
...
+    mirrorURLsListView.visibleProperty().bind(Bindings.createBooleanBinding(() -> !mirrorURLsListView.getItems().isEmpty(), mirrorURLsListView.getItems()));
```
Sheikah45: If this is ever updated not through the UI then this needs to be a listener and not a binding

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -589,5 +600,21 @@ public void onPatchGamePrefsForMultipleInstances() {
       notificationService.addImmediateErrorNotification(e, "settings.fa.patchGamePrefsFailed");
     }
   }
+
+  public void onAddMirrorURL() {
+    URI uri = URI.create(mirrorURITextField.getText());
+    if (mirrorURITextField.getText().isEmpty() || mirrorURLsListView.getItems().contains(uri)) {
```
Sheikah45:
```suggestion
    if (mirrorURITextField.getText().isBlank() || mirrorURLsListView.getItems().contains(uri)) {
```

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -589,5 +600,21 @@ public void onPatchGamePrefsForMultipleInstances() {
...
+      return;
+    }
+    try {
+      uri.toURL();
+    } catch (Exception e) {
```
Sheikah45:
```suggestion
    } catch (MalformedURLException e) {
```

Sheikah45 (changes requested):
Do not make changes to the messages_XX.properties files those are autoupdatedby a third party. So all those changes should be reverted

src/test/java/com/faforever/client/patch/DownloadServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.net.URI;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.when;
+
+public class DownloadServiceTest extends ServiceTest {
```
Sheikah45: Need a test that the downloadFileWithMirrors rotates through the mirrors properly as well

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
Askaholic: No, I had that and then it threw an IllegalArgumentException

Sheikah45: Then this should onlt check for MalformedURLException and IllegalArgumentException

Also the try should really be around the whole method body since URI.create can also throw the IllegalArgumentException if it doesn't match the spec. If you want to check both the URL and URI at the same time I would just do
`URI uri = new URL(text).toURI()`
and add the try catch around that

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
Askaholic: Yea idk, I just copy/pasted from the autojoinchannel listview

src/main/java/com/faforever/client/io/DownloadService.java
Sheikah45: This still needs to be changed

src/main/java/com/faforever/client/io/ChecksumMismatchException.java
```diff
@@ -0,0 +1,17 @@
+package com.faforever.client.io;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor
+public class ChecksumMismatchException extends Exception {
+  private final Object item;
```
Sheikah45: Can this be more specific than object?

src/main/java/com/faforever/client/io/ChecksumMismatchException.java
```diff
@@ -0,0 +1,17 @@
...
+  private final String expected;
+  private final String actual;
```
Sheikah45: Also sorry just realized they are all private final 
```suggestion
@Value
public class ChecksumMismatchException extends Exception {
  Object item;
  String expected;
  String actual;
```

src/main/resources/i18n/messages_cs.properties
```diff
@@ -758,4 +758,7 @@ chat.userContext.report = Nahlásit hráče
 userInfo.ratingHistory.lastYear = Poslední rok
 userInfo.ratingHistory.lastMonth = Poslední měsíc
 game.generateMap.less = Méně
-game.generateMap.more = Více
\ No newline at end of file
+game.generateMap.more = Více
+settings.data.mirrorURLs=Download Mirrors
+settings.data.mirrorURLs.description=Add any mirrors here that you wish to use for downloading game patches instead of downloading from the main FAF server. This can help prevent extremely long download times.
+settings.data.mirrorURLPrompt=https://faf-mirror.example.com/
```
Sheikah45: These need to be removed from all but the message.properties files

Sheikah45 (changes requested):
Need to revert the changes to the other messages_x.properties files

src/main/resources/i18n/messages_cs.properties
Askaholic: Weird I thought I did that already… will take another look.

Sheikah45: Ah you probably did I wiped some changes with the rebase but thought I fixed them all. So I will go ahead and fix this

Sheikah45: Since I caused it

src/main/java/com/faforever/client/patch/GameUpdaterImpl.java
```diff
@@ -82,18 +83,22 @@ public GameUpdater addFeaturedModUpdater(FeaturedModUpdater featuredModUpdater)
...
-            throw new UnsupportedOperationException("Unable to patch Forged Alliance to the required version " +
-                "due to conflicting version running", throwable);
-          } else if (throwable instanceof UnsupportedOperationException) {
-            throw new CompletionException(throwable);
-          } else if (!allowReplaysWhileInGame) {
+            if (throwable.getCause() instanceof AccessDeniedException) {
+              throw new UnsupportedOperationException("Unable to patch Forged Alliance to the required version " +
+                  "due to conflicting version running", throwable);
+            } else if (throwable.getCause() instanceof ChecksumMismatchException) {
+              throw new CompletionException(throwable.getCause());
```
Sheikah45: Ah just realized this will only rethrow when the player has allowReplaysWhileinGame set which is probably not intended.

I can fix it though as there are a few things to clean up I have been meaning to do

src/main/java/com/faforever/client/io/DownloadService.java
```diff
@@ -1,49 +1,119 @@
...
+  /*
+   * Download file first trying each mirror in order. If all mirrors fail then try the original URL.
+   */
+  public void downloadFileWithMirrors(URL url, Path targetFile, ByteCountListener progressListener, String md5sum) throws IOException, NoSuchAlgorithmException, ChecksumMismatchException {
+    for (URL mirrorUrl : getMirrorsFor(url)) {
+      try {
+        downloadFile(mirrorUrl, targetFile, progressListener, md5sum);
+        return;
+      } catch (FileNotFoundException e) {
+        log.warn("Could not download file at {}", mirrorUrl);
```
Sheikah45: log the exception here please. Also this can be combined with the IOexception catch no need to have 2 that are the same

src/main/java/com/faforever/client/io/DownloadService.java
```diff
@@ -1,49 +1,119 @@
...
           .totalBytes(urlConnection.getContentLength())
           .listener(progressListener)
           .copy();
 
+      // NOTE: It is crucial that we verify the checksum before using the file when downloading from mirrors! We don't
+      // want to be running unverified executables!
+      String checksum = DatatypeConverter.printHexBinary(messageDigest.digest()).toLowerCase();
+      if (!Objects.equals(md5sum, checksum)) {
+        throw new ChecksumMismatchException(url, checksum, md5sum);
+      }
```
Sheikah45, line 89: Can we add a test that exercises this path?

src/main/java/com/faforever/client/io/DownloadService.java
```diff
@@ -1,49 +1,119 @@
...
+  }
+
+  public Optional<URL> getMirrorURL(URI mirror, URL url) {
+      URI uri = mirror.resolve(url.getPath());
+      try {
+        return Optional.of(uri.toURL());
+      } catch (MalformedURLException e) {
+        log.warn("Failed to create URL from URI: {}", uri);
+        return Optional.empty();
+      }
```
Sheikah45: Can we add tests for these as well?

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -589,5 +602,26 @@ public void onPatchGamePrefsForMultipleInstances() {
...
+      log.debug("Failed to add invalid URL: {}", text, e);
+    }
+
+    if (mirrorURLsListView.getItems().contains(uri)) {
+      return;
+    }
+    preferencesService.getPreferences().getMirror().getMirrorURLs().add(uri);
+    preferencesService.storeInBackground();
+    mirrorURITextField.clear();
+  }
```
Sheikah45: Test please

Sheikah45 (changes requested):
A few tests to add then we should be good

src/main/java/com/faforever/client/io/DownloadService.java
Askaholic: I already added tests for these

Askaholic: Not sure why it says that line is uncovered because I definitely already have a test that should hit the 'empty' path here. Maybe my test class isn't being picked up for some reason?

Sheikah45: Could be I will double check and look at the tests. Codecov can be spotty at times

Askaholic: I do get this error when trying to run the test through IDEA:
```
> No tests found for given includes: [com.faforever.client.io.DownloadServiceTest.testDownloadFile](filter.includeTestsMatching)
```

src/main/java/com/faforever/client/io/DownloadService.java
Askaholic: Oh yea the weird thing here is that URL will throw this `FileNotFoundException` if it encounters an http error when trying to open the URL. That's why I separated it out, because it's specifically catching the http errors. I should probably make it log on INFO or something though because I would expect this to happen all the time given that the mirrors may be out of date and not contain all needed files.

Sheikah45: Maybe but if the error message and log levels are the same then there is no need to separate them out since they are performing the same action

src/test/java/com/faforever/client/io/DownloadServiceTest.java
```diff
@@ -129,4 +133,35 @@ public void testDownloadWithMirrors() throws Exception {
...
+    URLConnection urlConnection = mock(URLConnection.class);
+
+    byte[] DATA = "some content".getBytes();
+    when(url.openStream()).thenReturn(new ByteArrayInputStream(DATA));
+    when(urlConnection.getContentLength()).thenReturn(DATA.length);
+    when(url.openConnection()).thenReturn(urlConnection);
+
+    Path temp = Files.createTempFile("download", ".dat");
+    instance.downloadFile(url, temp, (a, b) -> {}, "00000000000000000000000000000000");
+  }
```
Sheikah45: It is better to wrap the instance.downloadFile inside of an Assertions.assertThrows() so that we are checking the specific line we expect to throw the exception rather than the whole method

Askaholic: Yea I wasn’t sure how to do it in Java. I’ll update it

src/main/java/com/faforever/client/io/DownloadService.java
Askaholic: I changed it to info

src/test/java/com/faforever/client/io/DownloadServiceTest.java
```diff
@@ -0,0 +1,167 @@
+package com.faforever.client.io;
+
+import com.faforever.client.preferences.PreferencesService;
+import com.faforever.client.test.ServiceTest;
+import com.faforever.commons.io.ByteCountListener;
+import javafx.collections.FXCollections;
+import org.assertj.core.api.Assertions;
+import org.junit.Test;
```
Sheikah45: This is the wrong test annotation

src/test/java/com/faforever/client/io/DownloadServiceTest.java
```diff
@@ -0,0 +1,167 @@
...
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class DownloadServiceTest extends ServiceTest {
+
+  @Mock(answer = Answers.RETURNS_DEEP_STUBS)
+  private PreferencesService preferencesService;
```
Sheikah45: There is no need to use a deep stub here. We have a preferencesBuilder for testing so we can pass complete data

Sheikah45 (changes requested):

Sheikah45 (commented):
I fixed the tests so that they run now but it looks like some of them are failing so that should be looked into

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -310,11 +316,18 @@ public void initialize() {
...
-    autoChannelListView.managedProperty().bind(autoChannelListView.visibleProperty());
+    JavaFxUtil.bindManagedToVisible(autoChannelListView);
     autoChannelListView.visibleProperty().bind(Bindings.createBooleanBinding(() -> !autoChannelListView.getItems().isEmpty(), autoChannelListView.getItems()));
 
+    mirrorURLsListView.setSelectionModel(new NoSelectionModel<>());
+    mirrorURLsListView.setFocusTraversable(false);
+    mirrorURLsListView.setItems(preferencesService.getPreferences().getMirror().getMirrorURLs());
+    mirrorURLsListView.setCellFactory(param -> uiService.<RemovableListCellController<URI>>loadFxml("theme/settings/removable_cell.fxml"));
+    mirrorURLsListView.getItems().addListener((ListChangeListener<URI>) c -> preferencesService.storeInBackground());
+    mirrorURLsListView.getItems().addListener((ListChangeListener<URI>) c -> mirrorURLsListView.setVisible(!mirrorURLsListView.getItems().isEmpty()));
```
Sheikah45: rather than two listeners it is probably better to just make both into a function and then call that function in a single listener.

Also will want to use JavaFxUtil.addListener for thread safety just in case.

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -589,5 +602,30 @@ public void onPatchGamePrefsForMultipleInstances() {
...
+    if (!text.endsWith("/")) {
+      text = text + "/";
+    }
+
+    URI uri = null;
+    try {
+      uri = new URL(text).toURI();
+    } catch (URISyntaxException | MalformedURLException e) {
+      log.debug("Failed to add invalid URL: {}", text, e);
+      return;
```
Sheikah45: Probably want a user notification here

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -589,5 +602,30 @@ public void onPatchGamePrefsForMultipleInstances() {
...
+    }
+
+    if (mirrorURLsListView.getItems().contains(uri)) {
+      return;
+    }
+    preferencesService.getPreferences().getMirror().getMirrorURLs().add(uri);
+    preferencesService.storeInBackground();
+    mirrorURITextField.clear();
```
Sheikah45: Since you have a return in the catch anyway everything after the catch can just be moved into the try statement, this removes the possibility of uri being null as well since you can move the declaration there as well.

src/test/java/com/faforever/client/io/DownloadServiceTest.java
```diff
@@ -0,0 +1,171 @@
...
+    verify(downloadServiceSpy).downloadFile(new URL("https://mirror1.example.com/bar"), targetFile, progressListener, md5sum);
+    verify(downloadServiceSpy).downloadFile(new URL("https://mirror2.example.com/bar"), targetFile, progressListener, md5sum);
+    verify(downloadServiceSpy).downloadFile(new URL("https://mirror3.com/foo/bar"), targetFile, progressListener, md5sum);
+    verify(downloadServiceSpy).downloadFile(new URL("https://example.com/bar"), targetFile, progressListener, md5sum);
+  }
+
+  @Test
+  public void testDownloadFile() throws Exception {
+    Path temp = Files.createTempFile("download", ".dat");
+    instance.downloadFile(SAMPLE_FILE.getURL(), temp, (processed, total) -> {}, "b53227da4280f0e18270f21dd77c91d0");
```
Sheikah45: Can we make the correct checksum into a constant as well

Sheikah45 (changes requested):
Just a few things I missed before

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
Askaholic: It would be nice if there were a way to flash an error message but I think a notification would be a bit overkill. You can tell that there is an error by the fact that the text field is not cleared but it would also be nice to show a small tooltip like message. Not sure if there’s a way to do that in JavaFX.

Sheikah45: I would just use the addimmeadiateerrornotification

Sheikah45: Err actually probably just addimmeadiatewarnnotification

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
Askaholic: I switched it to use JavaFxUtil, but I didn't see anywhere in the code that used a method rather than a lambda. Lambda is fine IMO. It even makes it easier to read because the code for the listener is right there where the listener is added.

Sheikah45 (approved):

PR #2475, opened 2021-12-16
league ui fixes

BlackYps:
This fixes some ui issues with the league system, most notably the division images, but also some annoyances when using background images.
Requires https://github.com/FAForever/faf-java-commons/pull/78 to be merged so I can bump the faf commons version to the appropriate commit.
With this we should be able to make a new prerelease that has no known bugs, so I can make a forum post to promote it for testing.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -298,7 +298,7 @@ dependencies {
   implementation("org.springframework:spring-web")
   implementation("org.springframework:spring-websocket")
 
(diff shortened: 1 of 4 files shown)


Sheikah45 (dismissed):
Just update to the merged commons versiion

Sheikah45 (approved):

Issue #1227, opened 2019-05-07
Allow deleting local replays


Geosearchef:
http://forums.faforever.com/viewtopic.php?f=2&t=17068&start=60#p173729

fultonm:
Hi, I'd like to take this on. Are there any special considerations to keep in mind? My initial thought for where to place the "Delete Replay" button would be on the Replay Details modal, to the right of the "Report Game" button.

As a side note, in the linked topic, they also talk about the grid based interface for FAF Replays and Local Replays. I wonder if supporting a list interface such as the one used for Live Games would be a good UX improvement?

Sheikah45:
Yeah feel free to take this on. With regards to the grid that is not something I think is as interesting or good as it requires supporting two different formats for all future changes which we want to avoid.

PR #2490, opened 2021-12-24
Allow deleting local replay file from disk.

fultonm:
Addresses #1227 

There are some new strings to translate:

- delete = Delete
- replay.deleteTooltip = Delete local replay file
- replay.delete = Delete Replay
- replay.deleteLocal = Delete local replay
- replay.deleteWarn = This action will permanently delete the replay from your hard drive.

Let me know what you think!


src/main/java/com/faforever/client/main/event/DeleteLocalReplayEvent.java
```diff
@@ -0,0 +1,12 @@
...
+import javafx.event.Event;
+import javafx.event.EventType;
+
+public class DeleteLocalReplayEvent extends Event {
+  public static final EventType<DeleteLocalReplayEvent> DELETE_LOCAL_REPLAY_EVENT_TYPE = new EventType<>(ANY);
+
+  public DeleteLocalReplayEvent() {
+    super(DELETE_LOCAL_REPLAY_EVENT_TYPE);
+  }
+}
```
Sheikah45: For events we actually use the Google EventBus if you search for it in the source you should see a few examples. This allows arbitrary objects to be events

src/main/java/com/faforever/client/replay/LocalReplayVaultController.java
```diff
@@ -50,10 +51,15 @@ public void initialize() {
...
     JavaFxUtil.assertApplicationThread();
     replayDetailController.setReplay(replay);
+    replayDetailController.getRoot().addEventHandler(DeleteLocalReplayEvent.DELETE_LOCAL_REPLAY_EVENT_TYPE, event -> onLocalReplayDeleted());
     replayDetailController.getRoot().setVisible(true);
     replayDetailController.getRoot().requestFocus();
   }
 
+  private void onLocalReplayDeleted() {
+    onPageChange(searchController.getLastSearchConfig(), false);
+  }
```
Sheikah45: Will want to convert this to use the google eventBus and so will need to annotate this method with @ Subscribe

src/main/java/com/faforever/client/replay/LocalReplayVaultController.java
```diff
@@ -50,10 +51,15 @@ public void initialize() {
   protected void onDisplayDetails(ReplayBean replay) {
     JavaFxUtil.assertApplicationThread();
     replayDetailController.setReplay(replay);
+    replayDetailController.getRoot().addEventHandler(DeleteLocalReplayEvent.DELETE_LOCAL_REPLAY_EVENT_TYPE, event -> onLocalReplayDeleted());
```
Sheikah45: Use google eventbus, might need to be added as a dependency through spring injection and also registration added in the initialize function

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -425,6 +434,20 @@ public void onReport() {
...
+    notificationService.addNotification(new ImmediateNotification(
+        String.format("%s %s", i18n.get("replay.deleteLocal"), replay.getTitle()), i18n.get("replay.deleteWarn"), Severity.INFO, Arrays.asList(
+        new Action(i18n.get("cancel")),
+        new Action(i18n.get("delete"), event -> deleteReplay())
+    )));
+  }
+
+  private void deleteReplay() {
+    replayService.deleteReplayFile(replay.getReplayFile());
+    getRoot().fireEvent(new DeleteLocalReplayEvent());
```
Sheikah45: use `eventbus.post`

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -266,6 +267,14 @@ private void moveCorruptedReplayFile(Path replayFile) {
     ));
   }
 
+  public void deleteReplayFile(Path replayFile) {
+    try {
+      delete(replayFile);
+    } catch (IOException e) {
+      log.error("Failed to delete local replay file {}", replayFile, e);
```
Sheikah45, line 243: probably want to add a user notification about the failure here

src/main/resources/i18n/messages.properties
```diff
@@ -766,12 +767,16 @@ replay.replayDurationTooltip = Replay length
 replay.playerCountTooltip = Players
 replay.ratingTooltip = Average rating
 replay.qualityTooltip = Quality
+replay.deleteTooltip = Delete local replay file
+replay.delete = Delete Replay
+replay.deleteLocal = Delete local replay
```
Sheikah45: why delete and deleteLocal I would think we only need one

Sheikah45 (changes requested):
Looking good. One note that you will see throughout is that we actuall use the google event bus for publsihing and subscribing to events. So if you convert to using that we should be in a good place.

Also as a thought I wonder if a delete button would fit on the local replay card as well.

src/main/java/com/faforever/client/main/event/DeleteLocalReplayEvent.java
fultonm: It seems I need to change a several classes constructors to have access to that EventBus. Does that sound correct or is there a better way to inject this dependency into LocalReplayVaultController? Any hint appreciated :)

fultonm: @Sheikah45 when you get a chance, not sure if you saw my reply buried in here

src/main/resources/i18n/messages.properties
fultonm: `replay.delete` is used as a label for the Delete Replay button, while `replay.deleteLocal` is used as part of the confirmation notification title. I updated the translation entry keys to be more descriptive of their purpose. To me, the difference in capitalization and verbosity make sense in each of their context. Let me know if we need to minimize the number of translations.

src/main/java/com/faforever/client/main/event/DeleteLocalReplayEvent.java
Sheikah45: Yes you should add the eventbus to the constructor

src/main/java/com/faforever/client/main/event/DeleteLocalReplayEvent.java
```diff
@@ -0,0 +1,12 @@
...
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+import java.nio.file.Path;
+
+@RequiredArgsConstructor
+public class DeleteLocalReplayEvent {
+  @Getter
+  private final Path replayFile;
```
Sheikah45:
```suggestion
@Value
public class DeleteLocalReplayEvent {
  Path replayFile;
```

Use @ Value here as it does everything you need with one annotation

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -75,7 +86,10 @@ public void setReplay(ReplayBean replay) {
     } else {
       onMapLabel.setText(i18n.get("game.onUnknownMap"));
     }
-
+    if (replay.getReplayFile() != null) {
+      deleteButton.setManaged(true);
+      deleteButton.setVisible(true);
```
Sheikah45: Instead of setting managed and visible true create an initialize function that uses the JavafxUtil.bindManagedToVisible() method so that we only need to set the visible flag as we do in other places

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -277,6 +286,8 @@ public void setReplay(ReplayBean replay) {
       chatTable.setItems(replay.getChatMessages());
       optionsTable.setItems(replay.getGameOptions());
       moreInformationPane.setVisible(true);
+      deleteButton.setVisible(true);
+      deleteButton.setManaged(true);
```
Sheikah45: Same bindManagedToVisible comment as above. Here you can can just add it to the already existing call

fultonm:
At the top of ReplayController.java, here is a todo: `// TODO change to spring event bus`

Sheikah45 (changes requested):
Looking good just a few changes

fultonm:
> At the top of ReplayController.java, here is a todo: `// TODO change to spring event bus`

@Sheikah45 should this be resolved or is it OK?

Sheikah45:
Yes that can be removed everything was already converted to the event bus

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -75,7 +86,10 @@ public void setReplay(ReplayBean replay) {
     } else {
       onMapLabel.setText(i18n.get("game.onUnknownMap"));
     }
-
+    JavaFxUtil.bindManagedToVisible(deleteButton);
```
Sheikah45: This should be done in an initialize function as it only needs to be done once

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -75,7 +86,10 @@ public void setReplay(ReplayBean replay) {
...
+    if (replay.getReplayFile() != null) {
+      deleteButton.setVisible(true);
+    }
```
Sheikah45:
```suggestion
  deleteButton.setVisible(replay.getReplayFile() != null);
```

src/main/resources/theme/vault/replay/replay_card.fxml
```diff
@@ -84,11 +84,24 @@
...
+    <HBox spacing="10.0" alignment="BOTTOM_RIGHT" GridPane.halignment="RIGHT" GridPane.valignment="BOTTOM"
+          GridPane.columnIndex="2" GridPane.rowIndex="8"
+          GridPane.hgrow="ALWAYS" GridPane.vgrow="ALWAYS">
+        <Button fx:id="deleteButton" mnemonicParsing="false"
+                visible="false" managed="false"
+                onAction="#onDeleteButtonClicked"
+                styleClass="delete-button">
+            <graphic>
+                <Region styleClass="icon,delete-icon"/>
+            </graphic>
```
Sheikah45: This should be the same delete icon as the ones used for the maps and mods delete buttons

src/main/resources/theme/style.css
```diff
@@ -19,6 +19,45 @@
...
+
+.main-navigation-button:hover {
+  -fx-border-color: -fx-selection-bar-non-focused;
+  -fx-background: #FFFFFF12;
+}
+
+.main-navigation-button:armed, .main-navigation-button:selected, .main-navigation-button:focused:selected {
+  -fx-background-color: #FFFFFF33;
+  -fx-border-color: -fx-accent;
+}*//
```
Sheikah45: Remove the commented out code

Sheikah45 (changes requested):

src/main/resources/theme/style.css
fultonm: Oops, I was using this to template the delete button icon styling.

src/main/resources/theme/vault/replay/replay_card.fxml
fultonm: Updated this one to `uninstall-icon` like the maps and mods icons.

fultonm:
I think the buttons on the replay, maps, and mods section should match in style as much as possible, so I added color to those as well.

- The colored buttons will highlight the shortcut to perform these actions directly from the card without the need to navigate into the detail view.
- Blue and Red make it clear what the buttons do without needing to recognize and process the meaning of the icon (less brain cells.)
- I think it will encourage players to actually use these buttons more.

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -160,4 +176,19 @@ public void onShowReplayDetail() {
   public void onWatchButtonClicked() {
     replayService.runReplay(replay);
   }
+
+  public void onDeleteButtonClicked() {
+    notificationService.addNotification(new ImmediateNotification(
+      String.format("%s %s",
+          i18n.get("replay.deleteNotification.heading"),
+          replay.getTitle()),
```
Sheikah45: Rather than use string.format you should add the argument to the string in the messages.properties file with the placeholder {0} then pass the replay title as an argument to the i18n.get call. This avoids the inherent locale dependency of String.format
```suggestion
          i18n.get("replay.deleteNotification.heading", replay.getTitle())
```

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -160,4 +176,19 @@ public void onShowReplayDetail() {
...
+      i18n.get("replay.deleteNotification.warning"),
+      Severity.INFO, Arrays.asList(
```
Sheikah45: If the severity is just info then the message shouldn't be call warning

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -425,6 +435,23 @@ public void onReport() {
     reportDialogController.show();
   }
 
+  public void onDeleteButtonClicked() {
+    notificationService.addNotification(new ImmediateNotification(
+      String.format("%s %s",
+        i18n.get("replay.deleteNotification.heading"),
+        replay.getTitle()),
```
Sheikah45: Same i18n comment as above

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -425,6 +435,23 @@ public void onReport() {
...
+        i18n.get("replay.deleteNotification.warning"),
+        Severity.INFO, Arrays.asList(
```
Sheikah45: same naming comment

Sheikah45 (changes requested):
Looking good just a few small changes left

src/main/java/com/faforever/client/replay/ReplayCardController.java
fultonm: Nice, didn't realize this was a thing

Sheikah45:
Hmm it looks like some of the rebasing onto develop didn't happen properly as some commits are duplicated here

fultonm:
> Hmm it looks like some of the rebasing onto develop didn't happen properly as some commits are duplicated here

I used merge instead of rebase, so I had to back up it up and reorder some commits and rebase instead. I force pushed to fix the history

Sheikah45 (dismissed):

Sheikah45:
Although it looks like some tests are failing

Sheikah45 (approved):

PR #2501, opened 2022-01-03
Update FafClientApplication.java

Evgene-Kopylov:
Remove default Windows admin check.
It should not bozer regular users

Screenshot 2022-01-03 190818
.

### DIFF
--- a/src/main/java/com/faforever/client/FafClientApplication.java
+++ b/src/main/java/com/faforever/client/FafClientApplication.java
@@ -87,19 +87,6 @@ public class FafClientApplication extends Application {
 
   @Override
   public void init() throws InterruptedException, NoSuchFieldException, IllegalAccessException {
-    if (org.bridj.Platform.isWindows() && WindowsUtil.isAdmin()) {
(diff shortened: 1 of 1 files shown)


Sheikah45:
The admin check warning is there to ensure users do not accidentally run the client as admin and mess up the file permissions on the files the client uses and writes to. It will not be removed.

Evgene-Kopylov:
> The admin check warning is there to ensure users do not accidentally run the client as admin and mess up the file permissions on the files the client uses and writes to. It will not be removed.

it disturbs gaming experience. that is critical issue.

bukajsytlos:
can you elaborate? it's one time check. you should not run applications under admin as 99% of users do.

Evgene-Kopylov:
it is not one-time. it pops up every time. 
and there is no easy way, nor reason to fix Win sys settings.

the correct way seems to add "/AdminCheck" as command line parameter instead.

bukajsytlos:
maybe there could be checkbox I understand the risk and I will not bother technical support for breaking client

PR #2521, opened 2022-01-11
Check if password is ascii

### LINKED ISSUE #2517: Cannot host custom games if the password contains "ø"
**Describe the bug**
Cannot host custom games if the password contains "ø".

After using a password containing "ø" once, the client fails to host games, even when the password does not contain "ø".

After a FAF Client restart, passords not containg "ø" work again.

**To Reproduce**
1. Go to "Play", then "Custom games"
2. "Create game"
3. Enter a password with "ø" in it
4. Press "Create".
5. Observe that the game does not start - nothing is happening.

On 1.5.0 and before, it was possible to host games with "ø" in the password.

**Log or error message**
No error message.

Log example: https://subcons.teod.eu/static/2021-12-30-debug.log.txt

More details & discussion: https://forum.faforever.com/topic/3114/unable-to-host-custom-games-any-advice/16

**Expected behavior**
Pressing "Create" starts up the game, and we enter a lobby.

**OS**
Windows

**Workaround**

Don't use "ø" in passwords.

This might be an issue for other non ANSI characters too.

Sheikah45:
Closes #2517

### DIFF
--- a/src/main/java/com/faforever/client/game/CreateGameController.java
+++ b/src/main/java/com/faforever/client/game/CreateGameController.java
@@ -180,6 +180,7 @@ public class CreateGameController implements Controller<Pane> {
 
     JavaFxUtil.addAndTriggerListener(userService.connectionStateProperty(), new WeakInvalidationListener(createButtonStateListener));
     JavaFxUtil.addListener(titleTextField.textProperty(), new WeakInvalidationListener(createButtonStateListener));
+    JavaFxUtil.addListener(passwordTextField.textProperty(), new WeakInvalidationListener(createButtonStateListener));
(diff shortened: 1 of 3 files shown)


Askaholic:
Passwords can be arbitrary utf8 
https://github.com/FAForever/server/blob/2f39b9785726f6659c49fd26c559354cbcc4f607/server/lobbyconnection.py#L979

teodorlu:
That was quick, thanks!

Sheikah45:
> Passwords can be arbitrary utf8 
> https://github.com/FAForever/server/blob/2f39b9785726f6659c49fd26c559354cbcc4f607/server/lobbyconnection.py#L979

Hmm okay maybe something else was causing the game not to launch before. But for now I will just keep the check in.

PR #2524, opened 2022-01-13
Fix possible memory leak in CSS Classes of cells

Marc-Spector:
I connected the profiler to the client.
As you can see from the graph, memory consumption increases dramatically during each minute, until the garbage collector starts its work. It looks like a memory leak.
After analyzing the code, I came to the conclusion that the `styleClasses` object is not a hash collection, which is why `styleClasses` contains many identical class names, which in turn triggers many monotonous actions (unexplained CPU spikes)

Before fix:
image

After fix:
image

### DIFF
--- a/src/main/java/com/faforever/client/fx/DualStringListCell.java
+++ b/src/main/java/com/faforever/client/fx/DualStringListCell.java
@@ -28,13 +28,13 @@ public class DualStringListCell<T> extends ListCell<T> {
       } else {
         if (dualStringListCellController == null) {
           dualStringListCellController = uiService.loadFxml("theme/dual_string_list_cell.fxml");
+          // copy font styles
+          dualStringListCellController.applyFont(getFont());
+          dualStringListCellController.applyStyleClass(styleClasses);
         }
         dualStringListCellController.setLeftText(functionLeft.apply(item));
         dualStringListCellController.setWebViewToolTip(functionWebViewToolTip.apply(item));
         dualStringListCellController.setRightText(functionRight.apply(item));
-        // copy font styles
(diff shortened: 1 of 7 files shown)


src/main/java/com/faforever/client/fx/DualStringListCell.java
```diff
@@ -28,13 +28,13 @@ protected void updateItem(T item, boolean empty) {
       } else {
         if (dualStringListCellController == null) {
           dualStringListCellController = uiService.loadFxml("theme/dual_string_list_cell.fxml");
+          dualStringListCellController.applyStyleClass(styleClasses);
+          dualStringListCellController.applyFont(getFont());
         }
         dualStringListCellController.setLeftText(functionLeft.apply(item));
         dualStringListCellController.setWebViewToolTip(functionWebViewToolTip.apply(item));
         dualStringListCellController.setRightText(functionRight.apply(item));
         // copy font styles
```
Sheikah45: This comment should be moved as well

src/main/java/com/faforever/client/fx/IconCell.java
```diff
@@ -12,17 +12,11 @@
...
-
-  public IconCell(Function<T, String> iconCssClassFunction) {
-    this(iconCssClassFunction, new String[0]);
-  }
 
   @Override
   protected void updateItem(T item, boolean empty) {
     super.updateItem(item, empty);
 
-    setText(null);
```
Sheikah45, line 25: Why was this removed? We should make sure the text is null for icon cells

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #2529, opened 2022-01-20
Add Smart live replays

Marc-Spector:
Closes #82 
Closes #2525 
Marks #1959 as outdated because the `Live Replayes` table interaction is disabled (no selection)

PR #2544, opened 2022-01-26
Show avatar next to the player's name

### LINKED ISSUE #2530: Increase the importance of avatars, show them wherever possible

Marc-Spector:
Closes #2530


Marc-Spector:
example1
example2

Askaholic:
Shouldn't they be to the left of the text?

Marc-Spector:
I corrected it only for the team card. In the rest, it is to the right of the player's name.

Marc-Spector:
image

Sheikah45:
It should be to the left everywhere to keep consistency

src/main/java/com/faforever/client/game/GameTileController.java
```diff
@@ -78,6 +83,10 @@ private void onGamePropertyChanged() {
     JavaFxUtil.runLater(() -> {
       gameTitleLabel.setText(StringUtils.normalizeSpace(game.getTitle()));
       hostLabel.setText(game.getHost());
+      playerService.getPlayerByNameIfOnline(game.getHost()).map(PlayerBean::getAvatar).ifPresent(avatar -> {
+        avatarImageView.setImage(avatarService.loadAvatar(avatar));
+        Tooltip.install(avatarImageView, new Tooltip(avatar.getDescription()));
+      });
```
Sheikah45: Do not load the avatar images in the application thread it should be moved outside the run later

Marc-Spector:
image

Does it look normal?

Marc-Spector:
Ok, no problem if that's what the maintainer wants. =)

Sheikah45:
I would probably not have them in the live replay list at all. And if I did we should add the full flag there as well. But because of how cluttered it is I think it is overkill.

src/main/java/com/faforever/client/avatar/AvatarService.java
```diff
@@ -29,12 +33,26 @@
...
+  public Optional<Image> getCurrentAvatarByPlayerName(String playerName) {
+    return Optional.ofNullable(loadAvatar(avatarsByPlayerName.get(playerName)));
+  }
+
+  public void addPlayerAvatarToMap(PlayerBean playerBean) {
+    if (playerBean.getAvatar() != null) {
+      avatarsByPlayerName.putIfAbsent(playerBean.getUsername(), playerBean.getAvatar());
+    }
+  }
+
```
Sheikah45: Right now this completely avoids the current caching system which uses the @ Cacheable annotation. 
As put here this will load the image from the file system everytime the getCurrentAvatarByPlayerName is called. Rather than storing your own cache map of string to AvatarBean you should probably use the playerService to get the playerBean by player name and then get the avatarBean from the player bean and load it using the loadAvatar function.

The other issue with this method is that the avatar is not updated in the map you have if the player changes avatar while online. and it wouldn't even get updated if the player logs out and back in.

As it stands this is just a little redundant information and causes too much coupling.

Sheikah45 (changes requested):
Main issue is that the additions to the avatarService cause a tight coupling of playerName to avatar when we already have the player service that can give that information. As well as avoids using the caching system so will result in lots of loads from the disk for the images. 

The loads happens because the cache is only activated when a method is called from outside its own class. If a method is called from inside its own class the cache is not checked.

Marc-Spector:
> The loads happens because the cache is only activated when a method is called from outside its own class. If a method is called from inside its own class the cache is not checked.

I did not know. Could you please provide a link where it says this?

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/cache/annotation/Cacheable.html
https://www.baeldung.com/spring-cache-tutorial
There is no such information in these links.

src/main/java/com/faforever/client/avatar/AvatarService.java
Marc-Spector: >  you should probably use the playerService to get the playerBean by player name

You forget about performance. What if the `playersByName` map contains 10000, 100,000, or more elements, of which 1000 players have avatars? Calling `map.get()` method for the map will be expensive. We still need a separate map.

Sheikah45:
It is because spring uses a proxy object in order to intercept the call and check the cache. If you are calling from the object itself there is no proxy used.  See self-invocation here https://code.google.com/archive/p/ehcache-spring-annotations/wikis/UsingCacheable.wiki

And no worries I only found this out after I ended up releasing a client that overloaded the server because of all the requests that were made because a cache wasn't being used.

src/main/java/com/faforever/client/avatar/AvatarService.java
Sheikah45: it doesn't matter how many players are inside of a map. Calling get on the hash map is a O(1) operation. It is the same speed no matter how large the map is.

The objects are stored in buckets where each bucket is reference by the hash of the object. With the implementation there are going to be practically no collisions. So there is no need to search the map. Given the key we can just directly retrieve the player. There is no need to search throughout the whole map.

Marc-Spector: Thanks very much for explaining it.

Marc-Spector:
`testInvisibleItemIfPlayerIsSelf` test is already fixed in #2529

Sheikah45:
It would be good to still fix the test in this pr as well so that the main branch isn't in a state of failing checks as that prevents things like releases and what not

src/main/java/com/faforever/client/player/PlayerService.java
```diff
@@ -144,6 +147,11 @@ public boolean areFriendsInGame(GameBean game) {
         .collect(Collectors.toList());
   }
 
+  public Optional<Image> getCurrentAvatarByPlayerName(String name) {
+    PlayerBean player = playersByName.getOrDefault(name, null);
+    return Optional.ofNullable(avatarService.loadAvatar(player != null ? player.getAvatar() : null));
```
Sheikah45:
```suggestion
    return Optional.ofNullable(player.getAvatar()).map(avatarService::loadAvatar).orElse(null);
```
If you are already using an optional no need to use a tertiary operator just use the map function.

Sheikah45 (changes requested):
Just a small thing and then the test

src/main/java/com/faforever/client/game/PlayerCardTooltipController.java
```diff
@@ -43,18 +46,21 @@ public void setPlayer(PlayerBean player, Integer rating, Faction faction) {
...
-
-    String playerInfoLocalized;
-    if (rating != null) {
-      playerInfoLocalized = i18n.get("userInfo.tooltipFormat.withRating", player.getUsername(), rating);
-    } else {
-      playerInfoLocalized = i18n.get("userInfo.tooltipFormat.noRating", player.getUsername());
-    }
+    countryFlagService.loadCountryFlag(player.getCountry()).ifPresentOrElse(image ->
+        countryImageView.setImage(image), () -> countryImageView.setVisible(false));
+    playerService.getCurrentAvatarByPlayerName(player.getUsername()).ifPresent(avatar -> {
```
Sheikah45:
```suggestion
    Optional.ofNullable(player.getAvatar()).ifPresent(avatar -> {
```

Sheikah45 (changes requested):
And just fix the conflicts

Sheikah45 (approved):

PR #2565, opened 2022-02-16
Close login ports and send failure html

Sheikah45:
Ensures that the websockets are closed if the user hasn't logged in within 5 minutes of requesting the login, making the ports available for reuse and are not left open indefinitely.

Also removes the statefulness from OAuthValuesReceiver since it was only being used for testing


src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
```diff
@@ -38,17 +35,14 @@
...
 @Slf4j
 public class OAuthValuesReceiver {
 
   private static final Pattern CODE_PATTERN = Pattern.compile("code=([^ &]+)");
   private static final Pattern STATE_PATTERN = Pattern.compile("state=([^ &]+)");
   private static final List<String> ALLOWED_HOSTS = List.of("localhost");
-
-  @VisibleForTesting
-  CompletableFuture<Integer> listenPort = new CompletableFuture<>();
+  private static final int MILLISECONDS_TO_WAIT = 300000;
```
Brutus5000: Is this still used?

Brutus5000 (commented):
5 minutes seems **very** short. I'd go for 15.

src/main/java/com/faforever/client/config/ClientProperties.java
```diff
@@ -224,5 +224,6 @@ public void updateFromEndpoint(ServerEndpoints serverEndpoints) {
     private URI redirectUri = URI.create("");
     private String clientId;
     private String scopes;
+    private int timeout;
```
Brutus5000: nobody knows this is supposed to be milliseconds

Sheikah45: Good point

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
Sheikah45: Nope can be removed

Sheikah45:
> 5 minutes seems **very** short. I'd go for 15.

Must be a slow clicker XD

PR #2580, opened 2022-02-22
ICE adapter debug window

### LINKED ISSUE #2579: Allow launching ice adapter debug window

1-alex98:


1-alex98:
Fixes #2579

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
```diff
@@ -154,10 +155,16 @@ public void onGpgGameMessage(GpgOutboundMessageEvent event) {
 
       Path workDirectory = Path.of(nativeDir).toAbsolutePath();
 
+      var classpath = getBinaryName(workDirectory);
```
Sheikah45: We don't really use `var` anywhere else so please use the full type 😝

Sheikah45 (commented):
Saw something quick will do a full review later

Geosearchef:
Euhm. Wow. Wtf? This works?
Thanks for single-handidly dispersing all my worriers about the ice adapter debug capabilities without modifying the actual adapter.

I tried upgrading the java version of it to integrate JFX, but ran into quite some issues.

Does this workaround work, given that the ice adapter compiled with java 8 and the jfx you use is from the client?

Sheikah45:
I suspect that since the client java is running the ice adapter it is running at the jvm15 level so as long as the class names are the same it should run.

1-alex98:
I tried it works flawlessly

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
1-alex98: Ok... I did use it a lot lately. And I like it. Especially now Intellij always shows the type of vars

1-alex98:
> Euhm. Wow. Wtf? This works? Thanks for single-handidly dispersing all my worriers about the ice adapter debug capabilities without modifying the actual adapter.
> 
> I tried upgrading the java version of it to integrate JFX, but ran into quite some issues.
> 
> Does this workaround work, given that the ice adapter compiled with java 8 and the jfx you use is from the client?

jvm can execute older version byte code without any issues and also mix it with new byte code

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
```diff
@@ -237,6 +244,12 @@ public void onGpgGameMessage(GpgOutboundMessageEvent event) {
     return iceAdapterClientFuture;
   }
 
+  private String javaFXClassPathJars() {
+    final String classPathOfClient = System.getProperty("java.class.path");
+    final var split = List.of(classPathOfClient.split(";"));
```
Sheikah45:
```suggestion
    final List<String> split = List.of(classPathOfClient.split(";"));
```

Sheikah45 (changes requested):

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
1-alex98: meh XD

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
Sheikah45: Yeah understood although I don't know how I feel about it yet

Sheikah45 (approved):

PR #2581, opened 2022-02-22
Remove friend and foe icons and replace them with a background

Marc-Spector:
Makes the design cleaner.

You can close the pull request if you think there is no reason to update the design.

image


Sheikah45:
Probably would be good to keep the icons for people who are red green colorblind

Askaholic:
I think there still needs to be something other than the color that shows who is a friend and who is a foe. Especially considering that red-green colorblindness is the most common form of colorblindness.

Brutus5000:
How about bold or underlined for friend and strikethrough for foe

Marc-Spector:
Maybe add the option "Colorblind mode" in the settings of the client?

Marc-Spector:
> How about bold or underlined for friend and strikethrough for foe

Nooo, it is even worse than icons. 😄

ghost:
Just color the text. Bg should be clear :P

Sheikah45:
For now I think the icons are better than the color background in terms of visibility

PR #2587, opened 2022-02-27
For Linux this launches proton using steam ID 9420 to launch FA when protontricks-launch is detected.

### LINKED ISSUE #2586: linux, replays, need proton/steamcmd love.  also, need something extensible that's not hardcoded for other channels.  what other channels though?
this appears to be an attempt to parse the executable using the shell.

given this is on ubuntu and not windows, there may be some proton magic required to perform a .cmd file +-  `protontricks 9420 shell `  with some drive letter normalization, or steamcmd which i am even less familiar with.  

I am running this locally in intellij.  If unifying the lobby/match and replay code is a good idea, I am happy to help.  if it's not a good idea, more info about this is also greatly appreciated!

details below:

2022-02-27 15:41:17.097  INFO 11776 --- [onPool-worker-3] c.f.client.fa.ForgedAllianceService      : Starting Forged Alliance with command: [%USER_PROFILE%/.faforever/bin/ForgedAlliance.exe, /init, init.lua, /nobugreport, /log, %USER_PROFILE%/.faforever/logs/, /replay, %USER_PROFILE%/.faforever/cache/temp.scfareplay, /replayid, 14167823] in directory: %USER_PROFILE%/.faforever/bin
2022-02-27 15:41:17.123  INFO 11776 --- [onPool-worker-3] com.faforever.client.game.GameService    : Forged Alliance terminated with exit code 2
/home/jim/.faforever/bin/ForgedAlliance.exe: 1: MZ����@0��: not found
/home/jim/.faforever/bin/ForgedAlliance.exe: 1: Syntax error: word unexpected (expecting ")")

image

jnorthrup:
addresses/fixes #2586 on my linux faf install.  faf has always been linked to Steam to my recollection, so no further indirection seems relevant for linux.  standard faf installation wiki(s), and FA guides all involve protontricks, again, no=op relating to this patch.
--

there is OS profile information in main with no obvious access other than the static methods written to run the bridj library calls.

instead of calling the bridj API I stuck to the PROFILE_STRING convention written.

I also noted there is no code setting the executableDecorator and it is unclear if there is DI that intellij is not hinting for me.  I took no liberies to bridge the unreachable PROFILE strings with the single-constant-instance of the decorator in the hopes that I can be shown any hierarchies I have missed with this patch.

### DIFF
--- a/src/main/java/com/faforever/client/FafClientApplication.java
+++ b/src/main/java/com/faforever/client/FafClientApplication.java
@@ -53,13 +53,14 @@ public class FafClientApplication extends Application {
   public static final String PROFILE_PROD = "prod";
   public static final String PROFILE_TEST = "test";
   /**
-   * Does always reload root tabs in the MainController. This is useful if you do hot swap and you want to see your
+   * Does always reload root tabs in the MainController. This is useful if you do hot swap, and you want to see your
    * changes.
    */
   public static final String PROFILE_RELOAD = "reload";
   public static final String PROFILE_LOCAL = "local";
   public static final String PROFILE_WINDOWS = "windows";
   public static final String PROFILE_LINUX = "linux";
+  public static final String PROFILE_PROTON = "proton";
   public static final String PROFILE_MAC = "mac";
   public static final int EXIT_STATUS_RAN_AS_ADMIN = 3;
 
@@ -70,13 +71,23 @@ public class FafClientApplication extends Application {
     launch(args);
   }
 
-    public static String[] getAdditionalProfiles() {
+  public static String[] getAdditionalProfiles() {
     List<String> additionalProfiles = new ArrayList<>();
 
     if (org.bridj.Platform.isWindows()) {
       additionalProfiles.add(PROFILE_WINDOWS);
     } else if (org.bridj.Platform.isLinux()) {
       additionalProfiles.add(PROFILE_LINUX);
+      try {
+        Process start = new ProcessBuilder("protontricks-launch", "-h").start();
+          start.getInputStream().readAllBytes();
+         int i= start  .waitFor();
+        if (0 == i) {
+          additionalProfiles.add(PROFILE_PROTON);
+        }
+      } catch (IOException | InterruptedException e) {
+        e.printStackTrace();
+      }
     } else if (org.bridj.Platform.isMacOSX()) {
       additionalProfiles.add(PROFILE_MAC);
     }
@@ -195,4 +206,5 @@ public class FafClientApplication extends Application {
     timeoutThread.setDaemon(true);
     timeoutThread.start();
   }
+
 }
--- a/src/main/java/com/faforever/client/FafClientApplication.java
+++ b/src/main/java/com/faforever/client/FafClientApplication.java
@@ -35,6 +35,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 
+import java.io.IOException;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
--- a/src/main/java/com/faforever/client/FafClientApplication.java
+++ b/src/main/java/com/faforever/client/FafClientApplication.java
@@ -70,7 +70,7 @@ public class FafClientApplication extends Application {
     launch(args);
   }
 
-  private static String[] getAdditionalProfiles() {
+    public static String[] getAdditionalProfiles() {
     List<String> additionalProfiles = new ArrayList<>();
 
     if (org.bridj.Platform.isWindows()) {
--- a/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
+++ b/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
@@ -16,12 +16,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Stream;
 
 import static com.faforever.client.util.Assert.checkNullIllegalState;
 
--- a/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
+++ b/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
@@ -1,5 +1,6 @@
 package com.faforever.client.fa;
 
+import com.faforever.client.FafClientApplication;
 import com.faforever.client.preferences.ForgedAlliancePrefs;
 import com.faforever.commons.lobby.Faction;
 import com.google.common.base.Strings;
@@ -15,8 +16,12 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import static com.faforever.client.util.Assert.checkNullIllegalState;
 
@@ -303,7 +308,9 @@ public class LaunchCommandBuilder {
   }
 
   public LaunchCommandBuilder executableDecorator(String executableDecorator) {
-    this.executableDecorator = Strings.isNullOrEmpty(executableDecorator) ? DEFAULT_EXECUTABLE_DECORATOR : executableDecorator;
+    boolean linuxProton = Arrays.asList(FafClientApplication.getAdditionalProfiles()).contains(FafClientApplication.PROFILE_LINUX);
+    String profileLauncher = (linuxProton ? "protontricks-launch --appid 9420 " : "") + DEFAULT_EXECUTABLE_DECORATOR;
+    this.executableDecorator = Strings.isNullOrEmpty(executableDecorator) ? profileLauncher : executableDecorator;
     return this;
   }
 }
--- a/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
+++ b/src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
@@ -304,7 +304,7 @@ public class LaunchCommandBuilder {
   }
 
   public LaunchCommandBuilder executableDecorator(String executableDecorator) {
-    boolean linuxProton = Arrays.asList(FafClientApplication.getAdditionalProfiles()).contains(FafClientApplication.PROFILE_LINUX);
+    boolean linuxProton = Arrays.asList(FafClientApplication.getAdditionalProfiles()).contains(FafClientApplication.PROFILE_PROTON);
     String profileLauncher = (linuxProton ? "protontricks-launch --appid 9420 " : "") + DEFAULT_EXECUTABLE_DECORATOR;
     this.executableDecorator = Strings.isNullOrEmpty(executableDecorator) ? profileLauncher : executableDecorator;
     return this;


Sheikah45:
This solution is very specific to a certain linux distribution and system set up and should not be hard coded into the client. When i talked about adding the command line decorator I mean in the client settings in the application itself not the code. The Forged Alliance forever settings allows you specify the decorator as seen here 

image

Brutus5000:
Not everybody plays through Steam on Linux! I'm not doing it neither does @p4block . (However I'm not using the installer thus I can't really help fixing that)

jnorthrup:
> Not everybody plays through Steam on Linux! I'm not doing it neither does @p4block . (However I'm not using the installer thus I can't really help fixing that)

I have been playing since the first beta, and owned the downloadable media, but i still purchased a steam key for a dollar many many years later.  

mind if i ask how you run it ?

jnorthrup:
> This solution is very specific to a certain linux distribution and system set up and should not be hard coded into the client. When i talked about adding the command line decorator I mean in the client settings in the application itself not the code. The Forged Alliance forever settings allows you specify the decorator as seen here 

as long as steam and protontricks are on the path this would apply to any steam or flatpack install. I'm aware that FAF ran a cracked patch update system, but whatever the case let's make a combobox for the nice users! if the complexity is limitted to a string, it won't be that hard to reuse the code written for os profiles rather than delegating a documentation-based and issue-tracker based scavenger hunt

I'm happy to support all of the above concerned users for the foreseeable future.

Sheikah45:
If you want to add options for linux users and fix the linux bugs that is fine but it should not be done by hard coding as this is not applicable for all users on linux and will cause errors in those systems

jnorthrup:
> If you want to add options for linux users and fix the linux bugs that is fine but it should not be done by hard coding as this is not applicable for all users on linux and will cause errors in those systems

I'm not in disagreement, but I am fixing the windows hard-coding I've been dealt by the defaults, and I'm lacking more info, and seeing a lack of architectural continuity in the areas.  Telling the users to go figure out the plumbing and type in a string is also a hostile attitude.

please enumerate all the scenarios where users are using faf on linux and not launching the steam app?  I'm happy to come up with a solution that is as much indirection as the job requires.  

the fact that vulkan is valve software and steam is driving proton which was the foundation of vulkan and none of the other game clearinghouses are doing anythihng with this title leads one to beleive that none of the constants I have gone with are outside of the majority of linux users and is potentially going to increase with the improved compatibility if the game sales hit an up-tick.

Sheikah45:
> I'm not in disagreement, but I am fixing the windows hard-coding I've been dealt by the defaults, and I'm lacking more info, and seeing a lack of architectural continuity in the areas. Telling the users to go figure out the plumbing and type in a string is also a hostile attitude.

I would not really call it hostile since there is no claim of official linux support from the client.

> please enumerate all the scenarios where users are using faf on linux and not launching the steam app? I'm happy to come up with a solution that is as much indirection as the job requires.

I personally do not know all the ways I do not work on linux systems so you would have to ask other linux users.

Sheikah45:
> the fact that vulkan is valve software and steam is driving proton which was the foundation of vulkan and none of the other game clearinghouses are doing anythihng with this title leads one to beleive that none of the constants I have gone with are outside of the majority of linux users and is potentially going to increase with the improved compatibility if the game sales hit an up-tick.

Also there is some work going on to support the GOG version of the game as well so it cannot be assumed that it will be a steam only environment going forward.

jnorthrup:
> > I'm not in disagreement, but I am fixing the windows hard-coding I've been dealt by the defaults, and I'm lacking more info, and seeing a lack of architectural continuity in the areas. Telling the users to go figure out the plumbing and type in a string is also a hostile attitude.
> 
> I would not really call it hostile since there is no claim of official linux support from the client.
> 
> > please enumerate all the scenarios where users are using faf on linux and not launching the steam app? I'm happy to come up with a solution that is as much indirection as the job requires.
> 
> I personally do not know all the ways I do not work on linux systems so you would have to ask other linux users.

right, I was making an educated guess that FAF has continued linking users to steam ID's, among those ID's is the game's 9420 set in stone and the number of casual linux desktop users vs. code literate ones may increase over time, and going on the main linux benchmark of _running steam_ as distribution viability and the main badge of entry into the pantheon of distributions.  we even see a non-glibc Void linux supporting steam.  

IIUC i have fixed a majority linux deficiency in the client and made every offer to continue doing so, but I'm not able to find any particular linux support requests that contradict anything I've contributed here.

in NOT choosing to use my local config files, I get the feeling no guidance on the worthiness or criterion for rejection is going to be made plain, but it doesn't look like you are going to work with me on this.

jnorthrup:
> 



> > the fact that vulkan is valve software and steam is driving proton which was the foundation of vulkan and none of the other game clearinghouses are doing anythihng with this title leads one to beleive that none of the constants I have gone with are outside of the majority of linux users and is potentially going to increase with the improved compatibility if the game sales hit an up-tick.
> 
> Also there is some work going on to support the GOG version of the game as well so it cannot be assumed that it will be a steam only environment going forward.

ok, well, i use steam, I've been playing since 2006 beta.  I'm happy to lend a hand wherever.  

the PR i have submitted fixes my problem and would enable same for steam users on linux, period.  should i patch the README and put my name as the linux steam FAF launch support guy so you stop proclaiming the lack of linux support from every mountain-top?

jnorthrup:
> > the fact that vulkan is valve software and steam is driving proton which was the foundation of vulkan and none of the other game clearinghouses are doing anythihng with this title leads one to beleive that none of the constants I have gone with are outside of the majority of linux users and is potentially going to increase with the improved compatibility if the game sales hit an up-tick.
> 
> Also there is some work going on to support the GOG version of the game as well so it cannot be assumed that it will be a steam only environment going forward.

https://www.gog.com/wishlist/games/supreme_commander_forged_alliance looks promising.  eyup, full rewrite coming.

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
```diff
@@ -303,7 +304,9 @@ public LaunchCommandBuilder gameOptions(Map<String, String> gameOptions) {
   }
 
   public LaunchCommandBuilder executableDecorator(String executableDecorator) {
-    this.executableDecorator = Strings.isNullOrEmpty(executableDecorator) ? DEFAULT_EXECUTABLE_DECORATOR : executableDecorator;
+    boolean linuxProton = Arrays.asList(FafClientApplication.getAdditionalProfiles()).contains(FafClientApplication.PROFILE_LINUX);
+    String profileLauncher = (linuxProton ? "protontricks-launch --appid 9420 " : "") + DEFAULT_EXECUTABLE_DECORATOR;
+    this.executableDecorator = Strings.isNullOrEmpty(executableDecorator) ? profileLauncher : executableDecorator;
     return this;
   }
```
Sheikah45: This should not really be in the launch command builder it should be moved up to where the builder is constructed.

src/main/java/com/faforever/client/FafClientApplication.java
```diff
@@ -70,7 +70,7 @@ public static void applicationMain(String[] args) {
     launch(args);
   }
 
-  private static String[] getAdditionalProfiles() {
+    public static String[] getAdditionalProfiles() {
```
Sheikah45: This should remain private. These profiles are just added to the Spring profiles. So any profile information should be gotten through spring and this should remain private.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/FafClientApplication.java
jnorthrup: where is that variable accessed from then?  there is no accessor

jnorthrup: you launch main, you launch the class in main, and you call this method and the results are not available to any other part of the code and not used.  this class has no instance other than a local in main.

jnorthrup: i welcome a rewrite to put the pieces in the right places, no ego here.  

the premise was: 
1) if there was a setter for the override, i didn't find it.
2) if the code was running on linux, and the value was blank, it's not going to (succeed in) launching a windows executable anyways, using defaults. that's hard-coded.
3) Instead of just checking the library that informs the PROFILE var i made every attempt to give a purpose to the profile var, even if, well, DRY is being violated.

Sheikah45:
Also just generally it would be better to modify the default executable decorator based on system that way a user could override it if necessary

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
jnorthrup: there are multiple entry points, 3 or 4 iiuc which arrive here ultimately.  I'm unclear on a single choke-point that does this cleaner.  I/Intellij is blind to the spring-boot framework injections if that's a thing here.

jnorthrup: the inward looking calls are boilerplate. I didn't break up the method chaining once i arrived at the constant in the code.  there is no setter in the code for this method, it is apparently happening, but by reflective or injected means.  

image

bukajsytlos:
sry but this simply does not work. I don't think client should be responsible for having correct command line arguments to run game which is very specific to given distribution and user preferences.

jnorthrup:
> sry but this simply does not work. I don't think client should be responsible for having correct command line arguments to run game which is very specific to given distribution and user preferences.

really? with a 15 year old windows binary and valve corp writing 100% of the wine, GPU drivers, and delivery, how many outliers are really going to push this rock up a hill and do something different? 

which distribution is protontricks (a python pip) and steam specific to on linux?  how many different behaviors does Steam support for compatibility mode?

when you find working faf installation instructions, the two things specific to getting FAF working are protontricks and PROTON vulkan  env variables in steam.  or WINED3D if that's your bag.

jnorthrup:
count the number of different distributions running identical recipes here:

https://www.protondb.com/app/9420

no two ina  row, and no variance between distros

jnorthrup:
>This should remain private. These profiles are just added to the Spring profiles. So any profile information should be gotten through spring and this should remain private.
>Also just generally it would be better to modify the default executable decorator based on system that way a user could override it if necessary
>This should not really be in the launch command builder it should be moved up to where the builder is constructed.

@Sheikah45 you are mentioning a boatload of spring conventions in a codebase which is not heavily annotated or providing many clues where spring adds value here.

this appears to be one of the first usecases where some kind of mono-culture default would be better servede by all of spring's convenient sledgehammer solutions, and none are currently in the code to draw from.

iiuc there would be a simple check for protontricks and steam on the path to add PROFILE_PROTON in addition to PROFILE_LINUX, somewhere.

the PR could be replaced by this profile and a check on the path. in order to provide a new default.


@Profile({FafClientApplication.PROFILE_~~LINUX~~PROTON,"protontricks-launch --appid 9420 %s"})

is appropriate, but these pojo's just aren't being used in the way you are suggesting yet.  since the diff is 4 lines, its hard to say it would be more concise, but maybe could get down to 2 or 3.

image

I haven't used spring in over 10 years, a lot has changed, hopefully for the better, I haven't continued with it for some really productive reasons.  some guidance here to replace this PR with however many linux profiles are necessary would be welcome.

p4block:
Hi,
[right now we use this badly written script as our game entrypoint](https://gist.github.com/p4block/8d947f85fa707b433af359a699b367fb).  It sets up a wine prefix if it's not there and launches the game. It's not fault tolerant at all and intentionally has to run an outdated proton binary distro, would need (probably a lot of) work to be integrated into the client in such a way that is reliable, but I think that's the endgame. 

We could also do a hackjob and just SHIP IT for sure and let the explosions begin. 

Just calling steam to do the deed is a quick way to make it work, I'm not completely against it esp. with the Deck coming in soon, but that's not ideal either. GoG support is coming up any day now. Plus, doing it ourselves saves up effort by not needing to touch steam at all (think: swapping proton versions or editing cli parameters) 

Additionally we need to be able to launch the game in any distro, can't rely on anything, especially not pip to be present and that's only half the story. We need a way to be able to flatpak install the client first, as the java requirement is a complete showstopper for many.

p4block:
The more I think of the Deck the more I am liking just "detecting" the game installed via Steam somehow and just telling Steam to launch it, but we still need to have the DIY wineprefix eventually for systems w/o Steam. 

I don't like the protontricks dependency though, we should do whatever it is doing instead.

jnorthrup:
> 



> The more I think of the Deck the more I am liking just "detecting" the game installed via Steam somehow and just telling Steam to launch it, but we still need to have the DIY wineprefix eventually for systems w/o Steam.
> 
> I don't like the protontricks dependency though, we should do whatever it is doing instead.

protontricks is a wrapper around winetricks.  protontricks delegates to one lazy-created ~/.steam/compat/ dir per appid, plus whatever other wineprefixes matter in your manual env life. 

the reason i went with that is that the launcher in steam has the blessed `PROTON_NOESYNC=1 %command%` and such-like which filters through the proton-installed dll's and then calls steam-launch on your new `%s`  `%command%`

but you have Void running side by side with opensuse, gentoo, fedora, and all the debian clones and whichever ones like flatpack, i had some things working on clear-linux last week.  

but right now, very definitely, running fa.exe does not parse correctly in bash for most users.

jnorthrup:
the latest commit detects protontricks on linux and creates a new PROFILE var for that override of the default only.

this could be done without  the current hack job if the launchercommands or some other example in the source was a coupled `@Configuration/@Component/@Profile/@Bean` annotations but right now this is not structured with profiles in the command logic and it's a non-trivial amount of changes to make these config objects using the annotations.

given the number of steam/proton tricks FA installs, this is 80/20 on the linux support with the least amount of surgery.  this detection code and override are like ten lines total so if or when there is work done on the design this is the least of the technical hurdles to overcome.

src/main/java/com/faforever/client/FafClientApplication.java
```diff
@@ -70,13 +72,23 @@ public static void applicationMain(String[] args) {
     launch(args);
   }
 
-  private static String[] getAdditionalProfiles() {
```
jnorthrup, line 73: @Sheikah45 it would be nice to make this `public` go away but the amount of annotation catchup-work looks enormous to achieve this.

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
jnorthrup: I can learn from example, but I cannot accomplish this upstream without making a bigger mess without knowing more.

jnorthrup (commented):
if there are xml or propteries files or things that are invisible to intellij, I can be of little immediate assistance, I also don't understand what "finish your review" means to the patch submitter, but i have 4 such pending requests.

bukajsytlos:
sry I wanted to discard my comment and instead closed this pr.
My question is, will this somehow reflect config params like PROTON_NO_ESYNC=1 PROTON_NO_FSYNC=1, d3d, and not having to run steam to run game?

jnorthrup:
> sry I wanted to discard my comment and instead closed this pr. My question is, will this somehow reflect config params like PROTON_NO_ESYNC=1 PROTON_NO_FSYNC=1, d3d, and not having to run steam to run game?

protontricks-launch will need to log steam in to get the steam parameters for %command% and find which version of proton fa is running.  

FAF is running something it calls "%s" which is used as %command% by protontricks-launch to stuff the env variables, regardless of which exe is being executed.

FAF doesn't use steam as %s but protontricks-launch will use steam.  this PR only overrides "%s" so if you change it yourself it will do whatever you tell it

Sheikah45:
> this could be done without the current hack job if the launchercommands or some other example in the source was a coupled `@Configuration/@Component/@Profile/@Bean` annotations but right now this is not structured with profiles in the command logic and it's a non-trivial amount of changes to make these config objects using the annotations.

This is exactly what should be done. Ideally the LaunchCommandBuilder would be annotated as a component with a prototype scope. then the default executor would be turned into an instance variable that is chosen depending on which profile is active

src/main/java/com/faforever/client/FafClientApplication.java
```diff
@@ -70,13 +72,23 @@ public static void applicationMain(String[] args) {
...
+      try {
+        Process start = new ProcessBuilder("protontricks-launch", "-h").start();
+          start.getInputStream().readAllBytes();
+         int i= start  .waitFor();
+        if (0 == i) {
+          additionalProfiles.add(PROFILE_PROTON);
+        }
+      } catch (IOException | InterruptedException e) {
+        e.printStackTrace();
+      }
```
Sheikah45, line 91: Just check the path variable rather than trying to run the application

Sheikah45 (changes requested):

jnorthrup:
> > this could be done without the current hack job if the launchercommands or some other example in the source was a coupled `@Configuration/@Component/@Profile/@Bean` annotations but right now this is not structured with profiles in the command logic and it's a non-trivial amount of changes to make these config objects using the annotations.
> 
> This is exactly what should be done. Ideally the LaunchCommandBuilder would be annotated as a component with a prototype scope. then the default executor would be turned into an instance variable that is chosen depending on which profile is active

im of the opinion that the running of FA.exe on linux would be the priority for linux steam, if the gui authors needed spring in the configs spring would already be there.  this is a different outcome than presently

Sheikah45:
> > > this could be done without the current hack job if the launchercommands or some other example in the source was a coupled `@Configuration/@Component/@Profile/@Bean` annotations but right now this is not structured with profiles in the command logic and it's a non-trivial amount of changes to make these config objects using the annotations.
> > 
> > 
> > This is exactly what should be done. Ideally the LaunchCommandBuilder would be annotated as a component with a prototype scope. then the default executor would be turned into an instance variable that is chosen depending on which profile is active
> 
> im of the opinion that the running of FA.exe on linux would be the priority for linux steam, if the gui authors needed spring in the configs spring would already be there. this is a different outcome than presently

I am not sure what you mean. What does the gui have to do with configuring the launch command as a component?

jnorthrup:
> > im of the opinion that the running of FA.exe on linux would be the priority for linux steam, if the gui authors needed spring in the configs spring would already be there. this is a different outcome than presently
> 
> I am not sure what you mean. What does the gui have to do with configuring the launch command as a component?

pragmatically speaking, it wasn't important to make a windows java gui, it shouldn't increase in priority now to the degree of becoming a blocker for enabling steam/linux by default either

Sheikah45:
The gui has nothing to do with what I am saying about how to access the profile and change the default command based on the spring profile. This is all backend

jnorthrup:
> The gui has nothing to do with what I am saying about how to access the profile and change the default command based on the spring profile. This is all backend

this is all a backend that needs to shed a LOT of wight to be more maintainable and straightforward.  adding new spring features to code that has dead boilerplate spring ballast lying around is again relatively hostile, or wishful thinking, or team politics, or any other euphamism one might come up with for being territorial.

this backend is some ini files for a windows binary.

jnorthrup:
> pragmatically speaking, it wasn't important to make a windows java gui, it shouldn't increase in priority now to the degree of becoming a blocker for enabling steam/linux by default either

i don't know, nor wish to invest time in learning all the spring component tricks that nobody ever bothered with before i added a 4 line fix, and increased it to a ten-line fix. 

what i see when i look for guidance here is:
image

jnorthrup:
what part of the backend deals with clicking a gui button and launching a
shell exec? is this a network protocol? is this embedded firmware? is this
authentication?

NO! it's a fantastically bloated button action handler which requires a
collection of 100-300 line classes to justify a framework.

bukajsytlos:
if I launch `protontricks-launch --appid 9420 ~/.faforever/bin/ForgedAlliance.exe` there is no steam log in involved, but all I get is a black screen, which I assume is due to missing PROTON_NO_ESYNC=1 PROTON_NO_FSYNC=1. So what am I doing wrong?

jnorthrup:
~/.faforever/bin/ForgedAlliance.exe is not Steram's blessed binaries/install.

this is faf's hacked patched network platform moho engine.   it probably has no script.

jnorthrup:
for whatever reason, my test of your commandline there launches my FA with the tropical island and gpg logo opening.  

im not sure what constructs this directory.  but if i run the steam app without proton env settings, i see black/0.0000001 FPS 

find ~/.faforever/
bin/           cache/         gamedata/      logs/          map_generator/ replays/       themes/        user/

jnorthrup:
PROTON_NO_ESYNC=1 PROTON_NO_FSYNC=1 PROTON_USE_WINED3D=0 PROTON_ENABLE_NVAPI=1   PROTON_USE_D9VK=2 %command%

my settings here... i had a wobbly start with nvidia-510 drivers and liquorix kernel, leading to the experiments you see above. 

image

BlackYps:
I don't understand what is happening here. It seems like you initially had problems because the launch options were not set up correctly. Now you are trying to hardcode them into the client so it works out of the box for your setup?
What is wrong with the previous approach of just putting in the launch options in the settings? This way everyone can easily modify them to suit their distribution or steam/gog install method.

jnorthrup:
>It seems like you initially had problems because the launch options were not set up correctly.

no, i had defaults, executing windows exe files on linux.

> Now you are trying to hardcode them into the client so it works out of the box for your setup?

no, again oblivious.   i am overriding the defaults when protontricks is on the path.  if the user isn't running defaults and isnt both running linux abnd protontricks, there is no change to functionality

> I don't understand what is happening here.

agreed


> What is wrong with the previous approach of just putting in the launch options in the settings? This way everyone can easily modify them to suit their distribution or steam/gog install method.

no change to that.  also no documentation, and a glib, enthusiastic declarations  of NO SUPPORT from @Sheikah45 

the patch works as mentioned above, and does not invent a spring configuration framework implementation for the gui which isn't there already, so the best guess i have is that maintainer #1 is camping/cherry picking issues he can get paid for, there has been resistance to working logical code from the minute it was submitted.

jnorthrup:
suggestion: put $20 on the issue

bukajsytlos:
I still don't understand how should this work or what am I doing wrong. As far as I understand what you are trying to accomplish here is that `protontricks-launch --appid 9420 "%s"` will be "default" if there are protontricks available. So I put it in client settings. I have those proton env flags in steam, but still it won't launch game via steam and all I have is FA window with black screen.

p4block:
FA window with black screen is caused by having something "wrong" but that something can be mostly anything. 
Testing should be done in a clean environment (ie. fresh ubuntu install, or at least a new user) because it's easy to get to a state where "it works" then ie. add more modifications that break a fresh prefix.  

I believe that all options here are very flawed, however...

In an ideal world, steam would have some way of running an appid w/ arguments from the cli. Maybe it _has_ and we just don't know. That would solve the Deck issue and a big chunk of the potential linux playerbase quickly. 

As for non-steam installs I think I should get the aforementioned script cleaned up. That would require, at minimum, this work:

- Shipping it alongside the client
- Making the default client launch command on linux to run it, at least as long as we don't find out how to make steam properly cooperate. 
- Pass the game directory to it somehow, as it is hardcoded now

@Brutus5000 also suggested to me that we could also port the wine binaries download & prefix prepping to java for, uh, _proper_ integration. 

Some random info now that I am writing a lot:
The script uses the last proton version that doesn't require steam libraries to run, and installs dx9 and xact audio, then dxvk on the prefix. That's enough to make the game work. 

Could test with normal vanilla wine, afaik should work no problem. We also have the lutris/bottles builds of "proton" that don't need steam. Swapping those is no problem, but we need to get the ball rolling somehow.

bukajsytlos:
I have dumped steam script which I have set up instead of what is proposed here and that works and is using "9420" prefix. So for this solution to be plausible I would at least expect that it will run with same config as is in steam game preferences. This at least what I understood should be happening, but in my case it does not seem so.

jnorthrup:
> FA window with black screen is caused by having something "wrong" but that something can be mostly anything. Testing should be done in a clean environment (ie. fresh ubuntu install, or at least a new user) because it's easy to get to a state where "it works" then ie. add more modifications that break a fresh prefix.
> 
> I believe that all options here are very flawed, however...
> 
> In an ideal world, steam would have some way of running an appid w/ arguments from the cli. Maybe it _has_ and we just don't know. That would solve the Deck issue and a big chunk of the potential linux playerbase quickly.
> 
> As for non-steam installs I think I should get the aforementioned script cleaned up. That would require, at minimum, this work:
> 
> * Shipping it alongside the client
> * Making the default client launch command on linux to run it, at least as long as we don't find out how to make steam properly cooperate.
> * Pass the game directory to it somehow, as it is hardcoded now
> 
> @Brutus5000 also suggested to me that we could also port the wine binaries download & prefix prepping to java for, uh, _proper_ integration.
> 
> Some random info now that I am writing a lot: The script uses the last proton version that doesn't require steam libraries to run, and installs dx9 and xact audio, then dxvk on the prefix. That's enough to make the game work.
> 
> Could test with normal vanilla wine, afaik should work no problem. We also have the lutris/bottles builds of "proton" that don't need steam. Swapping those is no problem, but we need to get the ball rolling somehow.

it looks like the blocker for this PR is not that we can't begin to iterate on configurations and launch techniques that work and can be detected and automated, but the PR review specifies that this requires a full spring rewrite of the missing parts of the spring profile system that was never implemented around the os detection code, now that we have something that works on as you say, a clean ubuntu system (mine to date).  

i also meant to add that without protontricks steam has a service client pass-thru URL that is being used on launch as well.  i didnt research this at all.

image

Sheikah45:
Just as a note we do already add the detected OS to the spring profiles. It is denoted as having the active profile windows, or linux. Since this is already done it is better to use that rather than redo it.

jnorthrup:
> Just as a note we do already add the detected OS to the spring profiles. It is denoted as having the active profile windows, or linux. Since this is already done it is better to use that rather than redo it.

those variables are detected but not used in the code.  gradle does a seperate unrelated report at build time about the OS. 

I am the first consumer of that dead detection code in the java codebase.  if there are dependency injection features in use, there is no comments in the code that handles any of this stuff giving a single pointer.  it's java all the way, with dead code in the spring init as a java coder would see it. 

as far as i can tell these are the issues raised with my PR to date:
1) im using specific distributions.  valve software programs. on ubuntu no less.  the supported platform by valve.
2) im not rewriting the affected class into spring multiple components to intercept the default value when proton is detected.
3) I didnt write the fix in the 4 calling sites of that method and i should move the fix upstream before the event.  there is no access to the program state needed in those places and no java code is touching those variables.
4) I should "check the path" for protontricks instead of run it and check the error code. afaict this is the best way to check the path for a working protontricks given all the potential installer variations and the potential for permissions violations.
5) linux is not supported.
6) i should undo the change to the private dead detection code where i make the function public and use the results.  or i can call sun.misc.unsafe and bypass private method access instead of marking the dead code public.  seems like a bad idea.

i think we just need to pony up some bounty reward money for linux support or ask different reviewers to take responsibility for the patch quality

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
bukajsytlos: isn't defaultLaunchCommand that choke point?

bukajsytlos:
you can add `private final Environment environment;` to ForgedAllianceService and access profiles like `final String[] activeProfiles = environment.getActiveProfiles();`

jnorthrup:
i took your advice at face value and attempted to relocate the constant and the decorator to the next caller up.

i started an hour ago (not without other distractions, but the fam is waiting to get some breakfast) and the number of methods i need to write and properties i need to add is going up rather than down to counter act the "private" choice of encapsulation in all the places.  I have to time box this endeavor as out of my league.

I have previously closed this issue and can't offer what i consider a simpler solution using the Environment in ForgedAllianceService;   maybe in the LaunchBuilder, but I leave that to the more skilled hands.  the important features of the proton detection and fixing the broken default appear to be harmless if  `!(linux&&proton)`

I'm hoping the profound discovery of `protontricks-launch` is a starting point for linux support on the valve support steam linux platform in faf.

I'm going to stick with my local copy of the codebase while it's convenient and let the seton's delegation or whatever came after it take the reins from here, or not.

BlackYps:
I can't really comment on the details of this specific pull request I just want to explain why linux is not officially supported:
It's not like we don't want to support linux it's just that we lack the manpower. We don't have anyone sufficiently familiar with the client that also runs linux to be able to fix the issues that pop up when using the client under linux. Instead of providing a product with known issues with no perspective of fixing it, we instead decided to declare linux not officially supported, so people know that there will be issues that we can't really help them with.

p4block:
The client runs fine(tm) both Brutus and I (and more people) happen to run it without much fuss. I would personally have no issue spending some time to get us closer to something that we could actively advertise, mostly because I don't think it's that much effort. 

In order to get FaF on Linux on the realm of "supportable", not saying supported just _working_ without extensive work on the user side we need to solve two roadblockers:

- Package the client in flatpak
- Get a working wine environment with the dependencies this game needs

The first one doesn't really need much input from any maintainer, just a matter of having the correct java version. The second one can be done quick and dirty by [shipping my script](https://gist.github.com/p4block/8d947f85fa707b433af359a699b367fb). I just place it alongside .faforever files and boom it works, the only gripe is how to get the game folder setting item passed to the script, which should be about half a line of java code.

If no one else is up to the task I'll just _do it myself_ hopefully this month. Get some bash linter to tidy up my crappy code and make a pr. With flatpak I tried to deal before and achieved not much, open to anyone with experience on that.

p4block:
_Disclaimer: the script's wine needs a lot of host libraries (esp 32 bit ones) that are not present on a freshly installed system. A production approach would be to DL ie. lutris' libraries and a lutris proton build that links against those libraries._

jnorthrup:
here's my pragmatism, opinions or otherwise:

tmfm - spring, install4j, large gradle constructs, +=lutris,+=flatpack{ += cgroups isolation},  .. growing this list instead of trimming it is a red flag in my book.  there's now golden handcuffs on jdk15 of all things because of the "test first?" policy.  and a silly API client with an exe launcer can't use the jdk graalvm executable as a direction to explore in all of this ambitious orchestration.

i have a steam asset supported by steam and suggested a fix here which traces back to Steam's magic, and they've got a pretty big and active dev team keeping billions of $REVENUE afloat, and in fact monopolizing it.  

it has been even more convenient to repurchase the game through steam than to go download my original licensed copy from a 404-site.

despite the FM, the steam fix for a windows native app was one line prepended, give or take the supported package manager on ubuntu as an enabler.

bukajsytlos:
just note: we need to run .faforever's exe instead of steam one so there are no desyncs

jnorthrup:
> just note: we need to run .faforever's exe instead of steam one so there are no desyncs

faf copies the steam dir including some dll's and other things and protontricks-launch is launching from that dir.

i did an awful lot of digging with dgvoodoo and came upon this through strace.  

fwiw, dgvoodoo 2.79 works when unzip into the faf bin directory with the executable but on my system the 32-bit dll breaks the lobby.  

the 32-bit dll shows the watermark, the 64 bit dll does not, but works fine in faf exe dir

something about coop replays was broken in latest update

jnorthrup:
> The client runs fine(tm) both Brutus and I (and more people) happen to run it without much fuss. I would personally have no issue spending some time to get us closer to something that we could actively advertise, mostly because I don't think it's that much effort.
> 
> In order to get FaF on Linux on the realm of "supportable", not saying supported just _working_ without extensive work on the user side we need to solve two roadblockers:
> 
> * Package the client in flatpak
> * Get a working wine environment with the dependencies this game needs
> 
> The first one doesn't really need much input from any maintainer, just a matter of having the correct java version. The second one can be done quick and dirty by [shipping my script](https://gist.github.com/p4block/8d947f85fa707b433af359a699b367fb). I just place it alongside .faforever files and boom it works, the only gripe is how to get the game folder setting item passed to the script, which should be about half a line of java code.
> 
> If no one else is up to the task I'll just _do it myself_ hopefully this month. Get some bash linter to tidy up my crappy code and make a pr. With flatpak I tried to deal before and achieved not much, open to anyone with experience on that.

@p4block on the topic of "my script", the protontricks is now part of the defacto how-to when the wiki is not pointing to dead links, and apparently the recipe du jour is using the proton_dump variable to create a /tmp/run script which is then copied into the recipe as part of the official advice (on discord at least)

as it turns out protontricks-launch is doing an identical amount of things to the degree that dgVoodoo dll's for nvidia shaders to replace wine d3d dll's;  

now apparently GOG is using proton also which means more potential viability for going with protontricks-launch 

the point i would like to make is that protontricks-launch seems like a good practice for faf or for troubleshooting and working through linux challenges on nearly any windows executable to reduce the amount of user steps relating to maintained wine.  this also works for Ada disassembler win32 as well, as another instance.

What we're seeing on discord is a willingness by those who are not really on top of script hacking to be official supporters, whatever that means.  it seems to cause anxiety among at least a few in casual conversation when you start to show your work in the form of unix console commands  while working through problems.  

So my advice for avoiding a lemming rush off of a cliff is again to minimize the surface area of wine scripts.   

cheers

PR #2591, opened 2022-03-01
Allow setting Player Notes and more interactive player objects

### LINKED ISSUE #2539: Add the ability to record (edit) a note for each player
The notes are stored in a local file. If there is a note for the player, display it as an icon next to the player's name. When you hover over the icon, the text is displayed

### LINKED ISSUE #2592: Show the user that it is possible to interact with the player card

Marc-Spector:
Closes #2539 
Closes #2592

This is still a draft version. I will add UI items in the future Tips and suggestions are welcome.

<details>
<summary>Demo</summary>

demo
</details>

<details>
<summary> More UI interaction with the player card</summary>

demo1
</details>


src/main/java/com/faforever/client/player/PlayerNoteService.java
```diff
@@ -0,0 +1,99 @@
...
+  private final ObjectMapper objectMapper;
+  private final EventBus eventBus;
+
+  private Path notesPath;
+  private Map<Integer, String> notes;
+
+  @Override
+  public void afterPropertiesSet() throws Exception {
+    eventBus.register(this);
+    notesPath = preferencesService.getPreferencesDirectory().resolve(PLAYER_NOTES_FILE_NAME);
```
Sheikah45: I think it might make more sense to just store the notes in the preferences object itself as an observable map, then you can also add a listener onto the map to add the note to the player. I don't see any advantages to having it in a separate file.

Marc-Spector:
It is also necessary to set a limit on the number of notes. I think 1500 is reasonable.

src/main/java/com/faforever/client/player/PlayerNoteService.java
Marc-Spector: Ok, I will try.

Sheikah45:
Why do we need a limit on the number of notes? Since it is all stored in local memory the only danger is that people fill up all their ram with notes but given it is all text that would be pretty hard to do.

Marc-Spector:
> Why do we need a limit on the number of notes? Since it is all stored in local memory the only danger is that people fill up all their ram with notes but given it is all text that would be pretty hard to do.

Okay. But still, do not underestimate people. 😄

Sheikah45:
> > Why do we need a limit on the number of notes? Since it is all stored in local memory the only danger is that people fill up all their ram with notes but given it is all text that would be pretty hard to do.
> 
> Okay. But still, do not underestimate people. 😄 

I am not ha, but at the end of the day it can have no affect on other users or the server so see little reason to limit it.

src/main/java/com/faforever/client/fx/contextmenu/AddPlayerNoteMenuItem.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Override
+  protected boolean isItemVisible() {
+    return object != null && !playerService.containsNote(object);
+  }
+
+  @Override
+  protected String getItemText() {
+    return i18n.get("chat.userContext.addNote");
```
Sheikah45: Why are this and EditPlayerNote two separate objects? Why not just have a single class that gives different text depending on whether there is a player note that already exists for this player, as otherwise they are identical.

src/main/java/com/faforever/client/fx/contextmenu/EditPlayerNoteMenuItem.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Override
+  protected boolean isItemVisible() {
+    return object != null && playerService.containsNote(object);
+  }
+
+  @Override
+  protected String getItemText() {
+    return i18n.get("chat.userContext.editNote");
```
Sheikah45: See comment on addPlayerNote

src/main/java/com/faforever/client/player/PlayerService.java
```diff
@@ -194,6 +210,26 @@ void createOrUpdatePlayerForPlayerInfo(@NotNull com.faforever.commons.lobby.Play
     resetIdleTime(player);
   }
 
+  public boolean containsNote(PlayerBean player) {
+    return playerNotes.containsKey(player.getId());
+  }
```
Sheikah45: Couldn't we just check if player.getNote() is blank since the player note should be set on all playerBean initialization?

src/main/java/com/faforever/client/player/PlayerService.java
```diff
@@ -194,6 +210,26 @@ void createOrUpdatePlayerForPlayerInfo(@NotNull com.faforever.commons.lobby.Play
...
+  public void updateNote(PlayerBean player, String text) {
+    int id = player.getId();
+
+    if (StringUtils.isBlank(text)) {
+      removeNote(player);
+    } else if (playerNotes.containsKey(id)) {
+      playerNotes.replace(id, text);
+    } else {
+      playerNotes.put(id, text);
+    }
```
Sheikah45: Why is this necessary seems like we don't need the else if

src/main/java/com/faforever/client/preferences/Preferences.java
```diff
@@ -89,6 +89,7 @@
   IntegerProperty cacheLifeTimeInDays = new SimpleIntegerProperty(30);
   BooleanProperty gameDataCacheActivated = new SimpleBooleanProperty(false);
   BooleanProperty debugLogEnabled = new SimpleBooleanProperty(false);
+  MapProperty<Integer, String> playerNotes = new SimpleMapProperty<>(observableHashMap()); // key - player ID
```
Sheikah45: Rather than the comment just name the variable `notesByPlayerId` like the other maps are named.

Also this should go into one of the sub preference objects or a new one. I am trying to move away from having things in the top level object

Sheikah45 (changes requested):

src/main/java/com/faforever/client/fx/contextmenu/DoPlayerNoteMenuItem.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.config.ConfigurableBeanFactory;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Component;
+
+@Component
+@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+@RequiredArgsConstructor
+public class DoPlayerNoteMenuItem extends AbstractMenuItem<PlayerBean> {
```
Sheikah45: Can just stay as EditPlayerNoteMenuItem

src/main/java/com/faforever/client/player/PlayerNoteController.java
```diff
@@ -0,0 +1,79 @@
...
+    textArea.setTextFormatter(formatter);
+    JavaFxUtil.addListener(textArea.lengthProperty(), observable ->
+        charactersCountLabel.setText(String.format("%d / %d", textArea.getLength(), maxSymbolLimit)));
+  }
+
+  private boolean isSymbolLimitReached(Change change) {
+    return change.getControlNewText().length() > maxSymbolLimit;
+  }
+
+  private boolean isMultipleSeparators(Change change) {
```
Sheikah45:
```suggestion
  private boolean hasMultipleSeparators(Change change) {
```

src/main/java/com/faforever/client/player/PlayerNoteController.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  public void okButtonClicked() {
+    playerService.updateNote(player, textArea.getText());
+    cancelButton.fire();
+  }
+
+  public void setCloseButtonAction(EventHandler<ActionEvent> handler) {
+    cancelButton.setOnAction(handler);
+  }
```
Sheikah45: When would this note be dialog.close()?

Sheikah45 (changes requested):
Almost there although tests are failing

src/main/java/com/faforever/client/player/PlayerNoteController.java
Marc-Spector: Sorry, I don't understand what you want.

Sheikah45: Sorry there was a typo. Although the main question is why does this function exist. The onClose action is always the same and I am not sure of a situation where it would be different.

Marc-Spector: Fixed.

Sheikah45 (approved):

PR #2598, opened 2022-03-02
Feature/#2597 rehost from replay

mrchris2000:
Enables a player to quickly host a game with the same mods and map as the replay they are looking at.


mrchris2000:
Shouldn't have re-used the commit message for the second change; wasn't paying attention. It was just to tidy up the warnings IntelliJ identified.

mrchris2000:
image
It could have a better icon - I tried the refresh-icon style but was concerned it could be confusing. I added tooltips to both icons to aid with differentiation as a potential resolution.

Sheikah45 (commented):
Will give a more detailed review later, but for now I just wanted to note that we shouldn't use System.out.println for logging. Each class has a dedicated logger which should be used. You can find examples in any class almost.

Also part of me thinks that the creation of the newGameInfo should be done in the game service rather than the replay service.

You can also see in the game service the better use of the completableFutures so that it is non blocking.

Sheikah45:
> image
> It could have a better icon - I tried the refresh-icon style but was concerned it could be confusing. I added tooltips to both icons to aid with differentiation as a potential resolution.

I would not change the behavior of the play button. If anything I am not sure if this button should go onto the card. I would just make it a button in the detailed view.

Sheikah45:
Also as a note we generally rebase prs rather than having a merge commit just to have a cleaner history and easier merge at the end.

mrchris2000:
Sorry, first time through this process. I'll try to get it right in future.

Sheikah45:
No worries, everyone starts somewhere that is why I mentioned it.

mrchris2000:
> Also part of me thinks that the creation of the newGameInfo should be done in the game service rather than the replay service.

It would be fairly trivial to add a method signature such the below to the game service if you think thats for the best:
 `rehostFromReplay(String title, FeaturedModBean featuredMod, String mapName, Set<String> simMods)`
Let me know if you want me to move that part of the implementation.

Sheikah45:
The retrieval of the featuredMod should happen inside of the game service, that way you could also perform the updates at a similar step

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -66,6 +66,7 @@
   public Label numberOfReviewsLabel;
   public HBox teamsContainer;
   public Label onMapLabel;
+  public Button rehostButton;
```
Sheikah45: I would not call it rehost but rather host

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -261,6 +265,19 @@ public void runReplay(ReplayBean item) {
     }
   }
 
+  public void runRehost(ReplayBean item) {
```
Sheikah45: runRehost doesn't make sense in this context as it is not running a file. I would do more hostFromReplay

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -261,6 +265,19 @@ public void runReplay(ReplayBean item) {
...
+    if (item.getReplayFile() != null) {
+      try {
+        runRehostFile(item.getReplayFile());
+      } catch (Exception e) {
```
Sheikah45: I would be more specific with the caught errors

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -330,6 +347,17 @@ public void runReplayFile(Path path) throws IOException, CompressorException {
     }
   }
 
+  public void runRehostFile(Path path) throws IOException, CompressorException {
+    log.info("Starting replay file: `{}`", path.toAbsolutePath());
```
Sheikah45: This is not running the replay file so the log statement is misleading

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +401,94 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
     gameService.runWithReplay(tempSupComReplayFile, replayId, gameType, version, modVersions, simMods, mapName);
   }
 
+  private void runRehostOnlineReplay(int replayId) {
+    downloadReplay(replayId)
+        .thenAccept((path) -> {
+          try {
+            runRehostFile(path);
+          } catch (IOException | CompressorException e) {
+            throw new RuntimeException(e);
```
Sheikah45: wrap in CompletionException not RuntimeException

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +401,94 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
...
+          }
+        })
+        .exceptionally(throwable -> {
+          if (throwable.getCause() instanceof FileNotFoundException) {
```
Sheikah45, line 412: Make sure to unwrap the exception using something like ConcurrentUtil.unWrapException

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +401,94 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
...
+  }
+
+  private void runRehostFafFile(Path path) throws IOException, CompressorException {
+    ReplayDataParser replayData = replayFileReader.parseReplay(path);
+    byte[] rawReplayBytes = replayData.getData();
+
+    Path tempSupComReplayFile = preferencesService.getPreferences().getData().getCacheDirectory().resolve(TEMP_SCFA_REPLAY_FILE_NAME);
+
+    Files.createDirectories(tempSupComReplayFile.getParent());
+    Files.copy(new ByteArrayInputStream(rawReplayBytes), tempSupComReplayFile, StandardCopyOption.REPLACE_EXISTING);
```
Sheikah45: Why do we need to copy the file to disk if we are just extracting info from it?

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +401,94 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
...
+          Strings.emptyToNull(""),
+          featuredMod,
+          mapName,
+          simMods,
+          GameVisibility.PUBLIC,
+          null,
+          null,
+          false);
+
+      gameService.hostGame(newGameInfo);
```
Sheikah45: This all should be moved to the game service. Additionally don't make it blocking with the get function rather it should return the chained completable future.

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +401,94 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
...
+    }catch (ExecutionException | InterruptedException e) {
```
Sheikah45: space before catch

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +401,94 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
...
+      log.warn("Failed to launch rehost from replay", e);
+    }
```
Sheikah45: same as above

src/main/resources/theme/vault/replay/replay_card.fxml
```diff
@@ -97,8 +98,16 @@
...
+                onAction="#onRehostButtonClicked" >
+            <tooltip><Tooltip text="Rehost"/></tooltip>
+            <graphic>
+                <Region styleClass="icon,host-icon"/>
+            </graphic>
+        </Button>
         <Button fx:id="watchButton" mnemonicParsing="false" styleClass="primary-button"
-                onAction="#onWatchButtonClicked" >
+                onAction="#onWatchButtonClicked">
+            <tooltip><Tooltip text="Watch"/></tooltip>
```
Sheikah45: Move this to the detailController not the card controller

Sheikah45 (changes requested):

src/main/java/com/faforever/client/replay/ReplayService.java
Sheikah45: Actually scratch this. Looking at the current pattern we have it is fine to have this in replayService. However it should be made non blocking, by chaining the futures. Also it would be nice to inherit the things last rating limits and password from the last hosted game

mrchris2000: I specifically didn't bring over rating limits etc because if I'm hosting from someone else's replay I probably don't have the same requirements - as it happens I don't see the password or rating limits in the replay data, so it looks like I couldn't get that anyway.

src/main/resources/theme/vault/replay/replay_card.fxml
mrchris2000: I don't agree with this; it moves the functionality off the front page where its quick to access. Unlike requesting extra detail in the detailController such as chat messages, this is intended to be a 'quick hit' rehost option - equivalent to the delete option on the local replays view.

image
&
image

Sheikah45: The main concern is just clutter and clarity. I think if we can get a better icon that mean host then it would be okay. Although I am not sure what that icon is at the moment.

src/main/java/com/faforever/client/replay/ReplayService.java
Sheikah45: That is why I said not from the replay data but from the last game that the player hosted. In the game service there is a function to enhance with the last game password that could be extended

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -330,6 +346,17 @@ public void runReplayFile(Path path) throws IOException, CompressorException {
     }
   }
 
+  public void runHostFromReplayFile(Path path) throws IOException, CompressorException {
```
Sheikah45:
```suggestion
  public void hostFromReplayFile(Path path) throws IOException, CompressorException {
```

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +400,57 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
     gameService.runWithReplay(tempSupComReplayFile, replayId, gameType, version, modVersions, simMods, mapName);
   }
 
+  private void runHostFromOnlineReplay(int replayId) {
```
Sheikah45:
```suggestion
  private void hostFromOnlineReplay(int replayId) {
```

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +400,57 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
...
+    String gameType = replayMetadata.getFeaturedMod();
+    String mapName = parseMapFolderName(replayData);
+    Set<String> simMods = parseModUIDs(replayData);
+
+    modService.enableSimMods(simMods);
+    modService.getFeaturedMod(gameType)
+        .thenAccept(featuredModBean -> gameService.hostGame(new NewGameInfo(replayMetadata.getTitle(),
+            null, featuredModBean, mapName, simMods, GameVisibility.PUBLIC,null,
+            null,false)));
+  }
```
Sheikah45: This would be better done to return a completable future and chain the gameService.hostGame together so that it can properly wait until the game is hosted if necessary.
```suggestion
  private CompletableFuture<Void> runHostFromFafReplayFile(Path path) throws IOException, CompressorException {
    ReplayDataParser replayData = replayFileReader.parseReplay(path);

    ReplayMetadata replayMetadata = replayData.getMetadata();
    String gameType = replayMetadata.getFeaturedMod();
    String mapName = parseMapFolderName(replayData);
    Set<String> simMods = parseModUIDs(replayData);

    modService.enableSimMods(simMods);
    return modService.getFeaturedMod(gameType)
        .thenCompose(featuredModBean -> {
        NewGameInfo newGameInfo = new NewGameInfo(
            replayMetadata.getTitle(),
            null, featuredModBean, 
            mapName, 
            simMods, 
            GameVisibility.PUBLIC,
            null,
            null,
            false
        );
        return gameService.hostGame(newGameInfo);
    });
  }
```

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +400,57 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
...
+    String fileName = path.getFileName().toString();
+    String gameType = guessModByFileName(fileName);
+    Set<String> simMods = parseModUIDs(replayData);
+
+    modService.enableSimMods(simMods);
+    modService.getFeaturedMod(gameType)
+        .thenAccept(featuredModBean -> gameService.hostGame(new NewGameInfo(replayData.getMetadata().getTitle(),
+            null, featuredModBean, mapName, simMods, GameVisibility.PUBLIC,null,
+            null,false)));
+  }
```
Sheikah45: Same chaining as above

Sheikah45 (changes requested):
An action for this should also be added to the detail controller regardless.

Also I ran the checks so you can see which tests are failing now

src/main/java/com/faforever/client/replay/ReplayService.java
Sheikah45: Note I wrote this in github so would need to double check syntax is correct

src/main/resources/theme/vault/replay/replay_card.fxml
mrchris2000: So something like this:
image
I'd change it to be inline with the watch button, but conceptually, around there is ok?

src/main/java/com/faforever/client/replay/ReplayService.java
mrchris2000: Required test changes because the 'FakeException' doesn't match a more specific catch statement.

src/main/resources/theme/vault/replay/replay_card.fxml
mrchris2000: Updated in the latest commit.

src/main/java/com/faforever/client/replay/ReplayService.java
mrchris2000: Can we consider this a follow up to this feature perhaps? 
It opens questions like, what if the player hasn't hosted a game yet during this play session (I assume use the current default), but also what if they have played a game but don't want the password and rating limitations to apply to this hosting session? Should we offer UI to set password and rating limits when hosting from this gesture? (I'd prefer not as its extra friction for the user, but there's probably a conversation there).

src/main/java/com/faforever/client/os/FileOpeningHandler.java
```diff
@@ -49,11 +49,26 @@ private void runReplay(Path filePath) {
...
       runReplay(Path.of(sourceArgs[0]));
     }
   }
+
+  public void host(ApplicationArguments args) {
+    String[] sourceArgs = args.getSourceArgs();
+    if (sourceArgs.length > 0) {
+      hostReplay(Path.of(sourceArgs[0]));
+    }
+  }
```
Sheikah45: What is the purpose of this? If we open a replay from the file system it is more likely they want to watch it not host a game like it.

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +400,77 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
...
+            notificationService.addImmediateWarnNotification("replayNotAvailable", replayId);
+          } else {
+            log.error("Replay could not be started", throwable);
+            notificationService.addImmediateErrorNotification(throwable, "replayCouldNotBeStarted", replayId);
+          }
+          return null;
+        });
+  }
+
+  CompletableFuture<Void> runHostFromFafReplayFile(Path path) throws IOException, CompressorException {
```
Sheikah45:
```suggestion
  CompletableFuture<Void> hostFromFafReplayFile(Path path) throws IOException, CompressorException {
```

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +400,77 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
...
+            GameVisibility.PUBLIC,
+            null,
+            null,
+            false
+        );
+        return gameService.hostGame(newGameInfo);
+    });
+  }
+
+  CompletableFuture<Void> runHostFromSupComReplayFile(Path path) throws IOException, CompressorException {
```
Sheikah45:
```suggestion
  CompletableFuture<Void> hostFromSupComReplayFile(Path path) throws IOException, CompressorException {
```

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -373,6 +400,77 @@ private void runFafReplayFile(Path path) throws IOException, CompressorException
...
 
+  private void hostFromOnlineReplay(int replayId) {
+    downloadReplay(replayId)
+        .thenAccept((path) -> {
+          try {
+            runHostFromFafReplayFile(path);
+          } catch (IOException | CompressorException e) {
+            throw new CompletionException(e);
+          }
+        })
```
Sheikah45:
```suggestion
        .thenCompose((path) -> {
          try {
            return runHostFromFafReplayFile(path);
          } catch (IOException | CompressorException e) {
            throw new CompletionException(e);
          }
        })
```

Sheikah45 (changes requested):

src/main/java/com/faforever/client/os/FileOpeningHandler.java
mrchris2000: I'm sure there was a reason for it, but I can't for the life of me recall it now.
I've removed it.

src/main/java/com/faforever/client/replay/ReplayService.java
mrchris2000: Updated to use Compose approach rather than Accept in latest commit.

Sheikah45 (changes requested):
Should be good once it is fixed so tests pass

mrchris2000:
Any advice on how to get it to pass the tests? The switch from 'thenAccept' to 'theCompose' has caused the break but I can't see how to resolve it (not used CompletableFutures before so there's quite the learning curve). I can cheat and use thenComposeAsync which passes one test but fails the others. From what I can tell its the mock thats causing problems, but I'm stumped as to how to resolve it; I'll keep reading though :)

Sheikah45:
> Any advice on how to get it to pass the tests? The switch from 'thenAccept' to 'theCompose' has caused the break but I can't see how to resolve it (not used CompletableFutures before so there's quite the learning curve). I can cheat and use thenComposeAsync which passes one test but fails the others. From what I can tell its the mock thats causing problems, but I'm stumped as to how to resolve it; I'll keep reading though :)

It is probably a race condition as the test hits the verify before the completable future has finished. You should make the functions return the completableFutures and then call join on them in the tests. That way you are guaranteed it has finished running before you verify and assert anything.

Sheikah45:
I think something might have gone wrong with a merge or rebase as this now shows ~1400 commits

mrchris2000:
> I think something might have gone wrong with a merge or rebase as this now shows ~1400 commits

I mucked up the rebase with upstream to try to resolve the merge conflict.
Not sure how to resolve it now. Might just bin this fork and re-implement.

PR #2602, opened 2022-03-03
Pause all labels of queue time if you are not in matchmaker tab

Marc-Spector:
Prevents useless GPU usage if you are not in the "Matchmaker" tab

### DIFF
--- a/src/main/java/com/faforever/client/teammatchmaking/MatchmakingQueueItemController.java
+++ b/src/main/java/com/faforever/client/teammatchmaking/MatchmakingQueueItemController.java
@@ -13,7 +13,6 @@ import javafx.animation.KeyFrame;
 import javafx.animation.Timeline;
 import javafx.beans.InvalidationListener;
 import javafx.beans.WeakInvalidationListener;
-import javafx.beans.property.BooleanProperty;
 import javafx.beans.value.ChangeListener;
 import javafx.beans.value.WeakChangeListener;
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/teammatchmaking/MatchmakingQueueItemController.java
```diff
@@ -161,4 +163,14 @@ public void onJoinLeaveQueueClicked(ActionEvent actionEvent) {
...
   }
+
+  public void addIsInMatchmakerTabListener(BooleanProperty isInMatchmakerTab) {
+    JavaFxUtil.addListener(isInMatchmakerTab, observable -> {
+      if (isInMatchmakerTab.get()) {
+        queuePopTimeUpdater.play();
+      } else {
+        queuePopTimeUpdater.pause();
+      }
+    });
```
Sheikah45: Rather than listeners for every queue item I think it would be better to have a start and stop timer method. Then the controller that displays the timers calls this start and stop method when it is displayed/hidden

Sheikah45 (changes requested):

Sheikah45:
@Marc-Spector were you planning on continuing this?

Marc-Spector:
Of course. When I have free time and desire.

Marc-Spector:
The #2619 PR fixed the high GPU usage issue. So this PR is no longer relevant.

PR #2609, opened 2022-03-07
Redesign chat user list & add show/hide categories

### LINKED ISSUE #1224: Make chat user groups hideable
allow users to hide sections of the chat user list, e.g. moderators, friends, everybody else, irc, ...

### LINKED ISSUE #2601: Make chat user list more attractive and interactive with the user

### LINKED ISSUE #2605: Add ability to customize chat user list
Related with #2601

### LINKED ISSUE #2606: Do not block UI thread when opening chat channel
UI thread blocks or slows down when a chat channel has many players or you have many channels for autojoin

Marc-Spector:
Closes #2601, #2605, #2606, #1224


Marc-Spector:
It is final design. I don't plan to change or fix it anymore.
demo7

src/main/java/com/faforever/client/main/MainController.java
```diff
@@ -101,10 +103,13 @@
 @Slf4j
 // TODO divide and conquer
 public class MainController implements Controller<Node>, InitializingBean {
+
   private static final PseudoClass NOTIFICATION_INFO_PSEUDO_CLASS = PseudoClass.getPseudoClass("info");
   private static final PseudoClass NOTIFICATION_WARN_PSEUDO_CLASS = PseudoClass.getPseudoClass("warn");
   private static final PseudoClass NOTIFICATION_ERROR_PSEUDO_CLASS = PseudoClass.getPseudoClass("error");
   private static final PseudoClass HIGHLIGHTED = PseudoClass.getPseudoClass("highlighted");
+  private final static ReadOnlyObjectWrapper<NavigationItem> CURRENT_TAB = new ReadOnlyObjectWrapper<>();
+
```
Sheikah45: Why is this static it doesn't really make any sense to be static since it is a property of a specific instance not shared across instances

src/main/resources/i18n/messages.properties
```diff
@@ -132,6 +132,8 @@ chat.filter.username = Username
 chat.filter.minRating = Min. Rating
 chat.filter.maxRating = Max. Rating
 chat.operator = SYSTEM
+chat.userList.settings.showMapPreview = show map preview
+chat.userList.settings.showMapName = show map name
```
Sheikah45:
```suggestion
chat.userList.settings.showMapPreview = Show map preview
chat.userList.settings.showMapName = Show map name
```

src/test/java/com/faforever/client/chat/ChatUserItemControllerTest.java
```diff
@@ -164,82 +117,4 @@ public void testOnContextMenuRequested() {
...
-    assertThat(instance.chatUserItemRoot.isManaged(), is(true));
-
-    instance.setVisible(false);
-    assertThat(instance.chatUserItemRoot.isVisible(), is(false));
-    assertThat(instance.chatUserItemRoot.isManaged(), is(false));
-
-    instance.setVisible(true);
-    assertThat(instance.chatUserItemRoot.isVisible(), is(true));
-    assertThat(instance.chatUserItemRoot.isManaged(), is(true));
-  }
```
Sheikah45: Why were these just removed instead of replaced or made new?

src/test/java/com/faforever/client/chat/ChatUserItemControllerTest.java
```diff
@@ -98,42 +87,6 @@ public void testGetPlayer() {
...
-        .statusImage(new Image(UiService.UNKNOWN_MAP_IMAGE))
-        .get();
-    instance.setChatUser(chatUser);
-    WaitForAsyncUtils.waitForFxEvents();
-
-    assertTrue(instance.avatarImageView.isVisible());
-    assertTrue(instance.playerStatusIndicator.isVisible());
-    assertTrue(instance.playerMapImage.isVisible());
-    assertTrue(instance.countryImageView.isVisible());
-  }
```
Sheikah45: Why were these just removed instead of replaced or made new?

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -114,12 +124,27 @@
...
   private Popup filterUserPopup;
   private ChatUserFilterController chatUserFilterController;
   private MapChangeListener<String, ChatChannelUser> usersChangeListener;
+  private boolean initializedList = false;
+  private Future<?> initializeListFuture;
+  private final InvalidationListener selectedTabListener = observable -> initializeListIfNeed();
+  private final InvalidationListener currentTabListener = observable -> {
+    if (MainController.getCurrentNavigationItem() == NavigationItem.CHAT) {
+      initializeListIfNeed();
+    }
```
Sheikah45: I don't quite understand the purpose of the listener here

Sheikah45 (changes requested):

src/test/java/com/faforever/client/chat/ChatUserItemControllerTest.java
Marc-Spector: I want to make sure the code has been reviewed. I don't want to rewrite tests many times. This is time-consuming.

src/main/java/com/faforever/client/main/MainController.java
Marc-Spector: How do I know which tab was open?

src/main/java/com/faforever/client/chat/ChannelTabController.java
Marc-Spector: I created this listener for the reason of this code:
https://github.com/FAForever/downlords-faf-client/blob/1e2165bf8c1a13f8bfea03c8c1e9856c07d67191/src/main/java/com/faforever/client/main/MainController.java#L223

All channel tabs become visible, whether the "Chat" tab is selected or not. Unable to determine when to initialize the list

src/test/java/com/faforever/client/chat/ChatUserItemControllerTest.java
Marc-Spector: See https://github.com/FAForever/downlords-faf-client/pull/2609#discussion_r824810819

src/test/java/com/faforever/client/chat/ChatUserItemControllerTest.java
Sheikah45: I am not sure what you mean about making sure the code has been reviewed. And I think the tests are important especially for the chat user list as many things broke without being noticed before in the past

Marc-Spector: I know tests are important and I will write the tests. I mean that I have to rewrite the tests if the methods can be changed or updated during the review.

Marc-Spector: You don't like the implementation of the `example` method, I'm correcting it and I still have to correct or rewrite the `example` test method

Marc-Spector:
demo1

src/test/java/com/faforever/client/chat/ChatUserItemControllerTest.java
Sheikah45: Ah okay gotcha

src/main/java/com/faforever/client/chat/ChatCategoryItemController.java
```diff
@@ -30,39 +34,65 @@
...
+  public Node root;
+  public Label categoryLabel;
+  public Label visibilityStateLabel;
+
   private ChatUserCategory chatUserCategory;
+  private String channelName;
+  private ObservableMap<String, ObservableList<ChatUserCategory>> channelNameToHiddenCategoriesProperty;
 
-  void setChatUserCategory(@Nullable ChatUserCategory chatUserCategory) {
+  public void setChatUserCategory(ChatUserCategory chatUserCategory, String channelName) {
```
Sheikah45: I know this is a draft but I don't see how associating the channel with the chatUserCategory would help. I think it would be  better to just have the association at the channeltabcontroller level

src/main/java/com/faforever/client/chat/test/ChatUserListController.java
```diff
@@ -0,0 +1,279 @@
...
+  public VBox userListContainer;
+
+  private final Map<ChatUserCategory, List<ChatUserItem>> categoriesToUsers = new HashMap<>();
+  private final Map<String, List<ChatUserItem>> usernameToChatUserList = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+  private final ObservableList<ListItem> source = FXCollections.observableArrayList();
+  private final Object monitor = new Object();
+
+  private MapProperty<String, ObservableList<ChatUserCategory>> channelNameToHiddenCategories;
+  private ObservableList<ChatUserCategory> hiddenCategories;
+  private final ObservableMap<ChatUserCategory, Boolean> visibleCategories = FXCollections.observableHashMap();
```
Sheikah45: Couldn't this just be a set?

Sheikah45 (commented):
Just as a note be careful with the synchronization there have been some tricky issues in the past with concurrent modifications since the char user info comes from server messages

Marc-Spector:
Lol... Now the `WebView` initializes faster before the `WebView` patcher finishes its work.

src/main/java/com/faforever/client/main/MainController.java
Sheikah45: I meant you could make it a property of the MainController rather than static

Marc-Spector: It won't help, it won't be possible to take him from another class. I have already found another solution. https://github.com/FAForever/downlords-faf-client/pull/2609/commits/c29b16ea198628d05933fa4927183811bc620749

Marc-Spector:
@Sheikah45 Everything seems to be stable. Only tests remain. Please review before I write and fix tests.

src/main/java/com/faforever/client/chat/ChatCategoryItemController.java
Marc-Spector: It was a temporary change. Reverted.

src/main/java/com/faforever/client/chat/test/ChatUserListController.java
Marc-Spector: It is better option. Done

Sheikah45:
> @Sheikah45 Everything seems to be stable. Only tests remain. Please review before I write and fix tests.

Will do

Sheikah45 (commented):
Looks good so far

Sheikah45 (changes requested):
The code looks good. Although I have a few concerns about the design as detailed below

- The categories right now are a dark grey which fades into the background and makes it hard to see. They should be a brighter color and maybe a bigger font. Potentially even fully left justified
- The game status icons are currently tied to showing the map name this is somewhat unintuitive. It should be broken out into its own element
- The game status icon should not be below the flags. As it currently stands when you have map names off each column is for a distinct item. With map name on you now have flags and status icons in the same column which is confusing and makes the information harder to identify.
- If the color selection for group applies to the category as well then it should probably be the same color all the time. Relates to the first point.

Marc-Spector:
> If the color selection for group applies to the category as well then it should probably be the same color all the time. Relates to the first point.

Could you explain in more detail? I don't see any problems. It works fine. Or do you mean something else?

image

Marc-Spector:
I see the same behavior in the current version of the client.

image

Sheikah45:
I just mean that in the redesign the group headers are different color than the names. When you choose a color for the group then the color of the header changes. That is a little counterintuitive. Either the header color should stay the same, or it should always be the same color as the names.

Marc-Spector:
> The game status icons are currently tied to showing the map name this is somewhat unintuitive. It should be broken out into its own element
> The game status icon should not be below the flags. As it currently stands when you have map names off each column is for a distinct item. With map name on you now have flags and status icons in the same column which is confusing and makes the information harder to identify.

I do not know what to do with these. Attempts to choose the better design were unsuccessful. The only thing that can be done is to remove the map name, return the game status icon to its old place, as in the current version of the client. Maybe you don't like the design because you're not used to it? :)

Sheikah45:
If another design doesn't work then I don't think it is too big a deal. But I would change the checkbox label from `show map name` to `show game status`

PR #2618, opened 2022-03-17
Only set value on free text fields

### LINKED ISSUE #2616: Cannot select any item for some criteria
Known broken criteria
From Maps vault: `Ranked`
From Replays vault: `Player Faction`, `Map is ranked`, `Validity`, `Victory Condition`

Sent by Eternal from discord:
test

1-alex98:
I can not see a way to reproduce this thing in the tests.
We would need a way to test components in integration for this.
Two controllers get in each other's way here.
Fixes #2616

### DIFF
--- a/src/main/java/com/faforever/client/vault/search/SearchController.java
+++ b/src/main/java/com/faforever/client/vault/search/SearchController.java
@@ -177,14 +177,20 @@ public class SearchController implements Controller<Pane> {
     logicalNodeController.specificationController.propertyField.valueProperty().addListener(queryInvalidationListener);
     logicalNodeController.specificationController.operationField.valueProperty().addListener(queryInvalidationListener);
     logicalNodeController.specificationController.valueField.valueProperty().addListener(queryInvalidationListener);
-    logicalNodeController.specificationController.valueField.getEditor().textProperty().addListener(observable -> {
-      if (!logicalNodeController.specificationController.valueField.valueProperty().isBound()) {
-        JavaFxUtil.runLater(() -> logicalNodeController.specificationController.valueField.setValue(logicalNodeController.specificationController.valueField.getEditor().getText()));
-      }
-    });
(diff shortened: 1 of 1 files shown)


1-alex98:
To explain... this lambda did deselects everything the user types if the value list is filled with none string elements...

src/main/java/com/faforever/client/vault/search/SearchController.java
```diff
@@ -177,18 +177,24 @@ private void addInvalidationListener(LogicalNodeController logicalNodeController
...
-    });
+    selectValueBeforeLoosingFocusOnValueField(logicalNodeController);
     logicalNodeController.specificationController.valueField.setOnKeyReleased(event -> {
       if (event.getCode() == KeyCode.ENTER) {
         searchButton.fire();
       }
     });
   }
 
+  private void selectValueBeforeLoosingFocusOnValueField(LogicalNodeController logicalNodeController) {
```
1-alex98: Why we need this?
Well if you start typing something in a free text field, then you want to see it life in the search query string. Default behaviour would be only to take effect if the combo box looses focus.
What's the issue?
This hack only works on free text fields or if the items list contains raw strings as items. In case the list is say booleans, it calls setValue("false") instead of setValue(false). This will cause the combo box to deselect. This will also trigger whenever the user tries to select something and overwrite him. Anyways if there is something in the list to select this feature is not that interesting. It's only needed for free text fields anyways so I now avoid it triggering on none free text fields.

Marc-Spector (dismissed):

src/main/java/com/faforever/client/vault/search/SearchController.java
Sheikah45: Losing not loosing

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #2619, opened 2022-03-17
Remove dirtyopts

### LINKED ISSUE #2615: Try removing -Dprism.dirtyopts=false
The client has been using `-Dprism.dirtyopts=false` for a long time as a workaround for https://github.com/FAForever/downlords-faf-client/issues/442

The problem is that this causes the whole UI to be rerendered even if just one pixel changes. This is especially bad when animations are used.

Back in time, I was able to reproduce the bug using the code from https://bugs.openjdk.java.net/browse/JDK-8089369
Right now, however, I'm no longer able to reproduce it with, not even with JDK 8 and JavaFX 11 so maybe it's fixed.

If we can make the client small, scroll all the way down or horizontally in a web view, and enlarge the client without this bug occurring, then it's probably fixed.

1-alex98:
Better performance, original bug not reproducible any more
Fixes #2615

### DIFF
--- a/.idea/runConfigurations/Main.xml
+++ b/.idea/runConfigurations/Main.xml
@@ -5,7 +5,7 @@
     </envs>
     <option name="MAIN_CLASS_NAME" value="com.faforever.client.Main" />
     <module name="faf-client.main" />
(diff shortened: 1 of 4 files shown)


Marc-Spector:
Please also remove the value from run configurations in files: [Main.xml](https://github.com/FAForever/downlords-faf-client/blob/develop/.idea/runConfigurations/Main.xml), [Main_with_Reload.xml](https://github.com/FAForever/downlords-faf-client/blob/develop/.idea/runConfigurations/Main_with_Reload.xml)

1-alex98:
uh true... that means  my test did not really execute it without that option

1-alex98:
Did test again all fine

Sheikah45 (approved):

PR #2633, opened 2022-03-31
Add league info to player info window

BlackYps:
grafik

The image for the unlisted league needs to be uploaded to the content server first


Sheikah45:
My main concern is that the faf league service isn't a core FAF service so if it ever breaks it could easily end up like the achievements. While this makes it more prominent than achievements. If you are willing to keep it up and also start all the new seasons then I am fine with this in principle but just wanted to let it be known.

BlackYps:
The league service will grow in importance even more when we show the divisions in the game scoreboard instead of the rating, even now we would lose the leaderboard functionality, so I think the league service won't suffer the same fate as the achievements.
I am very much willing to continue maintaining it in the future.

src/main/java/com/faforever/client/player/UserLeaderboardInfoController.java
```diff
@@ -0,0 +1,84 @@
...
+      divisionImage.setVisible(true);
+      divisionLabel.setText(divisionName);
+      divisionLabel.setVisible(true);
+    });
+  }
+
+  public void setUnrankedLeague() {
+    JavaFxUtil.runLater(() -> {
+      try {
+        divisionImage.setImage(leaderboardService.loadDivisionImage(new URL("https://content.faforever.com/divisions/icons/unranked.png")));
```
Sheikah45: Can this be not hard coded or at least made a client property and have the image be local to the client so we don't end up with 404s

Sheikah45 (commented):
When there are no images the space is still reserved so that should not be the case. Also this introduced a weird gap at the bottom of the windo
Untitled
w

src/main/java/com/faforever/client/player/UserLeaderboardInfoController.java
BlackYps: When it is local, what do we need a client property for?

BlackYps:
You see reserved space, because the image didn't load as it is not on the content server yet. I can change it to local, so we ship the image with the client

src/main/java/com/faforever/client/player/UserLeaderboardInfoController.java
Sheikah45: For the name of the image potentially. Although could be better as just a constant

Sheikah45:
Even then if the image isn't loaded it should be set to not visible so it is not managed. Also would it be the case that if there is no league that the images would not be reserved?

BlackYps:
the image is set to not visible in the fxml. There is a binding for the managed property. So it will only use space if it is explicitly set to be visible. If there is no league at all for a leaderboard it will never call the functions to set the league info, so it will stay hidden.
I noticed that when I have a slow connection and it takes a split second for the league info to load, it only makes room for the images once they appear. So it is working.

Sheikah45 (commented):
Can you add tests for the new parts of the PlayerInfoWindowController

Sheikah45:
Also the png should be called unlisted_league_image or something not just unlisted

Sheikah45 (approved):

PR #2640, opened 2022-04-04
Fix button size changing on hover

### LINKED ISSUE #2639: Queue button size changes on hover
**Describe the bug**
Hover of the queue button, they will become smaller after hover

1-alex98:
Button size changed on hover. This is a strange thing.
 I do not why that is but it must have to do with priority of styles.
 By putting font size to the item's style the issue is mitigated.
 I would appreciate a proper solution but I fear nobody is gonna figure
 out that magical bug.

Fixes #2639


Sheikah45:
Hmm very weird. I ran into this in the migration to javafx18 and thought I had fixed it but guess not. Weird bug

Sheikah45 (approved):

PR #2659, opened 2022-04-13
Add informative icons to the `create ` button

### LINKED ISSUE #2645: Add an indicator for the host that his lobby is only for friends or not
https://forum.faforever.com/topic/3565/patch-3732/17?_=1649348814019 by master-chief

Marc-Spector:
Closes #2645 

To avoid when you create the game with the wrong parameters, which you did not notice


Sheikah45:
Hmm looking at this I am not sure how helpful the icons are as it isn't entirely intuitive to me on first glance what game setting each icon is related to. Also shouldn't there be one for enforced rating range as well?

Marc-Spector:
Even the tooltips didn't help you? 🤔

Marc-Spector:
It looks like we need another way to communicate it. Or just close the issue, since the issue is not common.

Marc-Spector:
Or, in general, it is better to redo the design of the upper part of the window, so that we can pay attention to significant things.

Sheikah45:
Hmm I didn't even see the tooltips really and I don't know if it is intuitive that tool tips would be there

Sheikah45:
I think making it so that the visibility changing options are more obvious when enabled might be better

Marc-Spector:
> I think making it so that the visibility changing options are more obvious when enabled might be better

Okay.

Marc-Spector:
I do not have an idea how to do it. I will open pr if an idea comes up.

> I think making it so that the visibility changing options are more obvious when enabled might be better

Marc-Spector:
I can not think of a better design than the current one. =)

PR #2662, opened 2022-04-14
Disable ICE adapter info window

putterer:
Shows only the debug window when requested, no reason to inform the user about what that is after they specifically enabled it in the settings.


Sheikah45:
Not everyone is aware what exactly the debug window is and that is should not be closed, so by default the info window which provides these details should remain in place as the reminder is not a bad thing.

PR #2668, opened 2022-04-19
Allow set a channel topic by moderators

### LINKED ISSUE #1474: allow tournament directors (TD) to set the topic
Topic has been brought back in #1335 (and it's helping a lot moderators, for pinning important information in the chat)
But it is not possible to change the topic. And allowing this, would permit TDs to display and pin important information in the tournament chat. Like the challonge link for example.
I know that you have implemented a tournament tab, but it's not easy to update information on it.

So i guess you can do it as python do : everyone has the right when they create a new channel (by being the first to join the channel : /join mychannel).
Or you give some kind of priviledge to some people TDs/moderators/councilor so they are the only one to update the topic.

Updating topic in python client is done by doing : /topic mynewmessage

### LINKED ISSUE #2669: After creating a new channel, you are not in the "Moderators" list

Marc-Spector:
Closes #1474 
Closes #2669


src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -183,6 +205,13 @@ private void updateChannelTopic() {
             }
           });
     }
+    topicTextField.setVisible(false);
+    topicPane.setDisable(false);
+    if (topicTriggeredByUser) {
+      topicTriggeredByUser = false;
+      String username = userService.getUsername();
+      chatService.sendMessageInBackground(channelName, username + " updated the channel topic");
```
Sheikah45: Why are we manually sending a notice that the topic has changed? Doesn't the irc server do this

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -115,17 +123,33 @@ public void afterPropertiesSet() throws Exception {
...
+
+    JavaFxUtil.addAndTriggerListener(topicTextField.textProperty(), new WeakInvalidationListener(changeTopicTextListener));
   }
 
   public void setChatChannel(ChatChannel chatChannel, BooleanBinding chatTabSelectedProperty) {
     this.chatChannel = chatChannel;
     this.channelName = chatChannel.getName();
+    Optional.ofNullable(chatChannel.getUser(userService.getUsername())).ifPresentOrElse(
+        user -> {
+          JavaFxUtil.bind(changeTopicTextButton.visibleProperty(), user.moderatorProperty().and(topicTextField.visibleProperty().not()));
```
Sheikah45: Can't use a bind here as moderatorProperty can be changed in any thread not just the application thread

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -115,17 +123,33 @@ public void afterPropertiesSet() throws Exception {
...
+          JavaFxUtil.bind(topicPane.visibleProperty(), chatChannel.topicProperty().isNotEmpty().or(user.moderatorProperty()));
```
Sheikah45: same issue with moderator property here

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -293,6 +322,40 @@ protected String getInlineStyle(String username) {
...
+  public void onCancelChangesTopicTextButtonClicked() {
+    String topic = chatChannel.getTopic();
+    topicTextField.setText(topic != null ? topic : "");
+    topicTextField.setVisible(false);
+    topicPane.setDisable(false);
+  }
+
+  public void onAcceptChangesTopicTextField() {
+    onAcceptTopicTextButtonClicked();
+  }
```
Sheikah45: What is the purpose of this method?

Sheikah45 (changes requested):

src/main/java/com/faforever/client/chat/ChannelTabController.java
Marc-Spector: It sometimes happens that players do not notice that the topic has been changed when they are in a different place (lobby, during the game). Especially when the TD often changes the topic

src/main/java/com/faforever/client/chat/ChannelTabController.java
Marc-Spector: It's the second way to accept the changes via the `Enter` button.

src/main/java/com/faforever/client/chat/ChannelTabController.java
Marc-Spector: We cannot know that the joined channel has an already established topic or not.

Marc-Spector: Hmm.. I think I understand what you mean. I'll change it.

Marc-Spector:
@Sheikah45 Fix tests on develop branch, please 😄

Marc-Spector:
Ah, I see it is alredy fixed

src/main/java/com/faforever/client/chat/ChannelTabController.java
Marc-Spector: And `topicProperty` also?

Marc-Spector:
There is some error in the code. The `changeTopicButton` button is still visible in edit mode.

src/main/java/com/faforever/client/chat/ChannelTabController.java
Sheikah45: Why not just use the onAcceptTopicTextButtonClicked there then? It seems a little superfluous to have a method that just directly calls another method with the exact same arguments

Sheikah45 (commented):
I would remove the duplicate method and just name the single method appropriately. No need to have two methods that d the same thing.

src/main/java/com/faforever/client/chat/ChannelTabController.java
Sheikah45: I missed this but yes

Sheikah45:
I was playing around with this and I think we can do better with the notification that the topic has changed. Rather than just telling the user that the topic has changed we should state in the message what the topic has changed to. Maybe a from to but at  least the to.

Also the topic change event from the kitteh chat service contains the actual setter of the topic so we should use that rather than our own user name for the message about the topic change.

Sheikah45:
The setter info is in `event.getNewValue().getSetter()` in the `ChannelTopicEvent`

Sheikah45 (changes requested):
See above comments

Marc-Spector:
> The setter info is in `event.getNewValue().getSetter()` in the `ChannelTopicEvent`

`event.getNewTopic().getSetter().map(Actor::getName).orElse("")` returns `MarcSpector!352305@7927089D.9209970C.27282DB6.IP` value.

Marc-Spector:
@Sheikah45 I cannot push. Do I need to recreate the token with workflow scope? It used to work without it.
image

Sheikah45:
It is probably because you rebased onto master where I updated it so yes remake the token with workflow scope.

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -183,6 +207,24 @@ private void updateChannelTopic() {
...
+
+  private void updateChannelTopic() {
+    String oldTopicContent = topicContent;
+    ChannelTopic topic = chatChannel.getTopic();
+    setChannelTopic(topic.getContent());
+    if (StringUtils.isNotBlank(topic.getAuthor())) {
+      onChatMessage(new ChatMessage(channelName, Instant.now(), topic.getAuthor(), i18n.get("chat.topicUpdated")));
+      onChatMessage(new ChatMessage(channelName, Instant.now(), topic.getAuthor(), i18n.get("chat.oldTopicContent", oldTopicContent)));
+      onChatMessage(new ChatMessage(channelName, Instant.now(), topic.getAuthor(), i18n.get("chat.newTopicContent", topic.getContent())));
```
Sheikah45: This should all be sent as a single message so that it doesn't get broken up by other messages

src/main/java/com/faforever/client/chat/KittehChatService.java
```diff
@@ -288,8 +290,10 @@ private void onChatUserQuit(UserQuitEvent event) {
 
   @Handler
   private void onTopicChange(ChannelTopicEvent event) {
-    Channel channel = event.getChannel();
-    getOrCreateChannel(channel.getName()).setTopic(event.getNewTopic().getValue().orElse(""));
+    String author = event.getNewTopic().getSetter().map(Actor::getName)
+        .map(name -> name.replaceFirst("!.*", "")).orElse("");
```
Sheikah45, line 294: This works, we could also check the type of the actor and use methods based on that but this is more general given it is part of the irc spec.

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #2681, opened 2022-04-29
Restore Game Session and Lobby connection properly

Sheikah45:
LoginSuccessResponse is not a part of the events stream from lobbyClient

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -297,7 +297,7 @@ dependencies {
   implementation("org.springframework:spring-websocket")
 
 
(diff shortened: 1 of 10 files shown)


Sheikah45:
Uses https://github.com/FAForever/faf-java-commons/pull/88 and moves retry logic to commons library

Sheikah45:
This moves the reconnect logic into the commons library and ensures that lobby disconnects are discovered quickly when in game to recover them

PR #2691, opened 2022-05-05
fix copy map name on create game window

### LINKED ISSUE #2687: Map name in Game Creation dialog map thumbnail is no longer right-clickable since the introduction of map zoom
Before we had [map zoom](https://github.com/FAForever/downlords-faf-client/pull/2641) in the Game creation dialog, one could right click the map name displayed on top of the thumbnail to copy it to clipboard, when that map was a generated one. This was useful in case i had to restart faf because my lobby broke (since restarting faf clears generated maps, i have to get that map name into the clipboard to be able to set the lobby back up fast). With the addition of map zoom, the right click functionality is gone, and instead the right click now behaves exactly like the left click, zooming in the map. This leaves no option for copying the map name of a generated map that's not associated with a replay except for hosting a game on that map and alt-tabbing to the faf client, then highlighting the lobby and right-clicking the map name in the lobby details on the right. Since sometimes broken lobbies disappear from the custom game list, even this option might not be available precisely when it's needed.

Marc-Spector:
Closes #2687

### DIFF
--- a/src/main/java/com/faforever/client/game/CreateGameController.java
+++ b/src/main/java/com/faforever/client/game/CreateGameController.java
@@ -11,6 +11,7 @@ import com.faforever.client.fx.DualStringListCell;
 import com.faforever.client.fx.JavaFxUtil;
 import com.faforever.client.fx.StringListCell;
 import com.faforever.client.fx.contextmenu.ContextMenuBuilder;
+import com.faforever.client.fx.contextmenu.CopyLabelMenuItem;
 import com.faforever.client.i18n.I18n;
 import com.faforever.client.map.MapService;
 import com.faforever.client.map.MapService.PreviewSize;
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -522,10 +525,20 @@ public void onMapFilterButtonClicked() {
...
+          .map(MapVersionBean::getFolderName)
+          .ifPresent(mapName -> PopupUtil.showImagePopup(mapService.loadPreview(mapName, PreviewSize.LARGE)));
+
+      // No other way to make same design without using StackPane class therefore we use this dirty hack
+    } else if (button == MouseButton.SECONDARY && mapNameLabel.getBoundsInParent().contains(mouseEvent.getX(), mouseEvent.getY())) {
+      contextMenuBuilder.newBuilder()
+          .addItem(CopyLabelMenuItem.class, mapNameLabel)
+          .build()
+          .show(mapNameLabel, mouseEvent.getScreenX(), mouseEvent.getScreenY());
+    }
```
Sheikah45: Rather than complicating the onMapPreviewImageClicked, add a new function and assign it to onContextMenuRequested of the stackPane perhaps

Sheikah45 (approved):

PR #2692, opened 2022-05-06
Fix map preview height in live replays tab

Marc-Spector:
image

### DIFF
--- a/src/main/resources/theme/style.css
+++ b/src/main/resources/theme/style.css
@@ -1428,7 +1428,8 @@
 /***************** Live Replay Tab ***************************/
 
 .live-replay-tab .table-row-cell {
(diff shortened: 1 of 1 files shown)


Marc-Spector:
In the current version of client, size of map preview is 36x30px. Now it is 36x36 as specified in the fxml file.
https://github.com/FAForever/downlords-faf-client/blob/baab841e723d22bd0af5333ec1182ea7d6f01fb4/src/main/resources/theme/vault/map/map_preview_table_cell.fxml#L5-L8

Sheikah45 (dismissed):

Marc-Spector:
There was an issue with no separation between the watch buttons. Fixed.

Sheikah45 (approved):

PR #2696, opened 2022-05-11
Add winrate column

### LINKED ISSUE #2526: Add more data to the leaderboards, like rating and win percentage
From forum https://forum.faforever.com/topic/3206/small-suggestions-topic/40?_=1642362790786

Marc-Spector:
Closes #2526

### DIFF
--- a/src/main/java/com/faforever/client/leaderboard/SubDivisionTabController.java
+++ b/src/main/java/com/faforever/client/leaderboard/SubDivisionTabController.java
@@ -67,11 +67,8 @@ public class SubDivisionTabController implements Controller<Tab> {
     gamesPlayedColumn.setCellValueFactory(param -> param.getValue().gamesPlayedProperty());
     gamesPlayedColumn.setCellFactory(param -> new StringCell<>(count -> i18n.number(count.intValue())));
 
-    winRateColumn.setCellValueFactory(param -> Bindings.createFloatBinding(() -> {
-      LeagueEntryBean entry = param.getValue();
-      float winRate = (float) entry.getScore() / entry.getGamesPlayed() * 100;
-      return winRate > 100 ? 100 : winRate;
-    }));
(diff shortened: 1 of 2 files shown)


Marc-Spector:
image
Is the winrate formula correct? I'm not sure looking at what winrate  `ZLO` player has. 🤔

BlackYps:
No, that is not correct. The score does not show how many games you have won.

BlackYps:
I'm not sure if we can calculate the winrate for a season in a sensible manner. We would have to look up the games during that time for every player individually and calculate the winrate from that. That would need a lot of api calls

BlackYps:
I think the best solution is to add games lost and games won columns to the league score journal table in the league service. Then we can easily access this information and even decide how we want to deal with draws.

src/main/java/com/faforever/client/leaderboard/SubDivisionTabController.java
```diff
@@ -65,6 +67,10 @@ public void initialize() {
     gamesPlayedColumn.setCellValueFactory(param -> param.getValue().gamesPlayedProperty());
     gamesPlayedColumn.setCellFactory(param -> new StringCell<>(count -> i18n.number(count.intValue())));
 
+    winRateColumn.setCellValueFactory(param -> Bindings.createFloatBinding(() ->
+        Math.min((float) param.getValue().getScore() / param.getValue().getGamesPlayed() * 100, 100)));
+    winRateColumn.setCellFactory(param -> new StringCell<>(winRate -> i18n.rounded(winRate.doubleValue(), 1)));
```
Sheikah45, line 72: As BlackYPS said the score does not equal the number of games won so this is not accurate

Sheikah45 (changes requested):

Marc-Spector:
Ok. I closing the PR.

PR #2704, opened 2022-05-18
Show only unique players and add date column in coop tab

### LINKED ISSUE #2703: COOP Leaderboards Update
Update Leaderboard to only display users fastest time & the date of when this was completed on

- [ ] Add Date of completion
- [ ] Only display usres fasted time

image

### LINKED ISSUE #2705: Repair broken delimiter of list players in coop table
Try to open list with count of players more than 1
image

### LINKED ISSUE #2715: Repair broken delimiter of mods in live games
image

Marc-Spector:
At the request of `Rowey` Promotions Team Leader
Closes #2703
Closes #2705 
Closes #2715 

image

### DIFF
--- a/src/main/java/com/faforever/client/coop/CoopController.java
+++ b/src/main/java/com/faforever/client/coop/CoopController.java
@@ -178,7 +178,9 @@ public class CoopController extends AbstractViewController<Node> {
   }
 
   private String commaDelimitedPlayerList(CoopResultBean coopResult) {
-    return coopResult.getReplay().getTeams().values().stream().flatMap(List::stream).collect(Collectors.joining(i18n.get("textSeparator")));
+    return coopResult.getReplay().getTeams().values().stream()
+        .flatMap(List::stream)
+        .collect(Collectors.joining(i18n.get("textSeparator")));
   }
 
   private void onReplayButtonClicked(ReplayButtonController button) {
--- a/src/main/java/com/faforever/client/coop/CoopController.java
+++ b/src/main/java/com/faforever/client/coop/CoopController.java
@@ -21,6 +21,8 @@ import com.faforever.client.util.TimeService;
 import com.faforever.commons.lobby.GameStatus;
 import com.faforever.commons.lobby.GameType;
 import com.google.common.base.Strings;
+import javafx.beans.Observable;
+import javafx.beans.binding.Bindings;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.property.SimpleStringProperty;
 import javafx.collections.ObservableList;
@@ -47,8 +49,14 @@ import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
 import java.time.Duration;
+import java.time.OffsetDateTime;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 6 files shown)


src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -172,9 +177,7 @@ private String coopMissionFromFolderNamer(List<CoopMissionBean> coopMaps, String
   }
 
   private String commaDelimitedPlayerList(CoopResultBean coopResult) {
-    return coopResult.getReplay().getTeams().values().stream()
-        .flatMap(List::stream)
-        .collect(Collectors.joining(i18n.get("textSeparator")));
+    return String.join(", ", coopResult.getReplay().getTeams().values().stream().flatMap(List::stream).toList());
```
Sheikah45: Why was this changed to not use the internationalized separator? Also the Collectors.joining is the preferred method since we are already using the stream api.

src/main/java/com/faforever/client/domain/CoopResultBean.java
```diff
@@ -18,6 +18,7 @@
 public class CoopResultBean {
   @EqualsAndHashCode.Include
   ObjectProperty<Integer> id = new SimpleObjectProperty<>();
+  // TODO: Why returns `null` value?
```
Sheikah45: Probably is just old data that doesn't have playerNames

src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -232,6 +236,12 @@ private void loadLeaderboard() {
         });
   }
 
+  private List<CoopResultBean> filterOnlyUniquePlayers(List<CoopResultBean> result) {
+    Set<String> uniquePlayerNames = new HashSet<>();
+    result.removeIf(coopResult -> !uniquePlayerNames.add(commaDelimitedPlayerList(coopResult)));
+    return result;
+  }
```
Sheikah45: This doesn't work if the players are in a different order. I would use something like an unmodifiable set as the key

Sheikah45 (changes requested):

src/main/java/com/faforever/client/coop/CoopController.java
Marc-Spector: > Why was this changed to not use the internationalized separator?

Usually this only applies to numbers, doesn't it?

> Also the Collectors.joining is the preferred method since we are already using the stream API.

Weird. The IDE suggested that it's better to use this instead. Ok, I will fix it

src/main/java/com/faforever/client/domain/CoopResultBean.java
Marc-Spector: Can I remove it?

src/main/java/com/faforever/client/coop/CoopController.java
Marc-Spector: Yes, it won't work if we remove the sort rule:
https://github.com/FAForever/downlords-faf-client/blob/ccef5b08309528993f4ee2a4fdc6c778015d8413/src/main/java/com/faforever/client/coop/CoopService.java#L51

Sheikah45: That is sorting on the replay not on the players or the teams

src/main/java/com/faforever/client/domain/CoopResultBean.java
Sheikah45: is it used anywhere?

src/main/java/com/faforever/client/coop/CoopController.java
Sheikah45: the text separator applies to text numbers is something different.

src/main/java/com/faforever/client/domain/CoopResultBean.java
Marc-Spector: Nope

Sheikah45 (approved):

PR #2709, opened 2022-05-19
WIP: Mute/unmute the user

### LINKED ISSUE #2214: Allow muting people without foeing
@**Thomas Hiatt** [said](https://faforever.zulipchat.com/#narrow/stream/203511-python-server/topic/github/near/235401419):
```quote
i think hiding someones chat from aeolus, and preventing them from joining your games should be 2 different features really
```

Sometimes people just don't want to read other's chat messages, but they don't mind playing custom games together. The mute feature could be implemented completely client side since the lobby server doesn't know anything about the irc chat.

Marc-Spector:
Closes #2214 

The icons have become more interactive
demo55111


Marc-Spector:
Replaced mute icon with blocked eye icon 
image

src/main/java/com/faforever/client/chat/ChatService.java
```diff
@@ -63,4 +64,12 @@ public interface ChatService {
...
 
   String getDefaultChannelName();
+
+  boolean isUserMuted(String username);
+
+  void muteUser(String username);
+
+  void unmuteUser(String username);
+
+  ObservableSet<Integer> getMutedUserIds();
```
Sheikah45, line 74: The muted should not only be userIds as chat only users will not have consistent ids associated with them

src/main/java/com/faforever/client/chat/KittehChatService.java
```diff
@@ -605,6 +612,30 @@ public String getDefaultChannelName() {
...
+    mutedUserIds.remove(getPlayerId(username));
+  }
+
+  @Override
+  public ObservableSet<Integer> getMutedUserIds() {
+    return mutedUserIds;
+  }
+
+  private Integer getPlayerId(String username) {
+    return getOrCreateChatUser(username, username, false).getPlayer().map(AbstractEntityBean::getId).orElse(null);
```
Sheikah45, line 636: This creates a lot of chatUsers that might not actually exist. If all we want is the id then we should just use the playerService directly to get the player rather than unecessarily creating a chatUser. Especially since this is called for every user's message through isUserMuted so it will think we have a private channel open with every user who sends a message and also won't have a chance to be closed.

src/main/java/com/faforever/client/chat/KittehChatService.java
```diff
@@ -341,13 +347,14 @@ private void onPrivateMessage(PrivateMessageEvent event) {
     User user = event.getActor();
     ircLog.debug("Received private message: {}", event);
 
-    ChatChannelUser sender = getOrCreateChatUser(user.getNick(), user.getNick(), false);
+    String username = user.getNick();
+    ChatChannelUser sender = getOrCreateChatUser(username, username, false);
     if (sender.getPlayer().map(PlayerBean::getSocialStatus).filter(status -> status == SocialStatus.FOE).isPresent()
-        && preferencesService.getPreferences().getChat().getHideFoeMessages()) {
-      ircLog.debug("Suppressing chat message from foe '{}'", user.getNick());
+        && preferencesService.getPreferences().getChat().getHideFoeMessages() || isUserMuted(username)) {
```
Sheikah45, line 353: For clarity I would group the first two conditions with parentheses

Sheikah45 (changes requested):
At the moment I think the behavior as is, is a little misleading. Right now you can only mute and un-mute the user from a private chat window. However when muted they also have their messages hidden in all other channels but this is not clear from the UI. Would be good to allow people to mute from the context menu as well

Additionally there is no way to know in total which users you have muted. You would have to individually go through each user to see if you had them muted or not. I think it would be good to have a separate property like isModerator so we can properly display who is muted.

Additionally right now this is only capable of muting players since we use ids, however a large impetus behind this feature was the ability to also mute chat only users so we should primarily use ids but have a fallback of nicks to check as well, otherwise this is just entirely parallel to the foe option.

Marc-Spector:
I closing the PR. I think it's better to wait for new fields to appear in the DB. https://github.com/FAForever/downlords-faf-client/issues/2214#issuecomment-1027266196

PR #2729, opened 2022-06-18
Enable more sort options

hahn-kev:
Currently you can only sort by name in mod and map search. This change enables sorting by additional properties. 

In making this change I noticed that if I just come to the map vault and change the sort by then the search button is still disabled, but if I want to search all maps but with a custom sort by then I'd like the search button enabled. This change will also enable the search button when it's not the first (or default) search option.

### DIFF
--- a/src/main/java/com/faforever/client/vault/VaultEntityController.java
+++ b/src/main/java/com/faforever/client/vault/VaultEntityController.java
@@ -131,9 +131,6 @@ public abstract class VaultEntityController<T> extends AbstractViewController<No
 
     initSearchController();
 
-    BooleanBinding inSearchableState = Bindings.createBooleanBinding(() -> state.get() != State.SEARCHING, state);
-    searchController.setSearchButtonDisabledCondition(inSearchableState);
-
     pagination.currentPageIndexProperty().addListener((observable, oldValue, newValue) -> {
       if (!oldValue.equals(newValue)) {
         SearchConfig searchConfig = searchController.getLastSearchConfig();
--- a/src/main/java/com/faforever/client/vault/VaultEntityController.java
+++ b/src/main/java/com/faforever/client/vault/VaultEntityController.java
@@ -131,6 +131,9 @@ public abstract class VaultEntityController<T> extends AbstractViewController<No
 
     initSearchController();
 
+    BooleanBinding inSearchableState = Bindings.createBooleanBinding(() -> state.get() != State.SEARCHING, state);
+    searchController.setSearchButtonDisabledCondition(inSearchableState);
+
     pagination.currentPageIndexProperty().addListener((observable, oldValue, newValue) -> {
       if (!oldValue.equals(newValue)) {
         SearchConfig searchConfig = searchController.getLastSearchConfig();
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/client/vault/search/SearchController.java
```diff
@@ -429,7 +430,14 @@ public void setRootType(Class<?> rootType) {
...
   public void setSearchButtonDisabledCondition(BooleanBinding inSearchableState) {
-    searchButton.disableProperty().bind(queryTextField.textProperty().isEmpty().or(inSearchableState.not()));
+    Optional<Property> firstSortProperty = sortPropertyComboBox.getItems().stream().findFirst();
+    BooleanBinding isSortChanged = firstSortProperty.isEmpty()
+        ? Bindings.selectBoolean(false)
+        : sortPropertyComboBox.valueProperty().isNotEqualTo(firstSortProperty.get());
+    searchButton.disableProperty().bind(
+        queryTextField.textProperty().isEmpty().and(isSortChanged.not())
+        .or(inSearchableState.not())
+    );
```
Sheikah45: I don't understand why this change is necessary. If this is just checking if the sort property is populated that is easier done by setting the combo box to not have editable or null values which I believe is already done.

Additionally right now this is only binding on the first item in the combobox rather than the selected item of the combo box so will have unintended behavior. And you will never be able to sort by the first property in the list.

As you can see in the image below it disables the search when sort is the first value which shouldn't happen. I don't think this needs to be changed at all.
image

Sheikah45 (changes requested):
For the most part this is good although the binding change is confusing. I left more details at the code point.

src/main/java/com/faforever/client/vault/search/SearchController.java
hahn-kev: the reason I did this is so that you can do a search with no filters and a different sort by.

Sheikah45: Then it would be better to just explicitly allow for searching with no filters than do this as it doesn't make sense why you just can't search with one of the sort options

hahn-kev: that makes sense

src/main/java/com/faforever/client/vault/VaultEntityController.java
```diff
@@ -131,9 +131,6 @@ public void initialize() {
 
     initSearchController();
 
-    BooleanBinding inSearchableState = Bindings.createBooleanBinding(() -> state.get() != State.SEARCHING, state);
-    searchController.setSearchButtonDisabledCondition(inSearchableState);
-
```
Sheikah45: We should still have the button disabled when searching sorry if that wasn't clear

src/main/java/com/faforever/client/vault/search/SearchController.java
```diff
@@ -428,10 +429,6 @@ public void setRootType(Class<?> rootType) {
     initialLogicalNodeController.specificationController.setRootType(rootType);
   }
 
-  public void setSearchButtonDisabledCondition(BooleanBinding inSearchableState) {
-    searchButton.disableProperty().bind(queryTextField.textProperty().isEmpty().or(inSearchableState.not()));
-  }
-
```
Sheikah45: just remove the queryTextField.textProperty().isEmpty() and only bind on inSearchableState.not()

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #2736, opened 2022-06-26
Rename 4v4 Full Share to plain 4v4

BlackYps:
With the removal of the 4v4 No Share queue and leaderboard we can use the short name again like we do for the other ones.


Sheikah45 (approved):
Should wait until it is actually turned off to merge

PR #2742, opened 2022-07-04
Show avatar instead of default host icon in game tiles

Marc-Spector:
To be consistent

image

image

### DIFF
--- a/src/main/java/com/faforever/client/game/GameTileController.java
+++ b/src/main/java/com/faforever/client/game/GameTileController.java
@@ -19,6 +19,7 @@ import javafx.scene.image.Image;
 import javafx.scene.image.ImageView;
 import javafx.scene.input.MouseButton;
 import javafx.scene.input.MouseEvent;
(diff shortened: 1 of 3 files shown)


src/main/java/com/faforever/client/game/GameTileController.java
```diff
@@ -53,6 +54,7 @@ public class GameTileController implements Controller<Node> {
   public Label avgRatingLabel;
   public Label hostLabel;
   public ImageView avatarImageView;
+  public Region defaultHostIconImageView;
```
Sheikah45: This isn't an imageView so I would just name it defaultHostIcon

Sheikah45 (changes requested):

src/main/java/com/faforever/client/game/GameTileController.java
Marc-Spector: Good catch

PR #2748, opened 2022-07-06
Show warning when vault base path contains non ascii characters

### LINKED ISSUE #2613: Non ascii characters in vault path
**Describe the bug**

The game can not load in maps / mods when there are unknown characters in the path of the map / mod. The client can detect that and choose a different location, but in this instance that didn't work.

**To Reproduce**

image

Any user profile with this type of character, it is polish according to the user.

**Expected behavior**

According to Zlo the client can already detect non-ascii characters in some situations. This would need to be extended.

**OS**

Windows

Marc-Spector:
Closes #2613


src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -283,12 +292,30 @@ public CompletableFuture<Void> hostGame(NewGameInfo newGameInfo) {
...
+          i18n.get("vaultBasePath.warning.description"),
+          WARN,
+          List.of(new CancelAction(i18n), new Action(i18n.get("vaultBasePath.warning.hostGame"), (event) -> createGame(newGameInfo))),
+          checkBox
+      ));
+      return completedFuture(null);
+    }
+
+    return createGame(newGameInfo);
+  }
```
Sheikah45: We shouldn't be creating ui elements in a service.

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -310,6 +337,23 @@ public CompletableFuture<Void> joinGame(GameBean game, String password) {
...
+      JavaFxUtil.bindBidirectional(checkBox.selectedProperty(), forgedAlliancePrefs.dontShowNonAsciiVaultBasePathWarning());
+      notificationService.addNotification(new ImmediateNotification(
+          i18n.get("vaultBasePath.warning.title"),
+          i18n.get("vaultBasePath.warning.description"),
+          WARN,
+          List.of(new CancelAction(i18n), new Action(i18n.get("vaultBasePath.warning.joinGame"), (event) -> joinGame_(game, password))),
+          checkBox
+      ));
+      return completedFuture(null);
+    }
```
Sheikah45: Same as above about UI elements

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -310,6 +337,23 @@ public CompletableFuture<Void> joinGame(GameBean game, String password) {
...
+
+    return joinGame_(game, password);
+  }
+
+  private CompletableFuture<Void> joinGame_(GameBean game, String password) {
```
Sheikah45: The underscore isn't a convention in this repo if used like this it should be more descriptive

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -854,6 +898,10 @@ public CompletableFuture<Boolean> isGamePrefsPatchedToAllowMultiInstances() thro
     return completedFuture(GAME_PREFS_ALLOW_MULTI_LAUNCH_PATTERN.matcher(gamePrefsContent).find());
   }
 
+  public boolean checkVaultBasePathIsValidForAcsii() {
+    return US_ASCII.newEncoder().canEncode(preferencesService.getPreferences().getForgedAlliance().getVaultBaseDirectory().toString());
+  }
```
Sheikah45: Don't think we want this in the game service, or called before every game launch since it is pretty static, better to be in the preferencesService

src/main/java/com/faforever/client/preferences/ForgedAlliancePrefs.java
```diff
@@ -40,6 +40,7 @@ public class ForgedAlliancePrefs {
   ObjectProperty<Path> installationPath = new SimpleObjectProperty<>(STEAM_FA_PATH);
   ObjectProperty<Path> preferencesFile = new SimpleObjectProperty<>(LOCAL_FA_DATA_PATH.resolve("Game.prefs"));
   ObjectProperty<Path> vaultBaseDirectory = new SimpleObjectProperty<>(DEFAULT_VAULT_DIRECTORY);
+  BooleanProperty dontShowNonAsciiVaultBasePathWarning = new SimpleBooleanProperty(false);
```
Sheikah45: Rename to warnNonAsciiBasePath

Sheikah45 (changes requested):
I recognize this is WIP but saw a few things I wanted to comment on. Generally I don't think we want the check to be in the GameService or performed before every game launch. It would probably be better to treat it similarly to the gamePathValid check and maybe only do it on start up and have a separate class/event which creates and handles the UI.

Marc-Spector:
Thanks for comments. Should we also add this check before running a replay and a live replay? Or is this an unnecessary action?

Sheikah45:
I think just once during client startup is fine

Marc-Spector:
Done. I will add tests soon.

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -227,6 +228,18 @@ public void afterPropertiesSet() {
...
+    if (preferencesService.isVaultBasePathInvalidForAscii()) {
+      log.warn("Vault base path contains non ASCII characters: {}", forgedAlliancePrefs.getVaultBaseDirectory());
+      if (forgedAlliancePrefs.getWarnNonAsciiVaultPath()) {
+        notificationService.addPersistentWarnNotification("vaultBasePath.nonAscii.warning.title",
+            List.of(new Action(i18n.get("vaultBasePath.nonAscii.warning.ignore"), event -> {
+              forgedAlliancePrefs.setWarnNonAsciiVaultPath(false);
+              preferencesService.storeInBackground();
+            })));
+      }
+    }
```
Sheikah45: I would move this to an event listener in the PreferencesService that is triggered on a log in event.

Also probably would be better to have a modal dialog like an immediate notification and option to change the path like we do with the game path. As persistent notifications are pretty minor and can be overlooked.

Sheikah45 (changes requested):
The check shouldn't really be in the afterProperties set as that is during the initialization of the spring beans so there might be some weird interactions. I would make it listen for the login event and check after that. We should also make it more prevalent than a persistent notification so that users know it can cause more issues as persistent notifications can be ignored easily.

Similar to the game path dialog also would be nice to provide a way to change it there.

src/main/java/com/faforever/client/game/GameService.java
Marc-Spector: Are you sure? A newbie running the client for the first time after logging in will see this dialog instead of getting to know the client?

Sheikah45: Yes. I think it is most important for people new to the client just like how we ask them to specify the installed game path first thing.

Since it can be a hard bug to track down that can cause launches to fail.

Sheikah45 (commented):
Looks good to me is this done from your perspective?

Marc-Spector:
I did as you asked (requested). Either I misunderstood your thoughts. :(

Sheikah45:
No this is good you just didn't request a review so wasn't sure if it was done from your perspective

Sheikah45:
Just wanted to make sure I didn't merge before you were finished

Marc-Spector:
I need time to make sure that the pull request is working perfectly. I need to test all things because we don't have integration tests. I always request a review only when I consider it finished)

Sheikah45:
Sounds good then I shall wait but the structure is good.

src/main/java/com/faforever/client/preferences/ui/SettingsController.java
```diff
@@ -184,43 +186,10 @@ public class SettingsController implements Controller<Node> {
...
-            controller.setLocale(locale);
-            controller.setOnSelectedListener(this::onLanguageSelected);
-            controller.setSelected(locale.equals(currentLocale));
-            return controller.getRoot();
-          })
-          .collect(Collectors.toList());
-      languagesContainer.getChildren().setAll(nodes);
-    };
-  }
+  private InvalidationListener availableLanguagesListener;
```
Sheikah45, line 189: Small nit is just that even though it doesn't allow for a nice lombok constructor this should probably still be final but not gonna hold up this PR because of that

Sheikah45 (approved):

PR #2759, opened 2022-07-30
WIP: Fix inability to run old coop replays

### LINKED ISSUE #2747: coop replay failures in latest update
**Describe the bug**
image

loading prothyon coop map replay from leaderboard.  

**To Reproduce**
linux platform, steam installation, launf faf, launch coop replay from net. 

**OS**
Linux

Marc-Spector:
Closes #2747 
more info in https://github.com/FAForever/downlords-faf-client/issues/2747#issuecomment-1200133251

### DIFF
--- a/src/main/java/com/faforever/client/patch/GameUpdaterImpl.java
+++ b/src/main/java/com/faforever/client/patch/GameUpdaterImpl.java
@@ -69,7 +69,7 @@ public class GameUpdaterImpl implements GameUpdater {
     if (!NAMES_OF_FEATURED_BASE_MODS.contains(featuredMod.getTechnicalName())) {
       featuredModUpdateFuture = simModsUpdateFuture.thenCompose(aVoid -> modService.getFeaturedMod(FAF.getTechnicalName()))
           .thenCompose(baseMod -> updateFeaturedMod(baseMod, null))
-          .thenCompose(patchResult -> updateGameBinaries(patchResult.getVersion()))
+          .thenCompose(patchResult -> updateGameBinaries(new ComparableVersion(version.toString())))
           .thenCompose(aVoid -> updateFeaturedMod(featuredMod, version));
     } else {
       featuredModUpdateFuture = simModsUpdateFuture.thenCompose(aVoid -> updateFeaturedMod(featuredMod, version))


Marc-Spector:
I need to test that old or current replays with featured mods don't break

Marc-Spector:
But no. They are broken.

src/main/java/com/faforever/client/patch/GameUpdaterImpl.java
```diff
@@ -69,7 +69,7 @@ public CompletableFuture<Void> update(FeaturedModBean featuredMod, Integer versi
     if (!NAMES_OF_FEATURED_BASE_MODS.contains(featuredMod.getTechnicalName())) {
       featuredModUpdateFuture = simModsUpdateFuture.thenCompose(aVoid -> modService.getFeaturedMod(FAF.getTechnicalName()))
           .thenCompose(baseMod -> updateFeaturedMod(baseMod, null))
-          .thenCompose(patchResult -> updateGameBinaries(patchResult.getVersion()))
+          .thenCompose(patchResult -> updateGameBinaries(new ComparableVersion(version.toString())))
```
Sheikah45, line 72: This won't work because this is just the binary. The lua files are still the wrong version as indicated by the updateFeaturedMod above. Really the version should go where that null is but I forget if there is a reason why it is currently null

Sheikah45: Oh now I remember why it is because allegedly the version corresponds to the version of the featured mod and not necessarily the base game. At least at the time this was written. So there wasn't really any info on the base game version so just always set it to latest. But that could have been misinterpreted and we can use it for the base game version

Sheikah45: But i think now this version number always corresponds to the base faf version but I don't know where the featuredmod version should be gotten from anymore. There is a mapping of numbers to version number in the replay header but I am unable to find the mapping itself to which featuredmod those correspond to

Sheikah45: Never mind I realized those correspond to the individual files. I will make a new pr to fix it all

Sheikah45:
Superceded by https://github.com/FAForever/downlords-faf-client/pull/2760

PR #2761, opened 2022-08-01
Set process priority to high on Windows

### LINKED ISSUE #1965: Start game with high process priority
Reduces interference of other programs.

Maybe make it configurable in case this causes problems for someone.

micheljung:
Fixes #1965


src/main/java/com/faforever/client/fa/ProcessUtils.java
```diff
@@ -0,0 +1,45 @@
...
+    HANDLE hProcess = getProcessHandle(process);
+    DWORD dwPriorityClass = priority.dword();
+    BOOL success = Kernel32Ex.INSTANCE.SetPriorityClass(hProcess, dwPriorityClass);
+    if (!success.booleanValue()) {
+      log.warn("Could not set priority of process {}", process.pid());
+    }
+  }
+
+  @SneakyThrows
+  private HANDLE getProcessHandle(Process process) {
```
Sheikah45: Rather than sneaky throws can this catch and either log the exception and cancel setting the priority or catch and rethrow  using some subclass of MajorNotifiableException so that there is a better message given to the user for the error.

Right now if this errors due to some systems permission or something else the user will not have a clear idea what went wrong. And it is likely it may fail on some systems due to security settings. So would be good to either move forward without setting the priority or let the user know with a more informative message.

It would also make it easier to debug so that I know if anyone has an issue here rather than having to dig through the logs they give me

Sheikah45 (changes requested):

src/main/java/com/faforever/client/fa/ProcessUtils.java
micheljung: This code can only throw an exception if `ProcessImpl.handle` doesn't exist anymore or if `--add-opens=java.base/java.lang=ALL-UNNAMED` is not set. System permissions would lead to `BOOL success` being false which is already covered. But your point is valid, I refactored it.

src/main/java/com/faforever/client/fa/ProcessUtils.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  private HANDLE getProcessHandle(Process process) throws Exception {
+    Field f = process.getClass().getDeclaredField("handle");
+    f.setAccessible(true);
+    long handle = f.getLong(process);
+
+    WinNT.HANDLE hProcess = new WinNT.HANDLE();
+    hProcess.setPointer(Pointer.createConstant(handle));
+    return hProcess;
```
Sheikah45, line 46: Also is the handle just the pid? Why not just use the toHandle() method on the process object to get a ProcessHandle?

micheljung: Handle is different from the PID. I searched for a way to get the handle for a PID but Reflection is the only way I found to get the process handle. AFAIK there are also permissions attached to a handle so it makes sense that you can't get handles by PIDs.

micheljung:
@Sheikah45 can the conversations be resolved?

src/main/java/com/faforever/client/fa/ProcessUtils.java
micheljung: btw. `ProcessHandle` is a Java API and doesn't allow you to get the native process handle.

Sheikah45:
Is this all set?

micheljung:
I gave you my responses so, yes

PR #2769, opened 2022-08-12
New filter system

### LINKED ISSUE #1946: Custom filter for displaying lobbies in play tab

### LINKED ISSUE #2767: Add filters in `Live Games` tab
- `Coop` games
- Modded games
- Matchmaker
- Custom games

Marc-Spector:
Closes #2767 
Closes #1946


Marc-Spector:
I removed `LiveReplayControllerTest` class because there is nothing to test there

src/main/java/com/faforever/client/vault/replay/LiveReplayFilterController.java
```diff
@@ -0,0 +1,77 @@
...
+    return root;
+  }
+
+  public void setGameList(FilteredList<GameBean> gameList) {
+    this.gameList = gameList;
+    JavaFxUtil.addListener(customGamesCheckbox.selectedProperty(), new WeakInvalidationListener(selectedCheckboxListener));
+    JavaFxUtil.addListener(matchmakerCheckbox.selectedProperty(), new WeakInvalidationListener(selectedCheckboxListener));
+    JavaFxUtil.addListener(coopGamesCheckbox.selectedProperty(), new WeakInvalidationListener(selectedCheckboxListener));
+    JavaFxUtil.addAndTriggerListener(moddedGamesCheckbox.selectedProperty(), new WeakInvalidationListener(selectedCheckboxListener));
+  }
```
Sheikah45: Rather than this taking in a game list can we have it just have a getPredicate() method and then also the ability to listen to changes in the predicate so that the class using it still controls the list.

Also we can just name this class GameFilterController since we could also use this other places.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/vault/replay/LiveReplayFilterController.java
Marc-Spector: I am having trouble with that. I can not figure out how to do it. Could you give a pseudocode?

Sheikah45: Could be something like this
```
class GameFilterController {
    private ObjectProperty<Predicate<GameBean>> predicateProperty;

    public void initalize() {
      predicateProperty.setValue(game -> {
        boolean playing = game.getStatus() == GameStatus.PLAYING;
        boolean custom = customGamesCheckbox.isSelected() || game.getGameType() != GameType.CUSTOM;
        boolean matchmaker = matchmakerCheckbox.isSelected() || game.getGameType() != GameType.MATCHMAKER;
        boolean coopGame = coopGamesCheckbox.isSelected() || game.getGameType() != GameType.COOP;
        boolean moddedGame = moddedGamesCheckbox.isSelected() || game.getSimMods().isEmpty();
        return playing && custom && matchmaker && coopGame && moddedGame;
      });

      public ObjectProperty<> getPredicateProperty() {
        return predicateProperty;
      }

      public Predicate<> getPredicate() {
        return predicateProperty.get();
      }
}

Where used

gameList.setPredicate(gameFilterController.getPredicate());
```

Here you don't even really need the listener beyond setting if the filter is actually being applied or not. But here you would just get the value of the predicate from the controller and could apply it to whatever game list. Could even use it in multiple places.

This would also be easy to extend for things like #1946. Also the predicateProperty is only needed if the predicate actually changes in this case nothing will change the predicate but could make extension easy.

Marc-Spector: > Here you don't even really need the listener

Sorry for my stupidity.Then how do I tell the list to start filtering items when the checkbox changes its state without `predicate` listener?

Marc-Spector: Wait...

Sheikah45: Yeah unless I misunderstand how the predicate works I think it should automatically pick up the change. If it doesn't then, you can add a filterStateChange property and listen to that to refresh the predicate

Marc-Spector: > I think it should automatically pick up the change

Nope. It doesn't work like that

Marc-Spector: > This would also be easy to extend for things like https://github.com/FAForever/downlords-faf-client/issues/1946. Also the predicateProperty is only needed if the predicate actually changes in this case nothing will change the predicate but could make extension easy.

OK. I will try to make the class extensible.

Sheikah45: > > I think it should automatically pick up the change
> 
> Nope. It doesn't work like that

Hmm unfortunate but in that case just add a predicate changed listener or just update the predicate property value and attach a listener to that and you can detect when the predicate is changed

Sheikah45:
Looking good with the structure. One thing I don't quite understand though is what are the FilterNames used for? And what is the idea behind the primary and secondary filters?

Marc-Spector:
> what are the FilterNames used for?

Filter names are necessary because different places use a specific set of filters where the same data type will be filtered. They will also be displayed in the order you specify.
For example:
In custom games (used type: game bean) we will use following set of filters:
`PRIVATE_GAME, WITH_MODS, FEATURED_MOD, PLAYER_NAME, ...`
In live replays (same type: game bean) we will use another set of filter: 
`GAME_TYPE, FEATURED_MOD, MAP_NAME, PLAYER_NAME, ...`

The feature is that you don't need to duplicate an extra class to create a new set of filters of the same type.

> what is the idea behind the primary and secondary filters?

Primary filters are always visible, we filter some properties that are often used.
Secondary filters are initially hidden, they can be displayed or hidden using the button.
I just implemented it, because if there are a lot of filters in one place, then you can hide some filters by making them secondary

Sheikah45:
Got it so the names are mostly there to differentiate between the primary and secondary filters.

Also we should make sure to differentiate these filters enough from the filters in the query package that are used for the api. As things are named right now they might be easy to confuse, or question why both exist.

Marc-Spector:
I did not complicate everything and removed the concepts of "primary" and "secondary" filters. All the specified filters are always visible

Marc-Spector:
I think it would be nice to add a counter of how many elements are hidden for each filter. I can make it optional.

This complicates the code. I won't do it

Marc-Spector:
I will also replace `Show` word with`Hide` in the `Show private games` and `Show modded games` filters to maintain consistency in other places

Marc-Spector:
I did not expect the PR to be big. :)

Sheikah45 (changes requested):
I will review this more in depth over the weekend but had some high level thoughts.

For the country filters is there anyway you think to limit the countries shown only to those where players are actually in the list?

I noticed that the filtering is all done in the application thread I wonder if there is a way to move some of the heavy lifting the a background thread.

I am still not convinced of the need for the FilterNames especially since it looks like each class that uses them creates a custom implementation for them anyway so it isn't like there is any unity between them especially with the filterBuilder which makes it easier. As it is now I would assume that any filter can take any of the filter names but that isn't the case. Because of that I think it might be better to just construct the filters in place and add them without the filter names but I could be convinced otherwise if there is a good reason I am not seeing. And at most it looks like you are just using the names to bind the preferences which can still be done at construction. Also I think it is just better for these filter controller to be a bit more static since the pieces they are made of are very modular to anyway.

And yeah in depth PRs like this one have a way of growing ha XD

Also for records I would just make the closing bracket on the same line

src/main/java/com/faforever/client/chat/ChatUserListController.java
```diff
@@ -435,6 +436,7 @@ List<ChatChannelUser> getUserList() throws Exception {
 
   @VisibleForTesting
   void waitForUsersEvent() throws Exception {
-    usersEventQueueExecutor.submit(() -> {  }).get();
+    usersEventQueueExecutor.submit(() -> {
+    }).get();
```
Sheikah45: one line please

src/main/java/com/faforever/client/filter/AbstractFilterController.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  public boolean getFilterState() {
+    return filterStateProperty.get();
+  }
+
+  public void setDefaultPredicate(Predicate<T> defaultPredicate) {
+    this.defaultPredicate = defaultPredicate;
+    predicateProperty.setValue(defaultPredicate);
+  }
```
Sheikah45, line 79: Maybe check if the current predicate is the old default before overwriting

src/main/java/com/faforever/client/filter/AbstractFilterController.java
```diff
@@ -0,0 +1,120 @@
...
+
+  public void completeSetting() {
+    setFilterContent();
+    filters.forEach(filter -> JavaFxUtil.addListener(filter.getPredicateProperty(), observable -> invalidated()));
+    externalFilters.addListener((InvalidationListener) observable -> invalidated());
+    invalidated();
+  }
+
+  private synchronized void invalidated() {
```
Sheikah45: would rename this to maybe just invalidate or a more specific action since it is doing something rather than a property.

src/main/java/com/faforever/client/filter/AbstractFilterController.java
```diff
@@ -0,0 +1,120 @@
...
+  public VBox filtersContent;
+  public Button resetAllButton;
+
+  protected final UiService uiService;
+  protected final I18n i18n;
+
+  private final ObservableMap<Property<?>, Predicate<T>> externalFilters = FXCollections.observableHashMap();
+  private final List<AbstractFilterNodeController<?, ? extends ObservableValue<?>, T>> filters = new ArrayList<>();
+  private Predicate<T> defaultPredicate = t -> true;
+  private boolean onResetAllTriggered = false;
```
Sheikah45: Looks like you use this to detect when a reset is occuring so might name it something like resetInProgress to be more clear

src/main/java/com/faforever/client/filter/AbstractFilterNodeController.java
```diff
@@ -0,0 +1,36 @@
...
+
+  protected abstract U getValue();
+
+  public void registerListener(BiFunction<U, T, Boolean> filter) {
+    JavaFxUtil.addAndTriggerListener(getObservable(), observable -> predicate.set(item -> filter.apply(getValue(), item)));
+  }
+
+  public ObjectProperty<Predicate<T>> getPredicateProperty() {
+    return predicate;
+  }
```
Sheikah45:
```suggestion
  public ObjectProperty<Predicate<T>> predicateProperty() {
    return predicate;
  }
```

Fits better with the other property naming conventions

src/main/java/com/faforever/client/filter/ChatUserFilterController.java
```diff
@@ -0,0 +1,115 @@
...
+@Component
+@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+public class ChatUserFilterController extends AbstractFilterController<ListItem> {
+
+  private final CountryFlagService countryFlagService;
+  private final LeaderboardService leaderboardService;
+
+  private final static int MIN_RATING = -1000;
+  private final static int MAX_RATING = 4000;
+
```
Sheikah45: We should probably standardize these across the replay filter and here

src/main/java/com/faforever/client/filter/ChatUserFilterController.java
```diff
@@ -0,0 +1,115 @@
...
+    public String toString(LeaderboardBean object) {
+      String rating = i18n.getOrDefault(object.getTechnicalName(), object.getNameKey());
+      return i18n.get("leaderboard.rating", rating);
+    }
+
+    @Override
+    public LeaderboardBean fromString(String string) {
+      throw new UnsupportedOperationException("Not supported");
+    }
+  };
```
Sheikah45, line 68: Just a comment but I feel like I should really just write a functioninterface that is a one way string converter for how much we use this pattern

src/main/java/com/faforever/client/filter/CustomGamesFilterController.java
```diff
@@ -0,0 +1,56 @@
...
+
+    filterBuilder.multiCheckbox(i18n.get("featuredMod.displayName"), modService.getFeaturedMods(),
+        new FeaturedModConverter(), new FeaturedModFilterFunction());
+
+    mapFolderNameBlackListFilter = filterBuilder.mutableList(i18n.get("blacklist.mapFolderName"), i18n.get("blacklist.mapFolderName.promptText"),
+        (folderNames, game) -> folderNames.isEmpty() || folderNames.stream()
+            .noneMatch(name -> game.getMapFolderName().contains(name)));
+  }
+
+  public BooleanProperty getPrivateGamesProperty() {
```
Sheikah45:
```suggestion
  public BooleanProperty privateGamesProperty() {
```

src/main/java/com/faforever/client/filter/CustomGamesFilterController.java
```diff
@@ -0,0 +1,56 @@
...
+    return privateGameFilter.getObservable();
+  }
+
+  public BooleanProperty getSimsModsProperty() {
```
Sheikah45:
```suggestion
  public BooleanProperty simModsProperty() {
```

src/main/java/com/faforever/client/filter/CustomGamesFilterController.java
```diff
@@ -0,0 +1,56 @@
...
+    return simModsFilter.getObservable();
+  }
+
+  public ListProperty<String> getMapFolderNameBlackListProperty() {
```
Sheikah45:
```suggestion
  public ListProperty<String> mapFolderNameBlackListProperty() {
```

src/main/java/com/faforever/client/filter/FilterBuilder.java
```diff
@@ -0,0 +1,85 @@
...
+    controller.setText(text);
+    controller.setMinValue(minValue);
+    controller.setMaxValue(maxValue);
+    controller.registerListener(filter);
+    onFilterBuilt.accept(controller);
+    return controller;
+  }
+
+  public <I> RangeSliderWithChoiceFilterController<I, T> rangeSliderWithCombobox(String text, CompletableFuture<List<I>> future, StringConverter<I> converter, double minValue, double maxValue, BiFunction<ItemWithRange<I, Integer>, T, Boolean> filter) {
+    RangeSliderWithChoiceFilterController<I, T> controller = uiService.loadFxml("theme/filter/range_slider_filter.fxml", RangeSliderWithChoiceFilterController.class);
```
Sheikah45, line 63: I don't particularly like that this takes a future as input but I see why you did and can't think necessarily of a better way besides blocking the construction of the filter which might be better

src/main/java/com/faforever/client/filter/FilterMultiCheckboxController.java
```diff
@@ -0,0 +1,177 @@
...
+
+  private final I18n i18n;
+
+  public MenuButton root;
+  public VBox contentVBox;
+  public TextField searchTextField;
+  public CheckListView<String> listView;
+
+  private final ObservableList<U> selectedItems = FXCollections.observableArrayList();
+  private final ListProperty<U> property = new SimpleListProperty<>(selectedItems);
```
Sheikah45:
```suggestion
  private final ObservableList<U> selectedItems = FXCollections.observableArrayList();
  private final ListProperty<U> selectedItemList = new SimpleListProperty<>(selectedItems);
```

Maybe a better name?

src/main/java/com/faforever/client/filter/FilterMultiCheckboxController.java
```diff
@@ -0,0 +1,177 @@
...
+        String.join(", ", listView.getCheckModel().getCheckedItems())), listView.getCheckModel().getCheckedItems()));
+  }
+
+  private void addSearchBar() {
+    searchTextField = TextFields.createClearableTextField();
+    searchTextField.setPromptText(i18n.get("chat.searchPrompt"));
+    contentVBox.getChildren().add(0, searchTextField);
+  }
+
+  private void invalidated(Change<? extends String> change) {
```
Sheikah45:
```suggestion
  private void invalidate(Change<? extends String> change) {
```

src/main/java/com/faforever/client/filter/FilterMultiCheckboxController.java
```diff
@@ -0,0 +1,177 @@
...
+        String uncheckedItem = change.getRemoved().get(0);
+        if (selectedStringItems.contains(uncheckedItem)) {
+          selectedStringItems.remove(uncheckedItem);
+          selectedItems.remove(sourceList.get(convertedSourceList.indexOf(uncheckedItem)));
+        }
+      }
+    }
+  }
+
+  private void invalidated(List<Integer> selectedIndices) {
```
Sheikah45:
```suggestion
  private void invalidate(List<Integer> selectedIndices) {
```
Visible for testing should only need package private

src/main/java/com/faforever/client/filter/FilterMultiCheckboxController.java
```diff
@@ -0,0 +1,177 @@
...
+  private String convertToString(U item) {
+    return converter != null ? converter.toString(item) : item.toString();
+  }
+
+  private List<String> convertToStringItems(List<U> items) {
+    return items.stream().map(this::convertToString).toList();
+  }
+
+  @VisibleForTesting
+  protected List<U> getSelectedItems() {
```
Sheikah45:
```suggestion
  List<U> getSelectedItems() {
```

src/main/java/com/faforever/client/filter/MutableListFilterController.java
```diff
@@ -0,0 +1,96 @@
...
+
+  private final UiService uiService;
+  private final I18n i18n;
+
+  public MenuButton root;
+  public ListView<String> listView;
+  public TextField addItemTextField;
+
+  private final ObservableList<String> items = FXCollections.observableArrayList();
+  private final ListProperty<String> property = new SimpleListProperty<>(items);
```
Sheikah45: clearer name needed

src/main/java/com/faforever/client/fx/JavaFxUtil.java
```diff
@@ -488,4 +490,29 @@ public static void runLater(Runnable runnable) {
   public static void bindManagedToVisible(Node... nodes) {
     Arrays.stream(nodes).forEach(node -> node.managedProperty().bind(node.visibleProperty()));
   }
+
+  public static void bindTextFieldAndRangeSlide(TextField textField, RangeSlider rangeSlider, boolean highValue) {
```
Sheikah45: Rather than a boolen highValue just pass both the high and lo value text fields
```suggestion
  public static void bindTextFieldAndRangeSlide(TextField lowValueTextField, TextField highValueTextField, RangeSlider rangeSlider) {
```

Sheikah45 (changes requested):
Overall looking good just a few minor things.

Also I had changed the name of some methods but then realized they were overriding abstract ones. Thought I removed all those comments but you can ignore them if I forgot to remove them.

Marc-Spector:
>I am still not convinced of the need for the FilterNames especially since it looks like each class that uses them creates a custom implementation for them anyway so it isn't like there is any unity between them especially with the filterBuilder which makes it easier. As it is now I would assume that any filter can take any of the filter names but that isn't the case. Because of that I think it might be better to just construct the filters in place and add them without the filter names but I could be convinced otherwise if there is a good reason I am not seeing. And at most it looks like you are just using the names to bind the preferences which can still be done at construction. Also I think it is just better for these filter controller to be a bit more static since the pieces they are made of are very modular to anyway.

Done.

> For the country filters is there anyway you think to limit the countries shown only to those where players are actually in the list?

I created a local class. He works. I will add it.

> I noticed that the filtering is all done in the application thread I wonder if there is a way to move some of the heavy lifting the a background thread.

Could you point out where  processing the heavy code?

To be continued...

src/main/java/com/faforever/client/filter/CustomGamesFilterController.java
Marc-Spector: Replaced with `bindBidirectionalTo...(Property)`

src/main/java/com/faforever/client/filter/CustomGamesFilterController.java
Marc-Spector: Replaced with `bindBidirectionalTo...(Property)`

src/main/java/com/faforever/client/filter/CustomGamesFilterController.java
Marc-Spector: Replaced with `bindBidirectionalTo...(Property)`

src/main/java/com/faforever/client/filter/ChatUserFilterController.java
Marc-Spector: They are needed for `view.setConverter(StringConverter)` method. The functional interface is not suitable.

src/main/java/com/faforever/client/filter/ChatUserFilterController.java
Marc-Spector: Sorry, I do not understand what you mean by `replay filter`?

src/main/java/com/faforever/client/filter/AbstractFilterController.java
Marc-Spector: There is no need. I do not see that it is possible to call the method twice.

src/main/java/com/faforever/client/filter/FilterBuilder.java
Marc-Spector: Uhhh, I do not want to redo the class for the nth time. :D

src/main/java/com/faforever/client/filter/ChatUserFilterController.java
Marc-Spector: https://github.com/FAForever/downlords-faf-client/blob/b79be72f1852f8c0da635d3cd57e49acc67adbd7/src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java#L144-L146
Do you want to replace these values?

Sheikah45: yes just for consistency in terms of what we consider the max and min rating

Sheikah45:
> > I noticed that the filtering is all done in the application thread I wonder if there is a way to move some of the heavy lifting the a background thread.
> 
> Could you point out where processing the heavy code?
> 
> To be continued...

I can't find it anymore so it is all good.

Do you have anything more you wanted to do with this?

Sheikah45 (approved):

Marc-Spector:
> For the country filters is there anyway you think to limit the countries shown only to those where players are actually in the list?

The only thing I did not have time to do. I can do this in a separate PR in the future.

Sheikah45:
> > For the country filters is there anyway you think to limit the countries shown only to those where players are actually in the list?
> 
> The only thing I did not have time to do. I can do this in a separate PR in the future.

Works for me

BlackYps:
Great work on this @Marc-Spector, I love it!

PR #2773, opened 2022-08-14
Update game before queue and add better game start error logging

### LINKED ISSUE #2245: Game crashes when matchmaker finds game while game update not done
When there is a new game update, and you queue for matchmaker, and the update doesnt fully get applied before the matchmaker starts, the game itself just crashes because its started while its files are updated.

Sheikah45:
Fixes #2245

### DIFF
--- a/src/main/java/com/faforever/client/coop/CoopController.java
+++ b/src/main/java/com/faforever/client/coop/CoopController.java
@@ -3,6 +3,7 @@ package com.faforever.client.coop;
 import com.faforever.client.domain.CoopMissionBean;
 import com.faforever.client.domain.CoopResultBean;
 import com.faforever.client.domain.GameBean;
+import com.faforever.client.exception.NotifiableException;
 import com.faforever.client.fx.AbstractViewController;
 import com.faforever.client.fx.JavaFxUtil;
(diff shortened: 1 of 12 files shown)


Katharsas:
Nice!
(btw. the autoformatter misformatted a bunch of lines, see https://github.com/FAForever/downlords-faf-client/pull/2773/files#diff-f5937835360916cbea30c94859678e3b9f008621c5a10476720a4e6d47a1f316R121)

Sheikah45:
> Nice! (btw. the autoformatter misformatted a bunch of lines, see [#2773 (files)](https://github.com/FAForever/downlords-faf-client/pull/2773/files#diff-f5937835360916cbea30c94859678e3b9f008621c5a10476720a4e6d47a1f316R121))

That is partially how it is supposed to be somewhat. It prioritizes chopping down the chained method calls

Katharsas:
I know ("misformat" might have been the wrong word, but its definitely horrible to read now).
Maybe putting each condition in parenthesis would prevent it being cut appart like that but not sure.

This is why i generally deactivate reorganization of line breaks for single statements in my own project's formatter settings. Haven't seen a Java formatter yet that is able to always choose the linebreaks that make the logic as clear as possible.

PR #2790, opened 2022-09-07
#2789 change default data location

### LINKED ISSUE #2789: Use AppData\Roaming folder instead of ProgramData as default Data Location
Pros:
- No mess with access permissions
- No warning about ProgramData if running as admin
- No space use of C: drive if users folder moved to another drive

Cons:
- Data will not be shared between multiple users on same system

Or maybe we should use %user%/.faforever as on linux?

vladislemon:
Fixes #2789


Sheikah45:
As I mentioned in the linked issue we don't really want defaults under the user directory.  Also the admin check is not only about program data. In general the client should not be run as admin.

vladislemon:
@Sheikah45 you're right.

PR #2793, opened 2022-09-11
Send division name to game on start for matchmaker games

BlackYps:
When starting a matchmaker game this sends the name of the division the player has for this leaderboard to the game.
This enables us to display the division instead of rating in the game score board, bringing us closer to establishing the league system as the default skill representation.

### DIFF
--- a/src/main/java/com/faforever/client/fa/ForgedAllianceService.java
+++ b/src/main/java/com/faforever/client/fa/ForgedAllianceService.java
@@ -3,11 +3,9 @@ package com.faforever.client.fa;
 import com.faforever.client.domain.LeaderboardRatingBean;
 import com.faforever.client.domain.PlayerBean;
 import com.faforever.client.fa.Kernel32Ex.WindowsPriority;
-import com.faforever.client.leaderboard.LeaderboardService;
 import com.faforever.client.logging.LoggingService;
 import com.faforever.client.player.PlayerService;
 import com.faforever.client.preferences.PreferencesService;
-import com.faforever.commons.lobby.GameLaunchResponse;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.jetbrains.annotations.NotNull;
@@ -38,7 +36,6 @@ public class ForgedAllianceService {
 
   public static final String DEBUGGER_EXE = "FAFDebugger.exe";
 
-  private final LeaderboardService leaderboardService;
   private final PlayerService playerService;
   private final PreferencesService preferencesService;
   private final LoggingService loggingService;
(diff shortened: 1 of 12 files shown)


src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -591,7 +595,7 @@ public boolean isInMatchmakerQueue() {
    * Actually starts the game, including relay and replay server. Call this method when everything else is prepared
    * (mod/map download, connectivity check etc.)
    */
-  private CompletableFuture<Void> startGame(GameLaunchResponse gameLaunchMessage) {
+  private CompletableFuture<Void> startGame(GameLaunchResponse gameLaunchMessage, boolean isMatchmakerGame) {
```
Sheikah45: This boolean is unecessary the game launch message contains the rating_type which is the leader board to be used, we already use that to retrieve the correct rating anyway

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -609,11 +613,25 @@ private CompletableFuture<Void> startGame(GameLaunchResponse gameLaunchMessage)
...
+                      bean.getSubdivision().getDivision().getNameKey() + " " + bean.getSubdivision().getNameKey())
+                  .orElse("unlisted");
+              try {
+                process = forgedAllianceService.startGameOnlineWithDivision(gameLaunchMessage,
+                    adapterPort, localReplayPort, rehostRequested, divisionName);
+              } catch (IOException e) {
+                throw new GameLaunchException("Could not start game", e, "game.start.couldNotStart");
+              }
+            });
+          } else {
```
Sheikah45: Since this is only specific to matchmaker it should go in the startMatchmaker function not the generic start game function

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -179,6 +179,13 @@ public CompletableFuture<Optional<LeagueEntryBean>> getActiveLeagueEntryForPlaye
             .findFirst());
   }
 
+  public CompletableFuture<Optional<LeagueEntryBean>> getActiveLeagueEntryForPlayer(PlayerBean player, String leaderboardName) {
+    return getActiveLeagueEntriesForPlayer(player)
+        .thenApply(leagueEntries -> leagueEntries.stream()
+            .filter(leagueEntry -> Objects.equals(leagueEntry.getLeagueSeason().getLeaderboard().getTechnicalName(), leaderboardName))
+            .findFirst());
+  }
```
Sheikah45: Can't we filter by leaderboard in the query? That is what this call should do rather than filter java side

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -609,11 +613,25 @@ private CompletableFuture<Void> startGame(GameLaunchResponse gameLaunchMessage)
...
-          try {
-            process = forgedAllianceService.startGameOnline(gameLaunchMessage,
-                adapterPort, localReplayPort, rehostRequested);
-          } catch (IOException e) {
-            throw new GameLaunchException("Could not start game", e, "game.start.couldNotStart");
+          if (isMatchmakerGame) {
+            leaderboardService.getActiveLeagueEntryForPlayer(playerService.getCurrentPlayer(), gameLaunchMessage.getLeaderboard()).thenAccept(leagueBeanOptional -> {
+              String divisionName = leagueBeanOptional.map(bean ->
+                      bean.getSubdivision().getDivision().getNameKey() + " " + bean.getSubdivision().getNameKey())
+                  .orElse("unlisted");
```
Sheikah45: are passing the nameKey or the name? Seems like this would have to go through i18n somewhere

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
```diff
@@ -51,6 +53,18 @@ public Process startGameOffline(String map) throws IOException {
...
+    LaunchCommandBuilder launchCommandBuilder = prepareLaunchCommand(gameLaunchMessage, gpgPort, localReplayPort, rehost);
+
+    return launch(launchCommandBuilder.build());
+  }
+
+  public Process startGameOnlineWithDivision(GameLaunchResponse gameLaunchMessage, int gpgPort, int localReplayPort, boolean rehost, String divisionName) throws IOException {
+    LaunchCommandBuilder launchCommandBuilder = prepareLaunchCommand(gameLaunchMessage, gpgPort, localReplayPort, rehost);
+
+    return launch(launchCommandBuilder.division(divisionName).build());
+  }
```
Sheikah45: I wouldn't make a new function and name. Ideally the info would be contained in the gameLaunchMessage. Maybe it might be better to have a new POJO that holds all the information necessary for launching the game that is populated in the game service, rather than using the DTO object.

Sheikah45 (changes requested):
As we add more info to the game launch it is probably better to create a new POJO that represents an enhanced GameLaunchResponse. This allows us to decrease the parameters needed in startGameOnline. Also there was a lot of logic added to startGame in the gameService that is better fit for startMatchmaker

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
BlackYps: Maybe we can use the existing LaunchCommandBuilder for that? We can set information there in multiple steps if we pass the Builder and only call build() in the ForgedAllianceService when everything is ready. Then we don't need an entirely new class.

src/main/java/com/faforever/client/game/GameService.java
BlackYps: We're passing the name key here, so the game can then handle the localization.

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
Sheikah45: No I wouldn't use the builder for that. It's purpose is for creating the proper command and setting the variables. Creating a domain object fits the rest of the pattern as well with things like player->PlayerBean

src/main/java/com/faforever/client/game/GameService.java
Sheikah45: Also shouldn't they be different options then to make localization easier?

BlackYps:
I broke two tests with the latest changes. There seems to be a problem with mocking, as the game process is null when it shouldn't be, but I couldn't figure out what exactly the problem is.

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
BlackYps: I changed it, but why is that better?

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -172,11 +171,24 @@ public CompletableFuture<Optional<LeagueEntryBean>> getHighestActiveLeagueEntryF
...
+            .string("leagueSeason.leaderboard.technicalName").eq(leaderboardName)
+            .and()
+            .instant("leagueSeason.startDate").before(OffsetDateTime.now().toInstant(), false)
+            .and()
+            .instant("leagueSeason.endDate").after(OffsetDateTime.now().toInstant(), false)
+        );
+    return fafApiAccessor.getMany(navigator)
+        .next()
+        .map(dto -> leaderboardMapper.map(dto, player, new CycleAvoidingMappingContext()))
+        .filter(leagueEntryBean -> leagueEntryBean.getSubdivision() != null)
```
Sheikah45, line 189: Couldn't this be made into an isnull check on the filter

Sheikah45 (changes requested):

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
BlackYps: Do you mean like this? `.filter(leagueEntryBean -> Objects.nonNull(leagueEntryBean.getSubdivision()))`
Not sure if this really improves readability

Sheikah45: No you are filtering java side, when you could just add the null check to the API call

BlackYps: `.and().string("leagueSeasonDivisionSubdivision").exists()` doesn't work. It seems to break the api call.

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
```diff
@@ -231,6 +243,16 @@ public List<String> build() {
       command.add(String.valueOf(deviation));
     }
 
+    if (division != null) {
+      command.add("/division");
+      command.add(division);
+    }
+
+    if (StringUtils.hasText(subdivision)) {
```
Sheikah45: Why is this one not a null check?

Sheikah45: Also spring StringsUtils is marked more for internal use so better to use Apache StringUtils

BlackYps: The grandmaster division has an empty string as subdivision name

Sheikah45: Got it then I would still switch the library used away from the internal spring one

Sheikah45 (approved):

PR #2807, opened 2022-09-24
Tournament games

1-alex98:


1-alex98:
https://github.com/FAForever/server/issues/675

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -519,16 +519,32 @@ public CompletableFuture<Void> startSearchMatchmaker() {
       return matchmakerFuture;
     }
 
+    matchmakerFuture =  listenToServerInitiatedGame(FAF.getTechnicalName());
+    return matchmakerFuture;
+  }
+
+  public CompletableFuture<Void> startListeningToTournamentGame(String featuredModTechnicalName) {
```
Sheikah45:
```suggestion
  public CompletableFuture<Void> startListeningForTournamentGame(String featuredModTechnicalName) {
```

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -519,16 +519,32 @@ public CompletableFuture<Void> startSearchMatchmaker() {
...
+    if (isRunning()) {
+      log.info("Game is running, ignoring tournament search request");
+      notificationService.addImmediateWarnNotification("game.gameRunning");
+      return completedFuture(null);
+    }
+
+    return listenToServerInitiatedGame(featuredModTechnicalName);
+  }
+
+  private CompletableFuture<Void> listenToServerInitiatedGame(String featuredModTechnicalName) {
```
Sheikah45:
```suggestion
  private CompletableFuture<Void> listenForServerInitiatedGame(String featuredModTechnicalName) {
```

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -519,16 +519,32 @@ public CompletableFuture<Void> startSearchMatchmaker() {
...
     if (!preferencesService.isGamePathValid()) {
       CompletableFuture<Path> gameDirectoryFuture = postGameDirectoryChooseEvent();
       return gameDirectoryFuture.thenCompose(path -> startSearchMatchmaker());
     }
 
-    log.info("Matchmaking search has been started");
+    log.info("Listening to server made game has been started");
```
Sheikah45:
```suggestion
    log.info("Started listening for game launch message");
```

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -519,16 +519,32 @@ public CompletableFuture<Void> startSearchMatchmaker() {
...
 
-    matchmakerFuture = modService.getFeaturedMod(FAF.getTechnicalName())
+    final var gameLaunchMessageFuture = fafServerAccessor.getGameLaunchMessage();
+    final var matchFuture = modService.getFeaturedMod(featuredModTechnicalName)
```
Sheikah45: No vars please XD

Sheikah45: Also more generally why did the gameLaunchMessageFuture need to be pulled out?

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -539,25 +555,25 @@ public CompletableFuture<Void> startSearchMatchmaker() {
...
-          log.info("Matchmaking search has been cancelled");
+          log.info("Listening to server made game has been cancelled");
           if (isRunning()) {
             notificationService.addServerNotification(new ImmediateNotification(i18n.get("matchmaker.cancelled.title"), i18n.get("matchmaker.cancelled"), Severity.INFO));
             gameKilled = true;
             process.destroy();
           }
         } else {
-          log.warn("Matchmade game could not be started", throwable);
+          log.warn("Server initiated game could not be started", throwable);
```
Sheikah45:
```suggestion
          log.warn("Game could not be started", throwable);
```

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -173,6 +174,9 @@ public CompletableFuture<Optional<LeagueEntryBean>> getHighestActiveLeagueEntryF
 
   @Cacheable(value = CacheNames.LEAGUE_ENTRIES, sync = true)
   public CompletableFuture<Optional<LeagueEntryBean>> getActiveLeagueEntryForPlayer(PlayerBean player, String leaderboardName) {
+    if (Strings.isNullOrEmpty(leaderboardName)) {
```
Sheikah45, line 178: Just for consistency most other places in the Code use StringUtils from Apache Commons

src/main/java/com/faforever/client/remote/FafServerAccessor.java
```diff
@@ -212,7 +212,7 @@ public void requestMatchmakerInfo() {
     lobbyClient.requestMatchmakerInfo();
   }
 
-  public CompletableFuture<GameLaunchResponse> startSearchMatchmaker() {
+  public CompletableFuture<GameLaunchResponse> getGameLaunchMessage() {
```
Sheikah45:
```suggestion
  public CompletableFuture<GameLaunchResponse> getGameLaunchMessageFuture() {
```

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingController.java
```diff
@@ -312,7 +313,7 @@ public void onFactionButtonClicked() {
 
   private void selectFactionsBasedOnParty() {
     List<Faction> factions = teamMatchmakingService.getParty().getMembers().stream()
-        .filter(member -> member.getPlayer().getId() == player.getId())
+        .filter(member -> Objects.equals(member.getPlayer().getId(), player.getId()))
```
Sheikah45: just curious but was this really null at times?

src/main/java/com/faforever/client/ui/progress/RingProgressIndicator.java
```diff
@@ -0,0 +1,135 @@
...
+      STYLEABLES = Collections.unmodifiableList(styleables);
+    }
+  }
+
+  @Override
+  public List<CssMetaData<? extends Styleable, ?>> getControlCssMetaData() {
+    return StyleableProperties.STYLEABLES;
+  }
+
+  public StringConverter<Integer> getProgressLableStringConverter() {
```
Sheikah45:
```suggestion
  public StringConverter<Integer> getProgressLabelStringConverter() {
```

src/main/java/com/faforever/client/ui/progress/RingProgressIndicator.java
```diff
@@ -0,0 +1,135 @@
...
+    return progressLableStringConverter.get();
+  }
+
+  public ObjectProperty<StringConverter<Integer>> progressLableStringConverterProperty() {
```
Sheikah45:
```suggestion
  public ObjectProperty<StringConverter<Integer>> progressLabelStringConverterProperty() {
```

src/main/java/com/faforever/client/ui/progress/RingProgressIndicator.java
```diff
@@ -0,0 +1,135 @@
...
+    return progressLableStringConverter;
+  }
+
+  public void setProgressLableStringConverter(StringConverter<Integer> progressLableStringConverter) {
```
Sheikah45:
```suggestion
  public void setProgressLabelStringConverter(StringConverter<Integer> progressLableStringConverter) {
```

src/main/java/com/faforever/client/ui/progress/RingProgressIndicator.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.List;
+
+
+/**
+ * Progress indicator showing a filling arc.
+ *
+ * @author Andrea Vacondio
+ */
+public class RingProgressIndicator extends ProgressCircleIndicator {
+  public ObjectProperty<StringConverter<Integer>> progressLableStringConverter = new SimpleObjectProperty<>(new StringConverter<>() {
```
Sheikah45:
```suggestion
  public ObjectProperty<StringConverter<Integer>> progressLabelStringConverter = new SimpleObjectProperty<>(new StringConverter<>() {
```

src/main/java/com/faforever/client/tournament/game/IsReadyForGameController.java
```diff
@@ -0,0 +1,101 @@
...
+  @Setter
+  private Runnable dismissCallBack;
+
+
+  @Override
+  public void initialize() {
+    progressIndicator.setProgressLableStringConverter(new StringConverter<>() {
+      @Override
+      public String toString(Integer object) {
+        return i18n.number(timeLeft);
```
Sheikah45, line 52: I find it a little odd that this is not just using the variable passed in. Why pass timeLeft directly here instead of setting the value of the progessIndicator

src/main/java/com/faforever/client/tournament/game/IsReadyForGameController.java
```diff
@@ -0,0 +1,101 @@
...
+    this.timeLeft = (int) (responseTimeSeconds - timeGone.toSeconds());
+    progressIndicator.setProgress((int) (percent * 100));
+    if (timeLeft <= 0 && queuePopTimeUpdater != null) {
+      end();
+    }
+  }
+
+  private void end() {
+    queuePopTimeUpdater.stop();
+    if (isReadyButton.isDisable()) {
```
Sheikah45: Might be more robust to have a dedicated ready boolean rather than checking the disabled property, or even using a checkbox rather than a button

src/main/java/com/faforever/client/tournament/game/IsReadyForGameController.java
```diff
@@ -0,0 +1,101 @@
...
+    OffsetDateTime start = OffsetDateTime.now();
+
+    queuePopTimeUpdater = new Timeline(1, new KeyFrame(javafx.util.Duration.seconds(0), (ActionEvent event) -> {
+      updateQueue(responseTimeSeconds, start);
+    }), new KeyFrame(javafx.util.Duration.seconds(1)));
+    queuePopTimeUpdater.setCycleCount(Timeline.INDEFINITE);
+    queuePopTimeUpdater.play();
+  }
+
+  private void updateQueue(int responseTimeSeconds, OffsetDateTime start) {
```
Sheikah45:
```suggestion
  private void updateTimer(int responseTimeSeconds, OffsetDateTime start) {
```
Maybe better?

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
```diff
@@ -0,0 +1,122 @@
...
+    notificationService.addImmediateInfoNotification(
+        "isReady.matchCanceled"
+    );
+  }
+
+  private void onReadRequest(IsReadyRequest isReadyRequest) {
+    log.info("Tournament game is ready, asking user.");
+    if (notification != null) {
+      log.warn("Tournament ready request ignored because tournament is already in progress.");
+      respondToReadyRequest(isReadyRequest.getRequestId(), isReadyRequest);
```
Sheikah45: Wouldn't this fail if the game had actually started and the notification was dismissed?

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
```diff
@@ -0,0 +1,122 @@
...
+      return;
+    }
+    uiService.bringMainStageToFront();
+    final var controller = initializeIsReadyController(isReadyRequest);
```
Sheikah45: No vars please

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
```diff
@@ -0,0 +1,122 @@
...
+    notification =
+        new ImmediateNotification(i18n.get("isReady.title"), i18n.get("isReady.message", isReadyRequest.getGameName()),
+            Severity.INFO, null, List.of(), controller.getRoot(), false);
+    notificationService.addNotification(notification);
```
Sheikah45, line 70: maybe add a new NotificationService Method instead of creating it if this is common?

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
```diff
@@ -0,0 +1,122 @@
...
+    safetyCloseMechanism();
+  }
+
+  private void safetyCloseMechanism() {
```
Sheikah45:
```suggestion
  private void ensureNotificationCloses() {
```

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
```diff
@@ -0,0 +1,122 @@
...
+    notification.dismiss();
+    notification = null;
+  }
+
+  @NotNull
+  private IsReadyForGameController initializeIsReadyController(IsReadyRequest isReadyRequest) {
+    IsReadyForGameController controller = uiService.loadFxml("theme/tournaments/is_ready_for_game.fxml");
+    controller.setTimeout(isReadyRequest.getResponseTimeSeconds());
+    controller.setReadyCallback(() -> respondToReadyRequest(isReadyRequest.getRequestId(), isReadyRequest));
+    controller.setDismissCallBack(() -> JavaFxUtil.runLater(this::dismissNotification));
```
Sheikah45: Might be better to just do the JavaFxUtil wrapping in dismissNotification, to make it more error proof

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
```diff
@@ -0,0 +1,122 @@
...
+    } catch (Exception e) {
+      dismissNotification();
+      cancelMatch();
+      notificationService.addImmediateErrorNotification(e, "isReady.readyUpFailed");
+    }
+  }
+
+  private void cancelMatch() {
+    if (matchFuture != null) {
+      matchFuture.cancel(false);
```
Sheikah45, line 119: Just a technical note the match future is not interruptible so only non started stages will be cancelled. This means that the game launch can still happen in some cases. Active problem with matchmaker cancellation as well

src/main/java/com/faforever/client/ui/progress/ProgressCircleIndicator.java
```diff
@@ -0,0 +1,155 @@
...
+import javafx.css.converter.SizeConverter;
+import javafx.scene.control.Control;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+
+/**
+ * Base class for the progress indicator controls represented by circualr shapes
```
Sheikah45:
```suggestion
 * Base class for the progress indicator controls represented by circular shapes
```

src/main/java/com/faforever/client/ui/progress/ProgressCircleIndicator.java
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * @author Andrea Vacondio
+ */
+abstract class ProgressCircleIndicator extends Control {
```
Sheikah45, line 40: Interesting will have to look at it but is a circular indicator that much better than the linear one we already have?

Also have to make sure it doesn't overload the client XD

src/main/resources/application-local.yml
```diff
@@ -33,7 +33,6 @@ faf-client:
     base-url: http://localhost:4444
     scopes: openid offline public_profile upload_map upload_mod lobby
     client-id: faf-java-client
-    timeout-milliseconds: 1500000
```
Sheikah45, line 36: why remove this?

Sheikah45 (changes requested):
Will have to see the indicator in action XD

src/main/java/com/faforever/client/game/GameService.java
1-alex98: Sorry I am a little used to it and my IDE is also set to use them :sweat_smile:

1-alex98: Imagine the case where the GameLaunchResponse is sent before the update of the featured mod is done. Then the GameLaunchResponse is missed. That does btw also apply for TMM and is fixed here as well.

src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
1-alex98: ok just took it from somewhere else

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingController.java
1-alex98: Never compare "Integer" with ==. That is what I am fixing here. I do not think Player ids can be null but I just took the safe option

src/main/java/com/faforever/client/tournament/game/IsReadyForGameController.java
1-alex98: Yes I know it is strange I know XD

The idea is the following:
The Controller calculates the timeLeft till timeout. It then calculates the progress in percent and sets the progress. The passed number is the progress in percent. Why calculate the time left in seconds back from the percentage if we already have it. Also might cause rounding errors :thinking:

src/main/java/com/faforever/client/tournament/game/IsReadyForGameController.java
1-alex98: Hmm don't know. I mean I thought about it.

src/main/java/com/faforever/client/tournament/game/IsReadyForGameController.java
1-alex98: Yes that is copy pasta :D

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
1-alex98: yes this is a left over when i was sending negative responses as well. Good catch. 
This should never happen anyways because server should make sure but who knows.

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
1-alex98: This is the only point we do such a notification that can not be canceled. Or what do you mean?

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
1-alex98: Hmm ok. Yes mechanism is the same as for tmm. Would be good to fix here as well should it ever be fixed.

src/main/java/com/faforever/client/ui/progress/ProgressCircleIndicator.java
1-alex98: It is only updated every second. So no overload possible

1-alex98: Screenshot_20220915_192658

src/main/resources/application-local.yml
1-alex98: no idea

1-alex98: It is a deprecated property that is read nowhere

1-alex98:
Thx for the review should have addressed almost all of it. And commented the rest.

src/main/java/com/faforever/client/game/GameService.java
Sheikah45: ah yeah that is a good point

src/main/resources/application-local.yml
Sheikah45: Hmm weird thought it was used but you are correct

src/main/java/com/faforever/client/tournament/game/IsReadyForGameController.java
Sheikah45: I guess so but then it seems like it would be best for the widget to have two properties and components. The label and the progress bar. Seems a little odd they are combined to be the same

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
Sheikah45: ah didn't see the not cancellable nature of it. Although still might be good to make it easier to use in the future but not necessary I suppose

src/main/java/com/faforever/client/tournament/game/TournamentGameService.java
Sheikah45: Yeah it is mostly caused by the CompletableFutures not having an interrupt mechanism so would have to go more low level to fix it in a nice way

src/main/java/com/faforever/client/ui/progress/ProgressCircleIndicator.java
Sheikah45: Yeah that does look pretty nice

Sheikah45:
Will look at this again soon

Sheikah45 (commented):
Seems like it is in a pretty good state. Does this need any changes on the server?

1-alex98:
> Seems like it is in a pretty good state. Does this need any changes on the server?

yes but it will not hurt having it in the client before that is done

PR #2808, opened 2022-09-24
Bugfix/code cov report

1-alex98:


1-alex98:
due to a deprecation the way the code cov is uploaded needs to change

PR #2821, opened 2022-10-04
Set `no image` image instead of empty background

### LINKED ISSUE #2770: Display `No image` image instead of empty background if an image failed to load or for some other reason

Marc-Spector:
Closes #2770


Marc-Spector:
image
image
image

Marc-Spector:
It seems to work well. No memory leaks. :)

src/main/java/com/faforever/client/fx/ImageViewHelper.java
```diff
@@ -0,0 +1,63 @@
...
+          JavaFxUtil.removeListener(imageView.imageProperty(), this);
+        }
+      }
+    };
+
+    public void setDefaultImage(ImageView imageView, Image defaultImage, boolean onlyOnce) {
+      this.imageView = imageView;
+      this.defaultImage = defaultImage;
+      this.onlyOnce = onlyOnce;
+      JavaFxUtil.addListener(imageView.imageProperty(), onlyOnce ? imageListener : new WeakInvalidationListener(imageListener));
```
Sheikah45: I don't quite understand the only once parameter. If the listener is always  removed when completed why do we need to specify a weak listener?

Sheikah45: Also wouldn't we want to addAndTrigger all these listeners?

Sheikah45: And might be nice if possible if the imageViews started with the default image maybe.

Marc-Spector: I did a small refactoring of the class. Please review the code again.

Marc-Spector: > Also wouldn't we want to addAndTrigger all these listeners?

No need. It is extra.

Marc-Spector: > And might be nice if possible if the imageViews started with the default image maybe.

Done for some image views.

Sheikah45:
Ah yeah that makes more sense now. 

The only thing I wonder now is if the image placeholder class should be made into a bean and have the UI service injected since it seems to be called the same way in a bunch of places.

Marc-Spector:
Ok. This is the good suggestion, the code will become much cleaner.

Sheikah45 (approved):

PR #2831, opened 2022-10-29
Add support for disabling leagues

BlackYps:
This removes the disabled 4v4 no share league from the leaderboard in the client.
At the moment this breaks the loading of the leaderboard. It will only become functional when the new api version is deployed that exposes the "enabled" field. So be careful with merging.


BlackYps:
With the latest api update this works now

Sheikah45 (approved):

PR #2833, opened 2022-10-29
Display number of filtered out games

### LINKED ISSUE #2832: Display number of filtered out games
For the newly introduced game filter it would be awesome if the client would display the number of filtered out games in the lobby and live games tab. This acts as a feedback and reminder to the user that the filter is active.

Marc-Spector:
Closes #2832 

image

### DIFF
--- a/src/main/java/com/faforever/client/filter/CustomGamesFilterController.java
+++ b/src/main/java/com/faforever/client/filter/CustomGamesFilterController.java
@@ -54,9 +54,9 @@ public class CustomGamesFilterController extends AbstractFilterController<GameBe
   protected void afterBuilt() {
     Preferences preferences = preferencesService.getPreferences();
 
-    privateGameFilter.bindBidirectional(preferences.hidePrivateGamesProperty());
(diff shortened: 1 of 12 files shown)


BlackYps:
Looks good, but it shows the number of still displayed games instead of the filtered out games. In your screenshot I would expect the text to read "3 of 11 items are filtered out"

BlackYps:
I just tested it and it works.
I noticed two issues that are not directly related to the displaying of the game numbers though:
- The clear button only works correctly for the live replays. In the custom games tab it only becomes active when filtering game mods and then it also only clears the game mod setting.
- Only the filters not affected by the clear button survive a restart of the client.

I guess both things are related and the underlying cause is that the client stores these settings in the preferences.
It would be nice if the client stored the settings but the button still allowed you to clear all settings easily

Marc-Spector:
>I guess both things are related and the underlying cause is that the client stores these settings in the preferences.

That is right.

Marc-Spector:
> It would be nice if the client stored the settings but the button still allowed you to clear all settings easily

Done

Sheikah45 (approved):

PR #2843, opened 2022-11-10
Disable download updates button on Linux

Brutus5000:
As long as there is no unified installer, we can't offer automatic update installation. Thus, the button was simply removed for all OS except Windows.


Sheikah45:
I wonder if we should actually make use of the os spring profiles for this

Brutus5000:
I see no benefit of this. It's not like we're replacing whole beans with different implementations. Also it sounds silly to tell the app on startup on which OS it's running on.

Sheikah45:
I mainly bring it up because we already run the is windows check on startup and add the Linux profile based on that.

It is all done programmatically

Brutus5000:
How would this impact the code change here?

Sheikah45:
It would just change it from running the isWindows check to injecting the profiles and checking if linux is in the list

Also theoretically would make it easier to test as well

Brutus5000:
In that case I'd prefer a platform bean that has methods similar to the bridge4j Platform class (isLinux, isWindows)

Sheikah45 (approved):

PR #2847, opened 2022-11-13
Notify that a lobby is full when you watching a replay

### LINKED ISSUE #2836: The client does not notify that a lobby is full when you watching a replay
**To Reproduce**
1. Allow watching replays while in a game via `Settings -> Forget Alliance Forever`
2. Host a game or join to any game
3. Watch any replay
4. Wait when a lobby is full

**Log or error message**
No need

**Expected behavior**
Notify an user that the lobby is full

**Screenshots**
No need

**OS**
Windows

**Client Version**
2022.10.0-alpha-2

Marc-Spector:
Closes #2836

### DIFF
--- a/src/main/java/com/faforever/client/fx/PlatformService.java
+++ b/src/main/java/com/faforever/client/fx/PlatformService.java
@@ -188,7 +188,7 @@ public class PlatformService {
   }
 
   public void stopFlashingWindow(String windowTitle) {
-    startFlashingWindow(windowTitle, null);
+    stopFlashingWindow(windowTitle, null);
   }
 
   public void stopFlashingWindow(String windowTitle, @Nullable Long processId) {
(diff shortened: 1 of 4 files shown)


Brutus5000:
Wouldn't it be easier to remember the process ids, when we launch a game of replay instead of fiddling around with window titles?

Marc-Spector:
No need to remember process ids of game replay. It is enough to know the process id, where the lobby is located.

src/main/java/com/faforever/client/fx/PlatformService.java
```diff
@@ -154,13 +187,21 @@ public void startFlashingWindow(String windowTitle) {
     User32.INSTANCE.FlashWindowEx(flashwinfo);
   }
 
-
   public void stopFlashingWindow(String windowTitle) {
+    startFlashingWindow(windowTitle, null);
+  }
```
Sheikah45: is this supposed to be calling stopFlashingWindow?

Sheikah45 (changes requested):

src/main/java/com/faforever/client/fx/PlatformService.java
Marc-Spector: Oh, good catch

Sheikah45 (approved):

PR #2867, opened 2022-12-05
Fix disappearance of players in the team list

### LINKED ISSUE #2678: A random player disappears from the team list
**Describe the bug**
In the details of the game, a random player disappears from the team list. The number of players in the lobby does not change. There's nothing wrong with the player in the lobby.

**To Reproduce**
Don't know.

**Log or error message**
No any error messages in log file.

**Expected behavior**
Return the player to the team list.

**Screenshots**
image

**OS**
Windows

Client Version: 2022.4.0-alpha0

Marc-Spector:
Closes #2678 

It happens when the player leaves `#aeolus` channel

### DIFF
--- a/src/main/java/com/faforever/client/chat/KittehChatService.java
+++ b/src/main/java/com/faforever/client/chat/KittehChatService.java
@@ -417,7 +417,8 @@ public class KittehChatService implements ChatService, InitializingBean, Disposa
     synchronized (chatChannelUsersByChannelAndName) {
       chatChannelUsersByChannelAndName.remove(mapKey(username, channelName));
     }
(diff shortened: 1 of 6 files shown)


Marc-Spector:
I see no reason to remove players from the map. It won't get any worse.

Marc-Spector:
damn it...

Sheikah45:
Wouldn't this potentially lead to a memory leak as players are only added and never removed from the map?

Marc-Spector:
It's possible. I need to find another solution. My change is not suitable as it breaks a notification that a friend has logged into the client.

Sheikah45 (approved):

PR #2872, opened 2022-12-19
Allow players to toggle ipv6 for the ice adapter

Sheikah45:


src/main/resources/i18n/messages.properties
```diff
@@ -921,6 +921,8 @@ settings.fa.changeProcessPriority = Use high process priority
 settings.fa.changeProcessPriority.description = (Windows only) Set the game's process priority to "high". Some antivirus programs might not like this.
 settings.fa.executableDecorator.description = How the client will execute the game executable. Leave it blank unless you know what you're doing.
 settings.fa.executionDirectory.description = In which directory the game executable should be executed. Leave it blank unless you know what you're doing.
+settings.fa.allowIpv6 = Allow the ice adapter to use ipv6
```
micheljung:
```suggestion
settings.fa.allowIpv6 = Allow the ICE adapter to use IPv6
```

src/main/resources/i18n/messages.properties
```diff
@@ -921,6 +921,8 @@ settings.fa.changeProcessPriority = Use high process priority
...
+settings.fa.allowIpv6.description = Ipv6 causes connection issues for some players. Turn this on if you do not have any ipv6 issues with connections
```
micheljung:
```suggestion
settings.fa.allowIpv6.description = IPv6 causes connection issues for some players. Turn this on if you do not have any IPv6 issues with connections.
```

src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
```diff
@@ -196,28 +197,37 @@ List<String> buildCommand(Path workDirectory, int adapterPort, int gpgPort, int
 
     String classpath = getBinaryName(workDirectory) + JavaUtil.CLASSPATH_SEPARATOR + getJavaFXClassPathJars();
 
-    List<String> cmd = Lists.newArrayList(
+    List<String> cmd = new ArrayList<>(List.of(
```
Brutus5000: Can we flatten the list nesting here? If its a mutable list, just cmd.add()

PR #2885, opened 2022-12-29
add `add/edit/remove note` items to context menu of chat user. Replace underline with note icon

### LINKED ISSUE #2884: Add `add/edit/remove note` items to context menu of chat user

Marc-Spector:
Closes #2884


Marc-Spector:
The PR is ready for review

src/main/java/com/faforever/client/chat/ChatChannelUser.java
```diff
@@ -55,6 +55,7 @@ public class ChatChannelUser {
   private final ObjectProperty<Image> gameStatusImage = new SimpleObjectProperty<>();
   private final StringProperty statusTooltipText = new SimpleStringProperty();
   private final BooleanProperty displayed = new SimpleBooleanProperty(false);
+  private final StringProperty playerNote = new SimpleStringProperty();
```
Sheikah45: Why add a separate player note attribute to the chatChannelUser instead of using the one already in the associated player?

Seems like you could avoid a binding by just returning the actual playernote property in the getter and setter.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/chat/ChatChannelUser.java
Marc-Spector: > in the getter.

Do you mean in `ChatChannelUser` class?

If there is no player? What should I return?

Sheikah45: If there is no player we should just return an empty property or null.

We can't truly track irc users so it doesn't make sense to set notes on them since they will likely get lost

Sheikah45 (approved):

Issue #746, opened 2018-02-04
Implement sounds for notifications and events


20seven:
Sounds for certain notifications/events aren't implemented yet. 

In AudioServiceImpl.java:

```
    // TODO implement
//    friendOnlineSound = loadSound(FRIEND_ONLINE_SOUND);
//    friendOfflineSound = loadSound(FRIEND_OFFLINE_SOUND);
//    friendJoinsGameSound = loadSound(FRIEND_JOINS_GAME_SOUND);
//    friendPlaysGameSound = loadSound(FRIEND_PLAYS_GAME_SOUND);
```

Sheikah45:
@axel1200 is this still relevant?

1-alex98:
I think it is

andrewtvuong:
Is this still relevant? Have no friends on FAF to test lol.. would be interested in feature request for sound to alert host or something, with cooldowns to avoid spamming. This would be helpful to wake up afk host waiting on 1/12 Dual Gap

Sheikah45:
Yes, still relevant. The sounds have still not been implemented

magge-faf:
I have no authorization to push features, so I just created an own GitHub project to showcase my implementation by following the GitHub branching standards & conventions:

https://github.com/magge-faf/downlords-faf-client/commit/841b57174ba8a1b7d5e05d31851261ea3e489b31

---

This was hard-coded:

`private static long SILENCE_PERIOD_AFTER_SOUND = 30000;
`

Best case scenario would be to let the user decide which threshold he wants:

I would add another fancy sub-menu under the existing "Enable sounds" where the user can have an input field for seconds to set his preferred silence:

![image](https://user-images.githubusercontent.com/101107758/212303742-ea8fb207-7a7d-4c53-8d5f-7405e161268f.png)

Is this a solution or had the developers a different approach in mind?

Sheikah45:
I think that is a solution. Although I might try to find a better name than silence after sound.

Also you can make a PR with the sounds from your fork to develop

magge-faf:
If you give me permission to PR, I would request those changes: https://github.com/magge-faf/downlords-faf-client/commit/841b57174ba8a1b7d5e05d31851261ea3e489b31

I would create later another issue/feature request about the users-input-threshold-menu. The foundation for the new notification sounds must have green light in the first place.

Sheikah45:
Should be good to make the pr now

Sheikah45:
Fixed in #2889

PR #2889, opened 2023-01-14
Implement sounds for notifications and events #746

magge-faf:
I kept the old sounds, and just added new ones which sounded like a user would expect to hear for certain events (all are under https://pixabay.com/service/license/ - see sound folder)

Normalized all sounds as well and made them all the same loudness. Finally, no ear bleeding anymore for PM sounds.


Sheikah45 (approved):
Nice and the sounds sound pretty good too

PR #2890, opened 2023-01-14
Upgrade to javafx 19

Sheikah45:
This is nice because it introduces .map, .flatMap, and .orElse to ObservableValue. This allows us to remove alot of the Bindings.create usage and use more readable syntax with null safety.

Also had to reimplement the SVGLoader since the signature of the underlying fx classes changed.


Sheikah45:
I will also be looking to simplify some of the game, player, and ui listeners using the new map functions in a later pr

Marc-Spector (approved):

PR #2898, opened 2023-01-22
Fix: Handling file encoding for preferencesFile

### LINKED ISSUE #2897: MalformedInputException related to auto update mods and maps bricks FAF
The error has traces back to [24 Nov 2022](https://forum.faforever.com/post/37930), but recently the "java.nio.charset.MalformedInputException: Input length = 1" is an occurring phenomena for several users or at least got reported more often than usual. The MalformedInputException basically bricks their FAF, and can only be solved temporary (for some users, not all apparently) with a complete wipe of all files and a reinstallation - until the Game.prefs (not 100% sure on that) breaks again.

Follow user report from Discord, and some are here in this thread: https://forum.faforever.com/topic/5186/mod-error-without-any-mod/24

image

**To Reproduce**
I can not reproduce it on my machine, I would guess the users with the issue have a specific map(s) or mod(s) XY which malformed their file when they auto update - Is there any way to enclose the issue further by asking the users which mods/maps they have installed, or about their Game.prefs ? Their installation and path profile looked clean so far, no OneDrive or Cyrillic letters - Usually the generic basic stuff to check first for.

**Log or error message**
Log file of a user:

[client (12).log](https://github.com/FAForever/downlords-faf-client/files/10472936/client.12.log)

magge-faf:
As discussed, a solution draft for following issue - feel free to comment/edit to your liking, thanks in advance: https://github.com/FAForever/downlords-faf-client/issues/2897

### DIFF
--- a/src/main/java/com/faforever/client/mod/ModService.java
+++ b/src/main/java/com/faforever/client/mod/ModService.java
@@ -54,7 +54,8 @@ import reactor.util.function.Tuple2;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
+import java.nio.charset.Charset;
+import java.nio.charset.MalformedInputException;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/mod/ModService.java
```diff
@@ -307,45 +308,63 @@ public void overrideActivatedMods(Collection<ModVersionBean> modVersions) {
...
-
-        if (Boolean.parseBoolean(activeModMatcher.group(2))) {
-          activeMods.add(modUid);
+    List<Charset> preferredEncodings = new ArrayList<>(List.of());
+    Map<String, Charset> map = Charset.availableCharsets();
+    for (Map.Entry<String, Charset> entry : map.entrySet()) {
+      Charset charset = entry.getValue();
+      preferredEncodings.add(charset);
+    }
+    for (Charset charset : preferredEncodings) {
```
Sheikah45:
```suggestion
    for (Charset charset : Charset.availableCharsets().values()) {
```

The preferredEncodings List is not necessary since we can just use the values view on a map.

src/main/java/com/faforever/client/mod/ModService.java
```diff
@@ -307,45 +308,63 @@ public void overrideActivatedMods(Collection<ModVersionBean> modVersions) {
...
+            if (Boolean.parseBoolean(activeModMatcher.group(2))) {
+              activeMods.add(modUid);
+            }
+          }
         }
+        break;
+      } catch (MalformedInputException e) {
+        throw new AssetLoadException("Could not read mod state", e, "mod.errorReadingMods");
       }
     }
```
Sheikah45: Right now this try catch loop still throws an exception rather than moving onto the next encoding. It would be better to just have the extraction of the preferencesContent in the for loop with the try catch and then throw an assetloadexception if the preferencesContent is null. So more like below
```suggestion
      try {
        String preferencesContent = Files.readString(preferencesFile, charset);
        break;
      } catch (MalformedInputException e) {
        log.warn(message);
      }
        
        Matcher matcher = ACTIVE_MODS_PATTERN.matcher(preferencesContent);
        if (matcher.find()) {
          Matcher activeModMatcher = ACTIVE_MOD_PATTERN.matcher(matcher.group(0));
          while (activeModMatcher.find()) {
            String modUid = activeModMatcher.group(1);
            if (Boolean.parseBoolean(activeModMatcher.group(2))) {
              activeMods.add(modUid);
            }
          }
        }
```

src/main/java/com/faforever/client/mod/ModService.java
```diff
@@ -307,45 +308,63 @@ public void overrideActivatedMods(Collection<ModVersionBean> modVersions) {
...
+      preferredEncodings.add(charset);
+    }
+    for (Charset charset : preferredEncodings) {
+      try {
+        String preferencesContent = Files.readString(preferencesFile, charset);
+        String currentActiveModsContent = null;
+        Matcher matcher = ACTIVE_MODS_PATTERN.matcher(preferencesContent);
+        if (matcher.find()) {
+          currentActiveModsContent = matcher.group(0);
+        }
```
Sheikah45: same as above. In fact might be best to make the reading of the pref file a function

Sheikah45 (changes requested):

magge-faf:
Pushed to the wrong branch, instead to [magge-faf/downlords-faf-client](https://github.com/magge-faf/downlords-faf-client). I will push it there later.

magge-faf:
I know how to write test cases for public methods, but I am not familiar with tests for private methods - After some reading material, I got the info that usually you do not write tests for private methods, unless it is needed by the design of the software. 

And no other private method has any test in ModServiceTest.java, so I assume the recently added private method readPreferencesFile does not require it as well.

The public method enableSimMods uses the private method readActiveMods and inevitably the private method readPreferencesFile.  ⇒ There is already a test case for enableSimMods, which would fail if there would be a bad return from these private methods, so they are indirectly tested, aren't they?

I have pushed it at my forked project, but it gets declined by codecov/patch, please have a look and tell me your thoughts @Sheikah45 

https://github.com/magge-faf/downlords-faf-client/commit/c26d3e0094154cefe0311f494859490f00f17141

1. Is there anything on my end I need to do to resolve the failed codecov/patch?
2. What are your thoughts about the changes?

- Created a method to read the preferences file
- throw AssetLoadException when preferencesContent is null

Sheikah45 (approved):
This all looks good to me. The null preferencesContent would throw an error later anyway so good to fail fast.

And as far as codecov it is a bit more of a loose guide than law so should all be good. And the method is covered by the sim mods test

magge-faf:
Thank you very much.

Issue #1333, opened 2019-06-29
Client fails to start on non-latin named user accounts (platform windows)


norraxx:
```
java.lang.UnsatisfiedLinkError: org.bridj.Platform.sizeOf_ptrdiff_t()I
	at org.bridj.Platform.sizeOf_ptrdiff_t(Native Method)
	at org.bridj.Platform.<clinit>(Platform.java:232)
	at com.faforever.client.preferences.PreferencesService.<clinit>(PreferencesService.java:78)
	at com.faforever.client.FafClientApplication.main(FafClientApplication.java:55)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at com.exe4j.runtime.LauncherEngine.launch(LauncherEngine.java:85)
	at com.exe4j.runtime.WinLauncher.main(WinLauncher.java:94)
	at com.install4j.runtime.launcher.WinLauncher.main(WinLauncher.java:25)
```

norraxx:
Fails on `org.bridj.Platform.isWindows()`.
via @axel1200 `import com.sun.jna.Platform;` has `isWindows` method. So maybe we could change the method.

norraxx:
Fails now on:
```
 if (com.sun.jna.Platform.isWindows()) {
      FAF_DATA_DIRECTORY = Paths.get(Shell32Util.getFolderPath(ShlObj.CSIDL_COMMON_APPDATA), "FAForever");
    }
```

Problem is system class loader:
see more info on Java Native Access library pages:
https://github.com/java-native-access/jna/issues/124
https://github.com/java-native-access/jna/issues/238
https://bugs.openjdk.java.net/browse/JDK-8017274

1-alex98:
We wann display a pop up then and ask user to make a new windows account with ascii characters

PR #2899, opened 2023-01-24
Fix: Check APPDATA path for Cyrillic characters and show pop up warning #1333

magge-faf:
image

After some reading material, JDK18 has a bug with Cyrillic letters, and it won't get fixed, needs >JDK 19 (related https://bugs.openjdk.org/browse/JDK-8017274)

18< JDK assumes "System locale and User locale are set to the same language" for

`org.bridj.Platform.isWindows()
`

---
Unfortunately, the error message in FAF client is a bit bland and confusing on the user end.

`ava.lang.UnsatisfiedLinkError: org.bridj.Platform.sizeOf_ptrdiff_t()I
	at org.bridj.Platform.sizeOf_ptrdiff_t(Native Method)
	at org.bridj.Platform.<clinit>(Platform.java:232)
	at com.faforever.client.preferences.PreferencesService.<clinit>(PreferencesService.java:78)
	at com.faforever.client.FafClientApplication.main(FafClientApplication.java:55)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at com.exe4j.runtime.LauncherEngine.launch(LauncherEngine.java:85)
	at com.exe4j.runtime.WinLauncher.main(WinLauncher.java:94)
	at com.install4j.runtime.launcher.WinLauncher.main(WinLauncher.java:25)`

---
Workaround:

1. Check for Cyrillic characters in APPDATA when FAF starts
2. Show a pop-up message in EN/RU when it has found any and tell the user what the issue is

---
I have trouble with the test cases, because mocking APPDATA does not work. Junit 4 has PowerMockRunner, but needs different dependencies to make it run. Junit 5 has PowerMockExtension

Before I go even further that way, fiddling with all that test part:

1. Is the drafted code a suitable solution, or does it need to go back to the drawing board?
2. If suitable, what would be the best approach about the test cases?

---

Note: 

Check for Linux can be added later, but those cases are very, very rare.

In the project are 6 more instances with org.bridj.Platform.isWindows(), but they all should be already covered, when the first org.bridj.Platform.isWindows() works and the user has created a working environment.

@Sheikah45 Please, can I hear your thoughts on this draft, thanks in advance and take your time

### DIFF
--- a/src/main/java/com/faforever/client/FafClientApplication.java
+++ b/src/main/java/com/faforever/client/FafClientApplication.java
@@ -81,28 +81,23 @@ public class FafClientApplication extends Application {
 
     OperatingSystem operatingSystem = applicationContext.getBean(OperatingSystem.class);
 
-    try {
-      String preferencesDirectoryString = String.valueOf(operatingSystem.getPreferencesDirectory());
-      log.debug("Current preferences directory " + preferencesDirectoryString);
-      if (preferencesDirectoryString.matches(".*[а-яА-ЯёЁ].*")) {
-        I18n i18n = applicationContext.getBean(I18n.class);
-        String warningMessage = i18n.get("home.directory.warning.cyrillic");
-        String closeText = i18n.get("close");
-        ButtonType closeButton = new ButtonType(closeText, ButtonData.CANCEL_CLOSE);
-        CountDownLatch waitForUserInput = new CountDownLatch(1);
-
-        JavaFxUtil.runLater(() -> {
-          Alert alert = new Alert(AlertType.WARNING, warningMessage, closeButton);
-          Optional<ButtonType> buttonType = alert.showAndWait();
-          if (buttonType.filter(button -> button == closeButton).isPresent()) {
-            System.exit(EXIT_STATUS_CYRILLIC_CHARACTER_PREFERENCES_DIRECTORY);
-          }
-          waitForUserInput.countDown();
-        });
-        waitForUserInput.await();
-      }
-    } catch (InterruptedException e) {
-      e.printStackTrace();
+    String preferencesDirectoryString = String.valueOf(operatingSystem.getPreferencesDirectory());
+    log.debug("Current preferences directory " + preferencesDirectoryString);
+    if (preferencesDirectoryString.matches(".*[а-яА-ЯёЁ].*")) {
+      I18n i18n = applicationContext.getBean(I18n.class);
+      String warningMessage = i18n.get("home.directory.warning.cyrillic");
+      String closeText = i18n.get("close");
+      ButtonType closeButton = new ButtonType(closeText, ButtonData.CANCEL_CLOSE);
+      CountDownLatch waitForUserInput = new CountDownLatch(1);
+      JavaFxUtil.runLater(() -> {
+        Alert alert = new Alert(AlertType.WARNING, warningMessage, closeButton);
+        Optional<ButtonType> buttonType = alert.showAndWait();
+        if (buttonType.filter(button -> button == closeButton).isPresent()) {
+          System.exit(EXIT_STATUS_CYRILLIC_CHARACTER_PREFERENCES_DIRECTORY);
+        }
+        waitForUserInput.countDown();
+      });
+      waitForUserInput.await();
     }
 
     if (operatingSystem.runsAsAdmin()) {
--- a/src/main/java/com/faforever/client/FafClientApplication.java
+++ b/src/main/java/com/faforever/client/FafClientApplication.java
@@ -21,6 +21,7 @@ import javafx.application.Application;
 import javafx.application.Platform;
 import javafx.scene.control.Alert;
 import javafx.scene.control.Alert.AlertType;
+import javafx.scene.control.ButtonBar.ButtonData;
 import javafx.scene.control.ButtonType;
 import javafx.scene.text.Font;
 import javafx.stage.Stage;
@@ -60,6 +61,7 @@ public class FafClientApplication extends Application {
   public static final String PROFILE_LINUX = "linux";
   public static final String PROFILE_MAC = "mac";
   public static final int EXIT_STATUS_RAN_AS_ADMIN = 3;
+  public static final int EXIT_STATUS_CYRILLIC_CHARACTER_PREFERENCES_DIRECTORY = 4;
 
   private ConfigurableApplicationContext applicationContext;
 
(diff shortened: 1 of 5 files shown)


Sheikah45:
The Logging service I have been meaning to refactor to move away from static methods and initialization and into something that is more mockable and so that we don't have to use the isWindows function everywhere. Previously we added an OSBean which detects the os at startup so was going to use that.

As far as your implementation, I would have just made check user path take a string or path as input so it can be easily tested. But this is the best way to do it until the refactor occurs.

magge-faf:
Changes:

- path as Input for checkUserPath() and checkForCyrillicCharacters() to be able to mock for test units
- custom implementation of JOptionPane showMessageDialog with a 'testing' flag, so it does not block test units
- linux support added
- throw Unknown operating system
- unit tests with cases

@Sheikah45 Please, take a look, thanks in advance.

---
I will look into those 412 failed tests later - I guess a mock object is not cleaning itself properly after a run and causes the others to fail.

Sheikah45:
@magge-faf looking at this and the amount of static methods and classes this needs. I don't think this is a good way to go about the detection of the cyrillic letters.

It is probably better to have this trigger on the spring bean concepts we introduced. I can refactor it to use that and then you can look at the cyrillic detection if that works.

Also just generally why did you use the JOptionPane instead of the Alert like the admin checl

magge-faf:
I had removed the admin check in my environment and totally forgotten, that it contains a pop-up window as well, which I could have simply taken as inspiration, my bad.

I threw my first attempts over board - and by taken your recent refactoring (aka merged into my dev branch), I just added a check before the runsAsAdmin() - If you think an own method with test cases is better, I can adapt to it. Or if you have something else in mind for this, just point me in the direction.

@Sheikah45 How about that solution, based on your recent refactoring?

src/main/java/com/faforever/client/FafClientApplication.java
```diff
@@ -80,6 +80,30 @@ public void init() throws InterruptedException, NoSuchFieldException, IllegalAcc
         .run(getParameters().getRaw().toArray(new String[0]));
 
     OperatingSystem operatingSystem = applicationContext.getBean(OperatingSystem.class);
+
+    try {
+      @NotNull String preferencesDirectoryString = String.valueOf(operatingSystem.getPreferencesDirectory());
```
Sheikah45: Don't need NotNull on local variable

src/main/java/com/faforever/client/FafClientApplication.java
```diff
@@ -80,6 +80,30 @@ public void init() throws InterruptedException, NoSuchFieldException, IllegalAcc
...
+      log.debug("Current APPDATA path " + preferencesDirectoryString);
```
Sheikah45: Not necessarily Appdata so maybe saypreferences

src/main/java/com/faforever/client/FafClientApplication.java
```diff
@@ -80,6 +80,30 @@ public void init() throws InterruptedException, NoSuchFieldException, IllegalAcc
...
+
+      if (preferencesDirectoryString.matches(".*[а-яА-ЯёЁ].*")) {
+        CountDownLatch waitForUserInput = new CountDownLatch(1);
+        JavaFxUtil.runLater(() -> {
+          Alert alert = new Alert(AlertType.WARNING, """
+              Предупреждение: Кириллические символы в пути APPDATA приведут к проблемам. Пожалуйста, избегайте их использования.
+              Warning: Cyrillic characters in the APPDATA path will cause problems. Please, avoid using them.
```
Sheikah45: I think now that we load the context first we could i18n for this

Sheikah45 (changes requested):

Sheikah45:
Did you mean to close this?

magge-faf:
Nah, I just wanted to push a few minutes ago the latest changes, and you had pushed something as well - I wanted to merge your changes first before pushing and IntelliJ has automatically closed/force pushed it for whatever reason after I have merged your changes ... 

I push to my develop GitHub later with my changes, as soon I have synced it.

magge-faf:
I have sync and pushed to my develops in GitHub, not sure why IntelliJ wanted to force push anything after the local merge, I would have not click OK on that.

I have reopened and pushed this commit:

https://github.com/FAForever/downlords-faf-client/commit/59ef5160fbb9f5fb2303625eaf2033c1cb58b729

---

I have taken inspiration by your implemented `closeMainWindow(WindowEvent event)`method from the same class, the rest about I18n I got via reading material.

`operatingSystem.runsAsAdmin()` can be refactored your suggest way with i18n as well (I think so?), to have everything on par in the UI - but this is a task for another day.

I have added translations for GER, RU and UKR (.properties file is called "UK" ...). I got help from the RU speaking moderator Resistance to have a proper translation. If I have to adapt to a certain workflow with those languages '.properties' files - just point me in the direction.

@Sheikah45 Please, let me know what you think about this solution - thanks in advance

src/main/java/com/faforever/client/FafClientApplication.java
```diff
@@ -78,6 +80,31 @@ public void init() throws InterruptedException, NoSuchFieldException, IllegalAcc
...
+          if (buttonType.filter(button -> button == closeButton).isPresent()) {
+            System.exit(EXIT_STATUS_CYRILLIC_CHARACTER_PREFERENCES_DIRECTORY);
+          }
+          waitForUserInput.countDown();
+        });
+        waitForUserInput.await();
+      }
+    } catch (InterruptedException e) {
+      e.printStackTrace();
+    }
```
Sheikah45: The try catch is unnecessary as the method declares that it throws InterruptedException

Sheikah45 (changes requested):

Sheikah45:
Looks good now. Is it all done from your perspective @magge-faf

magge-faf:
Yes, I am done -  thanks for your patience.

Sheikah45 (approved):

Marc-Spector:
As far as I know, the client works fine with Cyrillic paths. I have Cyrillic and no problem with that. For the game, yes, there is this problem (loading maps or reading mods) but it is resolved in https://github.com/FAForever/downlords-faf-client/pull/2748.

> After some reading material, JDK18 has a bug with Cyrillic letters, and it won't get fixed, needs >JDK 19 (related https://bugs.openjdk.org/browse/JDK-8017274)

This has nothing to do with Cyrillic ways. This is a problem of reading non-standard rare characters. Which is readable if you use a different encoding that does not support Java.

magge-faf:
I saw that Cyrillic issue frequently in the tech channel, and to confirm it, I had created an environment W10, with Cyrillic char in user.home path, and it caused issues like in the first post.

Now that the user.home path is checked for all players, the people who had issues with it, getting a better warning description (instead of a cryptic error message), but it means, that people who had it successfully running get the error message as well.

Can you please provide more information about the situation where it worked with Cyrillic chars in user.home path, so I can recreate that environment as well?

Marc-Spector:
@magge-faf 

> with Cyrillic char in user.home path, and it caused issues like in the first post.

Could you write the full path, which you used for testing, please? I'll try to test it.

magge-faf:
I removed the W10 already, but I had the same issue with W11. If you like, I can install W10 again and recreate it.

Default installation for FAF client ~v2022.12.0 
Default Windows username: Программы

`C:\Users\Программы>`

image

Marc-Spector:
> If you like, I can install W10 again and recreate it.

No need. Thanks.

I cannot reproduce the issue. =(

test

magge-faf:
I want to remind you, about that the JDK18 bug:

https://bugs.openjdk.org/browse/JDK-8017274

`JDK's "official supported" use scenario for Windows is that the System locale and User locale are set to the same language. `

You need to have a different language for your OS with Cyrillic path. Which some players have. I should have probably underlined that more.

Same system language RU and Cyrillic  = works
Any different combination = does not work

magge-faf:
The 'Cyrillic check' could probably be extended to check for the OS language - if it has RU by default, it can ignore the Cyrillic path - However, I do not know if that will cause to other issues in the long run.

---

Edit: 

I gave it a quick glance, it could be done like that by checking Locale and System OS language for RU:

```
    Locale languageSystemLocaleUserOS = Locale.getDefault(Locale.Category.DISPLAY);
    String languageLocaleUserOS = Locale.getDefault().getLanguage();
    if (!languageLocaleUserOS.equals("ru") && !languageSystemLocaleUserOS.getLanguage().equals("ru") && preferencesDirectoryString.matches(".*[а-яА-ЯёЁ].*")) {...}
```

@Sheikah45 Your opinion about extending the check, or will it cause trouble in the long term?

Marc-Spector:
Oh, thank you for pointing that out.
Ok. I don't have the opportunity to test this, since my Windows doesn't support adding more than one language.

Sheikah45:
For the moment I will change this just to a soft warning that won't close the client while we figure it out

PR #2901, opened 2023-01-26
Remove uses of isWindows outside OS determination

Sheikah45:


src/main/java/com/faforever/client/fa/ProcessUtils.java
```diff
@@ -7,22 +7,14 @@
...
 
-@UtilityClass
 @Slf4j
 public class ProcessUtils {
 
-  void setProcessPriority(Process process, WindowsPriority priority) {
-    if (!Platform.isWindows()) {
-      log.debug("Setting process priority is only supported on Windows");
-      return;
-    }
```
Brutus5000, line 24: But the current code is still Windows only?

magge-faf (dismissed):
I have looked over your changes several times to get a better picture of how it works - I am a bit wiser than before. This is more understandable for a newcomer like me than the whole static variable logging stuff before.

With that knowledge, the Cyrillic characters issue is much easier to tackle - I appreciate the opportunity. Nice job on the refactoring.

src/main/java/com/faforever/client/fa/ProcessUtils.java
Sheikah45: Yes, I wrapped where it is called in a check for windows.

I didn't want to add injection to this class just for an os check.

Slightly not ideal since it isn't as low as it could be. So could be convinced.

Brutus5000: I suggest to put the Windows-exclusiveness into the method name then.

Issue #2905, opened 2023-01-31
Java bug - Unmappable Character Exception - handling file encoding part 2


magge-faf:
**Situation:**

Since committing following code to fix the occasionally file encoding issues, https://github.com/FAForever/downlords-faf-client/pull/2898 there were numerous reports in Discord/Forum about following new error which users encountered when they click a button like 'create game'. They have not changed anything and it worked before. with their 'Game.prefs' file:

<details>
  <summary>error text</summary>
  

`java.lang.Error: java.nio.charset.UnmappableCharacterException: Input length = 2
at java.base/java.lang.String.decodeWithDecoder(String.java:1212)
at java.base/java.lang.String.newStringNoRepl1(String.java:786)
at java.base/java.lang.String.newStringNoRepl(String.java:738)
at java.base/java.lang.System$2.newStringNoRepl(System.java:2390)
at java.base/java.nio.file.Files.readString(Files.java:3369)
at com.faforever.client.mod.ModService.readPreferencesFile(ModService.java:314)
at com.faforever.client.mod.ModService.readActiveMods(ModService.java:330)
at com.faforever.client.mod.ModService.enableSimMods(ModService.java:232)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:343)
at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:699)
at com.faforever.client.mod.ModService$$SpringCGLIB$$0.enableSimMods(<generated>)
at com.faforever.client.game.GameService.lambda$joinGame$10(GameService.java:332)
at java.base/java.util.concurrent.CompletableFuture$UniRun.tryFire(CompletableFuture.java:787)
at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
at java.base/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2147)
[...]

PR #2906, opened 2023-01-31
Fix: Avoid Java bug caused by readString for readPreferencesFile

magge-faf:
Related to https://github.com/FAForever/downlords-faf-client/issues/2905

### DIFF
--- a/src/main/java/com/faforever/client/mod/ModService.java
+++ b/src/main/java/com/faforever/client/mod/ModService.java
@@ -54,7 +54,12 @@ import reactor.util.function.Tuple2;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
 import java.nio.charset.Charset;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CodingErrorAction;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/mod/ModService.java
```diff
@@ -308,17 +313,26 @@ private String readPreferencesFile(Path preferencesFile) throws IOException {
...
+        log.info("Successfully read preferences file with charset: " + charset.displayName());
         break;
-      } catch (Exception e) {
-        log.warn("Could not read preferences file with charset: " + charset.displayName());
+      } catch (CharacterCodingException e) {
+        log.info("Failed to read preferences file with charset: " + charset.displayName());
+        // Continue and try a different character set
+      } catch (IOException e) {
+        // User has a malformed file
+        throw new RuntimeException(e);
```
Sheikah45: I would still somewhere just catch all exceptions just in case

Sheikah45 (changes requested):

magge-faf:
- Added catch-all-block just in case there are unforeseen issues.
- Changed comment into log.error to clarify that is an IOException issue and not only a "probably malformed file"

It worked on all the tests files, which I got from the forum - I converted a random Game.prefs to several commonly used char sets (https://www.wikiwand.com/en/Character_encoding), and it worked for them as well - I think this alternate solution to avoid the read-decoder-Java-bug, is the next best one. 

@Sheikah45 From my POV the solution is ready, I tested all successfully I could on my end. If I find some volunteer who have the current bug, I could test the new client with them first, if you think this is not wasted time.

Sheikah45 (approved):

PR #2907, opened 2023-01-31
Fix ru online and offline notification titles

PathKnower:
Fix RU online and offline notification titles.


src/main/resources/i18n/messages_ru.properties
```diff
@@ -518,9 +518,9 @@ settings.account.username = Имя аккаунта FAF
 settings.account.username.change = Изменить имя
 urlPreviewDescription = {0} ({1})
 achievement.unlockedTitle = Достижение получено\!
-friend.nowOnlineNotification.title = {0} вошел в лобби
+friend.nowOnlineNotification.title = {0} в сети
```
PathKnower, line 521: "вошел в лобби" sounds like "joined into a lobby", so it replaced with "в сети" which means "online".

src/main/resources/i18n/messages_ru.properties
```diff
@@ -518,9 +518,9 @@ settings.account.username = Имя аккаунта FAF
...
 friend.nowOnlineNotification.action = Нажмите, чтобы начать разговор.
-friend.nowOfflineNotification.title = {0} покинул лобби
+friend.nowOfflineNotification.title = {0} не в сети
```
PathKnower, line 523: "покинул лобби" same as for above. Sounds like "left a lobby", replaced with "не в сети" which basically means "offline".

Sheikah45:
The translation files actually not modified on github directly. We use the translation software project poe editor here https://poeditor.com/projects/view?id=126971. So you can register and make the changes there.

PR #2913, opened 2023-02-13
Use javafx mapping for player, game, and chat user properties

Sheikah45:
This moves a lot of the listener patterns in the chat world to use the new javafx lazy bindings.

Also changed the chat list items to be reusable which drastically improves scrollability. Additionally allowed for full removal of the chatuserservice as now the properties set there can be requested using the lazy bindings removing the need for a lot of the book keeping.


src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -130,24 +126,51 @@ public abstract class AbstractChatTabController implements Controller<Tab> {
...
+  };
+  @SuppressWarnings("FieldCanBeLocal")
+  private final ChangeListener<TabPane> tabPaneChangeListener = (tabPane, oldTabPane, newTabPane) -> addTabListeners(newTabPane);
+  private final ChangeListener<State> webViewStatChangeListener = (observable, oldValue, newValue) -> sendWaitingMessagesIfLoaded(newValue);
+
+  private final ChangeListener<Boolean> resetUnreadMessagesListener = (observable, oldValue, newValue) -> {
+    if (hasFocus()) {
+      setUnread(false);
+    }
+  };
```
Brutus5000: This is odd style and hard to read. Why not make these functions and reference them with this::<listenerMethode> where used?

src/main/java/com/faforever/client/domain/GameBean.java
```diff
@@ -12,47 +12,69 @@
...
+  ObjectProperty<OffsetDateTime> startTime = new SimpleObjectProperty<>();
+  BooleanProperty enforceRating = new SimpleBooleanProperty();
+  ObjectProperty<GameType> gameType = new SimpleObjectProperty<>();
   /**
    * Maps a sim mod's UID to its name.
    */
-  private final ObjectProperty<Map<String, String>> simMods = new SimpleObjectProperty<>(Map.of());
-  private final ObjectProperty<Map<Integer, List<PlayerBean>>> teams = new SimpleObjectProperty<>(Map.of());
+  ObjectProperty<Map<String, String>> simMods = new SimpleObjectProperty<>(Map.of());
+  ObjectProperty<Map<Integer, Set<PlayerBean>>> teams = new SimpleObjectProperty<>(Map.of());
```
Brutus5000: Map.of() is immutable. Does that even work?

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
Sheikah45: Yes I intend to go back and make as many of them functions where I can to simplify the reading of it.

Although in many places we can't use the this:: syntax since the function would only take the new value.

src/main/java/com/faforever/client/domain/GameBean.java
Sheikah45: Yes because for these properties we do not want to make changes directly to the maps. So they are designed to be immutable.

Instead we are observing when the map itself changes. Which is why this is an observable value wrapping a map.

Although later I changed this to a Map property which is the same thing with a few extra methods. But it is still wrapping an unmodifiable map.

PR #2942, opened 2023-03-12
Add single button for starting search

### LINKED ISSUE #2938: Add an explicit search button for matchmaker
This is a proposal to rework the matchmaking ui a little. Apparently some people simply refuse to read any text and are therefore unaware of the concept of multiqueueing. We could improve this by making all queues selected by default similar to the factions. For this we need an additional button that triggers the actual queueing. I think we can do this without any change of the protocol by simply changing the logic when the client sends the "start search" messages.

The new button has the additional benefit that the text changes that e.g. inform you about reasons why you are unable to queue are now in a prominent button instead of the gray on black text above the queues.

grafik

BlackYps:
closes #2938 

grafik


BlackYps:
The queue controller uses the queueJoined value to determine if it should request a queue join or leave. I am not sure why that is the case. Do you know @Sheikah45?

src/main/java/com/faforever/client/teammatchmaking/MatchmakingQueueItemController.java
```diff
@@ -140,17 +145,19 @@ private void setQueuePopTimeUpdater(MatchmakerQueueBean queue) {
...
-    if (queueJoined.getValue()) {
-      teamMatchmakingService.leaveQueue(getQueue());
-    } else {
-      teamMatchmakingService.joinQueue(getQueue()).thenAccept(success -> {
-        if (!success) {
-          joinLeaveQueueButton.setSelected(false);
-          refreshingLabel.setVisible(false);
-        }
-      });
+    if (teamMatchmakingController.searchButton.isSelected()) {
```
Sheikah45: This should be a property of the teamMatchmakingService not derived solely from the controller

src/main/resources/i18n/messages.properties
```diff
@@ -960,21 +960,21 @@ teammatchmaking.match.starting = Match launching...
...
-teammatchmaking.queueTitle = Join one or more queues when you are ready\:
-teammatchmaking.queueTitle.inGame = Can't join queues while you are in a game
-teammatchmaking.queueTitle.memberInGame = Can't join queues while someone is in a game
-teammatchmaking.queueTitle.inParty = The party leader will join the queues
-teammatchmaking.queueTitle.inQueue = Your party is now in queue
+teammatchmaking.searchButton = Play Forged Alliance Forever
+teammatchmaking.searchButton.inGame = Can't join matchmaking while you are in a game
+teammatchmaking.searchButton.memberInGame = Can't join matchmaking while someone is in a game
+teammatchmaking.searchButton.inParty = The party leader will join the matchmaking
+teammatchmaking.searchButton.inQueue = Your party is now searching for a game
```
Sheikah45: There isn't really a reason to change the key name I. Doing so just breaks translations that already exists

src/main/resources/theme/play/team_matchmaking.fxml
```diff
@@ -137,8 +137,10 @@
                         </columnConstraints>
                     </GridPane>
                     <Separator/>
-                    <Label fx:id="queueHeadingLabel" styleClass="h1-tmm" text="%teammatchmaking.queueTitle"/>
                     <FlowPane fx:id="queuePane" hgap="12.0"/>
+                    <ToggleButton fx:id="searchButton" mnemonicParsing="false" styleClass="tmm, start-search"
+                                  alignment="CENTER" maxWidth="1.7976931348623157E308"
+                                  text="%teammatchmaking.searchButton" onAction="#onSearchButtonClicked"/>
```
Sheikah45, line 142: Should be above the queue pane

Sheikah45 (changes requested):

Sheikah45:
> The queue controller uses the queueJoined value to determine if it should request a queue join or leave. I am not sure why that is the case. Do you know @Sheikah45?

queueJoined is directly bound to the MatchmakerQueue with null safety. So it is the same as checking the joined property of the queue but allows for reusability of the controller

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingController.java
```diff
@@ -345,15 +346,38 @@ public void onChatMessage(ChatMessageEvent event) {
...
+            }
+          });
+        }
+      });
+    } else {
+      controllers.forEach(controller -> {
+        if (controller.joinLeaveQueueButton.isSelected()) {
+          teamMatchmakingService.leaveQueue(controller.getQueue());
+        }
+      });
```
Sheikah45: Rather than using the controllers for this tracking it would be better if there was a separate property of either a queue or a list of preferences that were used. Then we could save and reload it on start like we do for factions.

Sheikah45: Then this whole operation could be a function in the teamMatchmakingService rather than the responsibility of the controller

src/main/resources/i18n/messages.properties
BlackYps: We are also changing the text here. If we don't change the key the old translations will still apply and the users of the client won't even notice that a retranslation is necessary

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingController.java
BlackYps: good idea

BlackYps:
>but allows for reusability of the controller

I don't understand this. Can you elaborate?

Sheikah45:
Looking through the whole implementation, this would probably better be done as handled wholly inside the teamMatchmakingService rather than by the individual controllers. 

It would make more sense for the teamMatchmakingService to have an observable set that represents the active queues. Then a method to start and stop the search for the queues in that set, as well as a listener to start the search for queues that are added in that set and stop for the ones removed from the set when in a searching state. 

Then the controllers just have to worry about adding queues to that set for the queue item controllers and starting and stopping the search for the teamMatchmaking controller

Sheikah45:
> > but allows for reusability of the controller
> 
> I don't understand this. Can you elaborate?

This is in a partial state of rework. My next step was to change the renderQueues function from destroying and recreating the controllers every time to reusing controllers and just changing which queue they represent if necessary.

Similar to the changes made to the vault entity cards

src/main/resources/i18n/messages.properties
Sheikah45: Perhaps although the rewording seems pretty minimal on all but the first I question if it is truly necessary

PR #2945, opened 2023-03-16
Allow watching replays while in queue

BlackYps:
We store game files for replay watching separately if the related option is enabled. This prevents the problems that would occur when the game gets launched with the wrong game version once a match is found.

I also deleted the canUpdate method that always returned true.


BlackYps:
Seems to be working now. Still needs tests

Garanas:
There are more places where you can have issues:

- (1) Shader cache
- (2) Preference file
- (3) Executable

(1) and (3) can be a problem when you were watching a replay from a different featured mod. (2) can be a problem where one game is writing to it (because you're closing it, for example) while the next game is trying to read it (because you're starting it).

Are these taken into account?

BlackYps:
No, can you elaborate how to take these into account?

Garanas:
For (1) and (2) the solution would be to delay starting the ladder instance until we're sure that the replay instance has no remaining file handles lingering anywhere. We kill the process, wait a second or two and then launch the game to join ladder. I assume this already happens to some degree.

(3) is more difficult, if I'm not mistaken then the files are updated as you join the queue. This includes the executable. Updating the executable would therefore always fail when it needs to be changed: either when you start the queue while in a replay or when you start the replay while in a queue. To overcome this you'd need to not just separate the game data files, but also separate the executable. This is not limited to watching replays of different featured mods, it can also happen between game versions (the few days after a release)

BlackYps:
I see there was a misunderstanding. The executable is copied over to a separate directory as well. So 3) shouldn't be an issue. Killing the replay before starting the match seems like a good idea.

Garanas:
I thought with game files you were referring to just the files in the `gamedata` directory, miss understanding indeed 😄

Sheikah45:
I think the pref file might still be a concern though. Would that be a problem if a replay is open when the game starts?

BlackYps:
Ideally people don't fiddle with their pref file when they watch a replay anyway, but I understand that this is not always the case. Shouldn't this be addressed by terminating the replay before starting the ladder match?

Sheikah45:
I didn't think you wanted to terminate the replay

BlackYps:
I didn't originally think of that and it's not implemented yet, but it sounds like the best solution. We circumvent these file issues, it frees up system resources and prevents that people miss that the game is starting because they are in the replay

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -592,7 +594,13 @@ public void startSearchMatchmaker() {
         .toFuture()
         .thenAccept(featuredModBean -> updateGameIfNecessary(featuredModBean, Set.of()))
         .thenCompose(aVoid -> fafServerAccessor.startSearchMatchmaker())
-        .thenCompose(gameLaunchResponse -> downloadMapIfNecessary(gameLaunchResponse.getMapName()).thenCompose(aVoid -> leaderboardService.getActiveLeagueEntryForPlayer(playerService.getCurrentPlayer(), gameLaunchResponse.getLeaderboard()))
+        .thenCompose(gameLaunchResponse -> downloadMapIfNecessary(gameLaunchResponse.getMapName()).thenCompose(aVoid -> {
+              if (isRunning()) {
+                gameKilled = true;
+                process.destroy();
+              }
```
BlackYps: I have no idea how I would test this in a unit test as isRunning() has to be false at the start of the method invocation.

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -451,7 +451,7 @@ public CompletableFuture<Void> runWithReplay(Path path, @Nullable Integer replay
         }))
         .thenRun(() -> {
           try {
-            Process processForReplay = forgedAllianceService.startReplay(path, replayId);
+            Process processForReplay = forgedAllianceService.startReplay(path, replayId, forgedAlliancePrefs.isAllowReplaysWhileInGame());
```
Sheikah45: Why pass the prefs value as a parameter? 

This is just the startReplay function so will only be used for replays so the isAllowReplaysWhileInGame should be entirely invisible to the caller.

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
```diff
@@ -97,8 +104,15 @@ public Process startReplay(Path path, @Nullable Integer replayId) throws IOExcep
   }
 
 
-  public Process startReplay(URI replayUri, Integer replayId) throws IOException {
-    List<String> launchCommand = defaultLaunchCommand().replayUri(replayUri)
+  public Process startReplay(URI replayUri, Integer replayId, boolean useReplayFolder) throws IOException {
```
Sheikah45: We already have ForgedAlliancePrefs injected. just read directly from that.

src/main/java/com/faforever/client/game/GameService.java
Sheikah45: Then that means that players cannot join the queue while a replay is running correct? 

So I would think the check would have to be refactored to allow for that.
Also this will have to account for when people might have multiple replays running at once. As of now this would only kill the last started one.

src/main/java/com/faforever/client/patch/GameBinariesUpdateTaskImpl.java
```diff
@@ -94,6 +93,9 @@ protected Void call() throws Exception {
     log.info("Updating binaries to `{}`", version);
 
     Path exePath = forgedAllianceService.getExecutablePath();
+    if (useReplayFolder) {
+      exePath = dataPrefs.getReplayBinDirectory().resolve(FORGED_ALLIANCE_EXE);
+    }
```
Sheikah45: The forgedAllianceService should be able to retrieve the normal or replay executable path

Sheikah45 (changes requested):
Still needs some changes so people can actually queue with one or more replays started.

Also I don't love the passing of the useReplaysFolder parameter but also not sure how to make it better at the moment.

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
BlackYps: We could do the same in `GameBinariesUpdateTaskImpl`. For the rest we would need to inject ForgedAlliancePrefs first. Should I refactor these as well?

src/main/java/com/faforever/client/game/GameService.java
BlackYps: I guess we would then have to differentiate running a replay and running a game. We still want to prevent joining the queue when you are in a custom game lobby. We could also restrict to only one replay running this way.
Should be easy to implement now that I think about it

BlackYps:
I realized that tracking when a replay is started separately from when an online game is started leads to complicated situations when we still want to make e.g. allowing to join a queue while a replay is running dependend on the preference from the settings.
I'm starting to wonder if it is even worth it to keep this option at all. Having this options introduces if statements all over the place. The additional files in the replay folder is about 200MB per featured mod. So in the worst case we have about 800MB of additional disk space, assuming people actually watch replays from fafdevelop and fafbeta. The cache folder can easily grow to multiple GBs. Each client version alone is just short of 200MB and there are possibly multiple game versions cached as well. We don't really take steps to keep this folder small either, so it seems inconsistent to warn about a feature that will actually have less impact on disk usage than the cache.
We could maybe store our game files for replay watching in the cache folder as well, so they can be easily cleaned and just enable the feature for all people, making the code much nicer.

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
Sheikah45: Well for the game binaries updater that function is actually used for both the regular game and the replay game. And the prefs value is not sufficient to use since it can be true when we are trying to update the regular game files.

Sheikah45:
Yes I was wondering as well if we should just always have the split folders. Then they could be truly separated in the code.

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
BlackYps: Right, good catch

BlackYps:
Sounds good. Why is the patch to the game prefs needed? Do we still need to do this then?
And one more thing I realized. When people are in a custom game lobby and watch a replay they already have the situation that two instances of the game are running. How would we prevent any game prefs issues? On the other hand people have used this feature for a while and so far I can't remember any complaints about issues with that.

Sheikah45:
Yes the patch to the game prefs is to allow multiple instances of the game to run at once.

And with the situation where people are already in a game the first game instance has a lock on the files so no other instance should be able to touch them. And that is the instance with the actual game they will play so has priority. When starting a game with a replay started the replay would have the lock which might not have the settings the player wants for the game itself which is the issue.

BlackYps:
When we remove the game.prefs edit from the settings we have to check and edit automatically. Where would be a good place for this check?

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
```diff
@@ -120,10 +124,21 @@ private LaunchCommandBuilder defaultLaunchCommand() {
...
 
+    return addDebugger(baseCommandBuilder);
+  }
+
+  private LaunchCommandBuilder replayLaunchCommand() {
+    LaunchCommandBuilder baseCommandBuilder = LaunchCommandBuilder.create()
+        .executableDecorator(forgedAlliancePrefs.getExecutableDecorator())
+        .executable(getReplayExecutablePath());
+
+    return addDebugger(baseCommandBuilder);
```
Sheikah45, line 135: We don't need to add the debugger to replays

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -725,35 +737,11 @@ private void onRecentlyPlayedGameEnded(GameBean game) {
     notificationService.addNotification(new PersistentNotification(i18n.get("game.ended", game.getTitle()), Severity.INFO, singletonList(new Action(i18n.get("game.rate"), actionEvent -> eventBus.post(new ShowReplayEvent(game.getId()))))));
   }
 
-  @VisibleForTesting
-  CompletableFuture<Void> spawnTerminationListener(Process process) {
-    return spawnTerminationListener(process, true);
-  }
```
Sheikah45, line 731: Why was this removed? It didn't need to be removed and allows easier method reference.

src/main/java/com/faforever/client/preferences/DataPrefs.java
```diff
@@ -37,6 +37,10 @@ public Path getBinDirectory() {
     return getBaseDataDirectory().resolve(BIN_SUB_FOLDER);
   }
 
+  public Path getReplayBinDirectory() {
+    return getBaseDataDirectory().resolve(CACHE_SUB_FOLDER).resolve(BIN_SUB_FOLDER);
```
Sheikah45: We should not be using the cache folder for in use files

src/main/java/com/faforever/client/patch/GameUpdaterImpl.java
```diff
@@ -144,19 +140,18 @@ private CompletableFuture<Void> downloadMissingSimMods(Set<String> simModUids) {
...
-    }
-    throw new UnsupportedOperationException("No updater available for featured mod: " + featuredMod
-        + " with version:" + version);
+  private CompletableFuture<PatchResult> updateFeaturedMod(FeaturedModBean featuredMod, Integer version, boolean forReplays) {
+    return featuredModUpdaters.stream()
+        .findFirst()
+        .map(updater -> updater.updateMod(featuredMod, version, forReplays))
+        .orElseThrow(() -> new UnsupportedOperationException("No updater available for featured mod: " + featuredMod
+            + " with version: " + version));
   }
```
Sheikah45: Without the canUpdate method this doesn't make much sense. So either add back in the canUpdate to make it clear there is the possibility that a featuredModUpdater can not update a featured mod

Sheikah45 (changes requested):

src/main/java/com/faforever/client/game/GameService.java
BlackYps: It made it harder for me to find all usages of termination listeners and made the code harder for me to understand. So I refactored it to make it more explicit

src/main/java/com/faforever/client/preferences/DataPrefs.java
BlackYps: So use the replay folder again?

src/main/java/com/faforever/client/patch/GameUpdaterImpl.java
BlackYps: It seems like the second half of the sentence is missing

Sheikah45: Oh yeah or just remove the list and only have one

src/main/java/com/faforever/client/preferences/DataPrefs.java
Sheikah45: more likely just a new replayBin folder

src/main/java/com/faforever/client/fa/ForgedAllianceService.java
Garanas: Is this the FAF Debugger?

BlackYps: I think it is

BlackYps: Jip said it's sometimes nice to have it on replays

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -592,7 +585,14 @@ public void startSearchMatchmaker() {
         .toFuture()
         .thenAccept(featuredModBean -> updateGameIfNecessary(featuredModBean, Set.of()))
         .thenCompose(aVoid -> fafServerAccessor.startSearchMatchmaker())
-        .thenCompose(gameLaunchResponse -> downloadMapIfNecessary(gameLaunchResponse.getMapName()).thenCompose(aVoid -> leaderboardService.getActiveLeagueEntryForPlayer(playerService.getCurrentPlayer(), gameLaunchResponse.getLeaderboard()))
+        .thenCompose(gameLaunchResponse -> downloadMapIfNecessary(gameLaunchResponse.getMapName()).thenCompose(aVoid -> {
+              // We need to kill the replay to free the lock on the game.prefs
+              if (isReplayRunning()) {
+                gameKilled = true;
```
Sheikah45: This shouldn't be gameKilled should it?

Sheikah45 (changes requested):

src/main/java/com/faforever/client/game/GameService.java
BlackYps: from my understanding gameKilled prevents the error message to pop up when the game - in this case the replay - terminates with non-zero exit code. This is what we want. Once the game starts gameKilled gets set to false again.

Sheikah45: But to me then this should be some other variable. As it is right now this just seems a little confusing since most of the other processes between game and replay are separated

BlackYps: do you want to have this variable renamed to something like faKilled or a second one introduced?

Sheikah45: A second one introduced would probably be better

BlackYps: I added a second variable, you can have a look but I am not really convinced that this is a good way to handle it. I think it would be more straightforward to have one variable that is used for games and replays (and named accordingly).

src/main/java/com/faforever/client/game/GameService.java
```diff
@@ -774,6 +764,37 @@ CompletableFuture<Void> spawnTerminationListener(Process process, Boolean forOnl
...
+    });
+
+    if (exitCode != 0 && !(gameKilled || replayKilled)) {
+      if (exitCode == -1073741515) {
+        notificationService.addImmediateWarnNotification("game.crash.notInitialized");
+      } else {
+        notificationService.addNotification(new ImmediateNotification(i18n.get("errorTitle"), i18n.get("game.crash", exitCode, logFile.map(Path::toString)
+            .orElse("")), WARN, List.of(new Action(i18n.get("game.open.log"), event -> platformService.reveal(logFile.orElse(operatingSystem.getLoggingDirectory()))), new DismissAction(i18n))));
+      }
+    }
```
Sheikah45: The main thing with breaking apart the gameKilled and replayKilled was so that replays would not interfere with this notification for games that crashed.

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #2951, opened 2023-04-02
Use ZGC for garbage collection

Sheikah45:
On my system noticed the OS reported memory  usage was halved

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -178,7 +178,7 @@ bootRun {
     jvmArgs = [
             "-DnativeDir=build/resources/native",
             "-Dprism.forceGPU=true",
-            "-Xmx712m",
+            "-Xmx1g",
             "-XX:MinHeapFreeRatio=15",
             "-XX:MaxHeapFreeRatio=33",
             "-XX:+HeapDumpOnOutOfMemoryError",
--- a/build.gradle
+++ b/build.gradle
@@ -182,7 +182,6 @@ bootRun {
             "-XX:MinHeapFreeRatio=15",
             "-XX:MaxHeapFreeRatio=33",
             "-XX:+HeapDumpOnOutOfMemoryError",
-            "-XX:+UseStringDeduplication",
             "-XX:+UseZGC",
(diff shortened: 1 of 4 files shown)


BlackYps:
Halved? That's huge!

magge-faf:
Yeah, that is amazing - have found no bugs so far, still testing it.

Sheikah45:
Yeah I ran through some tests with it. Still not entirely sure why. A lot of the memory used by the client is actually off heap, so it is rather hard to track down. So I was a little surprised when using the ZGC collector reported so much less memory usage. I am still investigating some as to why hence why this is not merged but figured I would send it out to be tested so others could confirm if it happened for them as well.

Sheikah45:
And previously this "unknown" memory segment was partly due to the Webviews. But using the ZGC the os reported memory actually correlates with the heap+nonheap memory. Where as before ~500MB were due to either the GC or some other native memory purpose. But I do not know why anything other than the memory of the GC would change just from the GC change.

build.gradle
```diff
@@ -167,15 +167,11 @@ bootRun {
     jvmArgs = [
             "-DnativeDir=build/resources/native",
             "-Dprism.forceGPU=true",
-            "-Xms128m",
-            "-Xmx712m",
+            "-Xmx1g",
```
micheljung, line 181: Why?

faf-client.install4j
```diff
@@ -57,10 +57,8 @@
       <vmOptionsFile mode="content" overwriteMode="3">
         <content># Enter one VM parameter per line
 # Adjust maximum memory usage to 712MiB
--Xmx712m
--XX:ConcGCThreads=1
--XX:ParallelGCThreads=1
--XX:+UseG1GC
+-Xmx1g
```
micheljung: Comment would also need to be adjusted (or just removed).

build.gradle
Sheikah45: With how ZGC operates it needs a little bit more heap overhead in order to operate properly without constantly performing concurrent GCs and kicking up CPU usage

PR #2988, opened 2023-06-09
Fix leaderboard order

BlackYps:
The 3v3 was listed last, because it has the highest id. So we have to switch to a different way to sort them, so they are in the order that you would expect.
grafik


Sheikah45:
So we just switch to sorting by the name? Wouldn't this potentially have the same issue in the future? Maybe we should add an order parameter to the league object

BlackYps:
I agree in principle, but adding an order parameter would be more than one order of magnitude more work. It would require changes to the league service, the commons repo and finally the client. All of this to fix a problem that we *might* have in the future. At the moment I don't know of any plans to add more leagues.
So I propose that for now we roll with this change and if we notice that it actually causes problems in the future, then we can implement the more elaborate solution. Adding a new league would require changes to the league service anyway, so I could then easily add a new order parameter as well.

PR #3003, opened 2023-06-29
Don't use ZGC on win7 or win8

Sheikah45:


micheljung:
Would be good if you explained the motivation behind it in the commit message.

bukajsytlos:
will this change affect Linux users?

Sheikah45:
Zgc is only supported on windows builds greater than win 10 1803 due to older versions lacking a newer windows memory API.

It does not affect Linux users because they don't use install4j and manage their own vmoptions

PR #3013, opened 2023-07-18
Closes #2992 Game, map, mod reviews average score filtering

### LINKED ISSUE #2992: Game, map, mod reviews average score filtering
https://github.com/FAForever/faf-java-api/issues/652

bukajsytlos:
Closes #2992 Game, map, mod reviews average score filtering

### DIFF
--- a/src/main/java/com/faforever/client/fx/JavaFxUtil.java
+++ b/src/main/java/com/faforever/client/fx/JavaFxUtil.java
@@ -36,7 +36,6 @@ import javafx.stage.Stage;
 import javafx.util.StringConverter;
 import javafx.util.converter.NumberStringConverter;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.controlsfx.control.RangeSlider;
 import org.springframework.util.Assert;
@@ -44,6 +43,8 @@ import org.springframework.util.Assert;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.text.DecimalFormat;
+import java.text.ParseException;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
@@ -444,6 +445,13 @@ public final class JavaFxUtil {
 
   public static void bindTextFieldAndRangeSlider(TextField lowValueTextField, TextField highValueTextField,
                                                  RangeSlider rangeSlider) {
+    DecimalFormat numberFormat = (DecimalFormat) DecimalFormat.getInstance();
+    numberFormat.setMaximumFractionDigits(0);
+    bindTextFieldAndRangeSlider(lowValueTextField, highValueTextField, rangeSlider, numberFormat);
+  }
+
+  public static void bindTextFieldAndRangeSlider(TextField lowValueTextField, TextField highValueTextField,
+                                                 RangeSlider rangeSlider, DecimalFormat format) {
     Map.of(
         lowValueTextField, Pair.of(rangeSlider.lowValueProperty(), rangeSlider.getMin()),
         highValueTextField, Pair.of(rangeSlider.highValueProperty(), rangeSlider.getMax())
@@ -454,7 +462,7 @@ public final class JavaFxUtil {
         @Override
         public String toString(Number number) {
           if (!number.equals(value)) {
-            return String.valueOf(number.intValue());
+            return format.format(number);
           } else {
             return "";
           }
@@ -462,10 +470,14 @@ public final class JavaFxUtil {
 
         @Override
         public Number fromString(String string) {
-          if (NumberUtils.isParsable(string)) {
-            return Double.parseDouble(string);
-          } else {
-            if (!string.equals("-") && !string.equals(".")) {
+          String decimalSeparator = Character.toString(format.getDecimalFormatSymbols().getDecimalSeparator());
+          try {
+            Number number = format.parse(string);
+            String decimalSeparatorSuffix = string.endsWith(decimalSeparator) && format.getMaximumFractionDigits() > 0 ? decimalSeparator : "";
+            textField.setText(format.format(number) + decimalSeparatorSuffix);
+            return number;
+          } catch (ParseException e) {
+            if (!string.equals("-") && !string.equals(decimalSeparator)) {
               textField.setText("");
             }
             return value;
--- a/src/main/java/com/faforever/client/query/RangeFilterController.java
+++ b/src/main/java/com/faforever/client/query/RangeFilterController.java
@@ -91,7 +91,7 @@ public class RangeFilterController implements FilterNodeController {
     menu.textProperty().bind(Bindings.createStringBinding(() -> i18n.get("query.rangeFilter", title, lowValue.getText(), highValue.getText()), lowValue.textProperty(), highValue.textProperty()));
   }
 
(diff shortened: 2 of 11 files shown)


src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
```diff
@@ -150,10 +150,12 @@ protected void initSearchController() {
...
     searchController.addRangeFilter("playerStats.ratingChanges.meanBefore", i18n.get("game.rating"),
-        MIN_RATING, MAX_RATING, 100, (value) -> value + 300);
+        MIN_RATING, MAX_RATING, 100, 0, value -> value + 300);
+
+    searchController.addRangeFilter("reviewsSummary.averageScore", i18n.get("reviews.averageScore"),0, 5, 0.5, 1);
 
     searchController.addDateRangeFilter("endTime", i18n.get("game.date"), 1);
-    searchController.addRangeFilter("replayTicks", i18n.get("game.duration"), 0, 60, 1, value -> (int) (value*60*10));
+    searchController.addRangeFilter("replayTicks", i18n.get("game.duration"), 0, 60, 6, 0, value -> (int) (value*60*10));
     searchController.addToggleFilter("validity", i18n.get("game.onlyRanked"), "VALID");
```
Sheikah45, line 159: Why change this to 6?

src/main/java/com/faforever/client/query/RangeFilterController.java
```diff
@@ -44,17 +51,20 @@ public void initialize() {
     valueTransform = (value) -> value;
   }
 
+  @SneakyThrows
   public Optional<List<Condition>> getCondition() {
     List<Condition> conditions = new ArrayList<>();
-    if (!lowValue.getText().isBlank() && rangeSlider.getLowValue() > rangeSlider.getMin()) {
+    double lowValueValue = numberFormat.parse(numberFormat.format(rangeSlider.getLowValue())).doubleValue();
```
Sheikah45: Why are we formating and parsing the text here? The text should be bound to the range slider so we only need to check it isn't blank and can get the range slider value

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
```diff
@@ -150,10 +150,12 @@ protected void initSearchController() {
 
     //TODO: Use rating rather than estimated mean with an assumed deviation of 300 when that is available
     searchController.addRangeFilter("playerStats.ratingChanges.meanBefore", i18n.get("game.rating"),
-        MIN_RATING, MAX_RATING, 100, (value) -> value + 300);
+        MIN_RATING, MAX_RATING, 100, 0, value -> value + 300);
```
Sheikah45: Why change to 250?

src/main/resources/i18n/messages.properties
```diff
@@ -1269,6 +1269,7 @@ game.create.generatedMap2 = Generate the map
 settings.fa.allowIpv6 = Allow the ICE adapter to use IPv6
 settings.fa.allowIpv6.description = Ipv6 causes connection issues for some players. Turn this on if you do not have any IPv6 issues with connections
 home.directory.warning.cyrillic = Warning\: Cyrillic characters in the home directory path may cause problems. Please, avoid using them.
+reviews.averageScore = Avg. reviews score
```
Sheikah45: I would spell out the full word

Sheikah45 (changes requested):

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
bukajsytlos: this is how it looks now
image
and this is how it looks with 1
image
I was thinking about making number of ticks configurable and not bound to increment

src/main/java/com/faforever/client/query/RangeFilterController.java
bukajsytlos: it's normalizing value to desired format. why sent different value to what is displayed (ignoring value transformer :))

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
bukajsytlos: what 250? :D I forgot to change it back, when I was learning/playing with how rangeslider is displaying ticks

src/main/resources/i18n/messages.properties
bukajsytlos: ok

src/main/java/com/faforever/client/query/RangeFilterController.java
Sheikah45: But the low value is set using the formatted value right. So outside double precision issues it should already be the same

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
Sheikah45: I think that would be better than decreasing the precision

bukajsytlos: and yeah 5 is more suitable

src/main/java/com/faforever/client/query/RangeFilterController.java
bukajsytlos: well I was developing most of the time without snapToTicks true and without this code, it looked like this
image
with snapToTicks true it will align after you release slider.
but I am not sure that we want to rely on that option.

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
bukajsytlos: fixed

src/main/java/com/faforever/client/query/RangeFilterController.java
Sheikah45: I would say that it was developed so that the range slider has the true value and the text is just for informing the user really

src/main/java/com/faforever/client/fx/JavaFxUtil.java
```diff
@@ -454,18 +462,22 @@ public static void bindTextFieldAndRangeSlider(TextField lowValueTextField, Text
...
-          } else {
-            if (!string.equals("-") && !string.equals(".")) {
+          String decimalSeparator = Character.toString(format.getDecimalFormatSymbols().getDecimalSeparator());
+          try {
+            Number number = format.parse(string);
+            String decimalSeparatorSuffix = string.endsWith(decimalSeparator) && format.getMaximumFractionDigits() > 0 ? decimalSeparator : "";
+            textField.setText(format.format(number) + decimalSeparatorSuffix);
+            return number;
+          } catch (ParseException e) {
+            if (!string.equals("-") && !string.equals(decimalSeparator)) {
```
Sheikah45, line 480: Is this all for locale handling? I though Decimal format could handle different locales

bukajsytlos: it was there before but hard-coded to US locale (dot as decimal separator). not sure if the main purpose was to be able to set for example 0.8 as `.8` but I left this as a possibility even in other locales

src/main/java/com/faforever/client/query/RangeFilterController.java
bukajsytlos: which text?

Sheikah45: The min an max text boxes

Sheikah45 (approved):

PR #3042, opened 2023-10-21
Show division in replays when available

BlackYps:
WIP
grafik


BlackYps:
The core functionality is there, but it is still a bit rough around the edges.
There is a slight delay because we have to fetch the division info from the api, so at the moment it behaves a bit ugly visually.
I appreciate input how this could be solved better.
~~Similarly  the team title still displays (0) because I didn't manage to properly set the teamrating to null to trigger the removal of the rating display alltogether.~~
Edit: Fixed, but I don't know if the way I did it is good style

src/main/java/com/faforever/client/domain/GamePlayerStatsBean.java
```diff
@@ -25,6 +25,7 @@ public class GamePlayerStatsBean {
   IntegerProperty team = new SimpleIntegerProperty();
   ObjectProperty<Faction> faction = new SimpleObjectProperty<>();
   ObjectProperty<OffsetDateTime> scoreTime = new SimpleObjectProperty<>();
+  ObjectProperty<GameOutcome> result = new SimpleObjectProperty<>();
```
Sheikah45: I would probably just name this outcome

src/main/java/com/faforever/client/game/PlayerCardController.java
```diff
@@ -98,10 +105,13 @@ public void initialize() {
...
         .bind(player.flatMap(PlayerBean::avatarProperty).map(avatarService::loadAvatar).when(showing));
+    divisionImageView.imageProperty()
+            .bind(division.flatMap(SubdivisionBean::smallImageUrlProperty).map(leaderboardService::loadDivisionImage).when(showing));
     playerInfo.textProperty()
         .bind(player.flatMap(PlayerBean::usernameProperty)
-            .flatMap(username -> rating.map(value -> i18n.get("userInfo.tooltipFormat.withRating", username, value))
-                .orElse(i18n.get("userInfo.tooltipFormat.noRating", username)))
+            .map(username -> (rating.get() != null && division.get() == null)
+                ? i18n.get("userInfo.tooltipFormat.withRating", username, rating.get())
+                : i18n.get("userInfo.tooltipFormat.noRating", username))
```
Sheikah45: This is not equivalent because before it was mapped off of rating so it would refresh based on changes to rating where as now we lost that property.

Also if we don't use the division why is it included in the if statement?

src/main/java/com/faforever/client/game/TeamCardController.java
```diff
@@ -43,6 +47,13 @@
 @Slf4j
 @RequiredArgsConstructor
 public class TeamCardController implements Controller<Node> {
+
+  private static final PseudoClass VICTORY = PseudoClass.getPseudoClass("victory");
+  private static final PseudoClass DEFEAT = PseudoClass.getPseudoClass("defeat");
+  private static final PseudoClass DRAW = PseudoClass.getPseudoClass("draw");
+  private static final PseudoClass UNKNOWN = PseudoClass.getPseudoClass("unknown");
+  private static final int UNSET = -1000000;
```
Sheikah45: No magic numbers please just use null

src/main/java/com/faforever/client/game/TeamCardController.java
```diff
@@ -51,19 +62,22 @@ public class TeamCardController implements Controller<Node> {
...
   private final ObjectProperty<RatingPrecision> ratingPrecision = new SimpleObjectProperty<>();
+  private final ObjectProperty<GameOutcome> teamResult = new SimpleObjectProperty<>();
   private final IntegerProperty teamId = new SimpleIntegerProperty();
   private final SimpleChangeListener<List<PlayerBean>> playersListener = this::populateTeamContainer;
   private final ObservableValue<Integer> teamRating = ratingProvider.flatMap(provider -> ratingPrecision.flatMap(precision -> players.map(playerBeans -> playerBeans.stream()
       .map(provider)
       .filter(Objects::nonNull)
       .map(rating -> precision == RatingPrecision.ROUNDED ? RatingUtil.getRoundedRating(rating) : rating)
-      .reduce(0, Integer::sum))));
+      .reduce(Integer::sum).orElse(UNSET))));
```
Sheikah45: null

src/main/java/com/faforever/client/game/TeamCardController.java
```diff
@@ -74,13 +88,18 @@ public void initialize() {
           case GameBean.OBSERVERS_TEAM -> i18n.get("game.tooltip.observers");
           default -> {
             try {
-              yield i18n.get("game.tooltip.teamTitle", id.intValue() - 1, teamRating);
+              if (teamRating == UNSET) {
```
Sheikah45: null

src/main/java/com/faforever/client/game/TeamCardController.java
```diff
@@ -107,6 +128,27 @@ private List<PlayerCardController> createPlayerCardControllers(List<PlayerBean>
       return controller;
     }).toList();
   }
+  
+  public void showGameResult() {
+    switch (teamResult.get()) {
```
Sheikah45: You should make sure the other pseudoClassStateChanges are set to false

src/main/java/com/faforever/client/game/TeamCardController.java
```diff
@@ -51,19 +62,22 @@ public class TeamCardController implements Controller<Node> {
...
   public Label teamNameLabel;
+  public Label gameResultLabel;
 
   private final ObjectProperty<List<Integer>> playerIds = new SimpleObjectProperty<>(List.of());
   private final ObjectProperty<List<PlayerBean>> players = new SimpleObjectProperty<>(List.of());
   private final ObjectProperty<Function<PlayerBean, Integer>> ratingProvider = new SimpleObjectProperty<>();
+  private final ObjectProperty<Function<PlayerBean, SubdivisionBean>> divisionProvider = new SimpleObjectProperty<>();
   private final ObjectProperty<Function<PlayerBean, Faction>> factionProvider = new SimpleObjectProperty<>();
   private final ObjectProperty<RatingPrecision> ratingPrecision = new SimpleObjectProperty<>();
+  private final ObjectProperty<GameOutcome> teamResult = new SimpleObjectProperty<>();
```
Sheikah45: Would name it teamOutcome

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -300,6 +305,15 @@ private void onReplayChanged(ReplayBean newValue) {
     if (newValue.getReplayFile() != null) {
       enrichReplayLater(newValue.getReplayFile(), newValue);
     }
+    
+    newValue.setLeagueScores(null);
+    replayService.getLeagueScoreJournalForReplay(newValue)
+        .thenAccept(scores -> Platform.runLater(() -> {
```
Sheikah45: Use the FxApplicationExecutor by passing it to the thenAcceptAsync method

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -424,6 +438,10 @@ private CompletableFuture<Void> enrichReplayLater(Path path, ReplayBean replay)
   private void populateTeamsContainer(Map<String, List<GamePlayerStatsBean>> newValue) {
     CompletableFuture.supplyAsync(() -> createTeamCardControllers(newValue)).thenAcceptAsync(controllers -> {
       teamCardControllers.clear();
+      if (controllers.stream()
+          .map(TeamCardController::getTeamResult).noneMatch(gameOutcome -> gameOutcome != GameOutcome.DEFEAT)) {
+        controllers.forEach(teamCardController -> teamCardController.setTeamResult(GameOutcome.DRAW));
+      }
```
Sheikah45: Why is this happening here instead of in createTeamCardControllers?

It seems like it should be done when the controllers are created rather than setting it to one value just to change it right after. I imagine we have all the information when the controllers are created.

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -300,6 +305,15 @@ private void onReplayChanged(ReplayBean newValue) {
...
       enrichReplayLater(newValue.getReplayFile(), newValue);
     }
+    
+    newValue.setLeagueScores(null);
+    replayService.getLeagueScoreJournalForReplay(newValue)
+        .thenAccept(scores -> Platform.runLater(() -> {
+          newValue.setLeagueScores(scores);
+          // This looks a bit ugly. Ideally we should wait with drawing the window until we have the league scores,
+          // then we don't need to trigger a redraw here
+          populateTeamsContainer(teams.getValue());
```
Sheikah45: As far as I can tell the league scores are not even used in the populateTeamsContainer method so why do we need to call it again here?

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -449,20 +469,53 @@ private List<TeamCardController> createTeamCardControllers(Map<String, List<Game
     }).toList();
   }
 
+  private GameOutcome calculateGameResult(Map<PlayerBean, GamePlayerStatsBean> statsByPlayer) {
```
Sheikah45: Just pass in a collection of GamePlayerStatsBean because we never use the keys

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -449,20 +469,53 @@ private List<TeamCardController> createTeamCardControllers(Map<String, List<Game
...
+        .map(gameOutcome -> (gameOutcome == GameOutcome.MUTUAL_DRAW) ? GameOutcome.DRAW : gameOutcome)
+        .collect(Collectors.groupingBy(gameOutcome -> gameOutcome, Collectors.counting()));
+
+    if (outcomeCounts.containsKey(GameOutcome.VICTORY)) {
+      return GameOutcome.VICTORY;
+    }
+
+    return outcomeCounts.entrySet()
+        .stream()
+        .max(Entry.comparingByValue()).map(Entry::getKey).orElse(GameOutcome.UNKNOWN);
```
Sheikah45: Why is victory a special case? Shouldn't it be covered by the comparison below. Or it seems like there should be more special cases.

At the moment it isn't exactly clear why the outcome is being decided this way.

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -449,20 +469,53 @@ private List<TeamCardController> createTeamCardControllers(Map<String, List<Game
...
     GamePlayerStatsBean playerStats = statsByPlayerId.get(player);
     return playerStats == null ? null : playerStats.getFaction();
   }
 
   private Integer getPlayerRating(PlayerBean player, Map<PlayerBean, GamePlayerStatsBean> statsByPlayerId) {
     GamePlayerStatsBean playerStats = statsByPlayerId.get(player);
-    return playerStats == null ? null : playerStats.getLeaderboardRatingJournals()
+    if ((!replay.get().getLeagueScores().isEmpty()) || playerStats == null) {
+      return null;
+    }
```
Sheikah45: Why do the league scores need to be empty for a player rating to be returned?

src/main/java/com/faforever/client/replay/ReplayDetailController.java
```diff
@@ -449,20 +469,53 @@ private List<TeamCardController> createTeamCardControllers(Map<String, List<Game
...
         .stream()
+        .filter(leagueScoreJournalBean -> leagueScoreJournalBean.getLoginId() == player.getId())
         .findFirst()
-        .filter(ratingJournal -> ratingJournal.getMeanBefore() != null)
-        .filter(ratingJournal -> ratingJournal.getDeviationBefore() != null)
-        .map(RatingUtil::getRating)
-        .orElse(null);
+        .map(LeagueScoreJournalBean::getDivisionBefore)
+        .orElse(null)
+    ).getValue();
```
Sheikah45: We shouldn't be using the replayProperty here because this just creates additional observables that aren't needed. We probably just want to pass in a map of leagueScoresByPlayer like we do for the rating.

src/test/java/com/faforever/client/builders/LeagueScoresListBuilder.java
Sheikah45, line 1: I don't think a list builder is really needed since all the methods here are exposed by the list class directly

Sheikah45 (changes requested):

src/main/java/com/faforever/client/domain/GamePlayerStatsBean.java
BlackYps: I named it this way, because that is how it is called in the api

Sheikah45: I would still just name it outcome since that will make it clearer in this code what it represents.

src/main/java/com/faforever/client/replay/ReplayDetailController.java
BlackYps: Sometimes not all players report victory in the team that won. I think it happens when a player dies before the end of the match, then they report "defeat" for themselves. From my empirical testing I concluded that as soon as we detect a victory, we can assume that that team has won, even if everyone else in the team reports defeat

Sheikah45: Wouldn't that be the case as well for draw? So does it make more sense rather than taking the mode taking the max reported result where Max is defined as priority?

src/main/java/com/faforever/client/game/PlayerCardController.java
BlackYps: when would the rating change? This is the rating associated with a particular replay, not the "live" rating.
 Losing the refresh was not a conscious decision though. These chained mappings elude me a little bit.
What do you mean we don't use the division? The intention is to hide the rating when there is a division instead.

src/main/java/com/faforever/client/game/TeamCardController.java
BlackYps: I can't use null here. If I do that the entire title disappears. Presumably because of a NPE when trying to map the teamrating.

src/main/java/com/faforever/client/replay/ReplayDetailController.java
BlackYps: They are used because we create the teamCardControllers again. Then we set the divison provider in line 462. If I don't do this the division info is not displayed, because the additionial api call needs a moment to complete.

src/main/java/com/faforever/client/replay/ReplayDetailController.java
BlackYps: I want to hide the player rating when we have league scores instead. The idea is that the league info is the new standard and rating is only used as a fallback

src/main/java/com/faforever/client/replay/ReplayDetailController.java
BlackYps: I don't really understand what you mean. Can you rephrase that?

src/main/java/com/faforever/client/replay/ReplayDetailController.java
BlackYps: I need to check if both controllers are reporting defeat. So I can't decide this in a single controller. Or do you mean to just move these lines as is to the end of the createTeamCardControllers function?

src/main/java/com/faforever/client/game/TeamCardController.java
BlackYps: I thought this was implicit because only one could be active. I can add setting the others to false

src/main/java/com/faforever/client/game/PlayerCardController.java
Sheikah45: The rating would change if we changed the player value of the card.

And if the intention is to hide the rating when we have the division then we should modify the visible property in that case and not the actual value.

src/main/java/com/faforever/client/game/TeamCardController.java
Sheikah45: Then it would still be better to handle that NPE where it occurs

src/main/java/com/faforever/client/replay/ReplayDetailController.java
BlackYps: So I change the start to this? `replay.get().getLeagueScores()` what exactly is the difference?

src/main/java/com/faforever/client/replay/ReplayDetailController.java
Sheikah45: See if anyone reported victory, draw, defeat, unknown and take the first one because the default is everyone reports defeat so we only need to check if someone reported something other than that

src/main/java/com/faforever/client/replay/ReplayDetailController.java
Sheikah45: As above then we should be modifying the visibility of the elements and not the values

src/main/java/com/faforever/client/replay/ReplayDetailController.java
Sheikah45: Using the property creates a bunch of intermediate objects that we never end up using.

src/main/java/com/faforever/client/game/PlayerCardController.java
Sheikah45: And at a high level the maps are kind of like streams where map is used for converting a value while flat map is used for converting an observable value.

They create derivative observables that are efficient to subscribe to and react to changes upstream

BlackYps: but there is no visibility property to modify. The formating is done through using a different i18n key

Sheikah45: Then I would probably change it to a rating label and a division label to make it clearer rather than trying to use the same thing for both

src/main/java/com/faforever/client/game/TeamCardController.java
Sheikah45: PsuedoClassStates are a list so you can have multiple active at the same time.

src/main/java/com/faforever/client/replay/ReplayDetailController.java
Sheikah45: Is draw not separately reported from defeat by the API?
And if that is the case then we shouldn't set the team result in the card controller because it is misleading

src/main/java/com/faforever/client/replay/ReplayDetailController.java
Sheikah45: We shouldn't have to recreate the teamCardControllers again. It should be handled like the showRatingChange where setting a value for the teamCardControllers will trigger the display rather than it needing to be recreated.

src/main/java/com/faforever/client/game/TeamCardController.java
Sheikah45: You would need to add an OrElse mapping after the flatmap where it is used

src/main/java/com/faforever/client/game/PlayerCardController.java
Sheikah45: I would also generally say that we shouldn't be returning null from methods like getPlayerRating just because the division is set. The display of the values should be happening at a higher level

Sheikah45:
@BlackYps were you ever going to get back to this?

There have also been a few changes to the underlying objects as an FYI.

BlackYps:
I don't plan to abandon this, on the other hand I am really not motivated to do the refactoring that is required here.
So I guess the honest answer is, I really don't know...

Sheikah45:
Then I might just close this in a week and it can be reopened when you revisit it.

BlackYps:
I'm picking this up again

src/main/java/com/faforever/client/mapstruct/ReplayMapper.java
```diff
@@ -136,8 +145,8 @@ default HashMap<String, List<GamePlayerStats>> mapTeamStatsFromParser(ReplayData
                                                                               null);
         Float factionFloat = (Float) armyInfo.get("Faction");
         Faction faction = Faction.fromFaValue(factionFloat.intValue());
-        GamePlayerStats stats = new GamePlayerStats(player, (byte) 0, team, faction, null,
-                                                    List.of(ratingJournal));
+        GamePlayerStats stats = new GamePlayerStats(false, faction, (byte) 0, team, (byte) 0, (byte) 0, null,
```
BlackYps, line 148: What is the purpose of the `mapTeamStatsFromParser` function? I noticed that the GamePlayerStats only used some of the fields that were provided by the api, so I expanded it, but this function creates a GamePlayerStats object and feeds it data from a parsed replay, but I don't know which of the additional fields can be extracted from the parser.

BlackYps:
The population of the values of the controllers and the displaying of these values are now nicely decoupled. There is one issue though. In a matchmaker game the people that don't have a division still display the rating and the team rating is displayed as well. The problem is that the player cards without a division don't know about the presence of divisions in different player cards. We need some way to check if we found any league stats for the replay (the api call) and then disable rating display for the whole thing somehow

src/main/java/com/faforever/client/mapstruct/ReplayMapper.java
Sheikah45: It is exactly as you say that it populates the team stats for local replay files. Local replays don't have all the data from the api so only fill in some fields

Sheikah45:
For displaying the whole thing you can just have a display type enum that has a property in the player card. Then you can set that property appropriately to display either rating or league. Could even just bind it to some property in the parent.

src/main/java/com/faforever/client/game/TeamCardController.java
```diff
@@ -70,18 +85,21 @@ public class TeamCardController extends NodeController<Node> {
...
             } catch (NumberFormatException e) {
               yield "";
             }
           }
-        })));
-
+        }));
+    JavaFxUtil.bindManagedToVisible(teamRatingLabel, gameResultLabel);
+    teamRatingLabel.textProperty().bind(teamRating.map(value -> i18n.get("game.tooltip.ratingFormat", value)));
+    teamRatingLabel.visibleProperty().bind(Bindings.createBooleanBinding(() -> teamRating.getValue() != null && teamRating.getValue() != 0, teamRating));
```
Sheikah45:
```suggestion
    teamRatingLabel.visibleProperty().bind(teamRating.map(value -> value != 0));
```

src/main/java/com/faforever/client/mapstruct/ReplayMapper.java
BlackYps: so how should I deal with this situation? Put in some dummy data like I did here?

BlackYps:
I guess a boolean would even be enough, something like `gameHasLeagueScore`. I don't have to differentiate multiple states, just this binary decision

src/main/java/com/faforever/client/replay/ReplayDetailController.java
BlackYps: Ah yeah, I see now what you mean. I think you are correct. So that means I don't even need to count, but I just need to check the presence of results in that order.

src/main/java/com/faforever/client/replay/ReplayDetailController.java
BlackYps: Months ago when I was starting to develop this, I found a game that reported me that both teams were defeated. I didn't save the replay Id, so I can't check now. It might have been because of my own faulty calculations at that time. E.g. only a minority reported a draw, so the majority of defeat results got chosen instead.
There is no way to know now, so I will remove this check and if we ever run into a replay that reports both teams as defeated we can investigate again.

Sheikah45:
Yeah I was just thinking of an enum to make it clearer what it signifies. Since hasLeagueScore doesn't imply as much that ratings won't be shown.

Where as something like display type would make that clearer.

BlackYps:
yeah, I can see that.

Sheikah45:
Was testing this out and one thing that stuck out as a little bit odd to me. Why do we require pressing the show game result to show the game outcome on the team card. Maybe we should just display it always on load?

src/main/java/com/faforever/client/mapstruct/ReplayMapper.java
Sheikah45: Yeah that is fine

BlackYps:
I think this kind as introduced to make it possible to watch a game without being spoiled on which side wins

Sheikah45:
Sure but now you can also play directly from the search result page

BlackYps:
Hmm, that's true. Maybe we should discuss this with more users to better understand how they use this button at the moment. It's definitely an interesting suggestion, but imo it is unrelated to this PR

Sheikah45:
What is the status of this?

BlackYps:
From my pov it is ready

Sheikah45 (commented):
The only consequential thing I see is a race condition between populateTeamContainers and loading the league score results.

BlackYps:
Can I do something about that race condition?

Sheikah45:
You could make sure that you don't try to load the league data until after the team containers are populated

BlackYps:
I think I first need to understand the race condition a little better. As I see it we run the risk that the league info is applied to the old teamcardcontrollers, so they get cleared right after and new controllers for the new replay get added in populateTeamsContainer if this happens after the application of the league info.
Is this the correct one, or did you have a different race condition in mind?
And why is createTeamCardControllers called asyncronously? I do not immediately see any operation in that method that would make this necessary.

Sheikah45:
They are created asynchronously so they don't block the application thread and cause the ui to freeze. And to cure the race condition you would just load the league info after create team controllers is called or make it apart of create team controllers

Sheikah45 (approved):
There was an issue with the team card controllers being added twice but I fixed it.

RoraRaven:
Showing division is nice, but was it necessary to remove the rating change? It means you need to check your profile after every match and keep a spreadsheet if you want to track changes rather than it being plainly displayed in the replays.

BlackYps:
Yes, the divisions are supposed to replace the rating. The rating system is considered to be part of the backend, it's not necessary to show it to the user when the division system exists

RoraRaven:
Are we going to get transparency in how divisions are calculated? Currently divisions are more or less meaningless, every other system (ingame player lists, lobby entry requirements, matchmaking, game quality rating) works on rating.

Divisions are really large and imprecise, you could win 15 games in a row and see no change. Even systems that obfuscate Elo / MMR (eg LoL) show points to measure how far up or down you are inside a division.

BlackYps:
What exactly is the problem you are having? I'm trying to und erstand it so I can hopefully address it in my answer.
And did you have a look at the leaderboard tab? Your comment gives me the impression that you didn't, because you can see points there and it's not possible to win 15 games in a row without changing divisions, because each division is only 10 points big.

RoraRaven:
I wasn't aware about the Leaderboards or Score system at all. I was under the impression that division was based on rating (e.g. Gold V being 800 - 900).

This does answer my questions.

I still don't understand why you would hide information from the player, even if you consider rating to be back-end, but that's your prerogative.

PR #3043, opened 2023-10-28
Load the ICE servers from the new faf-icebreaker API

Brutus5000:
Right now this is an MVP approach for fast testing. The turn selection window still loads the old API and no longer makes sense, because we might want to use Xirsys for everyone. Needs more thought once DDoS is resolved.

### DIFF
--- a/src/main/java/com/faforever/client/api/FafApiAccessor.java
+++ b/src/main/java/com/faforever/client/api/FafApiAccessor.java
@@ -157,9 +157,15 @@ public class FafApiAccessor implements InitializingBean {
         .doOnNext(object -> log.trace("Retrieved {} from /me with type MeResult", object));
   }
 
+  public Flux<IceServer> getIceServers() {
+    return retrieveMonoWithErrorHandling(IceServerResponse.class, apiWebClient.get().uri("/ice/server"))
+        .flatMapIterable(IceServerResponse::servers)
+        .doOnNext(object -> log.trace("Retrieved {} from /ice/server with type IceServer", object));
+  }
+
   public Mono<IceSession> getIceSession(int gameId) {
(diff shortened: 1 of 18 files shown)


src/main/java/com/faforever/client/api/FafApiAccessor.java
```diff
@@ -157,6 +157,11 @@ public Mono<MeResult> getMe() {
         .doOnNext(object -> log.trace("Retrieved {} from /me with type MeResult", object));
   }
 
+  public Mono<IceSession> getIceSession(int gameId) {
+    return retrieveMonoWithErrorHandling(IceSession.class, apiWebClient.get().uri("/ice/session/game/" + gameId))
+        .doOnNext(object -> log.trace("Retrieved {} from /ice/session/game/{} with type MeResult", object, gameId));
```
Sheikah45: We can't use the retrieveMonoWithErrorHandling since that expects JsonApiDocument errors

src/main/java/com/faforever/client/api/IceSession.java
```diff
@@ -0,0 +1,11 @@
+package com.faforever.client.api;
+
+import com.faforever.commons.api.dto.CoturnServer;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
Sheikah45: This is already set at the mapper level so shouldn't be needed

src/main/java/com/faforever/client/api/FafApiAccessor.java
Brutus5000: What can we use instead?

Sheikah45: RetrieveWithErrorHandling is just a wrapper around common error handling so we would just make the call with the required custom error handling

PR #3052, opened 2023-11-20
Tweaked CSS Code for Main Header

MrRowey:
@Sheikah45 This is just a Simple CSS change increasing the Top header padding from 9 -> 20 in windowed mode and 0 -> 20px when in full-screen


Sheikah45:
I took a look deeper further into this and found a better solution of using the title bar as intended so that we get the full intended interactions. See #3056

Issue #2880, opened 2022-12-23
Add search by player name in coop tab 


Marc-Spector:
Leaderboard contains 1000 players. It's hard to understand at all whether you are in the table or not

PR #3053, opened 2023-11-21
Add search bar for coop leaderboard

Kimo-s:
#2880

### DIFF
--- a/src/main/java/com/faforever/client/coop/CoopController.java
+++ b/src/main/java/com/faforever/client/coop/CoopController.java
@@ -58,9 +58,11 @@ import java.time.OffsetDateTime;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -96,10 +98,12 @@ public class CoopController extends AbstractViewController<Node> {
   public Pane gameViewContainer;
   public TextField titleTextField;
   public Button playButton;
+  public TextField leaderboardSearchTextField;
   public PasswordField passwordTextField;
   public ImageView mapPreviewImageView;
   public Region leaderboardInfoIcon;
   public TableView<CoopResultBean> leaderboardTable;
+  public FilteredList<CoopResultBean> leaderboardFilteredTable;
   public ComboBox<Integer> numberOfPlayersComboBox;
   public TableColumn<CoopResultBean, Integer> rankColumn;
   public TableColumn<CoopResultBean, Integer> playerCountColumn;
@@ -124,6 +128,8 @@ public class CoopController extends AbstractViewController<Node> {
     numberOfPlayersComboBox.setCellFactory(param -> numberOfPlayersCell());
     numberOfPlayersComboBox.getSelectionModel().select(0);
     numberOfPlayersComboBox.getSelectionModel().selectedItemProperty().addListener(observable -> loadLeaderboard());
+
+    leaderboardSearchTextField.textProperty().addListener((observable, oldValue, newValue) -> onSearchFieldChange(newValue));
 
     rankColumn.setCellValueFactory(param -> param.getValue().rankingProperty().asObject());
     rankColumn.setCellFactory(param -> new StringCell<>(String::valueOf));
@@ -194,6 +200,20 @@ public class CoopController extends AbstractViewController<Node> {
     });
   }
 
+  private void onSearchFieldChange(String newvalue){
+    leaderboardFilteredTable.setPredicate(coopResultBean -> {
+      String teamstring = "";
+      for(Map.Entry<String, List<String>> entry : coopResultBean.getReplay().getTeams().entrySet()){
+        teamstring = entry.getValue().stream().collect(Collectors.joining(i18n.get("textSeparator")));
+      }
+      if(newvalue.isEmpty() || teamstring.equalsIgnoreCase("")){
+        return true;
+      }
+      return teamstring.toLowerCase().contains(newvalue.toLowerCase());
+    });
+    leaderboardTable.setItems(observableList(leaderboardFilteredTable));
+  }
+
   private String coopMissionFromFolderNamer(List<CoopMissionBean> coopMaps, String mapFolderName) {
     return coopMaps.stream()
         .filter(coopMission -> coopMission.getMapFolderName().equalsIgnoreCase(mapFolderName))
@@ -243,7 +263,10 @@ public class CoopController extends AbstractViewController<Node> {
         .thenAccept(coopLeaderboardEntries -> {
           AtomicInteger ranking = new AtomicInteger();
           coopLeaderboardEntries.forEach(coopResult -> coopResult.setRanking(ranking.incrementAndGet()));
-          fxApplicationThreadExecutor.execute(() -> leaderboardTable.setItems(observableList(coopLeaderboardEntries)));
+          fxApplicationThreadExecutor.execute(() -> {
+            leaderboardTable.setItems(observableList(coopLeaderboardEntries));
+            leaderboardFilteredTable = new FilteredList<>(observableList(coopLeaderboardEntries));
+          });
         })
         .exceptionally(throwable -> {
           log.warn("Could not load coop leaderboard", throwable);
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -243,7 +263,10 @@ private void loadLeaderboard() {
         .thenAccept(coopLeaderboardEntries -> {
           AtomicInteger ranking = new AtomicInteger();
           coopLeaderboardEntries.forEach(coopResult -> coopResult.setRanking(ranking.incrementAndGet()));
-          fxApplicationThreadExecutor.execute(() -> leaderboardTable.setItems(observableList(coopLeaderboardEntries)));
+          fxApplicationThreadExecutor.execute(() -> {
+            leaderboardTable.setItems(observableList(coopLeaderboardEntries));
+            leaderboardFilteredTable = new FilteredList<>(observableList(coopLeaderboardEntries));
```
Sheikah45: Instead of creating the list here we should create the list as a private final field and then add the elements to it.

This would allow us to call the setitems of the leaderboard table only once instead of multiple times.

src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -96,10 +98,12 @@ public class CoopController extends AbstractViewController<Node> {
   public Pane gameViewContainer;
   public TextField titleTextField;
   public Button playButton;
+  public TextField leaderboardSearchTextField;
   public PasswordField passwordTextField;
   public ImageView mapPreviewImageView;
   public Region leaderboardInfoIcon;
   public TableView<CoopResultBean> leaderboardTable;
+  public FilteredList<CoopResultBean> leaderboardFilteredTable;
```
Sheikah45: This should not have table in the name since it is just the list and not the table object

src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -125,6 +129,8 @@ public void initialize() {
     numberOfPlayersComboBox.getSelectionModel().select(0);
     numberOfPlayersComboBox.getSelectionModel().selectedItemProperty().addListener(observable -> loadLeaderboard());
 
+    leaderboardSearchTextField.textProperty().addListener((observable, oldValue, newValue) -> onSearchFieldChange(newValue));
```
Sheikah45:
```suggestion
    leaderboardSearchTextField.textProperty().subscribe(this::onSearchFieldChange);
```
JavaFX has better syntax for this now

src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -194,6 +200,20 @@ public void initialize() {
...
+    leaderboardFilteredTable.setPredicate(coopResultBean -> {
+      String teamstring = "";
+      for(Map.Entry<String, List<String>> entry : coopResultBean.getReplay().getTeams().entrySet()){
+        teamstring = entry.getValue().stream().collect(Collectors.joining(i18n.get("textSeparator")));
+      }
+      if(newvalue.isEmpty() || teamstring.equalsIgnoreCase("")){
+        return true;
+      }
+      return teamstring.toLowerCase().contains(newvalue.toLowerCase());
+    });
```
Sheikah45: We are looking for a team member in the list of teams. So it would be more efficient to use the anyMatch of a stream rather than concatenating all the names together and doing a contains.

Sheikah45: Additionally we can set the predicate in the initialize function by binding it to the text field value.

src/main/java/com/faforever/client/coop/CoopController.java
Sheikah45: Although really we can better set the predicate with a binding to the text field

Sheikah45 (changes requested):

src/main/java/com/faforever/client/coop/CoopController.java
Sheikah45: Also it looks like this only checks the last team in the coop result bean instead of all teams

src/main/java/com/faforever/client/coop/CoopController.java
Kimo-s: > Although really we can better set the predicate with a binding to the text field

I don't know how to do this. Would be helpful if you give an example.

src/main/java/com/faforever/client/coop/CoopController.java
Kimo-s: > Instead of creating the list here we should create the list as a private final field and then add the elements to it.
> 
> This would allow us to call the setitems of the leaderboard table only once instead of multiple times.

I tried to use the setAll method on the list but it gives null error. Something like this: 

```leaderboardFilteredList.setAll(observableList(coopLeaderboardEntries));```

I initialized the list as ```private FilteredList<CoopResultBean> leaderboardFilteredList = new FilteredList<>(FXCollections.observableArrayList());```

Comment (Sheikah45) [reply src/main/java/com/faforever/client/coop/CoopController.java]: What did it say was null?

File: src/main/java/com/faforever/client/coop/CoopController.java [hunk 5]
```diff
@@ -194,6 +204,22 @@ public void initialize() {
...
+      boolean foundmatch = false;
+      for(Map.Entry<String, List<String>> entry : coopResultBean.getReplay().getTeams().entrySet()){
+        foundmatch =  entry.getValue().stream().anyMatch((s) -> s.toLowerCase().contains(newValue.toLowerCase()));
+        if(foundmatch)
+          break;
+      }
+      if(newValue.isEmpty()){
+        return true;
+      }
+      return foundmatch;
```
Sheikah45: This is still more verbose than necessary. You call entryset but we are only using the values.
Also if newValue.isempty() we don't care about looping through the teams so we should check that first.

I would use the .values() method and the flatMap method of a stream and you should be able to make this much more readable.

src/main/java/com/faforever/client/coop/CoopController.java
Sheikah45: bindings are used to auto listen to changes in another property in javaFX. Coupled with the map and flatmap methods on observable values that are similar to those on the java streams they can create easy listener structures.

This is already done with a few other predicateProperty in the code so if you search for predicateProperty().bind you should find some examples. Then you just need a function here that takes the text and turns it into the desired predicate and can use that as `predicateProperty().bind(textProperty().map(this::createSearchPredicate))`

src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -194,6 +201,24 @@ public void initialize() {
     });
   }
 
+  private Predicate<CoopResultBean> onSearchCreatePredict(String newValue){
+    leaderboardTable.setItems(observableList(leaderboardFilteredList));
```
Sheikah45: This should be moved into the initialize function.

You also don't need to wrap the filtered list in an observablelist

Sheikah45 (changes requested):

src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -125,6 +131,21 @@ public void initialize() {
     numberOfPlayersComboBox.getSelectionModel().select(0);
     numberOfPlayersComboBox.getSelectionModel().selectedItemProperty().addListener(observable -> loadLeaderboard());
 
+    leaderboardSearchTextField.textProperty().subscribe(s -> {
+      leaderboardTable.setItems(leaderboardFilteredList);
+    });
```
Sheikah45: The set items shouldn't be in a listener. It just needs to be called once in the initialize function

src/main/java/com/faforever/client/coop/CoopController.java
```diff
@@ -243,7 +264,10 @@ private void loadLeaderboard() {
         .thenAccept(coopLeaderboardEntries -> {
           AtomicInteger ranking = new AtomicInteger();
           coopLeaderboardEntries.forEach(coopResult -> coopResult.setRanking(ranking.incrementAndGet()));
-          fxApplicationThreadExecutor.execute(() -> leaderboardTable.setItems(observableList(coopLeaderboardEntries)));
+          fxApplicationThreadExecutor.execute(() -> {
+            leaderboardUnFilteredList.setAll(coopLeaderboardEntries);
+            leaderboardTable.setItems(observableList(leaderboardFilteredList));
```
Sheikah45: We don't need to set the items again

Sheikah45 (changes requested):

Sheikah45 (approved):

Issue #2881, opened 2022-12-23
Handle clicking the "Up" button to insert the last message in chat channel


Marc-Spector:
Useful for correcting typos

PR #3050, opened 2023-11-18
Overhaul Client State Handling

Sheikah45:
Currently the client maintains most state in the various UI controllers. This results in the view elements needing to be cached so as to not lose state. This is suboptimal because it has lead to memory leaks where some ui elements cannot be garbage collected even when they are removed from the scene.

Additionally there is no cleanup of elements and their listeners when they are removed from the UI which also contributes to memory leaks.

This PR aims to overhaul the state maintenance of the client such that views are just displays and can be created and discarded at will.

Tasks
- [x] Remove Google Event Bus (Deprecated Global Event Bus)
- [x] Don't force chat controller to be initialized
- [x] Remove controller caching
- [x] Remove controller memory leaks
- [x] Introduce simple controller initialization and cleanup
- [x] Cleanup circular dependencies
- [x] Make UI population lazy

Closes #2994 
Closes #3016 
Closes #3018

PR #3059, opened 2023-11-27
Chat up and down arrow key scrolls over chat channel message history 

Kimo-s:
#2881

### DIFF
--- a/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
+++ b/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
@@ -119,7 +119,6 @@ public abstract class AbstractChatTabController extends TabController {
   private static final String ACTION_CSS_CLASS = "action";
   private static final String MESSAGE_CSS_CLASS = "message";
 
-  protected final List<String> messageHistory = new ArrayList<>();;
   protected final LoginService loginService;
   protected final ChatService chatService;
   protected final PlayerService playerService;
@@ -379,8 +378,6 @@ public abstract class AbstractChatTabController extends TabController {
     } else {
       sendMessage();
     }
-
-    messageHistory.add(text);
 
     hideEmoticonsWindow();
   }
--- a/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
+++ b/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
@@ -119,6 +119,7 @@ public abstract class AbstractChatTabController extends TabController {
   private static final String ACTION_CSS_CLASS = "action";
   private static final String MESSAGE_CSS_CLASS = "message";
 
+  protected final List<String> messageHistory = new ArrayList<>();;
   protected final LoginService loginService;
   protected final ChatService chatService;
   protected final PlayerService playerService;
@@ -378,6 +379,8 @@ public abstract class AbstractChatTabController extends TabController {
     } else {
       sendMessage();
     }
+
+    messageHistory.add(text);
 
     hideEmoticonsWindow();
   }
--- a/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
+++ b/src/main/java/com/faforever/client/chat/AbstractChatTabController.java
@@ -141,7 +141,7 @@ public abstract class AbstractChatTabController extends TabController {
    * Messages that arrived before the web view was ready. Those are appended as soon as it is ready.
    */
   private final List<ChatMessage> waitingMessages = new ArrayList<>();
-  private final List<ChatMessage> userMessageHistory = new ArrayList<>();
+  private final List<String> userMessageHistory = new ArrayList<>();
   private final IntegerProperty unreadMessagesCount = new SimpleIntegerProperty();
   protected final ObjectProperty<ChatChannel> chatChannel = new SimpleObjectProperty<>();
   protected final ObservableValue<String> channelName = chatChannel.map(ChatChannel::getName);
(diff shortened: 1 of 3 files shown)


Sheikah45 (commented):
I will probably put this one on hold until #3050 is merged because it reworks how the chat messages are handled and then this becomes easier to do in a consistent way.

Sheikah45:
Now that the state overhaul has been deployed you can use this with the chat messages kept in the chatChannel object once you rebase

src/main/java/com/faforever/client/chat/ChatUserItemController.java
```diff
@@ -101,6 +101,8 @@ public class ChatUserItemController extends NodeController<Node> {
   protected void onInitialize() {
     initializeTooltips();
     bindProperties();
+
+
```
Sheikah45: Unneeded whitespace

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -189,6 +208,22 @@ public AutoCompletionHelper getAutoCompletionHelper() {
                                                         .collect(Collectors.toList()));
   }
 
+  private void onUpOrDownArrowKeyClick(KeyEvent event){
+    if(event.getCode() == KeyCode.UP){
```
Sheikah45: Whitespace between if and brackets everywhere

src/main/java/com/faforever/client/chat/ChannelTabController.java
```diff
@@ -165,6 +174,16 @@ protected void onInitialize() {
...
 
+    chatChannel.when(attached).subscribe((oldValue, newValue) -> {
+      if(newValue != null){
+        newValue.addMessageListener(chatMessage -> {
+          if(chatMessage.username().equals(playerService.getCurrentPlayer().getUsername())){
+            userMessageHistory.add(chatMessage);
+          }
+        });
+      }
+    });
```
Sheikah45: Should also unsubscribe from the oldValue when changed

Sheikah45 (changes requested):
This is generally applicable across all chat channels so should be in the abstractTabController not the ChannelTabController

Sheikah45 (changes requested):
After testing there are a few things that should probably change to make this usable:

- There is no way currently to get back to a blank/in progress message
- You can easily lose the message you are typing by pressing the up or down arrow key accidentally
- when there is text it was taking two button presses to traverse the message history

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -139,12 +141,14 @@ public abstract class AbstractChatTabController extends TabController {
    * Messages that arrived before the web view was ready. Those are appended as soon as it is ready.
    */
   private final List<ChatMessage> waitingMessages = new ArrayList<>();
+  private final List<ChatMessage> userMessageHistory = new ArrayList<>();
+  private String currentUserMessage = "";
+  private int curMessageHistoryIndex = 0;
```
Sheikah45: Move these down with the other private non final variable declarations

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -431,6 +465,9 @@ private void sendAction(final TextInputControl messageTextField, final String te
   }
 
   protected void onChatMessage(ChatMessage chatMessage) {
+    if(chatMessage.username().equals(playerService.getCurrentPlayer().getUsername())) {
+      userMessageHistory.add(chatMessage);
```
Sheikah45: We should make sure to limit this to a max size. Lets just go with 50

Sheikah45 (commented):
Looking good just some minor things.

Additionally it would be good to place the cursor at the end of the text after the up or down arrow navigation

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -431,6 +466,14 @@ private void sendAction(final TextInputControl messageTextField, final String te
...
 
   protected void onChatMessage(ChatMessage chatMessage) {
+    if(chatMessage.username().equals(playerService.getCurrentPlayer().getUsername())) {
+      if(userMessageHistory.size() >= 50){
+        userMessageHistory.remove(0);
+        userMessageHistory.add(chatMessage);
+      } else {
+        userMessageHistory.add(chatMessage);
+      }
+    }
```
Sheikah45: One thing I realized is that this is probably better off put in the send message method. Then you don't have to check the username

src/main/java/com/faforever/client/chat/AbstractChatTabController.java
```diff
@@ -212,6 +221,32 @@ private void clearUnreadIfFocused() {
...
 
+  private void onUpOrDownArrowKeyClick(KeyEvent event){
+    if(event.getCode() == KeyCode.DOWN || event.getCode() == KeyCode.UP){
+      if(curMessageHistoryIndex == 0){
+        currentUserMessage = messageTextField().getText();
+      }
+      if(event.getCode() == KeyCode.DOWN)
+        curMessageHistoryIndex--;
+      if(event.getCode() == KeyCode.UP)
+        curMessageHistoryIndex++;
```
Sheikah45: Add brackets to the if statement

Sheikah45 (changes requested):

Issue #3023, opened 2023-08-17
Enhance Replay Card Information


BlackYps:
It would be nice to already see the factions that the players have in this replay card. This way it is easier (i.e. requires less clicking) to identify interesting replays.
Additionally the replay id could be displayed near the bottom, next to the play button, to enable copying of the replay id.

![grafik](https://github.com/FAForever/downlords-faf-client/assets/52536103/e382ebe3-6f06-4ebd-878d-02146956f307)

PR #3084, opened 2023-12-11
Display more player information for replay cards

obydog002:
https://github.com/FAForever/downlords-faf-client/issues/3023

This PR only addresses the first part and not the replay id. Adds player factions to replays and allows right click on player names


obydog002:
faf-replay-teams-demo

Sheikah45:
The cutoff of the player names is a bigger loss than the gain from the extra information from my point of view

obydog002:
Hmm, I agree. Would it then make sense to change the style such that more of the names are visible (since its left aligned) and give more space for the teams, or remove the rating for players? 

Or should I not bother?

Sheikah45:
I think the names shouldn't be cutoff really in any situation for a less than 2 team game.

As far as the styling outside of that I don't have great guidelines unfortunately

obydog002:
3teams

demo2

obydog002:
Theres a bit of a highlighting issue when mousing over the names to click

Brutus5000:
Your example screens will probably break if people have avatars. Or where these removed?

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -135,22 +145,31 @@ protected void onInitialize() {
...
       teamBox.getChildren().add(teamLabel);
-      team.forEach(player -> teamBox.getChildren().add(new Label(player)));
+
+      playerStats.forEach(player -> {
+        PlayerCardController controller = uiService.loadFxml("theme/player_card.fxml");
+        controller.setPlayer(player.getPlayer());
+        controller.setFaction(player.getFaction());
+        controller.avatarStackPane.setVisible(false);
+        controller.avatarStackPane.setManaged(false);
+        controller.avatarImageView.setManaged(false);
```
Sheikah45: The UI elements are only public to allow fxml to set them. If we want an external element to be able to control their state we should introduce a setter.

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -135,22 +145,31 @@ protected void onInitialize() {
...
-      String teamLabelText = id.equals("1") ? i18n.get("replay.noTeam") : i18n.get("replay.team", Integer.parseInt(id) - 1);
+      String teamLabelText = team.equals("1") ? i18n.get("replay.noTeam") : i18n.get("replay.team", Integer.parseInt(team) - 1);
       Label teamLabel = new Label(teamLabelText);
       teamLabel.getStyleClass().add("replay-card-team-label");
       teamLabel.setPadding(new Insets(0, 0, 5, 0));
       teamBox.getChildren().add(teamLabel);
-      team.forEach(player -> teamBox.getChildren().add(new Label(player)));
+
+      playerStats.forEach(player -> {
+        PlayerCardController controller = uiService.loadFxml("theme/player_card.fxml");
```
Sheikah45: Ideally we avoid loading fxml on the application thread so we might want to put this into a background thread where the results are used in the application thread.

Unless the player card controller creates a web view somewhere.

src/main/java/com/faforever/client/game/PlayerCardController.java
```diff
@@ -169,6 +171,7 @@ public void openContextMenu(MouseEvent event) {
           .build()
           .show(getRoot().getScene().getWindow(), event.getScreenX(), event.getScreenY());
     }
+    event.consume();
```
Sheikah45, line 174: Why do you consume the event here?

Sheikah45 (changes requested):

src/main/java/com/faforever/client/game/PlayerCardController.java
obydog002: Without this line clicking the player name also results in the detail card opening

obydog002:
@Brutus5000 thats a good point, didnt consider that

Sheikah45:
@obydog002 What is the status of this?

obydog002:
Had a holiday, but still progressing the PR. Will have some updates in the week

obydog002:
Theres a few problems:
- I still don't know what happens with avatars, I cant see any avatars in e.g. the replay detail screen even for players that have them (bugged?)
- There is a bit of lag in the player names/factions appearing when navigating to another page in the replays

src/main/java/com/faforever/client/game/PlayerCardController.java
```diff
@@ -213,6 +216,12 @@ private Integer getRatingChange(LeaderboardRatingJournalBean ratingJournal) {
     return null;
   }
 
+  public void removeAvatar() {
+    this.avatarImageView.managedProperty().set(false);
+    this.avatarStackPane.managedProperty().set(false);
+    this.avatarStackPane.visibleProperty().set(false);
```
Sheikah45: You set the avatar to invisible here.

Also just use the setVisible method rather than going through the property.

And in the initialize use the utility function JavafxUtil.bindManagedToVisible on the stack pane.

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -135,25 +153,46 @@ protected void onInitialize() {
...
+        Label teamLabel = new Label(teamLabelText);
+        teamLabel.getStyleClass().add("replay-card-team-label");
+        teamLabel.setPadding(new Insets(0, 0, 5, 0));
+        teamCard.getChildren().add(teamLabel);
+
+        playerStats.forEach(player -> {
+          PlayerCardController controller = uiService.loadFxml("theme/player_card.fxml");
+          controller.setPlayer(player.getPlayer());
+          controller.setFaction(player.getFaction());
+          controller.removeAvatar();
```
Sheikah45, line 178: And here is where you call it

Sheikah45 (commented):
The lag is natural because we have to download the images and load them into the UI

src/main/java/com/faforever/client/game/PlayerCardController.java
obydog002: Fair enough, I meant that without any changes I havent seen an avatar anyway (like on the 12.3 client), supposedly they should be visible in the replaydetailcards

obydog002:
> The lag is natural because we have to download the images and load them into the UI

Is there any way to preemptively download the images so they appear instantly?

Sheikah45:
It is the imageview object actually getting initialized with the pixels.

We already do cache all the images but every image and imageview is set to load in the background so as to not block the UI

obydog002:
What else needs to be done for this? @Sheikah45

Sheikah45:
I will run it and check out the visuals. Can you also remove some of the unused imports.

src/main/java/com/faforever/client/game/PlayerCardController.java
```diff
@@ -213,6 +216,11 @@ private Integer getRatingChange(LeaderboardRatingJournalBean ratingJournal) {
     return null;
   }
 
+  public void removeAvatar() {
+    this.avatarImageView.managedProperty().set(false);
```
Sheikah45: This isn't needed since avatarStackPane contains it

Sheikah45 (approved):
This all checks out to me. Is it ready for merge from your point of view?

obydog002:
Yes, Ill add ReplayCardController tests in a seperate PR.

PR #3085, opened 2023-12-13
Add a biome selector to the map generation dialog

BlackYps:
Requires https://github.com/FAForever/Neroxis-Map-Generator/pull/382 to be merged and released to actually be able to load the biomes list

grafik

### DIFF
--- a/src/main/java/com/faforever/client/game/CreateGameController.java
+++ b/src/main/java/com/faforever/client/game/CreateGameController.java
@@ -409,10 +409,9 @@ public class CreateGameController extends NodeController<Pane> {
   public void onGenerateMapButtonClicked() {
     GenerateMapController generateMapController = uiService.loadFxml("theme/play/generate_map.fxml");
     mapGeneratorService.getNewestGenerator()
-        .thenCompose(aVoid -> mapGeneratorService.getGeneratorStyles())
-        .thenAccept(generateMapController::setStyles)
-        .thenCompose(aVoid -> mapGeneratorService.getGeneratorBiomes())
-        .thenAccept(generateMapController::setBiomes)
+        .thenCompose(aVoid -> CompletableFuture.allOf(
+            mapGeneratorService.getGeneratorStyles().thenAccept(generateMapController::setStyles), 
+            mapGeneratorService.getGeneratorBiomes().thenAccept(generateMapController::setBiomes)))
         .thenRunAsync(() -> {
           Pane root = generateMapController.getRoot();
           generateMapController.setCreateGameController(this);
--- a/src/main/java/com/faforever/client/game/CreateGameController.java
+++ b/src/main/java/com/faforever/client/game/CreateGameController.java
@@ -411,6 +411,8 @@ public class CreateGameController extends NodeController<Pane> {
     mapGeneratorService.getNewestGenerator()
         .thenCompose(aVoid -> mapGeneratorService.getGeneratorStyles())
         .thenAccept(generateMapController::setStyles)
+        .thenCompose(aVoid -> mapGeneratorService.getGeneratorBiomes())
+        .thenAccept(generateMapController::setBiomes)
         .thenRunAsync(() -> {
           Pane root = generateMapController.getRoot();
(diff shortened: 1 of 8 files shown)


src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -399,6 +399,8 @@ public void onGenerateMapButtonClicked() {
     mapGeneratorService.getNewestGenerator()
         .thenCompose(aVoid -> mapGeneratorService.getGeneratorStyles())
         .thenAccept(generateMapController::setStyles)
+        .thenCompose(aVoid -> mapGeneratorService.getGeneratorBiomes())
+        .thenAccept(generateMapController::setBiomes)
```
Sheikah45: These should be two futures that can run in parallel rather than in series

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -236,7 +239,18 @@ private void initMapStyleComboBox() {
     mapStyleComboBox.disableProperty()
         .bind(previousMapName.textProperty().isNotEmpty()
             .or(generationTypeComboBox.valueProperty().isNotEqualTo(GenerationType.CASUAL))
-            .or(commandLineArgsText.textProperty().isNotEmpty()));
+            .or(commandLineArgsText.textProperty().isNotEmpty())
+            .or(biomeComboBox.valueProperty().isNotNull())
+                .and(biomeComboBox.valueProperty().isNotEqualTo(MapGeneratorService.GENERATOR_RANDOM_BIOME)));
```
Sheikah45: I am pretty sure the and should go directly on the isNotNull

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -236,7 +239,18 @@ private void initMapStyleComboBox() {
...
+  }
+
+  private void initBiomeComboBox() {
+    biomeComboBox.disableProperty()
+        .bind(previousMapName.textProperty().isNotEmpty()
+            .or(generationTypeComboBox.valueProperty().isNotEqualTo(GenerationType.CASUAL))
+            .or(commandLineArgsText.textProperty().isNotEmpty())
+            .or(mapStyleComboBox.valueProperty().isNotNull())
+               .and(mapStyleComboBox.valueProperty().isNotEqualTo(MapGeneratorService.GENERATOR_RANDOM_STYLE)));
```
Sheikah45: Same as above with the parentheses

Sheikah45 (changes requested):

BlackYps:
Now that the text for random style and biome is the same, should the two constants GENERATOR_RANDOM_STYLE and GENERATOR_RANDOM_BIOME be consolidated into one?

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -358,6 +373,20 @@ protected void setStyles(List<String> styles) {
     mapStyleLabel.setVisible(true);
   }
 
+  protected void setBiomes(List<String> biomes) {
+    biomes.add(0, MapGeneratorService.GENERATOR_RANDOM_BIOME);
```
Sheikah45: Looking at this now this and set styles really should make a copy of the list that is guaranteed to be mutable

Sheikah45 (changes requested):

src/main/java/com/faforever/client/game/GenerateMapController.java
BlackYps: how do I do that and why is that desirable?

Sheikah45: It is desirable because this method could be passed an immutable list so the add function would fail.

Or if a mutable list is passed this then modifies that list in ways that the caller may not be expecting so it breaks something higher up.

You would implement it by just creating a new list and then adding this list to that one.

Sheikah45 (approved):

PR #3106, opened 2023-12-30
Use non webview based chat

Sheikah45:


micheljung:
@Sheikah45 why was this done? I feel like the chat experience got way worse because of this.

Sheikah45:
This was done to allow for replies and reactions in the chat.

And to make the event handling and tooltips more uniform.

Some issues appeared for some users with scrolling where permanent fixes are still being investigated.

Additionally there were memory issues with the webview and how the chat state was handled so this was also used to clean all that up as well.

PR #3124, opened 2024-02-11
Replay card controller tests

obydog002:
Add tests for ReplayCardController

withThreeTeamsShowClickForMore is intermittently failing, think its an issue with a race condition but unsure how to fix


obydog002:
Removed the intermittent test, let me know if there is any else I should test

Sheikah45 (approved):

PR #3132, opened 2024-02-19
Keep map selection when searching

### LINKED ISSUE #3092: Typing in map search clears map selection and name preview
**Describe the bug**
Typing in the search box clears the map name in the preview on the right, and the map selection in the list.
Search Bug

**Expected behavior**
The map name and map selection should remain.

**OS**
Windows

obydog002:
Fix #3092 

Now retains the selected option


Sheikah45 (changes requested):
Instead of introducing a mutable variable that we have to manually remember to set and track, I think it would be better to use the biconsumer version of subscribe on the selectedItem property in initMapSelection. This provides the new and old selected values, then if the new value is null we can select the old value if it is contained by filtered maps.

Effectively moving that logic outside of the setSelectedMap function.

Which is better for the long term as well since most of those properties set in setSelected map will be changed to be bound to derived properties eventually.

Sheikah45 (dismissed):
The old item null check is superfluous but that is fine

src/main/java/com/faforever/client/game/CreateGameController.java
```diff
@@ -108,6 +108,8 @@ public class CreateGameController extends NodeController<Pane> {
 
   private final SimpleInvalidationListener createButtonStateListener = this::setCreateGameButtonState;
 
+  private MapVersionBean lastMapChosen = null;
+
```
Sheikah45: This field can be removed, also did you want to write a test for this?

obydog002: Oh yep Ill remove that and add some tests

Sheikah45 (approved):

PR #3134, opened 2024-02-22
Replay id visible on replay cards

obydog002:
Complete https://github.com/FAForever/downlords-faf-client/issues/3023

### DIFF
--- a/src/main/resources/theme/vault/replay/replay_card.fxml
+++ b/src/main/resources/theme/vault/replay/replay_card.fxml
@@ -97,7 +97,7 @@
           GridPane.columnIndex="2" GridPane.rowIndex="8"
           GridPane.hgrow="ALWAYS" GridPane.vgrow="ALWAYS">
         <TextField fx:id="replayIdField" alignment="CENTER_RIGHT" editable="false"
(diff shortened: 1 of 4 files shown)


src/main/resources/theme/vault/replay/replay_card.fxml
```diff
@@ -95,6 +96,8 @@
     <HBox spacing="10.0" alignment="BOTTOM_RIGHT" GridPane.halignment="RIGHT" GridPane.valignment="BOTTOM"
           GridPane.columnIndex="2" GridPane.rowIndex="8"
           GridPane.hgrow="ALWAYS" GridPane.vgrow="ALWAYS">
+        <TextField fx:id="replayIdField" alignment="CENTER_RIGHT" editable="false"
+                maxWidth="100.0" promptText="&lt;ReplayID&gt;" text="#123456789"/>
```
Sheikah45: Why is the prompt text set? It should never be shown since the value is always set.

If it is intended to be set it should be a translatable value.

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #3151, opened 2024-03-15
Feature/#2319 veto system: simple maplist popup

K-ETFreeman:
"Maplist" button on the Play tab now summons popup with maplist instead of navigating to the vault.
This is "Simple" version of the popup, shows maps without duplicates, with grayscaling of map tiles


src/main/java/com/faforever/client/map/MapService.java
```diff
@@ -627,37 +631,18 @@ public Mono<Void> downloadAllMatchmakerMaps(MatchmakerQueueInfo matchmakerQueue)
...
-                                                                                                   Comparator.comparing(
-                                                                                                       Map::displayName,
-                                                                                                       Comparator.nullsLast(
-                                                                                                           String.CASE_INSENSITIVE_ORDER)))))));
-    return Mono.zip(matchmakerMapsFlux.skip((long) (page - 1) * count).take(count).collectList(),
-                    matchmakerMapsFlux.count().map(size -> (int) (size - 1) / count + 1));
+
+    String customFilter = ((String) new QBuilder().and(conditions).query(new RSQLVisitor()));
+
+    return fafApiAccessor.getMany(navigator, customFilter)
```
Sheikah45: You don't need the custom filter since it is just empty.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingController.java
```diff
@@ -331,6 +335,24 @@ private void selectFactions(List<Faction> factions) {
...
+    dialogLayout.setBody(controller.getRoot());
+    Dialog dialog = new Dialog();
+    dialog.setContent(dialogLayout);
+    dialog.setTransitionType(DialogTransition.CENTER);
+    teamMatchmakingRoot.setOnKeyPressed(event -> {
+      if (event.getCode() == KeyCode.ESCAPE) {
+        dialog.close();
+      }
+    });
+    controller.init(queue, dialogLayout);
```
Sheikah45: This is done backwards. The controller should not need to know it is being placed in a dialog.

You should only need to use the getRoot method to get the parent element of the controller and add it to the dialog.

Additionally queue should have its own setter.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+public class TeamMatchmakingMapListController extends NodeController<Pane> {
+
+  public Pane root;
+  public FlowPane tilesContainer;
+  private final MapService mapService;
+  private final UiService uiService;
+  private final I18n i18n;
+  private final PlayerService playerService;
+  private final FxApplicationThreadExecutor fxApplicationThreadExecutor;
+  private final int TILE_SIZE = 125;
```
Sheikah45: This should be static

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+  public ScrollPane scrollContainer;
+  public VBox loadingPane;
+  private Map<MatchmakerQueueMapPool, List<MapVersion>> rawBrackets;
+  private SortedMap<MatchmakerQueueMapPool, List<MapVersion>> sortedBrackets;
+  private SortedMap<MatchmakerQueueMapPool, List<MapVersion>> sortedBracketsWithDuplicates;
+  private Integer playerBracketIndex = null;
+  private Pane dialogRoot;
```
Sheikah45: Please group the properties by constants, injected, UI, instance final, instance mutable.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+    JavaFxUtil.bindManagedToVisible(loadingPane);
+  }
+
+  @Override
+  public Pane getRoot() {
+    return root;
+  }
+
+
+  public void init(MatchmakerQueueInfo queue, Pane dialogRoot) {
```
Sheikah45: This should just be a setter for the queue

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+    this.dialogRoot = dialogRoot;
+    mapService.getMatchmakerBrackets(queue).subscribe(rawBrackets -> {
+      loadingPane.setVisible(false);
+      this.rawBrackets = rawBrackets;
+      this.sortedBracketsWithDuplicates = this.getSortedBrackets(rawBrackets);
+      this.sortedBrackets = this.removeDuplicates(this.getSortedBrackets(rawBrackets));
```
Sheikah45: You are doing the sorted work here twice when it needs to only be done once

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+
+      PlayerInfo player = playerService.getCurrentPlayer();
+      LeaderboardRating ratingBean = player.getLeaderboardRatings().get(queue.getLeaderboard().technicalName());
+      double rating = ratingBean.mean() - 3 * ratingBean.deviation();
```
Sheikah45: Rating could be null

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+      this.playerBracketIndex = this.getPlayerBracketIndex(this.sortedBrackets, rating);
+
+      List<MapVersion> values = this.sortedBrackets.values().stream().flatMap(List::stream).toList();
+      this.resizeToContent(values.size(), this.TILE_SIZE);
+
+      Flux.fromIterable(values)
+          .publishOn(fxApplicationThreadExecutor.asScheduler())
+          .doOnNext(this::addMapTile)
+          .subscribe();
```
Sheikah45: This doesn't need to be a flux you can just run this in the fx application thread executor.

And really this parent flux should be published on the fx application executor so this is unnecessary.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+                                                                           String.CASE_INSENSITIVE_ORDER);
+
+    Comparator<MatchmakerQueueMapPool> mapPoolComparator = Comparator.comparingDouble(MatchmakerQueueMapPool::minRating)
+                                                                     .thenComparingDouble(
+                                                                         MatchmakerQueueMapPool::maxRating);
+
+    SortedMap<MatchmakerQueueMapPool, List<MapVersion>> sortedMap = new TreeMap<>(mapPoolComparator);
+
+    brackets.forEach((mapPool, mapVersions) -> {
+      List<MapVersion> sortedList = mapVersions.stream().sorted(mapVersionComparator).collect(Collectors.toList());
```
Sheikah45: Use .toList instead of .collect(Collectors.toList())

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+      sortedMap.put(mapPool, sortedList);
+    });
+
+    return sortedMap;
+  }
+
+  private SortedMap<MatchmakerQueueMapPool, List<MapVersion>> removeDuplicates(SortedMap<MatchmakerQueueMapPool, List<MapVersion>> sortedBrackets) {
+    HashSet<String> usedMaps = new HashSet<>();
```
Sheikah45: Used maps should be initialized with HashSet but typed with just Set

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+    sortedBrackets.replaceAll((k, v) -> sortedBrackets.get(k).stream().filter(mapVersion -> {
+      String name = mapVersion.map().displayName();
+      if (usedMaps.contains(name))
+        return false;
+      usedMaps.add(name);
+      return true;
+    }).collect(Collectors.toList()));
+    return sortedBrackets;
```
Sheikah45: This is modifying the map in place but the function implies it is returning a new object. So this needs to either return a copy or just modify in place.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+      double min = pool.minRating();
+      double max = pool.maxRating();
+      if (min == 0.0) {min = Double.NEGATIVE_INFINITY;}
+      if (max == 0.0) {max = Double.POSITIVE_INFINITY;}
+
+      if (rating < max && rating > min)
+        return i;
+      i++;
+    }
+    return null;
```
Sheikah45: If the rating can be null they should be changed to Integer instead of int.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  private void addMapTile(MapVersion mapVersion) {
+    int i = 0;
+    List<Integer> indexes = new ArrayList<>();
+    for (var maps : this.sortedBracketsWithDuplicates.values()) {
```
Sheikah45: We don't use var in this code base

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+    int diff = Collections.min(indexes.stream().map(idx -> Math.abs(idx - this.playerBracketIndex)).toList());
+    double relevanceLevel = switch (diff) {
+      case 0 -> 1;
+      case 1 -> 0.2;
+      default -> 0;
+    };
+
+    TeamMatchmakingMapTileController controller = uiService.loadFxml(
+        "theme/play/teammatchmaking/matchmaking_map_tile.fxml");
+    controller.init(mapVersion, relevanceLevel);
```
Sheikah45: These should be independent setters

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,198 @@
...
+      i++;
+    }
+    return null;
+  }
+
+  private void addMapTile(MapVersion mapVersion) {
+    int i = 0;
+    List<Integer> indexes = new ArrayList<>();
+    for (var maps : this.sortedBracketsWithDuplicates.values()) {
+      if (maps.contains(mapVersion)) indexes.add(i);
```
Sheikah45: Use brackets for if statements

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
```diff
@@ -0,0 +1,109 @@
...
+  public Label sizeLabel;
+  public VBox authorBox;
+
+  @Override
+  public Pane getRoot() {
+    return root;
+  }
+
+
+  public void init(MapVersion mapVersion, double relevanceLevel) {
```
Sheikah45: These should be separate setters

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
```diff
@@ -0,0 +1,109 @@
...
+    thumbnailImageView.effectProperty().bind(entity.map(mapVersion -> {
+      ColorAdjust grayscaleEffect = new ColorAdjust();
+      grayscaleEffect.setSaturation(-1 + this.relevanceLevel);
+      return grayscaleEffect;
+    }));
+    ObservableValue<Map> mapObservable = entity.map(MapVersion::map);
+
+    nameLabel.textProperty().bind(mapObservable.map(map -> {
+      String name = map.displayName();
+      if(mapGeneratorService.isGeneratedMap(name))
```
Sheikah45: Brackets and white soace

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
```diff
@@ -0,0 +1,109 @@
...
+        return "map generator";
+      return name;
+    }));
+
+    authorBox.visibleProperty().bind(mapObservable.map(map -> (map.author() != null) || (mapGeneratorService.isGeneratedMap(map.displayName()))));
+    authorLabel.textProperty().bind(mapObservable.map(map -> {
+      if (map.author() != null) {
+        return map.author().usernameProperty().get();
```
Sheikah45: Just use getUsername

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
```diff
@@ -0,0 +1,109 @@
...
+      } else if (mapGeneratorService.isGeneratedMap(map.displayName())) {
+        return "Neroxis";
+      } else {
+        return "";
```
Sheikah45: There is a translation key for unknown author

Sheikah45 (changes requested):
This is a first pass review.

More generally there are lots of places where you set just a basic object in the controllers but this could lead to null pointer exceptions. It would be better to use a property to represent these fields like queue, pool, relevance. Then you can also use the binding and mapping on the properties so that the reactive changes are handled for you.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingController.java
K-ETFreeman: But i need dialog root in my controller to calculate width/height of the popup reliably (because i need to control vertical space which is taken by dialog itself, title, paddings, etc)
I can do it with cycled Node.getParent as alternative, but it is less reliable because it is dialog-structure-dependent

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: > This should be static

you mean TILE_SIZE only, right?

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Need to clarify the situation with dialogRoot dependency first to solve this

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: A small price for readability, as i see it (allows to avoid unnecessary code). But ok.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Not sure what did you mean by "And really this parent flux should be published on the fx application executor so this is unnecessary."

For now i replaced all these 4 lines with
`fxApplicationThreadExecutor.execute(()->values.forEach(this::addMapTile));`

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Can you explain why? Just wanna know more

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: to Double, you mean?
I also changed minRating and maxRated types from double to Double in MatchmakerQueueMapPool.java, otherwise no way to distinct null from 0.0

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
Sheikah45: yes

K-ETFreeman:
Hi @Sheikah45 , some adjustments were made, i think it can be reviewed once more

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,205 @@
...
+
+      this.sortedBracketsWithDuplicates = this.getSortedBrackets(rawBrackets);
+
+      this.sortedBrackets = new TreeMap<>(this.sortedBracketsWithDuplicates);
+      this.sortedBrackets.replaceAll((key, value) -> new ArrayList<>(value));
+      this.removeDuplicates(this.sortedBrackets);
+
+      PlayerInfo player = playerService.getCurrentPlayer();
+      LeaderboardRating ratingBean = player.getLeaderboardRatings().get(queue.getLeaderboard().technicalName());
+      double rating = (ratingBean != null) ? (ratingBean.mean() - 3 * ratingBean.deviation()) : 0;
```
Sheikah45: There is a util class RatingUtils I think for doing this

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,205 @@
...
+  public FlowPane tilesContainer;
+  public ScrollPane scrollContainer;
+  public VBox loadingPane;
+  private SortedMap<MatchmakerQueueMapPool, List<MapVersion>> sortedBrackets;
+  private SortedMap<MatchmakerQueueMapPool, List<MapVersion>> sortedBracketsWithDuplicates;
+  private Integer playerBracketIndex = null;
+  @Setter
+  private double maxWidth = 0;
+  @Setter
+  private double maxHeight = 0;
```
Sheikah45: These should be properties rather than just primitives so they can be responsive.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,205 @@
...
+    });
+
+  }
+
+
+  private SortedMap<MatchmakerQueueMapPool, List<MapVersion>> getSortedBrackets(Map<MatchmakerQueueMapPool, List<MapVersion>> brackets) {
+
+    Comparator<MapVersion> mapVersionComparator = Comparator.nullsFirst(Comparator.comparing(MapVersion::size))
+                                                            .thenComparing(mapVersion -> mapVersion.map().displayName(),
+                                                                           String.CASE_INSENSITIVE_ORDER);
```
Sheikah45: The comparator should be a constant

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,205 @@
...
+
+    Comparator<MatchmakerQueueMapPool> mapPoolComparator = Comparator.comparing(MatchmakerQueueMapPool::minRating, Comparator.nullsFirst(Double::compare))
```
Sheikah45: Same here should be a constant

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,205 @@
...
+
+  private void removeDuplicates(SortedMap<MatchmakerQueueMapPool, List<MapVersion>> sortedBrackets) {
+    Set<String> usedMaps = new HashSet<>();
+    sortedBrackets.replaceAll((k, v) -> sortedBrackets.get(k).stream().filter(mapVersion -> {
+      String name = mapVersion.map().displayName();
+      if (usedMaps.contains(name)) {
+        return false;
+      }
+      usedMaps.add(name);
+      return true;
```
Sheikah45: The sortedBrackets.get(k) is unnecessary you already have access to the value. Also please name the variables appropriately.

Sheikah45: Also rather than replacing it would be better to just return a new map. Then you don't have to copy all the lists as well when you create sorted brackets.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,205 @@
...
+  private final PlayerService playerService;
+  private final FxApplicationThreadExecutor fxApplicationThreadExecutor;
+
+  public Pane root;
+  public FlowPane tilesContainer;
+  public ScrollPane scrollContainer;
+  public VBox loadingPane;
+  private SortedMap<MatchmakerQueueMapPool, List<MapVersion>> sortedBrackets;
+  private SortedMap<MatchmakerQueueMapPool, List<MapVersion>> sortedBracketsWithDuplicates;
+  private Integer playerBracketIndex = null;
```
Sheikah45: The bracket index should also be a property

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,205 @@
...
+
+  @Override
+  public Pane getRoot() {
+    return root;
+  }
+
+
+  public void setQueue(MatchmakerQueueInfo queue) {
+    mapService.getMatchmakerBrackets(queue).subscribe(rawBrackets -> {
+      loadingPane.setVisible(false);
```
Sheikah45: Loading pane should be set visible in this method

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,205 @@
...
+      LeaderboardRating ratingBean = player.getLeaderboardRatings().get(queue.getLeaderboard().technicalName());
+      double rating = (ratingBean != null) ? (ratingBean.mean() - 3 * ratingBean.deviation()) : 0;
+      this.playerBracketIndex = this.getPlayerBracketIndex(this.sortedBrackets, rating);
+
+      List<MapVersion> values = this.sortedBrackets.values().stream().flatMap(List::stream).toList();
+      this.resizeToContent(values.size(), TILE_SIZE);
+
+
+      fxApplicationThreadExecutor.execute(() -> values.forEach(this::addMapTile));
+    });
```
Sheikah45: A lot of this method sets values that are better off as properties so they can be more responsive and in line with the rest of the code base.

I would suggest taking a look at how some of the map methods are used by properties in other controllers.

This allows for greater flexibility in the  controller design so the developer doesn't have to make sure things are called in a specific order.

Additionally when already in a mono or flux pipeline rather than using fxapplicationexecutor.execute directly it is better to return the value and then use the publishOn method of the publisher.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,205 @@
...
+    TeamMatchmakingMapTileController controller = uiService.loadFxml(
+        "theme/play/teammatchmaking/matchmaking_map_tile.fxml");
+    controller.setRelevanceLevel(relevanceLevel);
+    controller.setMapVersion(mapVersion);
+    this.tilesContainer.getChildren().add(controller.getRoot());
+  }
+
+  private void resizeToContent(int tilecount, int tileSize) {
+    double maxWidth = (this.maxWidth != 0) ? this.maxWidth :  Screen.getPrimary().getVisualBounds().getWidth();
+    double maxHeight = (this.maxHeight != 0) ? this.maxHeight : Screen.getPrimary().getVisualBounds().getHeight();
```
Sheikah45: Making these properties will remove the need to use the Screen static methods which could cause issues in tests.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
```diff
@@ -0,0 +1,114 @@
...
+  public Pane root;
+  public ImageView thumbnailImageView;
+  public Label nameLabel;
+  public Label authorLabel;
+  public Label sizeLabel;
+  public VBox authorBox;
+
+  protected final ObjectProperty<MapVersion> entity = new SimpleObjectProperty<>();
+  @Setter
+  private double relevanceLevel = 1;
```
Sheikah45: This should be a property

Sheikah45 (changes requested):

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingController.java
```diff
@@ -331,6 +335,15 @@ private void selectFactions(List<Faction> factions) {
     factionsToButtons.forEach((faction, toggleButton) -> toggleButton.setSelected(factions.contains(faction)));
   }
 
+  protected void onMapPoolClickedListener(MatchmakerQueueInfo queue) {
```
Sheikah45, line 338: This method can just be onMapPoolClicked

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,227 @@
...
+      PlayerInfo player = playerService.getCurrentPlayer();
+      Integer rating = RatingUtil.getLeaderboardRating(player, queue.getLeaderboard());
+
+      this.playerBracketIndex.set(this.getPlayerBracketIndex(this.sortedBrackets, rating));
+
+      List<MapVersion> values = this.sortedBrackets.values().stream().flatMap(List::stream).toList();
+      this.resizeToContent(values.size(), TILE_SIZE);
+
+
+      fxApplicationThreadExecutor.execute(() -> values.forEach(this::addMapTile));
```
Sheikah45: So now here the sortedBrackets, sortedBrackets with duplicates, playerBracketIndex, and the values can be made into properties.

This can be done in the initialize method and you can utilize the map methods on properties to create observable values.

Then you can also add listeners to these properties that call resizeContent. That should cover the bases with regards to this method.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,244 @@
...
+
+    this.queue.addListener((obs, oldVal, newVal) -> {
+      loadingPane.setVisible(true);
+
+      mapService.getMatchmakerBrackets(newVal).subscribe(rawBrackets -> {
+        loadingPane.setVisible(false);
+        this.sortedBracketsWithDuplicates.set(getSortedBrackets(rawBrackets));
+        this.sortedBrackets.set(removeDuplicates(this.sortedBracketsWithDuplicates.get()));
+      });
+    });
```
Sheikah45: Can be simplified with some of the new javafx apis. This would 

```suggestion
    this.queue.subscribe(value -> {
      loadingPane.setVisible(true);

      mapService.getMatchmakerBrackets(value).subscribe(rawBrackets -> {
        loadingPane.setVisible(false);
        this.sortedBracketsWithDuplicates.set(getSortedBrackets(rawBrackets));
      });
    });
```

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,244 @@
...
+
+
+  private final MapService mapService;
+  private final UiService uiService;
+  private final PlayerService playerService;
+  private final FxApplicationThreadExecutor fxApplicationThreadExecutor;
+  private final DoubleProperty maxWidth = new SimpleDoubleProperty(0);
+  private final DoubleProperty maxHeight = new SimpleDoubleProperty(0);
+  private final ObjectProperty<SortedMap<MatchmakerQueueMapPool, List<MapVersion>>> sortedBracketsWithDuplicates = new SimpleObjectProperty<>();
+  private final ObjectProperty<SortedMap<MatchmakerQueueMapPool, List<MapVersion>>> sortedBrackets = new SimpleObjectProperty<>();
```
Sheikah45:
```suggestion
  private final ObservableValue<SortedMap<MatchmakerQueueMapPool, List<MapVersion>>> sortedBrackets = sortedBracketsWithDuplicates.map(this::removeDuplicates);;
  private final ObservableValue<List<MapVersion>> maps = sortedBrackets.map(Map::values).map(values -> values.stream().flatMap(List::stream).toList());
```

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,244 @@
...
+        this.sortedBracketsWithDuplicates.set(getSortedBrackets(rawBrackets));
+        this.sortedBrackets.set(removeDuplicates(this.sortedBracketsWithDuplicates.get()));
+      });
+    });
+
+    this.sortedBrackets.addListener((obs, oldVal, newVal) -> {
+      PlayerInfo player = playerService.getCurrentPlayer();
+      Integer rating = RatingUtil.getLeaderboardRating(player, getQueue().getLeaderboard());
+      this.updatePlayerBracketIndex(newVal, rating);
+    });
```
Sheikah45:
```suggestion
    this.sortedBrackets.subscribe(value -> {
      PlayerInfo player = playerService.getCurrentPlayer();
      Integer rating = RatingUtil.getLeaderboardRating(player, getQueue().getLeaderboard());
      this.updatePlayerBracketIndex(value, rating);
    });
```

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,244 @@
...
+
+    this.playerBracketIndex.addListener((obs, oldVal, newVal) -> {
+      List<MapVersion> values = this.sortedBrackets.get().values().stream().flatMap(List::stream).toList();
+      this.resizeToContent(values.size(), TILE_SIZE);
+      fxApplicationThreadExecutor.execute(() -> values.forEach((mapVersion) -> this.addMapTile(mapVersion, newVal)));
+    });
```
Sheikah45:
```suggestion
    this.maps.subscribe(() -> updateContent());
    this.playerBracketIndex.subscribe(() -> updateContent());
```

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -0,0 +1,244 @@
...
+  }
+
```
Sheikah45, line 116:
```suggestion 
private void updateContent() {
      List<MapVersion> values = maps.get();
      Integer bracket = this.playerBracketIndex.get();
      fxApplicationThreadExecutor.execute(() -> {
          this.resizeToContent(values.size(), TILE_SIZE);
          values.forEach(mapVersion -> this.addMapTile(mapVersion, bracket));
      });
}
```

Sheikah45 (changes requested):
A few ways to use new method signatures to clean things up

Sheikah45:
Also I would make an observable value queue.map(Queue::getLeaderboard) so that null exceptions with queues can be avoided

Sheikah45 (approved):

PR #3157, opened 2024-04-06
Change text on search button

BlackYps:
Changes the text on the search button that is used in the matchmaker tab.
Apparently people regularly don't realize that they need to press this button to actually start searching, so the text is more explicit now.

Btw, do existing translations automatically get some sort of mark that the original text has changed and they need retranslation?


Sheikah45:
They do not that is why a new key should be used instead of reusing an old one.

Issue #3119, opened 2024-01-27
Add persistent preferences for replay search


Sheikah45:
This should allow the search parameters to be remembered between tab opens and client restarts

Issue #3118, opened 2024-01-27
Add preferences for persistent map vault search


Sheikah45:
This should allow the search parameters to be remembered between tab opens and client restarts

PR #3166, opened 2024-05-11
#3118 add persistent properties to replay search page

obydog002:
https://github.com/FAForever/downlords-faf-client/issues/3119

Every field except for the game date filter when searching are now remembered after closing/opening the client and navigating away/to the replay search page.


Sheikah45 (changes requested):
Instead of passing in a property to the add*Filter methods and adding setters to the Filter controllers it would be better to instead expose the relevant value properties of the Filter controllers with appropriate getters and setter so that the creators can then do with as they choose.

Additionally in the vaultPrefs object it is probably better to create sub objects like ReplaySearchPrefs to represent the search state.

src/main/java/com/faforever/client/preferences/ReplaySearchPrefs.java
```diff
@@ -0,0 +1,229 @@
...
+public class ReplaySearchPrefs {
+  private final StringProperty playerNameField = new SimpleStringProperty("");
+  private final StringProperty mapNameField = new SimpleStringProperty("");
+  private final StringProperty mapAuthorField = new SimpleStringProperty("");
+  private final StringProperty titleField = new SimpleStringProperty("");
+  private final StringProperty replayIDField = new SimpleStringProperty("");
+  private final ObjectProperty<ObservableList<String>> featuredModFilter = new SimpleObjectProperty<ObservableList<String>>(
+    FXCollections.emptyObservableList());
+  private final ObjectProperty<ObservableList<String>> leaderboardFilter = new SimpleObjectProperty<ObservableList<String>>(
+    FXCollections.emptyObservableList());
```
Sheikah45: These should be ListProperties

src/main/java/com/faforever/client/query/RangeFilterController.java
```diff
@@ -141,6 +141,14 @@ public void bind() {
     JavaFxUtil.bindTextFieldAndRangeSlider(lowValue, highValue, rangeSlider, numberFormat);
   }
 
+  public javafx.beans.property.DoubleProperty lowValueProperty() {
+    return rangeSlider.lowValueProperty();
+  }
+
+  public javafx.beans.property.DoubleProperty highValueProperty() {
+    return rangeSlider.highValueProperty();
+  }
```
Sheikah45, line 150: import DoubleProperty instead of using the full name

src/main/java/com/faforever/client/query/TextFilterController.java
```diff
@@ -81,4 +81,7 @@ public Node getRoot() {
     return textBox;
   }
 
+  public javafx.beans.property.StringProperty textFieldProperty() {
+    return textField.textProperty();
+  }
```
Sheikah45, line 86: import instead of full name

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
```diff
@@ -131,19 +138,29 @@ protected void initSearchController() {
...
+    TextFilterController textFilterController = searchController.addTextFilter("playerStats.player.login", i18n.get("game.player.username"), true);
+    textFilterController.textFieldProperty().bindBidirectional(replaySearchPrefs.playerNameFieldProperty());
+    textFilterController = searchController.addTextFilter("mapVersion.map.displayName", i18n.get("game.map.displayName"), false);
+    textFilterController.textFieldProperty().bindBidirectional(replaySearchPrefs.mapNameFieldProperty());
+    textFilterController = searchController.addTextFilter("mapVersion.map.author.login", i18n.get("game.map.author"), false);
+    textFilterController.textFieldProperty().bindBidirectional(replaySearchPrefs.mapAuthorFieldProperty());
+    textFilterController = searchController.addTextFilter("name", i18n.get("game.title"), false);
+    textFilterController.textFieldProperty().bindBidirectional(replaySearchPrefs.titleFieldProperty());
+    textFilterController = searchController.addTextFilter("id", i18n.get("game.id"), true);
+    textFilterController.textFieldProperty().bindBidirectional(replaySearchPrefs.replayIDFieldProperty());
```
Sheikah45: Rather than binding bidirectionally we should just set the initial value and then bind the searchPrefs property to the filter property. Additionally since the replaySearch properties are long lived we need to limit the binding to the lifetime of the VaultController by adding .when(showing). This will cause the binding to automatically stop whenever the vault is not showing. This will prevent memory leaks

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
```diff
@@ -131,19 +138,29 @@ protected void initSearchController() {
...
         i18n.get("featuredMod.displayName"), List.of());
 
     featuredModService.getFeaturedMods().map(FeaturedMod::displayName)
                       .collectList()
                       .publishOn(fxApplicationThreadExecutor.asScheduler())
-                      .subscribe(featuredModFilterController::setItems);
+                      .subscribe((mods) -> {
+                        featuredModFilterController.setItems(mods);
+                        replaySearchPrefs.featuredModFilterProperty().get().stream().forEach((item) -> featuredModFilterController.checkItem(item));
+                        replaySearchPrefs.featuredModFilterProperty().bind(Bindings.createObjectBinding(() -> featuredModFilterController.getCheckedItems()));
```
Sheikah45: There is a utility method Bindings.bindContent that can be used here. Although we have to be more careful about unbinding when the vault controller is closed. This could be easily done by adding the unsubscription with the addSubscription method of the controller

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
```diff
@@ -154,18 +171,32 @@ protected void initSearchController() {
...
+    rangeFilterController = searchController.addRangeFilter("reviewsSummary.averageScore", i18n.get("reviews.averageScore"),0, 5, 10, 4, 1);
+    rangeFilterController.lowValueProperty().bindBidirectional(replaySearchPrefs.averageReviewScoresMinProperty());
+    rangeFilterController.highValueProperty().bindBidirectional(replaySearchPrefs.averageReviewScoresMaxProperty());
 
-    searchController.addDateRangeFilter("endTime", i18n.get("game.date"), 1);
-    searchController.addRangeFilter("replayTicks", i18n.get("game.duration"), 0, 60, 12, 4, 0, value -> (int) (value * 60 * 10));
-    searchController.addToggleFilter("validity", i18n.get("game.onlyRanked"), "VALID");
+    DateRangeFilterController dateRangerFilterController = searchController.addDateRangeFilter("endTime", i18n.get("game.date"), 1);
+    dateRangerFilterController.beforeDateProperty().bindBidirectional(replaySearchPrefs.gameBeforeDateProperty());
+    dateRangerFilterController.afterDateProperty().bindBidirectional(replaySearchPrefs.gameAfterDateProperty());
```
Sheikah45: Same binding comments as above

Sheikah45 (changes requested):
One thing we should probably additionally do is set a max date range for the replays. We currently initialize the range to be one year from today due to server load and probably don't want the persistence to undo that.

src/main/java/com/faforever/client/query/RangeFilterController.java
obydog002: This one I had to write the full name as DoubleProperty clashes with the existing com.github.rutledgepaulv.qbuilders.properties.concrete.DoubleProperty import

src/main/java/com/faforever/client/query/TextFilterController.java
obydog002: same here as well

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
obydog002: I dont understand whats needed about unbinding

obydog002: ahh, to prevent memory leaks as above

src/main/java/com/faforever/client/query/RangeFilterController.java
Sheikah45: got it I missed that

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
obydog002: I don't understand how to add an unsubscription

obydog002:
> One thing we should probably additionally do is set a max date range for the replays. We currently initialize the range to be one year from today due to server load and probably don't want the persistence to undo that.

Any ideas on how you want this to look?

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
Sheikah45: There is a method addShownSubscription that you add a subscripton to where a subscription is just a function interface. So just pass in a function that performs the unbinding

Sheikah45:
> > One thing we should probably additionally do is set a max date range for the replays. We currently initialize the range to be one year from today due to server load and probably don't want the persistence to undo that.
> 
> Any ideas on how you want this to look?

Yeah I was thinking it would either be defaulting the preferences everytime or setting a limit on the range controller. Setting a limit might be easier.

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
```diff
@@ -131,21 +140,36 @@ protected void initSearchController() {
...
         i18n.get("featuredMod.displayName"), List.of());
 
     featuredModService.getFeaturedMods().map(FeaturedMod::displayName)
                       .collectList()
                       .publishOn(fxApplicationThreadExecutor.asScheduler())
-                      .subscribe(featuredModFilterController::setItems);
+                      .subscribe((mods) -> {
+                        featuredModFilterController.setItems(mods);
+                        replaySearchPrefs.featuredModFilterProperty().forEach((item) -> featuredModFilterController.checkItem(item));
+                        Bindings.bindContent(replaySearchPrefs.featuredModFilterProperty(), featuredModFilterController.getCheckedItems());
```
Sheikah45: I would do the actual binding of the content in the onShow method as well to ensure if it is made visible and invisible the binding will still be there

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
```diff
@@ -154,18 +178,37 @@ protected void initSearchController() {
                           leaderboard -> i18n.getOrDefault(leaderboard.technicalName(), leaderboard.nameKey()),
                           leaderboard -> String.valueOf(leaderboard.id())))
                       .publishOn(fxApplicationThreadExecutor.asScheduler())
-                      .subscribe(leaderboardFilterController::setItems);
+                      .subscribe((leaderboards) -> {
+                        leaderboardFilterController.setItems(leaderboards);
+                        replaySearchPrefs.leaderboardFilterProperty().forEach((item) -> leaderboardFilterController.checkItem(item));
+                        Bindings.bindContent(replaySearchPrefs.leaderboardFilterProperty(), leaderboardFilterController.getCheckedItems());
```
Sheikah45: Same binding comment above

Sheikah45 (changes requested):

src/main/java/com/faforever/client/replay/OnlineReplayVaultController.java
obydog002: If we do this then there is a race condition where the binding will sometimes get set to the empty controller items before checked items getting set here, therefore overwriting the preferences.

Sheikah45: We could create a copy of the list during initialization and use that in the subscribe.

obydog002: I dont love this

Sheikah45: Well as it is right now there are cases where the selection would not be saved to the prefs

Sheikah45:
The code looks good but there are some failing tests

Sheikah45:
Adding a null check is not the way to fix the tests.

Rather you need to properly mock the filter builder in the test code so it returns mock versions of the controllers

obydog002:
There is still a minor issue, if no preferences have been set (i.e. empty client.prefs file) the range slider low and high values default to 0

Sheikah45:
In that case if null would be better we can make them Object property<Integer> and that will be nullable

Sheikah45 (approved):
Is this done from your perspective?

obydog002:
Yes I believe so

PR #3181, opened 2024-05-26
Leaderboard improvements

BlackYps:
Adds the ability to select a different season than the currently active one.
Also fixes the sorting of the leaderboard tabs


src/main/java/com/faforever/client/leaderboard/LeaderboardService.java
```diff
@@ -102,26 +102,19 @@ public Flux<LeagueSeason> getActiveSeasons() {
...
-                                                                                                                       false))
+                                                                                                                   .eq(league.id()))
                                                                                                      .addSortingRule(
                                                                                                          "startDate",
                                                                                                          false);
-    return fafApiAccessor.getMany(navigator)
-                         .next().map(leaderboardMapper::map);
+    return fafApiAccessor.getMany(navigator).map(leaderboardMapper::map)
+                         .cache();
   }
```
Sheikah45, line 118: The only thing is that this list could get very long over time. So we may want some kind of pagination

BlackYps: If we want to access a season that is not in the first page we would have to be able to load earlier pages. That sounds very cumbersome from a ui perspective. I'm not sure how we could reflect this pagination in the client ui

BlackYps: besides, you are technically correct that this list could get long, but we have four seasons per year, so in ten years we will be at a total of around 50. I think this is still small enough to load it in one go

Sheikah45: It was less the loading I was thinking about and more how long the list in the UI could be and how that would be to navigate

BlackYps: It seems the combox only shows at max 10 entries and creates a scrollbar automatically if there are more

Issue #3176, opened 2024-05-22
Date language mixup


Backspace345:
**Describe the bug**
In leaderboards tab the current season period is shown in French, while client language is set to English. Changing client language doesn't change date language in leaderboards. Date format in settings is set to Day/Month/Year, if that's related.

**To Reproduce**
Go to settings > appearance, set date format to Day/Month/Year, go to Leaderboards tab.

**Expected behavior**
Dates changing according to client language. 

**Screenshots**
![1](https://github.com/FAForever/downlords-faf-client/assets/170477845/281779f5-42ca-4925-ad06-77c729ab9c86)

**OS**
Windows 10

Sheikah45:
This is an artifact of the time locale being able to be set independently of the language in settings.

magge-faf:
FYI: In the context of 'mai', 'juin' is most likely French.

Backspace345:
> FYI: In the context of 'mai', 'juin' is most likely French.

Thanks! Post updated

sulikdan:
@Sheikah45 Were you able to reproduce it by description? I tried to look into, but the date is following a user's setting of preferred language.  Tried multiple combination.

@Backspace345 If you change a language(to somethign different) and restart the client, is it still same? Do you still get french one? Even when you get back to origin language?

Tested on Win11.

Backspace345:
@sulikdan yes, tried multiple times now, for some reason it just stays there. Complete reinstall removed the bug though. I wonder if it's because my windows install broke once. Due to that i had to reformat drive with it, which apparently contained some of the metadata client files, but the client itself was installed on a separate drive, and after reinstalling windows i just went to the client folder and created a shortcut manually. Surprisingly enough it worked just fine. So i guess there never was any issue?

Sheikah45:
Yes it is because of this setting being not auto 
![image](https://github.com/FAForever/downlords-faf-client/assets/66929319/f5526208-4e02-4d4e-9ec3-66b6f4912f47)

Backspace345:
[...]

PR #3183, opened 2024-05-28
Leaderboard dates using format and language

sulikdan:
Simple fix for an issue - https://github.com/FAForever/downlords-faf-client/issues/3176

## Description

Took simpler approach, there are still functions that could be also refactored(DateTime), but would require more boiler plate code of how should specific format look like. Moreover Added support only for **Short**, **Medium** and **Long** style of **Date**

## Test
 * Created a few simple test - really not covering all cases, as there are ton of combination
 * Manual tested


src/main/java/com/faforever/client/util/TimeService.java
```diff
@@ -47,13 +47,24 @@ public String asDate(TemporalAccessor temporalAccessor, FormatStyle formatStyle)
...
+    DateInfo dateInfo = localizationPrefs.getDateFormat();
+    DateTimeFormatter formatter;
+
+    if (dateInfo.equals(DateInfo.DAY_MONTH_YEAR)) {
+      formatter = DateTimeFormatter.ofPattern(getDatePatternForDayMonthYear(formatStyle));
+    } else if (dateInfo.equals(DateInfo.MONTH_DAY_YEAR)) {
+      formatter = DateTimeFormatter.ofPattern(getDatePatternForMonthDayYear(formatStyle));
+    } else {
+      formatter = DateTimeFormatter.ofLocalizedDate(formatStyle);
+    }
```
Sheikah45: This can be a switch expression

src/test/java/com/faforever/client/util/TimeServiceTest.java
```diff
@@ -0,0 +1,89 @@
...
+import java.time.ZoneOffset;
+import java.time.format.FormatStyle;
+import java.util.Locale;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.when;
+
+@Execution(ExecutionMode.CONCURRENT)
+@ExtendWith({MockitoExtension.class})
+class TimeServiceTest {
```
Sheikah45, line 25: We should test with other locales as well

src/main/java/com/faforever/client/util/TimeService.java
```diff
@@ -80,6 +91,38 @@ private Locale getCurrentDateLocale() {
...
+      case SHORT -> {
+        return "d/M/yy";
+      }
+      case MEDIUM -> {
+        return "d MMM, yyyy";
+      }
+      default -> {
+        // for other styles extend this
+        return "d MMMM, yyyy";
+      }
```
Sheikah45: Why is default used instead of an exhaustive switch? 

Also use a switch expression instead of statement.

src/main/java/com/faforever/client/util/TimeService.java
```diff
@@ -80,6 +91,38 @@ private Locale getCurrentDateLocale() {
...
+      case SHORT -> {
+        return "M/d/yy";
+      }
+      case MEDIUM -> {
+        return "MMM d, yyyy";
+      }
+      default -> {
+        // for other styles extend this
+        return "MMMM d, yyyy";
+      }
```
Sheikah45: Same as above

Sheikah45 (changes requested):
This can be simplified by just using a switch statement,

Also if we no longer use the locale in date info it should be removed.

src/test/java/com/faforever/client/util/TimeServiceTest.java
sulikdan: So which combinations you would like to see? 2(locales) x 3(format styles Auto,Day,Month) ? Or you want to add additional FormatStyles as well?

src/main/java/com/faforever/client/util/TimeService.java
sulikdan: Well, because there are Long and Full FormatStyle that are not implemented -> in this case, default will be able to deal Both as Long style. From the code I could see usage of short and medium styles.

Additionally, I dont think it is worth it to throw exception and break application, if somebody passed weird style. Though, if you want it, I don't mind to change it. Just let me know.

src/main/java/com/faforever/client/util/TimeService.java
sulikdan: Is this requirement or suggestion?

For me it has same readability, moreover less code as no useless break statement and no default state(which shouldn't even exists in 1st place). But hey, if you want it that way. I don't mind.

Brutus5000: This is how it would look like
```
    DateTimeFormatter formatter = switch(dateInfo) {
      case DateInfo.DAY_MONTH_YEAR -> DateTimeFormatter.ofPattern(getDatePatternForDayMonthYear(formatStyle))
      case DateInfo.MONTH_DAY_YEAR -> DateTimeFormatter.ofPattern(getDatePatternForMonthDayYear(formatStyle));
      default ->
        DateTimeFormatter.ofLocalizedDate(formatStyle);
    }
 ```

- no dangling variable declaration
- only a single assignment
- no .equals calls

This is more readable.

Comment (sulikdan) [reply src/main/java/com/faforever/client/util/TimeService.java]: @Brutus5000 is default necessary? I mean from code-style perspective @Sheikah45 mentioned in other issue to use exhaustive switch .. I'm trying to find out, how the code-style should look like here ..

Comment (Brutus5000) [reply src/main/java/com/faforever/client/util/TimeService.java]: I didn't check how many values DateInfo has. Instead of default you might as well case AUTO, then you are exhaustive.

Comment (Sheikah45) [reply src/main/java/com/faforever/client/util/TimeService.java]: Yes I would add the auto case instead of just handling it in default.

File: src/main/java/com/faforever/client/util/TimeService.java (hunk 3 above)
Comment (Sheikah45) [reply src/main/java/com/faforever/client/util/TimeService.java]: I would not throw an exception but would just make explicit the long and full instead of using default.

File: src/main/java/com/faforever/client/util/TimeService.java (hunk 1 above)
Comment (Sheikah45) [reply src/main/java/com/faforever/client/util/TimeService.java]: So it just looks like this

```
    DateTimeFormatter formatter = switch(dateInfo) {
      case DAY_MONTH_YEAR -> DateTimeFormatter.ofPattern(getDatePatternForDayMonthYear(formatStyle))
      case MONTH_DAY_YEAR -> DateTimeFormatter.ofPattern(getDatePatternForMonthDayYear(formatStyle));
      case AUTO -> DateTimeFormatter.ofLocalizedDate(formatStyle);
    }
 ```

src/test/java/com/faforever/client/util/TimeServiceTest.java
Sheikah45: I would test with the us which is m/d/y and another locale which is default d/m/y for each of the three date info options

src/main/java/com/faforever/client/util/TimeService.java
Sheikah45: Also, we could just make the function return a datetimeformatter instead of just the pattern string.

Then we could also make those constants so we don't recreate them all the time.

sulikdan: @Sheikah45  fine for me.

> Also, we could just make the function return a datetimeformatter instead of just the pattern string.

I'm not quite sure if I understand you(could you elaborate more on it?), but I doubt there is other way than to use pattern. Already in just Date locale, every locale has its own way how is **short, medium, long, full** style written US is different to FR, etc.. And currently, what I have written (using date any custom format) will differ to auto format for at least FR.

Example of France FormatStyle.Medium auto:
> 9 mai 2022

Example of France FormatStyle.Medium DayMonthYear:
> mai 9, 2022

Example of France FormatStyle.Medium DayMonthYear:
> 9 mai, 2022

As u can see France auto format doesn't use comma, however US style use it. Across all other locale it is more inconsistent. Moreover, if you add more complex formats (long, full - this is reason I avoided full style, while it is not used anywhere), maybe even dateTime itself, the more inconsistencies arise.

A solution would be to agree, which format we should follow for every locale. It seems to me too much of trouble for minimal return.

Sheikah45: In the functions getDatePatternForX you are returning a string that you just feed into DateTimeFormatter.ofPattern

Instead we can just return the DateTimeFormatter from the getDatePatternForX and rename it to getFormatterForX.

And then instead of creating the datetimeformatter each time we can just define private static final members in the class and return those.

sulikdan: @Sheikah45  Then for every dateFormat change , he would need to restart the application. How you taken that into account?

ad1: ~~If you are fine with it, how do you trigger that notification to restart app?~~ I think I found it, would need to create additional message, but let me know, if you want to go that way.

Sheikah45: I am only talking about returning a date time format err from the getPatternForX methods. Those return constant strings so we can just make them return constant datetime formatters.

There would be no restart necessary and it would still work just changing the formatstyle during runtime

sulikdan: I hear and I obey.

src/main/java/com/faforever/client/util/DateFormatterUtil.java
```diff
@@ -0,0 +1,56 @@
...
+
+
+  public DateTimeFormatter getAutoFormatter(FormatStyle style) {
+    return switch (style) {
+      case SHORT -> FORMATTER_AUTO_SHORT;
+      case MEDIUM -> FORMATTER_AUTO_MEDIUM;
+      case LONG -> FORMATTER_AUTO_LONG;
+      case FULL -> FORMATTER_AUTO_FULL;
+    };
+  }
```
Sheikah45: This method is not needed. We only w needed it for the actual constants with patterns. For auto we just call the ofLocalized directly in line like we were

src/main/java/com/faforever/client/util/DateFormatterUtil.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public static final DateTimeFormatter FORMATTER_DMY_SHORT = DateTimeFormatter.ofPattern("d/M/yy");
+  public static final DateTimeFormatter FORMATTER_DMY_MEDIUM = DateTimeFormatter.ofPattern("d MMM, yyyy");
+  public static final DateTimeFormatter FORMATTER_DMY_LONG = DateTimeFormatter.ofPattern("d MMMM, yyyy");
+  public static final DateTimeFormatter FORMATTER_DMY_FULL = DateTimeFormatter.ofPattern("EEEE, d MMMM, yyyy");
+
+  public static final DateTimeFormatter FORMATTER_MDY_SHORT = DateTimeFormatter.ofPattern("M/d/yy");
+  public static final DateTimeFormatter FORMATTER_MDY_MEDIUM = DateTimeFormatter.ofPattern("MMM d, yyyy");
+  public static final DateTimeFormatter FORMATTER_MDY_LONG = DateTimeFormatter.ofPattern("MMMM d, yyyy");
+  public static final DateTimeFormatter FORMATTER_MDY_FULL = DateTimeFormatter.ofPattern("EEEE, MMMM d, yyyy");
```
Sheikah45: These should be private and can just be members of TimeSerivce rather than being in a separate class

src/main/java/com/faforever/client/util/DateFormatterUtil.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  public DateTimeFormatter getMDYFormatter(FormatStyle style) {
+    return switch (style) {
+      case SHORT -> FORMATTER_MDY_SHORT;
+      case MEDIUM -> FORMATTER_MDY_MEDIUM;
+      case LONG -> FORMATTER_MDY_LONG;
+      case FULL -> FORMATTER_MDY_FULL;
+    };
+  }
```
Sheikah45: Similar to the static members only needing to be in TimeService these methods can just be in TimeService

Sheikah45 (changes requested):

src/main/java/com/faforever/client/util/DateFormatterUtil.java
sulikdan: I don't see advantage to use them as members. From memory footprint, it will still be created anew( afterwards dateformatter is called by other methods that create its copy) and then we use new one all over again for AUTO?

Though i fixed as you asked.

Sheikah45: It also just helps communicate the intent of these as constants

Sheikah45 (approved):

.idea/runConfigurations/Main.xml
```diff
@@ -1,5 +1,7 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Main" type="Application" factoryName="Application" singleton="false" nameIsGenerated="true">
+    <option name="ALTERNATIVE_JRE_PATH" value="temurin-21" />
+    <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
```
BlackYps, line 4: Is this supposed to be here?

Sheikah45: It just specifies the jre for the run config if someone doesn't have it set

PR #3192, opened 2024-06-02
Implement the repository URL

Garanas:
Relates to:

- https://github.com/FAForever/faf-java-api/issues/564

It's shown as the 'Source' link, below the author. Visually it is not ideal, but we can improve that separately.

image

### DIFF
--- a/src/main/java/com/faforever/client/mod/ModService.java
+++ b/src/main/java/com/faforever/client/mod/ModService.java
@@ -337,6 +337,10 @@ public class ModService implements InitializingBean, DisposableBean {
     ModUploadTask modUploadTask = modUploadTaskFactory.getObject();
     modUploadTask.setModPath(modPath);
 
+    // retrieve information from the mod_info.lua to send to the API
+    ModVersion modVersion = extractModInfo(modPath);
+    modUploadTask.setRepositoryURL(modVersion.mod().repositoryURL());
+
     return taskService.submitTask(modUploadTask);
   }
 
--- a/src/main/java/com/faforever/client/mod/ModService.java
+++ b/src/main/java/com/faforever/client/mod/ModService.java
@@ -337,14 +337,6 @@ public class ModService implements InitializingBean, DisposableBean {
     ModUploadTask modUploadTask = modUploadTaskFactory.getObject();
     modUploadTask.setModPath(modPath);
 
-    // retrieve information from the mod_info.lua to send to the API
-    try {
-      ModVersion modVersion = extractModInfo(modPath);
-      modUploadTask.setRepositoryURL(modVersion.mod().repositoryURL());
-    } catch (Exception e) {
-      // ???, do nothing?
-    }
-
     return taskService.submitTask(modUploadTask);
   }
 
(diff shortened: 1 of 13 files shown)


src/main/java/com/faforever/client/mod/ModService.java
```diff
@@ -337,6 +337,10 @@ public CompletableTask<Void> uploadMod(Path modPath) {
     ModUploadTask modUploadTask = modUploadTaskFactory.getObject();
     modUploadTask.setModPath(modPath);
 
+    // retrieve information from the mod_info.lua to send to the API
+    ModVersion modVersion = extractModInfo(modPath);
+    modUploadTask.setRepositoryURL(modVersion.mod().repositoryURL());
```
Sheikah45: This can be moved into the ModUploadTask itself

Garanas: Processed in https://github.com/FAForever/downlords-faf-client/pull/3192/commits/26166cbea6053b7adadf9a7e0d04725516fd7942

src/main/java/com/faforever/client/mod/ModUploadTask.java
```diff
@@ -69,10 +78,20 @@ protected Void call() throws Exception {
             .zip();
       }
 
+      // retrieve information from the mod_info.lua to send to the API
+
+      ModVersion modVersionInfo = modService.extractModInfo(modPath);
+      URL repositoryURL = modVersionInfo.mod().repositoryURL();
```
Sheikah45: This is shadowing the field repositoryUrl now. So the field should be removed or this should set the field if it is null.

Garanas: Processed in: https://github.com/FAForever/downlords-faf-client/pull/3192/commits/57ff0b36dea0f583c9a3bf526364eb194051d2e4

src/test/java/com/faforever/client/mod/ModDetailControllerTest.java
```diff
@@ -157,6 +161,29 @@ public void testSetModWithNoUploader() {
...
   }
 
+  @Test
+  public void testWithRepositoryURL() throws MalformedURLException {
+    modVersion = Instancio.of(ModVersion.class)
+                                     .set(field(Mod::repositoryURL),
+                                          URI.create("https://github.com/FAForever/fa").toURL())
+                                     .create();
+    runOnFxThreadAndWait(() -> instance.setModVersion(modVersion));
+    WaitForAsyncUtils.waitForFxEvents();
```
Sheikah45: technically you don't need WaitForAsyncUtils.waitForFxEvents if you run things with runOnFxThreadAndWait

Garanas: Processed in https://github.com/FAForever/downlords-faf-client/pull/3192/commits/23ab26f875ea07ac10fe75181bc01b1b4813ba03

Sheikah45 (commented):
Looking at this I think we may want to check the repository url before uploading the mod. The API only accepts certain hosts as the repository url for a mod. And we could probably give a better error message if we detect it earlier.

Sheikah45:
Stale

PR #3195, opened 2024-06-03
basic UI scaling fix: 1.25x, 1.5x, 1.75x

K-ETFreeman:
Resolves issue when popup had wrong size when UI scaling is enabled:
image

I did not find any meaningful explanation for the adjustment values which needed to be made. Probably some rounding error in javaFX, but maybe not.
Anyways, this works on 1.25x, 1.5x, 1.75x scaling and does nothing in any other case, so should be good enough.

### DIFF
--- a/src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
+++ b/src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
@@ -19,7 +19,6 @@ import javafx.beans.property.SimpleIntegerProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.value.ObservableValue;
 import javafx.scene.control.ScrollPane;
-import javafx.scene.layout.Region;
 import javafx.scene.layout.TilePane;
 import javafx.scene.layout.Pane;
 import javafx.scene.layout.VBox;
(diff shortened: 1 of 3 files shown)


Sheikah45:
I don't like this solution for how brittle it is and it only covers I think windows machines for specific scaling values. And adds complexity and handles the UI in a way that isn't done anywhere else in the client.

K-ETFreeman:
> I don't like this solution for how brittle it is and it only covers I think windows machines for specific scaling values. And adds complexity and handles the UI in a way that isn't done anywhere else in the client.

I can test on non-windows machine if you want
Without deep understanding of the nature of the issue no better solution will be available any time soon, and i believe this is far better than broken UI-scaled popup which we have currently

Code which works for more users > code which works for less users

K-ETFreeman:
Now its working for all UI scales
about changes:
1) using tilePane instead flowPane allowed to fix the UI scaling issue and remove some code
2) using some styles magic allowed to remove scrollbar-fix code
3) when(showing) was deleted because it prevents resize function from launching when it should
4) changed the moment when we hide loading pane. This way allows to remove default 530 size from scrollpane while causing no popup srinking the moment loadingPane is hidden (because now its being hidden exactly when maps appears, no frames between removal and maps added like it was before)

Sheikah45 (approved):

PR #3198, opened 2024-06-04
Add lifetime of lobby in games table

### LINKED ISSUE #3054: Lobby Hosting Duration Indicator
**Issue:**

Sometimes a lobby is idling for many hours because the hosting player has forgotten about it for whatever reason.

**Causing:**

Players are waiting in that lobby until it fills, then they leave after host does not respond; The lobby still exists, and new people join the lobby again - new cycle of waiting and leaving begins.

**Solution:**

Indicate the hosting time to see which lobbies are idling for many hours/days which are most likely dead ends.

image

Marc-Spector:
Requests https://github.com/FAForever/faf-java-commons/pull/152
Closes #3054 
image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -230,7 +230,7 @@ dependencies {
   implementation("io.projectreactor.addons:reactor-extra")
   implementation("io.projectreactor:reactor-tools")
 
-  def commonsVersion = "1013d75fc9"
+  def commonsVersion = "fbaab3c"
 
   implementation("com.github.FAForever.faf-java-commons:faf-commons-data:${commonsVersion}") {
             exclude module: 'guava'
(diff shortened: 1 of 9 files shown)


Sheikah45 (commented):
The only thing that gives me a little pause is that the lifetime column is static so won't actually be updated unless we get an update from the game.

Part of me wonders if it should be guaranteed to be updated.

Marc-Spector:
> The only thing that gives me a little pause is that the lifetime column is static so won't actually be updated unless we get an update from the game.

Or we can switch to another tab and then back again to get the fresh result. Sorting the lifetime column. And much more

> Part of me wonders if it should be guaranteed to be updated.

That's up to you. If it's necessary, then I need to figure out how to do it.

Sheikah45:
Yeah I guess if it updates that often then we will just see if anyone complains about it

Sheikah45 (approved):

PR #3205, opened 2024-06-17
Remember last opened chat tab

Marc-Spector:
Since the main tabs are now recreated with each click, the last opened chat tab of the channel or player is no longer remembered.
This pull request will help bring this feature back.

### DIFF
--- a/src/main/java/com/faforever/client/chat/ChatNavigation.java
+++ b/src/main/java/com/faforever/client/chat/ChatNavigation.java
@@ -2,6 +2,7 @@ package com.faforever.client.chat;
 
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
+import org.jetbrains.annotations.Nullable;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Component;
 
@@ -23,6 +24,7 @@ public class ChatNavigation { // TODO: Not a very good name of class
     currentTabs.removeIf(chatTab -> chatTab.getId().equals(tabId));
   }
 
+  @Nullable
   public String getLastOpenedTabId() {
     return currentTabs.stream().filter(ChatTab::isSelected).findFirst().map(ChatTab::getId).orElse(null);
   }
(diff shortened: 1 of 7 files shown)


src/main/java/com/faforever/client/chat/ChatNavigation.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public void setLastOpenedTabId(String tabId) {
+    if (ChatController.ADD_CHANNEL_TAB_ID.equals(tabId)) {
+      return;
+    }
+
+    currentTabs.stream()
+               .peek(chatTab -> chatTab.setSelected(false))
+               .filter(chatTab -> chatTab.getId().equals(tabId))
```
Sheikah45: We can't have a side effect in peek and use filter with find first because the peek operation is not guaranteed to be used run for every element of the stream.

Sheikah45 (commented):
At a high level this makes sense.

I just wonder if we could just use a map or something in Chat navigation instead of a list of a new class.

Marc-Spector:
> At a high level this makes sense.
> 
> I just wonder if we could just use a map or something in Chat navigation instead of a list of a new class.

When creating the `ChatTab` class I assumed that in the future there would be other variables that could be added. For example, `lastVisibleMessageIndex` variable if the user scrolls up and goes to another main tab and then returns to the chat in the same place. This is just a rough example.

Sheikah45:
Fair I suppose. Although the last visible message is stored in the Chat channel object now just as FYI

src/main/resources/theme/chat/chat.fxml
```diff
@@ -14,7 +14,7 @@
     <children>
         <TabPane fx:id="tabPane" tabClosingPolicy="ALL_TABS" AnchorPane.bottomAnchor="0.0" AnchorPane.leftAnchor="0.0"
                  AnchorPane.rightAnchor="0.0" AnchorPane.topAnchor="0.0" visible="false">
-            <Tab text="+" closable="false">
+            <Tab fx:id="addChannelTab" text="+" closable="false">
```
Sheikah45: The fx:id value actually gets set as the id for the node. So you shouldn't need to set the id in the controller itself.

Sheikah45 (changes requested):

src/main/resources/theme/chat/chat.fxml
Marc-Spector: Oh! I'll know, thanks. But we don't need the id anymore. Found a more elegant solution using `isClosable` condition

Sheikah45 (approved):

PR #3216, opened 2024-06-25
Feature: Sound for match found

### LINKED ISSUE #2984: Better alerts when a game is found
I often leave the client searching for a game while doing something else & don't notice it has found a game until it is launching & I have seconds to react. This has sometimes been about half an hour after I started searching, and I've forgotten I'm searching so I'm suddenly warping into a game - the extra few seconds of warning would make a big difference.

As an option, the FAF client becomes the top window when you find a match - e.g this behaviour appears in dota 2 when you find a game (I believe it's optional there too).

**Describe alternatives you've considered**
There've also been suggestions on the discord of a sound playing when you find a game which is another good option, though I wouldn't say it's a total replacement.

DotNetSimon:
Added a new sound when the match-found notification triggers.
It is like a simple beep-beep-beep-beep race start sound, not too intrusive but clear.

- Added notification to enable/disable it (on by default).

- Added tests for sound playing as with other sounds.

Manual tests done:
✅  Check labels/checkboxes are ok in English.
✅  Check checkbox functions (on/off)
✅  Check sound plays when match is found
[ ] Check default settings show correctly in settings - notifications window (not sure how to reset settings...) 

Not sure how to add all the translations? Used the same words as for the "Match Found" string, so it should be easy to add these.

Fixes https://github.com/FAForever/downlords-faf-client/issues/2984


Sheikah45 (dismissed):

Sheikah45:
You can check the default settings by just renaming your client.prefs file

BlackYps:
Someone found a nice sound here: https://github.com/FAForever/downlords-faf-client/issues/2984
We should use that

Sheikah45:
Right now the tests are failing because the notificationPrefs object wasn't added to the test. 

Additionally the if check around the playMatchSound is a little redundant since it checks inside the audio service

DotNetSimon:
> Someone found a nice sound here: #2984 We should use that

I've used this new sound. I think it might be more intrusive/disruptive than the one I was using before but if anyone asks I'll send them to you 👍

DotNetSimon:
> Right now the tests are failing because the notificationPrefs object wasn't added to the test.
> 
> Additionally the if check around the playMatchSound is a little redundant since it checks inside the audio service

Fixed both.

DotNetSimon:
@BlackYps  can you try with that sound? It is quite loud and it will scare people shitless when it suddenly blares after 30 minutes of not paying attention :p

The sound I used originally is more kind to the ears.

https://pixabay.com/sound-effects/countdown-sound-effect-8-bit-151797/

It's Pixabay which we already use.

DotNetSimon:
I've reused the sound from Warren that was lowered by 8db and it's pretty good now. Let me know if there's anything else to adjust @Sheikah45

Sheikah45 (approved):

Issue #3093, opened 2023-12-22
Filters clear for "Live Replays" tab when switching


piisawheel:
**Describe the bug**
Filters in newest client (23.12.0) reset when tab changes in "Live Replays"

**To Reproduce**
Goto "Replays" tab.
Goto "Live Games"
Open Filters and check a box (I like to ignore games with 1 player, for example)
Click the "Play" tab (I'm usually on custom games, not sure if relevant)
When going back to "Replays" tab, the filters have reset, and I must recheck the "Ignore games with 1 player" option again.

**Log or error message**
None

**Expected behavior**
Previous version remembered that I checked the "Ignore games with 1 player" box and filtered accordingly when switching tabs.

**Screenshots**
None

**OS**
Windows 10 Home N using Faf Client 23.12.0

sulikdan:
@Sheikah45 any idea for an approach? I've check a few commits related to the LiveReplayController, but couldn't find anythign that would be related to saving filter values. The whole LiveReplayController is created anew with no "memories". Could create singleton wrapper service, that would remember these filter values during runtime. Are you okay with that?

Sheikah45:
The way to do it would be to create a new preferences object like is used for the online replay vault search and then you would just inject that prefs object and bind it to the relevant fields in a similar fashion.

That will also allow it to be stored in client.prefa as well.

PR #3230, opened 2024-08-07
Live Replay Setting

sulikdan:
Should solve issue https://github.com/FAForever/downlords-faf-client/issues/3093 , 
though I highly doubt it was working before.

@Sheikah45 Is it to your liking? While I was going through it thought that some refactoring is needed.

Currently there is VaultPrefs - Which contains:
* ReplaySearchPrefs 
* LiveReplaySearchPrefs ( newly created for this feature)

I see that VaultPrefs were firstly created then ReplaySea.. was added, though I'm not sure myself to which screen belongs which as there is FAF Replay and Local Replay (could debug it). Anyway, I think It would be good to get them all in the same level.

And when I would be doing it, I've have seen that there are prefs for filters in CustomGame(Filter)Controller , it could be also somehow nested and could increase readability.

The question is - should I proceed? Or are you fine with it being this way

### DIFF
--- a/src/main/java/com/faforever/client/filter/LiveGamesFilterController.java
+++ b/src/main/java/com/faforever/client/filter/LiveGamesFilterController.java
@@ -11,7 +11,7 @@ import com.faforever.client.fx.ToStringOnlyConverter;
 import com.faforever.client.i18n.I18n;
 import com.faforever.client.map.generator.MapGeneratorService;
 import com.faforever.client.player.PlayerService;
-import com.faforever.client.preferences.Preferences;
+import com.faforever.client.preferences.LiveReplaySearchPrefs;
 import com.faforever.client.social.SocialService;
 import com.faforever.client.theme.UiService;
 import com.faforever.commons.lobby.GameType;
@@ -33,7 +33,8 @@ public class LiveGamesFilterController extends AbstractFilterController<GameInfo
   private final PlayerService playerService;
   private final FeaturedModService featuredModService;
   private final MapGeneratorService mapGeneratorService;
-  private final Preferences preferences;
+
+  private final LiveReplaySearchPrefs liveReplaySearchPrefs;
 
   private FilterCheckboxController<GameInfo> hideModdedGamesFilter;
   private FilterCheckboxController<GameInfo> hideSingleGamesFilter;
@@ -51,13 +52,13 @@ public class LiveGamesFilterController extends AbstractFilterController<GameInfo
                                    MapGeneratorService mapGeneratorService,
                                    FxApplicationThreadExecutor fxApplicationThreadExecutor,
                                    SocialService socialService,
-                                   Preferences preferences) {
+                                   LiveReplaySearchPrefs liveReplaySearchPrefs) {
     super(uiService, i18n, fxApplicationThreadExecutor);
     this.featuredModService = featuredModService;
     this.playerService = playerService;
     this.mapGeneratorService = mapGeneratorService;
     this.socialService = socialService;
-    this.preferences = preferences;
+    this.liveReplaySearchPrefs = liveReplaySearchPrefs;
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


Sheikah45:
The problem with changing the nesting structure is that the Preferences POJO is serialiized and deserialized directly to and from the client.prefs file. So changing the nesting structure would break people's current preferences that are saved on their machines. Towards this end we do have the migrate preferences function in preferencesConfig which can be used for updating the new preferences from the old preferences structure. I have just been too lazy to do this in the past XD. That also means that you cannot simply just move objects but have to have a period where the old location is deprecated and the new location is used and the migrate function translates between the two.

src/main/java/com/faforever/client/filter/LiveGamesFilterController.java
```diff
@@ -90,4 +109,16 @@ public GameType fromString(String string) {
...
+
+  @Override
+  protected void afterBuilt() {
+    hideModdedGamesFilter.valueProperty().bindBidirectional(preferences.getVault().getLiveReplaySearch().getHideModdedGames());
+    hideSingleGamesFilter.valueProperty().bindBidirectional(preferences.getVault().getLiveReplaySearch().getHideSingleGames());
+    onlyGamesWithFriendsFilter.valueProperty().bindBidirectional(preferences.getVault().getLiveReplaySearch().getOnlyGamesWithFriends());
+    onlyGeneratedMapsFilter.valueProperty().bindBidirectional(preferences.getVault().getLiveReplaySearch().getOnlyGeneratedMaps());
+    gameTypeFilter.valueProperty().bindBidirectional(preferences.getVault().getLiveReplaySearch().getGameTypes());
+    featuredModFilter.valueProperty().bindBidirectional(preferences.getVault().getLiveReplaySearch().getModName());
+    playerFilter.valueProperty().bindBidirectional(preferences.getVault().getLiveReplaySearch().getPlayerName());
```
Sheikah45: To get around the nesting and chained method calls in PreferencesConfig the nested preferences are exposed as their own beans so they can be injected directly

src/main/java/com/faforever/client/preferences/LiveReplaySearchPrefs.java
```diff
@@ -0,0 +1,29 @@
...
+  private final BooleanProperty hideSingleGames = new SimpleBooleanProperty(false);
+  private final BooleanProperty onlyGamesWithFriends = new SimpleBooleanProperty(false);
+  private final BooleanProperty onlyGeneratedMaps = new SimpleBooleanProperty(false);
+
+  private final ListProperty<GameType> gameTypes = new SimpleListProperty<>(FXCollections.observableArrayList());
+  private final ListProperty<FeaturedMod> modName = new SimpleListProperty<>(FXCollections.observableArrayList());
+
+  private final StringProperty playerName = new SimpleStringProperty("");
+
+}
```
Sheikah45: Unfortunately we cannot use lombok getter and setter on these javafx property beans as they will not serialize and deserialize properly.

This is because the getter and setter are produced for the property object itself rather than the underlying representation, thus jackson will serialize the property object and not the value. To this end we follow javafx bean convention of 
```
T get*()
void set*(T)
Property<T> *Property()
```

You can see this pattern in the other pref objects

Sheikah45 (changes requested):

sulikdan:
> The problem with changing the nesting structure is that the Preferences POJO is serialiized and deserialized directly to and from the client.prefs file. So changing the nesting structure would break people's current preferences that are saved on their machines. Towards this end we do have the migrate preferences function in preferencesConfig which can be used for updating the new preferences from the old preferences structure. I have just been too lazy to do this in the past XD. That also means that you cannot simply just move objects but have to have a period where the old location is deprecated and the new location is used and the migrate function translates between the two.

That hurts, I didn't realize it, but it will be needed, cos its starting to be real mess, it took me a while to get sorted in it.
Gonna do fixes + fix tests, then will notify you. Ty for information

sulikdan:
@Sheikah45 Could you please, help me out with tests?
I fail already at set-up
> class com.faforever.client.filter.FilterTextFieldController cannot be cast to class com.faforever.client.filter.FilterCheckboxController

but i have no Idea why it is throwing this as it should pass FilterCheckbox.. not FilterText... I already burned like 3 days of commenting thing in and out ...

Sheikah45:
I can take a deeper look later, but one thing I know is that injectMocks can sometimes be flaky when you have two parameters that are the same type so we may need to just construct the object manually ourselves.

src/main/java/com/faforever/client/filter/LiveGamesFilterController.java
sulikdan: As requested, it is now injected as bean even though it is still nested.

src/main/java/com/faforever/client/preferences/LiveReplaySearchPrefs.java
sulikdan: Got rid of lombok and manually crfaeted methods.

sulikdan:
Finally fixed test, wouldn't believe that replacing 
> @mock
 for
 > @spy
  would solve issue, not in theory makes sense ... but still surprised.
  
  No test added, not sure if needed as it is all was implemented inside inherited method. And implementation is same as everywhre(? at least CustomGameFilterController )

Sheikah45 (commented):
Why did all the fields change to protected in the test?

Other than that looks good

sulikdan:
> Why did all the fields change to protected in the test?

Uff, ehm, I was playing around and wanted to do some inheritance, but gave it up and then thought it is not much of difference protected vs private, but will clean it in evening. My bad.

sulikdan:
Encapsulation members reverted.

Sheikah45 (approved):

PR #3233, opened 2024-08-16
Replay show game time

sulikdan:
Solves issue 3159 (Hope I got it right, extremely hungry, gonna get somethign ..) - https://github.com/FAForever/downlords-faf-client/pull/3233

Added 2 new icons:
* game-duration
* world(real)-duration

both now showing in:
* card_replay
* replay_detail

game-duration is being shown in card-replay.

@Sheikah45 
Any thoughts and prayers?


Sheikah45 (commented):
Looks good.

Could we add a test to make sure tick duration shows/doesn't show appropriately

sulikdan:
@Sheikah45  updated tests. Current implementation shows bot real-time and game-time, therefore not much of modification.

Slightly increased size of card, as there was issue that new values(with icons) were only partially visible, though there are still occur situation that second  may be cut out.(Check local replay) I'm not sure how to address this, or even if there should be done anythign against that as it is not really significant - for clear data the user shall open the card itself ...

Sheikah45 (approved):

PR #3235, opened 2024-08-19
Add indicator when someone in your rating range is searching in matchmaker

BlackYps:
Depends on https://github.com/FAForever/faf-java-commons/pull/170


BlackYps:
I don't plan to finish this.
We also decided that the server should not send the ratings. Related: https://github.com/FAForever/server/issues/1059

Issue #3022, opened 2023-08-09
The Mod Manager needs an update. please.


tot121212:
**Is your feature request related to a problem? Please describe.**
I have serious issues with the mod manager in place currently.

**Describe the solution you'd like**
1 - A better description section for more vivid summary of mods (I don't want to have to go in-game and find the github link to the description of the mod, please). 

2 - The ability to see a list of all of your currently installed mods (Manage active mods is not good enough, either change it or make a separate menu), similar to the system in place to find mods on the mod downloading list and of course the one in game. 

3 - A mod compatibility section for developers to put incompatible mods into (This should have hyperlinks with the names of the mods). Also, a system for players to elect incompatible mods (with date of election, to protect against misinformation about whether it was patched or not) and have it show up on a separate list saying "Players have found that these mods are incompatible with "Name of Mod": mod1, mod2, mod3" as well as a thumbs up/thumbs down system for players agree with a persons findings or disagree. This would help greatly with confusion about compatibility.

4 - Date updated and Date uploaded as well as a previous-update list with changelogs. Self explanatory

Thanks for reading, that is all.

MrRowey:
what you're asking for 3 is down the mod developer not Faf. for 3, the mod developer can specify which mods will not work alongside it.

tot121212:
not well from what ive seen, im talking about setting up a modular system for linking the mods.

Sheikah45:
4 is already present in the mod vault. And 1 is done by the mod developer not the client

tot121212:
> 4 is already present in the mod vault. And 1 is done by the mod developer not the client

For 4 - 
Who thought it was a good idea to put it at the bottom instead of the giant empty space at the top? Legitimately didn't see it because of that.

![image](https://github.com/FAForever/downlords-faf-client/assets/126296068/99d53d6e-cd5a-4acb-8387-ec0bea1c7b1b)

Look at all that empty space

![image](https://github.com/FAForever/downlords-faf-client/assets/126296068/042e2396-5484-41ab-b5cb-c1f2abe20efa)

You could easily condense this down into the top right corner.

For 1 - 
![image](https://github.com/FAForever/downlords-faf-client/assets/126296068/206252d9-2d82-4e52-8fdb-895cfac6ac97)

In my perception, there doesn't seem to be a lot of space for the description. If people were able to describe it more there would be clear evidence of that; The fact I haven't seen a single mod with a large description seems to not be a coincidence. 

[...]

Issue #2509, opened 2022-01-07
Inconsistent map and mod detail cards


BlackYps:
The map detail shows a description heading whereas the mod detail does not. In my opinion the heading can be removed as it is pretty self-explanatory that the following text is a description. In any way it should be consistent between the two.

I also just noticed that they don't use the rolling in window like the map/mod upload but instead still use the old way of displaying a window. This should probably also be fixed.

Sheikah45:
What do you mean by rolling in window?

BlackYps:
It's what happens when you use uiService.showInDialog()

Sheikah45:
ah well they aren't really an old way and a new way. They are just two different types of displays used for different things.

BlackYps:
I called it old and new, because the overlay has been changed from the "old" way to showing things in dialog some time ago. The issue I see with the "old" way is that the window has a fixed distance from the client border. So in fullscreen it blows the window up bigger than it needs to. This not only looks a bit bad, but is also annoying because you have to move your mouse all the way to the border to close the window. The showInDialog way only uses as much space as the elements need.

PR #3241, opened 2024-09-08
Feature/mods more info moved 2 top

sulikdan:
Somewhat related to issue https://github.com/FAForever/downlords-faf-client/issues/3022 .
Moved More info from Mods details to top, next to thumbnail. Should improve readibility.
+ Removed Progress bar/label as it was not shown
+ Mod Dependencies (container) was moved under description - is not visible, as feature is not yet implemented.

~ No test modification done as there was no special logic implemented

Though it looks better, I believe there could be more improvements, but feedback is needed.

P.s.: Translations - especially Turkish and 2 asians needs to be checked.


sulikdan:
Potentially also solves issue - https://github.com/FAForever/downlords-faf-client/issues/2509
But do we want to modify windows size to minimalistic as possible? Change would need to be done across - Replays, Maps and Mods as to satisfy consistency.

Sheikah45 (approved):

PR #3250, opened 2024-09-24
Don't select seasons that haven't started yet

BlackYps:
Currently the client will select the next season as soon as it exists, even if the currently ongoing season hasn't finished yet.

### DIFF
--- a/src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
+++ b/src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
@@ -25,8 +25,10 @@ import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 import reactor.core.publisher.Mono;
 
+import java.time.OffsetDateTime;
 import java.time.format.FormatStyle;
 import java.util.List;
+import java.util.Optional;
 
 
 @Slf4j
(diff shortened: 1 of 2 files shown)


BlackYps:
Since when are the tests flaky? I first had one unrelated test failing, now three. Locally I had one successful run, but also a fail.

Sheikah45:
They have been mildly flaky for a bit

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
```diff
@@ -66,7 +68,13 @@ protected void onInitialize() {
 
     leagueSeasons.map(FXCollections::observableList).when(showing).subscribe(seasons -> {
       seasonPicker.getItems().setAll(seasons);
-      seasonPicker.getSelectionModel().selectFirst();
+      Optional<LeagueSeason> currentSeason = seasons.stream()
+                                                    .filter(season -> season.startDate().isBefore(OffsetDateTime.now()))
+                                                    .findFirst();
```
Sheikah45: Why are we filtering here instead of in the api call?

Sheikah45 (changes requested):

src/main/java/com/faforever/client/leaderboard/LeaderboardController.java
BlackYps: Because I didn't think of doing it in the api call.
Theoretically this has the advantage that the future season is available in the picker, but I don't see any real advantage in that.

Sheikah45 (approved):

PR #3266, opened 2024-10-31
Extract oauth error information and fail early if it exists

### LINKED ISSUE #3257: [Bug]: Catch OAuth login failures before extracting code and state
### Describe the Bug

A common issue during login an exception due to failing regex parsing the OAuth redirect from Ory Hydra

### Log file or Error Message

Client Version: 2024.8.0
java.lang.IllegalStateException: Could not extract value with pattern 'code=([^ &]+)' from: GET /?error=request_forbidden&error_description=The+request+is+not+allowed.+No+CSRF+value+available+in+the+session+cookie.&state=6NPKAdf1l6gJAUOkFOiWY2ruTnEHrZyUqLEaatbXDjeQzpTAZofpQF9opJnPw04WYYmh7cWu8FvKSsg4ocxObMubbV38Kr HTTP/1.1
    at com.faforever.client.login.OAuthValuesReceiver.extractValue(OAuthValuesReceiver.java:145)
    at com.faforever.client.login.OAuthValuesReceiver.readValues(OAuthValuesReceiver.java:137)
    at com.faforever.client.login.OAuthValuesReceiver.readValues(OAuthValuesReceiver.java:92)
    at com.faforever.client.login.OAuthValuesReceiver.lambda$receiveValues$0(OAuthValuesReceiver.java:53)
    at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
    at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760)
    at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
    at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
    at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
    at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
    at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)

### Reproduce the bug

To reproduce most probably, either delete the CSRF cookie or wait until the CSRF token in login screens times out. (Haven't tested it).

### Expected Behavior

From the redirect url it is obvious that there is an `error` key in the request parameters. In that case fail early and show a different/more useful error to the user.

### Screenshots

_No response_

### Additional context

_No response_

### Which Operating System are you using?

Linux

obydog002:
Fix #3257 

After deleting the CSRF cookie:

error

After explicitly denying the oauth authorization:

error2

Logging in normally still works as expected


Brutus5000:
This is nice. Why not go one step further and map some technical known errors into actual valuable user errors.

Ideas:
scope_denied => Seems like you did not accept the scopes on the login page.
csrf token expired (i'm not sure how to extract this) => You probably took too long to perform the login please try again.

obydog002:
Do we want to display that in the error details view? It may be nice also to have some of that information populate instead of the generic "If this error continues..." message. That would be nicer for users as I imagine not everyone reads the stack trace, though I dont know much work it would be to do this

Brutus5000:
It's a question of confidence. Are we 100% sure that our extracted errors are really the error cause? I am not sure if there is ambiguity and we might still need the original stacktrace.

obydog002:
How about we leave the stack trace for exactly whatever the error is with as much detail as possible, but we allow the generic message to change based on the cases you describe. For a start anyway

We can also always leave in a general "If this does not resolve..." paragraph at the end for the specific cases

obydog002:
Wasnt sure how I wanted to go about this

Scopes:


No CSRF:

Sheikah45:
Only thing I would probably say is just if we can de URL encode the description when we print it

src/main/java/com/faforever/client/login/OAuthValuesReceiver.java
```diff
@@ -147,5 +151,19 @@ private String extractValue(String request, Pattern pattern) {
     return matcher.group(1);
   }
 
+  private void checkForError(String request) {
+    String decodedRequest = URLDecoder.decode(request, StandardCharsets.UTF_8);
```
obydog002: Instead of decoding twice I was thinking we can decode the request string once and change the regex patterns, thoughts @Sheikah45 ?

Sheikah45: Yes I think that would be best

obydog002:
@Sheikah45 this is ready for review. I decided against changing the regex as matching on the un-escaped request is easier. Instead we can just format the request string when we throw. As it is also easier to read the error description if its present I decided not to print it separately

I guess we need to confirm the exact wording of the i18n keys and translate?

src/main/java/com/faforever/client/login/LoginController.java
```diff
@@ -279,6 +283,17 @@ private Mono<Void> loginWithCode(String code, URI redirectUri, String codeVerifi
...
+    String cause = throwable.getMessage();
+    if (ERROR_SCOPE_DENIED.matcher(cause).find()) {
+      return "login.scopeDenied";
+    }
+    if (ERROR_NO_CSRF.matcher(cause).find()) {
+      return "login.noCSRF";
+    }
+    return "login.failed";
+  }
+
```
Sheikah45: The only thing I might say is instead of matching based off the string of the message if instead we just throw a specific type of exception for scope denied and no csrf. Then we just just use pattern matchiing

obydog002: Can you expand a bit? Where should these be throwing from? and then Ideally we catch them here and put the keys in that way?

I experimented a bit with exceptions in the OAuthReceivers file

Sheikah45: This is already in a catch statement so these exception messages are being created in the oauthValuesRetriever so instead of crafting a custom exception message we should instead create an instance of a custom exception type.

So more specifically it should be done in the checkForError function.

That is better because there it is easier to see how the error string has a limited set of options that can be matched against. 

While here we are just checking a generic string which is more fragile and error prone.

obydog002:
@Sheikah45 Alright, I think this is what you meant. Please review

This does look nicer

Sheikah45 (approved):

Issue #3222, opened 2024-07-16
Disable reclaim and resource sliders if the respective generator is set to random


BlackYps:
### Is your feature request related to a problem?

People use the reclaim sliders (for example setting it to low) while keeping the prop generator at random and are then frustrated when a high reclaim map comes out because a prop generator got selected that always produces a lot of reclaim.

### Describe the solution you'd like

We should disable the sliders when the respective generator is set to random.

### Describe alternatives you've considered

ALternatively having the slider restrict which generators can be chosen is the behaviour that we just went away from and it is too opaque for the user.

sulikdan:
@BlackYps Would you have any more description about relations between buttons/sliders? I could try to disable some of them and also add hover-hint message ...

Do I understand correctly: 
Resource Generation set Random -> disable Resource Density slider
Prop Generation set Random -> disable Reclaim Density slider

?

What about Symmetry? There is also random option. Are there any relations? E.g. Fixed seed or even map style?

Anything else?

BlackYps:
yeah, these two that you describe are correct. The symmetry has no implications for the other settings. I don't think we need any other. Some other disabling logic is already implemented and these two seem to be the only ones missing.

PR #3268, opened 2024-11-01
[3222] Disabling sliders when set to random

sulikdan:
Did some code clean-up, though naming may not be ideal - any suggestions welcomed.

added 3 tests.

Feature:
If propCombo set to Random -> slider disabled
If resourceCombo set to Random -> slider disabled

Solving ticket https://github.com/FAForever/downlords-faf-client/issues/3222

### DIFF
--- a/src/main/java/com/faforever/client/game/GenerateMapController.java
+++ b/src/main/java/com/faforever/client/game/GenerateMapController.java
@@ -13,6 +13,8 @@ import com.faforever.client.notification.NotificationService;
 import com.faforever.client.preferences.GeneratorPrefs;
 import com.google.common.annotations.VisibleForTesting;
 import javafx.beans.binding.BooleanBinding;
+import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.SimpleBooleanProperty;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import javafx.collections.transformation.FilteredList;
@@ -90,14 +92,15 @@ public class GenerateMapController extends NodeController<Pane> {
       IntStream.range(2, 17).boxed().collect(Collectors.toList()));
   private final FilteredList<Integer> selectableSpawnCounts = new FilteredList<>(validSpawnCount);
   public Spinner<Integer> numTeamsSpinner;
-  private BooleanBinding customizationAllowed;
+  private final BooleanProperty disableCustomization = new SimpleBooleanProperty();
 
   @Override
   protected void onInitialize() {
-    customizationAllowed = previousMapName.textProperty().isNotEmpty()
-                                          .or(generationTypeComboBox.valueProperty()
-                                                                    .isNotEqualTo(GenerationType.CASUAL))
-                                          .or(commandLineArgsText.textProperty().isNotEmpty());
+    disableCustomization.bind(previousMapName.textProperty()
+                                             .isNotEmpty()
+                                             .or(generationTypeComboBox.valueProperty()
+                                                                       .isNotEqualTo(GenerationType.CASUAL))
+                                             .or(commandLineArgsText.textProperty().isNotEmpty()));
 
     JavaFxUtil.bindManagedToVisible(commandLineLabel, commandLineArgsText);
     initCommandlineArgs();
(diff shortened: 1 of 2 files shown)


src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -269,31 +254,41 @@ private void initCustomStyleOptions() {
...
-  private void bindCustomStyleDisabledProperty(Node... nodes) {
+    seedTextField.disableProperty().bind(getExtendedFixedSeedNotBooleanBinding());
+    seedRerollButton.disableProperty().bind(getExtendedFixedSeedNotBooleanBinding());
+  }
+
+  private BooleanBinding getSharedBooleanBinding() {
+    return previousMapName.textProperty().isNotEmpty()
+                          .or(generationTypeComboBox.valueProperty().isNotEqualTo(GenerationType.CASUAL))
+                          .or(commandLineArgsText.textProperty().isNotEmpty());
+  }
```
Sheikah45: This should be renamed to like customizationAllowed or something like that

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -269,31 +254,41 @@ private void initCustomStyleOptions() {
...
+
+  private BooleanBinding getExtendedFixedSeedNotBooleanBinding() {
+    return getSharedBooleanBinding()
+        .or(fixedSeedCheckBox.selectedProperty().not());
+  }
```
Sheikah45: This function could just be a local variable since it is only used in one function

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -269,31 +254,41 @@ private void initCustomStyleOptions() {
...
+
+  private BooleanBinding getExtendedCustomStyleNotBooleanBinding() {
+    return getSharedBooleanBinding()
+        .or(customStyleCheckBox.selectedProperty().not());
+  }
+
+  private BooleanBinding getExtendedSliderBooleanBinding(ComboBox<String> comboBox) {
+    return getExtendedCustomStyleNotBooleanBinding()
+        .or(comboBox.valueProperty().isEqualTo(MapGeneratorService.GENERATOR_RANDOM_OPTION));
+  }
```
Sheikah45: This function is only used in one place so I think it should just be inlined

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -269,31 +254,41 @@ private void initCustomStyleOptions() {
...
+
+  private BooleanBinding getExtendedFixedSeedNotBooleanBinding() {
+    return getSharedBooleanBinding()
+        .or(fixedSeedCheckBox.selectedProperty().not());
+  }
+
+  private BooleanBinding getExtendedCustomStyleNotBooleanBinding() {
+    return getSharedBooleanBinding()
+        .or(customStyleCheckBox.selectedProperty().not());
+  }
```
Sheikah45: I would probably just inline this function as well even if it is used twice since I don't know if there is really a clear name for it and inlining would be easier to understand

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: And actually rather than having a function like this if it is going to be shared a new Property should be created and bound to this instead of using a function.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/game/GenerateMapController.java
sulikdan: This is used twice. It is cause of the whole mess. 2 separate sliders set..

src/main/java/com/faforever/client/game/GenerateMapController.java
sulikdan: will do, sounds much better.

src/main/java/com/faforever/client/game/GenerateMapController.java
Sheikah45: It is only called in the function bindCustomStyleDisabledPropertySlider

sulikdan: yeah, but the function is called twice and each time with dif. comboBox - resource slider(dedicated combobox) vs props/reclaim slider (+ dedicated combobox)

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -89,9 +90,15 @@ public class GenerateMapController extends NodeController<Pane> {
...
   public Spinner<Integer> numTeamsSpinner;
+  private BooleanBinding customizationAllowed;
 
   @Override
   protected void onInitialize() {
+    customizationAllowed = previousMapName.textProperty().isNotEmpty()
+                                          .or(generationTypeComboBox.valueProperty()
+                                                                    .isNotEqualTo(GenerationType.CASUAL))
+                                          .or(commandLineArgsText.textProperty().isNotEmpty());
+
```
Sheikah45: This is better as a private final BooleanProperty that then gets bound in onInitialize, also as is wouldn't the variable be better named noCustomizatonAllowed?

Sheikah45 (changes requested):

src/main/java/com/faforever/client/game/GenerateMapController.java
sulikdan: Ehm, not sure if I get you on the first part. if I make property final, it won't go as it will fall into category to be autowired (there is no bean to be autowire) additionally, it depends on previousMapName textField. So therefore no way to make it final. Alternative is to make it local and then pass it as param. Or am I thinking about something else? Is there Object BooleanBinding that can be specified at later time?

hmm, yeah or forbidCustomization / disableCustomization? In general, I'm lost in these logical operators and then their binding to field/property....

Sheikah45: The property wouldn't be auto wired because we should initialize it at declaration so just 'private final BooleanProperty disableCustomization = new Simple BooleanProperty();'

And then in onInit we would just disableCustomization.bind(...).

This way the property itself is final and we don't have to worry about it being reassigned. And it fits the pattern in most of the other controllers

sulikdan: Ouch, forgot about SimpleBooleanProperty ... jobs done.

Anyway, are these all <type>Property() classes used nowadays in projects? I've never came upon them till now, though I'm working only on web-applications ...

Sheikah45: They are used pretty heavily in JavaFX and are specific to it. But JavaFX is a desktop application library so if you have mainly done web application development you probably don't see them very often.

But mainly they are JavaFX's general solution to reactivity and event listeners. Somewhat like promises but with change detection and the like that you expect from fields.

Sheikah45 (approved):

PR #3275, opened 2024-11-22
[3277] Dynamically changing the timeout for Reconnect

GodFuper:
I suggest making the dynamic connection timeout depends on whether you are currently in the game.

During the game, you need to try to reconnect a little faster. It has a positive effect on reconnect in ICE and in the game itself.


Sheikah45:
My main concern here is that I am not sure how much effect the timeout has on the reconnect speed. And if the shorter time might make it harder for some players to reconnect as it constantly timesout

GodFuper:
Link https://github.com/FAForever/downlords-faf-client/issues/3277

Brutus5000:
Given that we already call `fafServerAccessor.setPingIntervalSeconds(5);` I think it makes sense to also decrease the timeout time.

Also long-term speaking this is the wrong solution. The correct solution is to extract the ICE signaling out of the lobby server and do it in the [icebreaker](https://github.com/FAForever/faf-icebreaker/pull/29)

Issue #3262, opened 2024-10-29
[Feature]: Setting option for opening URLs in browser


obydog002:
### Is your feature request related to a problem?

When clicking links in the news tab, it would be nice if instead of them opening in the FAF browser, they will open in the default OS browser .e.g. firefox. This could be controllable

### Describe the solution you'd like

An option in the settings that controls if links open in the faf browser versus the OS browser

### Describe alternatives you've considered

Not sure

1-alex98:
PR merged

PR #3276, opened 2024-11-24
Force news hubs links to open in default OS browser

obydog002:
Fix the underlying issue preventing links opening correctly in #3262

This PR doesn't add a setting to control it, I assume everyone want the links to open in the OS browser.


src/main/java/com/faforever/client/fx/WebViewConfigurer.java
```diff
@@ -58,19 +55,13 @@ public void configureWebView(WebView webView) {
...
-
-        link.setAttribute("href", "javascript:java.openUrl('" + href + "');");
-      }
+      engine.executeScript(
+        "document.onclick = function (elt) {" + 
+        "  document.querySelectorAll(\"a[target]\").forEach(e => {" + 
+        "    if (!e.href.includes(\"javascript\")) " + 
+        "      e.href = \"javascript:java.openUrl('\" + e.href + \"')\"" + 
+        "  });" +
+        "}");
```
Sheikah45: I don't quite understand this function. So everytime the document is clicked it is swapping out the href of all the links in the document right? When someone first clicks a link wouldn't this not work because the onclick event propagates from the inside out? Also this assumes the event did not consume the event and stop propagation right?

Sheikah45: I wonder if MutationObserver is supported by webview and we would rather use that. https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

obydog002: Yea your understanding is correct, its not great. it does seem to work on all the links in the news page, even on first click. Yes it assumes the event was not consumed. 

Im happy to explore if mutationobserver will work because I dont like this solution at all, it was a starting point. Another option I explored is finding the href of the clicked element which was infeasible since sometimes its the clicked element, or a direct parent or child, or a descendant of either.

obydog002:
@Sheikah45 Mutation observer works nicely to fix any links added to the page

Unsure but perhaps there is still a concern with a race condition between the java code and the js hook? Where a link can get added after the java code runs but before mutation observer is added.

Sheikah45:
> @Sheikah45 Mutation observer works nicely to fix any links added to the page
> 
> Unsure but perhaps there is still a concern with a race condition between the java code and the js hook? Where a link can get added after the java code runs but before mutation observer is added.

The race condition concern would be fixed by first adding the mutation observer and then going through and changing all the links already added to the page. This would guarantee that all the links are set since we observe and then change the preexisting ones.

src/main/java/com/faforever/client/fx/WebViewConfigurer.java
```diff
@@ -68,9 +67,26 @@ public void configureWebView(WebView webView) {
...
+
+      engine.executeScript("""
+          let obs = new MutationObserver((mutations, observer) => {
+            const addedNodes = mutations.flatMap(mut => Array.from(mut.addedNodes));
+            const links = [];
+            for (const node of addedNodes) {
+              if (node?.querySelectorAll) {
+                links.push(...Array.from(document.querySelectorAll("a")));
+              }
+            }
```
Sheikah45: Only thing besides switching this and the href setting above. But don't we only need to push node.querySelectorAll("a")?

obydog002: I dont think so

It returns a nodelist and pushing that alone didnt seem to explode it correctly

obydog002:
We could also prevent the race condition by writing the logic to handle pre-existing links in the js script

src/main/java/com/faforever/client/fx/WebViewConfigurer.java
Sheikah45: node of node list is an array? But you are already running queryselector on it

obydog002: Ill investigate

Sheikah45 (approved):

PR #3279, opened 2024-11-25
Allow copying the OAuth2 access token from user menu

Brutus5000:


src/main/java/com/faforever/client/headerbar/UserButtonController.java
```diff
@@ -57,6 +61,10 @@ public void onReport() {
     reportDialogController.show();
   }
 
+  public void onCopyAccessToken() {
+    Platform.runLater(() -> ClipboardUtil.copyToClipboard(tokenRetriever.getAccessToken()));
```
Sheikah45: The get Access token method should return a mono and not block. And then the fxApplicationExecutor should be used to execute on the platform thread instead of platform.runLater. you can use its as scheduler method and publishOn of the returned mono.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/headerbar/UserButtonController.java
Sheikah45: Although I am not sure if the clipboard has to be used on the application thread

Brutus5000: I tried without it and failed without running later.

src/main/java/com/faforever/client/headerbar/UserButtonController.java
```diff
@@ -57,6 +64,13 @@ public void onReport() {
     reportDialogController.show();
   }
 
+  public void onCopyAccessToken() {
+    tokenRetriever.getAccessToken().subscribe(accessToken -> {
+      log.info("Copied access token to clipboard");
+      fxApplicationThreadExecutor.execute(() -> ClipboardUtil.copyToClipboard(accessToken));
+    });
```
Sheikah45: tokenRetriever.getAccessToken().publishOn(fxApplicationThreadExecutor.asScheduler()).subscribe(accessToken -> ClipboardUtil.copyToClipboard(accessToken));

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #3281, opened 2024-11-27
Bring back force relay option

Brutus5000:
As soon as Cloudflare TURN is stable (https://github.com/FAForever/faf-icebreaker/pull/45), we should bring back this option.


Brutus5000:
Closed in favour of #3331

PR #3287, opened 2024-12-20
[Feature]: In the matchmaker map preview, add a symbol whether or not the map is enabled

### LINKED ISSUE #3269: [Feature]: In the matchmaker map preview, add a symbol whether or not the map is enabled.

It is difficult to distinguish between a black and white unavailable map and an available map that is white/grey colored.
For example in this screenshot only Niflheim is unavailable, but it is not apparent without looking closely.
image

Add a small symbol like a green checkmark and a red cross in the bottom right corner of the box to indicate the availability. There should always be enough space because the bottom line of the preview is occupied by the map size line of text.

- Crossing out the map name: Hides useful information
- Putting the symbol over the map: hides useful information
- Underlining available maps: not clear enough

Using a colored border to indicate availability is rather viable but I think it clashes a bit with the client's design and there is little space inbetween the preview cells for a gentle looking border.

Ivan-Shaml:
closes #3269


Sheikah45 (commented):
Unfortunately this is not really accurate to what is possible. This would imply to players that they can't get maps that are marked with an x. But that is not true. The maps are picked using the average rating of the players so any map with technically is possible some are just less likely.

Ivan-Shaml:
> Unfortunately this is not really accurate to what is possible. This would imply to players that they can't get maps that are marked with an x. But that is not true. The maps are picked using the average rating of the players so any map with technically is possible some are just less likely.

Hi,

I understand. I'm not fully into the client's UX perspective. If you have any suggestions for this PR - such as changes to the icons, colors, etc. - please feel free to share your thoughts.

Let's give this PR some time to settle in, and it can serve as a foundation once things are clarified :smiley:

Sheikah45:
Admittedly as well this whole screen is likely to change soon TM as the veto system for matchmaker gets put in place.

Sheikah45:
superceded by #3330

PR #3290, opened 2024-12-30
feat: warn players when XACT-errors are in their logs

### LINKED ISSUE #3142: Feature request: Warn players when XACT-errors are in their logs
Issues with surround sound are one of the most common causes for crashes. It would be nice if the client could scan for those in the logs, and then warn the player to disable surround sound.

Credits go to @4z0t for this idea.

Ivan-Shaml:
Checks for `info: Minimized true` and `XACT` log entries in the latest game log.
If present -> a message is displayed, when the game exits (only on exitCode !=0)

Closes #3142


src/main/java/com/faforever/client/game/GameRunner.java
```diff
@@ -422,6 +427,20 @@ private void handleTermination(Process finishedProcess) {
     }
   }
 
+  private void runLogAnalysis(String logContent) {
+    final var analysisResult = logAnalyzerService.analyzeLogContents(logContent);
```
Sheikah45: Stylistically we don't use var

src/main/java/com/faforever/client/logging/analysis/AnalysisResult.java
```diff
@@ -0,0 +1,11 @@
...
+
+import com.faforever.client.notification.Action;
+
+import java.util.Collection;
+
+public record AnalysisResult(Collection<String> analysisMessages, Collection<Action> actions) {
+  public boolean isOk() {
+    return null == analysisMessages || analysisMessages.isEmpty();
+  }
+}
```
Sheikah45: analysiisMessages and actions should be non null since they are collections

src/main/java/com/faforever/client/logging/analysis/LogAnalyzerService.java
```diff
@@ -0,0 +1,46 @@
...
+
+    if (StringUtils.contains(logContents, SND_WARNING_TRACE) && StringUtils.contains(logContents, SND_XACT_TRACE)) {
+      final String moreInfoButtonCaption = i18n.get("game.log.analysis.moreInfo");
+      analysisResult.add(i18n.get("game.log.analysis.snd", moreInfoButtonCaption));
+      actions.add(new Action(moreInfoButtonCaption,
+                 () -> platformService.showDocument(clientProperties.getLinks().get("linksSoundIssues"))));
+
+    }
+
+    return new AnalysisResult(analysisResult, actions);
```
Sheikah45: There should be some ability to ignore or suppress the warnings and pop up of the dialog since for some users they occur but have no impact on the game

src/main/java/com/faforever/client/logging/analysis/AnalysisResult.java
Sheikah45: Additionally how it is implemented the action is tied to a specific message but how you have it implemented that is not clear and they cannot be displayed in a manner that makes it clear to the use. One would probably want to tie the action and message more closely together.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/logging/analysis/AnalysisResult.java
```diff
@@ -0,0 +1,11 @@
...
+
+import com.faforever.client.notification.Action;
+
+import java.util.Map;
+
+public record AnalysisResult(Map<String, Action> result) {
+  public boolean isOk() {
+    return result.isEmpty();
+  }
+}
```
Sheikah45: It seems like a little overkill to have an object that just wraps a map

Ivan-Shaml: fixed

Sheikah45 (approved):

PR #3293, opened 2025-01-11
Feature/3240 mark watched replay with indicator

### LINKED ISSUE #3240: [Feature]: Mark Watched Replays with Visual Indicators

Losing track of which replays have already been watched.

It would be a nice QOL improvement if watched replays were marked with an icon, different shade of color, watched-date/time or an outline to indicate they have already been viewed.

E.g.:

image

A simple checkmark-icon or watched-tag

sulikdan:
A completed PR should solve/close https://github.com/FAForever/downlords-faf-client/issues/3240 - 3240 .

As of now this is **sort of PoC**(not ready to me merged), if I'm on the right track, what remains is:
* test regarding every modificaiton
* adding icon + time also to replayCardDetail (not sure with this border "highlighting" is need also in detail card.)

this is current state represented in gui:
image

@Sheikah45 thoughts?


Sheikah45 (commented):
The main things I see right now are just that the watched label seems a little in the wrong place. 

Currently it is shown like all the other  meta information about the replay but it is more information about the users action so maybe a different display like an overlay on the card or something might be better to be clearer.

Additionally just not sure how much value the date watched adds but that is minor.

BlackYps:
I'm also not convinced that the date you watched it is at all relevant. For the visuals: How about making the whole card a slightly darker shade? Giving it a brighter border goes in the right direction, but I think it makes sense to make watched replays less pronounced instead of more pronounced. Replays you have not yet watched are probably more interesting.

sulikdan:
Well, basic user it may be pointless(that's why even time is not shown), but maybe for replay casters would be helpful(Could imagine, that they want to find replay again and they know approximated dated, when they watched...)? Do we have any contact to any? Can we get their response? TBFH, My main motivation is that replay caster would do more casts :D.

> How about making the whole card a slightly darker shade? Giving it a brighter border goes in the right direction, but I think it makes sense to make watched replays less pronounced instead of more pronounced. Replays you have not yet watched are probably more interesting.

@BlackYps 
makes sense. Could do it. So no current borders, but to go with a whole card into darker shades if was watched?


What about icon? Do we want go with it? And where?
@Sheikah45 
> Currently it is shown like all the other meta information about the replay but it is more information about the users action so maybe a different display like an overlay on the card or something might be better to be clearer.

You want to move it under current specs? What about adding next to replay name? Or right next to Title at the end as separate column? Just pure icon?

BlackYps:
If you want to make life easier for casters then I strongly recommend that you first message some casters and get an understanding of what they need. You can find them if you look for people with the caster role on our discord.

If you want to use an icon I would suggest that you also add a tooltip that explains the icon. We have too many icons already where you have to guess what exactly it is supposed to represent, because there is no info other than the icon itself. But I am not sure if we need an icon here at all if we go the way of making the whole card a darker shade. On the other side I am not strongly against an icon. I just care about the feature being coherent and helpful.

sulikdan:
Sorry for being inactive( I have an issue with my job and looking for other ..)

Anyway, I went through some streamers/casters, to be brutally honest - I'm disgusted, though not sure if by my awful communication skills or peoples ability to read a message. I got response that they would welcome replay indicator(Huehuehue) and only 2 mentioned date (that I mainly asked about), however there was no throughout answer about it (that's why my self-esteem hit rock bottom).

> If you want to use an icon I would suggest that you also add a tooltip that explains the icon.
That's crystal clear, I'm counting on it(if we go down the road - icons)

I personally, am for Icon and Date(thought it would be helpful). Icon alone would be pointless as highlighting would have done much better job.

Lets make definite decision(Please pick one of these option - majority wins):
1. Only highlighting seen replays
2. Icon + Date + Highlighting card
3. Icon + date
4. Icon

@BlackYps , @Sheikah45  please pick one option, I'm excluding myself, though I think 1. is most straightforward and if we get real interest from players, we can implement dates/icons later.

P.s.: If we highlight cards (when seen replay), should card-detail be also highlighted to indicate that it was seen?

BlackYps:
No worries about being inactive. Yes, it can be hard to get good feedback on planned features. Do not let that bring you down.
I would say option 1. is best for now. It brings the main functionality and if people notice the feature and suggest improvements you can always add them at a later date.
With card-detail you mean the window after you have clicked on a replay card? I don't think we need an additional indicator there. If you are there you already know if you watched the replay because you saw it on the replay card that you had to click on. (Assuming people are not navigating the client with their eyes closed.)

sulikdan:
Here is the current state:
image
As agreed, remove everything else, just pure highlighting seen replays.

There is an one issue that I cannot deal with, The cards will not be highlighted right away after seeing replay, but when the cards are reloaded anew. I miss an item that I can bind these changes to ... like label ...

@BlackYps Thoughts?

P.s.: Will go through tests later ..

sulikdan:
Can I get review?

Added tests for developed parts.

What I'm not proud of is 2 subscriptions aaaand I have no idea how it is managed in JavaFX or at least in FAF client - soo possible mem leaks?.
Tried to use onShown/onAttached but then it wouldn't be triggered ...

src/main/java/com/faforever/client/preferences/ReplayHistoryPrefs.java
```diff
@@ -0,0 +1,21 @@
...
+  private final SetProperty<Integer> watchedReplayMap = new SimpleSetProperty<>(
+      FXCollections.observableSet());
+
+  public ObservableSet<Integer> getWatchedReplayMap() {return watchedReplayMap.getValue();}
+
+  public void setWatchedReplayMap(ObservableSet<Integer> watchedReplayMap) {
+    this.watchedReplayMap.setValue(watchedReplayMap);
+  }
+
+  public SetProperty<Integer> watchedReplayMapProperty() {return watchedReplayMap;}
```
Sheikah45: This can just be named watched replays

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -213,4 +229,10 @@ private void deleteReplay() {
       onDeleteListener.run();
     }
   }
+
+  private void applyWatchedReplayHighlight() {
+    replayTileRoot.setStyle("-fx-background-color: -card-watched-color;");
+    replayTileRoot.applyCss();
+  }
```
Sheikah45: Rather than applying the style directly a css pseudo class name should be used instead.

src/main/java/com/faforever/client/replay/ReplayWatchedService.java
```diff
@@ -0,0 +1,36 @@
...
+  void updateReplayWatchHistory(Integer replayId) {
+    if (!wasReplayWatched(replayId)) {
+      preferences.getReplayHistory().getWatchedReplayMap().add(replayId);
+      replayIdJustWatched.set(replayId);
+    }
+  }
+
+  public IntegerProperty replayIdJustWatchedProperty() {
+    return replayIdJustWatched;
+  }
```
Sheikah45: I don't think this service is needed since it just interacts with the preference object that is better done directly I would think.

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -136,13 +137,28 @@ protected void onInitialize() {
...
+        applyWatchedReplayHighlight();
+      } else {
+        // keep an eye on possible future view
+        replayWatchedService.replayIdJustWatchedProperty().subscribe(replayId -> {
+          if(replayId.equals(id)) {
+            applyWatchedReplayHighlight();
+          }
+        });
+      }
+    });
```
Sheikah45: This technically will create a subscription every time the replay is set if it is not watched.

It would probably be better to have the property be an observable contains check which I think can be done with a static methods on the Bindings class. Or just creating a single observable using the observableSet and the entity. Just probably have to add when(showing) to it to prevent memory leaks.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/preferences/ReplayHistoryPrefs.java
sulikdan: Do you mean change name of variable member **watchedReplayMap** -> **watchedReplays** ?

src/main/java/com/faforever/client/replay/ReplayCardController.java
sulikdan: I'm not sure how to do it.

I've have come up with some solution to deal with searching through already saved WatchedReplays:
``` java
 BooleanBinding binding = Bindings.createBooleanBinding( () -> {
      var id = entity.map(Replay::id).getValue();
      var replayHistory = preferences.getReplayHistory().getWatchedReplaySet();
      if( id != null && replayHistory != null ){
        return replayHistory.contains(id);
      } else {
        return false;
      }
    }, entity.map(Replay::id), preferences.getReplayHistory().getWatchedReplaySet());


    watchedProperty.bind(binding.map(aBoolean -> {
      log.debug("Watched: {}", aBoolean);
      if (aBoolean) {
        applyWatchedReplayHighlight();
        return true;
      } else {
        return false;
      }
    }).when(showing));
```
I may try to somehow simplify it, but trouble for me now is -> how to dynamically pass value/action (when a user clicks watch replay) to all/specific card to update its status to be "watched". In the reviewed code , i've used subscribe, I could go for listener, but I think this will cause same mem-leaks (please correct me if im wrong) ... Or is there way to incorporated it into the code above?

src/main/java/com/faforever/client/preferences/ReplayHistoryPrefs.java
Sheikah45: Yes

src/main/java/com/faforever/client/replay/ReplayCardController.java
Sheikah45: Applying the highlight should not be a side effect. It should be in it's own dedicated subscription directly on the watched property.

Additionally in the booloean binding you should just put entity as the second parameter instead of entity.map.

And the Boolean binding should get updated whenever watched replays changes so on user click just adding the replay to watched replays should update things.

sulikdan: uff, would this be fine?

``` java
    watchedProperty.bind(Bindings.createBooleanBinding( () -> {
      var id = entity.map(Replay::id).getValue();
      var replayHistory = preferences.getReplayHistory().getWatchedReplays();
      if( id != null && replayHistory != null ){
        return replayHistory.contains(id);
      } else {
        return false;
      }
    }, entity, preferences.getReplayHistory().getWatchedReplays()));


    watchedProperty.when(showing).subscribe((oldValue, newValue)-> {
      if(newValue) {
        applyWatchedReplayHighlight();
      }
    });
```

It seems it is working, even the scenerio, whenever an user click on watch replay, where I'm dumbfound, is it because binding registers whatever changes to mentioned dependencies/properties?

sulikdan:
@Sheikah45 how about now?

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -136,13 +148,31 @@ protected void onInitialize() {
                     .map(reviewsSummary -> reviewsSummary.score() / reviewsSummary.numReviews())
             .when(showing));
 
+
+    isReplayWatched.bind(Bindings.createBooleanBinding(() -> {
+      var id = entity.map(Replay::id).getValue();
+      var replayHistory = this.replayHistory.getWatchedReplays();
```
Sheikah45: Just remove the var usage here since it isn't used in this codebase

Sheikah45 (changes requested):

sulikdan:
@Sheikah45 Did not get any info since, anything bad with it?

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -136,13 +149,31 @@ protected void onInitialize() {
...
+        return false;
+      }
+    }, entity, replayHistory.getWatchedReplays()));
+
+
+    isReplayWatched.when(showing).subscribe((oldValue, newValue)-> {
+      if(newValue) {
+        applyWatchedReplayHighlight();
+      }
+    });
```
Sheikah45: Since we don't use the oldValue we can just use the single value version of subscribe
```suggestion
    isReplayWatched.when(showing).subscribe(value-> {
      if (value) {
        applyWatchedReplayHighlight();
      }
    });
```

src/main/java/com/faforever/client/replay/ReplayCardController.java
```diff
@@ -136,13 +149,31 @@ protected void onInitialize() {
...
+
+    isReplayWatched.bind(Bindings.createBooleanBinding(() -> {
+      Integer id = entity.map(Replay::id).getValue();
+      ObservableSet<Integer> replayHistory = this.replayHistory.getWatchedReplays();
+      if( id != null && replayHistory != null ){
+        return replayHistory.contains(id);
+      } else {
+        return false;
+      }
+    }, entity, replayHistory.getWatchedReplays()));
```
Sheikah45: The null checks can be simplified here a little bit as well as the dependencies
```suggestion
    isReplayWatched.bind(replayHistory.watchedReplaysProperty().flatMap(watchedReplays -> entity.map(Replay::id).map(watchedReplays::contains)).orElse(false));
```

Sheikah45 (changes requested):
Sorry this dropped off my radar a little bit. To things to just cleanup the properties

sulikdan:
Did put the suggested changes, checked locally tests and even check feature in the app itself - all seems to be surprisingly working. Thanks for feedback. 👍

Sheikah45 (approved):

PR #3304, opened 2025-02-09
fix(GameRunner): multiple client instances launched

### LINKED ISSUE #3301: [Bug]: Multiple client instances launched after incorrect password attempts
### Describe the Bug

Entering incorrect passwords for a protected lobby one or more times causes that many game instances to launch when successfully joining any lobby afterwards.

Bug was reported via tech-support (Link to [Discord Thread](https://discord.com/channels/197033481883222026/1159383906718588968))

I can reproduce the bug as well.

### Log file or Error Message

[ice-adapter.log](https://github.com/user-attachments/files/18546551/ice-adapter.log)

[client.log](https://github.com/user-attachments/files/18546555/client.log)

### Reproduce the bug

How to reproduce:

1. Choose a lobby protected with password, and try to enter multiple times with wrong password
2. Launch any public lobby or protected lobby with correct password
3. Several SC instances will start at the same time, as many times you have entered the incorrect PW before.

### Expected Behavior

Only one client instance should spawn when entering a lobby

### Screenshots

Image

Image

### Additional context

_No response_

### Which Operating System are you using?

Windows

Ivan-Shaml:
To be merged after: https://github.com/FAForever/faf-java-commons/pull/210 and https://github.com/FAForever/server/pull/1045

Closes #3301


Sheikah45 (commented):
Do we not need to notify the user that the join failed?

Ivan-Shaml:
> Do we not need to notify the user that the join failed?

Notification is handled by the `NoticeInfo` message, which is being sent into conjunction with the new one. There is an idea about the notifications from the server to get reworked (add error codes, don't use English-only hard-coded messages), and this fits right in there. https://github.com/FAForever/server/pull/1045#issuecomment-2676322535

PR #3306, opened 2025-02-17
Update messages_zh_TW.properties

Yuchenjimmy:
As title


Yuchenjimmy:
There're several strings are missing on poeditor, and several strings aren't properly set. Would be appretiated if they're fixed.

Issue #3041, opened 2023-10-21
Make interaction with player names consistent


BlackYps:
There are various places in the client where the names of other players show up and you can interact with them. Typically by bringing up the context menu where you can add friend/foe, show user info etc. However the interaction to bring up the menu is inconsistent.
The nicest implementation I found is in the replay window where the name gets highlighted and the cursor changes on mouseover and you bring up the menu with right or left click. We should do it this way everywhere.

Locations I found where it behaves differently:
clicking on the name in aeolus chat directly: just opens up a direct chat instead of bringing up the menu
name in the chat user list on the right: only right click works
leaderboard: only right click works
map or mod windod: you only have the option to copy the name on right click

The highlighting on mouseover to indicate that an interaction is possible is also different everywhere.

CelsoAntunes:
I'm going to try to tackle this problem.

PR #3328, opened 2025-03-30
Fix interactions with names

CelsoAntunes:
I've made significant changes to address issue #3041. While it's not finished yet, I wanted to get feedback to ensure I'm approaching this in a way that aligns with your expectations.

### Summary of Changes:
- Created Utils to avoid code duplication and improve maintainability.

- Implemented changes across most cited areas, but still need to apply them to the Leaderboard and Mods tabs.

### Current Issue:
The context menu remains open when opened via right-click, even after selecting an item that doesn’t trigger a tab switch.

### Next Steps:
- Implement Utils on the Leaderboard and Mods tabs.

- Address the context menu issue.

I probably should have opened this earlier, but I wanted to get things working first. A friend suggested submitting this now to avoid going too far in the wrong direction. Looking forward to your feedback!


Sheikah45:
At a high level I see you implemented the context menu on double click and I know the original issue stated opening the context menu on either click should work but that is probably the wrong choice.

UX is consistent that a right mouse click opens the context menu so we should stick with that pattern. That also removes the need for the MouseEventUtil and creation of fake context events.

Also the ContextMenuUtil shouldn't be static and instead should be a spring component so the UI service and contextmenubuulder can be injected into it.

Also you can look into using some of the native mapping functions of javafx properties to remove some null checks and add a bit more safety and remove Bindings.createObjectBinding.

Sheikah45:
Also as a note this code base typically doesn't import static methods and uses the top class qualification for inner classes.

CelsoAntunes:
I changed the behavior to only right-click as you mentioned, used direct injection instead of using static methods, and avoided some null checks. There is still an issue with the fact that some names in the mods contain spaces, so I cannot find the player associated to it.

I had to add a lot of code to handle proper closing of the context menu after closing a map detail and opening a new one. I think this is now solved. Maybe what I did was overkill.

Regarding the mod tab, it's still with the same problem that the map had previously, so before I try to do the same thing again, maybe it would be a good point to ask for some guidance once more.

Another small change I made was that now the context menu item message clan leader will only appear in case the player indeed has a clan.

Issue #2319, opened 2021-08-04
map veto for ladder


loxpooig:
Hello,
today i was thinking that players should be allowed to **veto ban at least 1 or 2 maps before playing ladder.**. that's because somethimes at the idea that I have to play for example open palms for the 1000th time I'm just "ok also today I will play ladder tomorrow"

This could incentivate a lot the increase of ladder players.. at least in my case! but I'm sure there are a lot of you that are like "at the idea that i could have to play that map.. naah i'll just play teamgames.."

I understood that this is not the first time we talk about this in FAF, but as far as I discovered **nobody as ever attempt to code it.** If I had known something about that I would have for sure tried to implement it myself, but I have no idea how to start.

So I'm opening this project, if anybody has the willingness to dedicate some time for it please let me know! Thank you

Sheikah45:
With the creation of the team matchmaker this got much more complicated to handle. And has not seen much dev traction in terms of will to implement so closing for now.

K-ETFreeman:
done with #3428

PR #3330, opened 2025-04-02
UI for veto system

K-ETFreeman:
Working UI for the veto system
No animations yet, will be added later
I didnt do any refactor yet so it can contain some logs and garbage AI comments which should be deleted


src/main/java/com/faforever/client/remote/FafServerAccessor.java
```diff
@@ -101,6 +109,12 @@ public void start() {
                                  .retry()
                                  .subscribe();
```
K-ETFreeman, line 104: I didnt know i needed fxApplicationThreadExecutor here, but Ai said:

> `.publishOn(fxApplicationThreadExecutor.asScheduler())` is required here because `onVetoesChanged()` calls `setAll()` on the `appliedVetoes` ObservableList, which has UI listeners subscribed to it
>   (TeamMatchmakingMapTileController:165).
> 
>   Without this, the reactive stream executes on the network I/O thread, causing `IllegalStateException: Not on FX application thread` when modifying the observable collection.
> 
Sounds like he has a point, probably you know what he's talking about

K-ETFreeman:
I just realized i need to also test and make proper UI for "no veto tokens" case. Will finish tomorrow.

src/main/java/com/faforever/client/preferences/MatchmakerPrefs.java
```diff
@@ -1,23 +1,43 @@
...
+    int index = IntStream.range(0, appliedVetoes.size())
+        .filter(i -> appliedVetoes.get(i).getMapPoolMapVersionId() == vetoData.getMapPoolMapVersionId() && appliedVetoes.get(i).getMatchmakerQueueMapPoolId() == vetoData.getMatchmakerQueueMapPoolId())
+        .findFirst()
+        .orElse(-1);
+    if (index == -1) {
+      appliedVetoes.add(vetoData);
+    } else {
+      appliedVetoes.set(index, vetoData);
+    }
+  }
```
Sheikah45: This should just be a map rather than finding and replacing the corrseponding value in a list

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -40,8 +54,14 @@
...
+  private static final Paint VETO_TOKEN_AVAILABLE_PAINT = Paint.valueOf("#ffffff");
+
+  public static final String VETO_ICON_SVG_PATH = "M 11.8068 1.8587 L 11.8067 2.4721 L 11.8058 9.8249 L 13.3528 8.6739 C 14.0631 8.1453 15.0404 8.1616 15.7326 8.7136 L 17.008 9.7306 L 17.4413 10.0761 L 17.1686 10.5586 L 14.2707 15.6856 L 13.7992 16.5846 C 12.4816 19.0965 9.7937 20.5816 6.9659 20.3599 C 3.245 20.0681 0.3874 16.9418 0.4302 13.2098 L 0.5402 3.6419 L 0.547 3.0517 L 1.1351 3.0017 L 3.831 2.7726 L 3.9211 0.9443 L 3.9506 0.347 L 4.5483 0.3267 L 7.8067 0.2164 L 8.4872 0.1933 L 8.4786 0.8742 L 8.4687 1.665 L 11.1944 1.8232 L 11.8068 1.8587 Z M 8.4523 2.9663 L 8.3681 9.6469 L 7.0682 9.6305 L 7.1609 2.2698 L 7.1702 1.5387 L 5.1901 1.6057 L 5.1017 3.3995 L 4.8833 9.6389 L 3.5841 9.5935 L 3.777 4.0819 L 1.8333 4.2471 L 1.7302 13.2247 C 1.6951 16.2725 4.0288 18.8256 7.0675 19.0639 C 9.3769 19.2449 11.572 18.0322 12.6479 15.9807 L 13.1242 15.0728 L 13.1289 15.0638 L 13.1339 15.0548 L 15.7641 10.4014 L 14.9221 9.73 C 14.6914 9.546 14.3656 9.5406 14.1289 9.7168 L 11.5437 11.6405 L 10.5057 11.1189 L 10.5067 3.0854 L 8.4523 2.9663 Z";
```
Sheikah45: This should be in a file not as a constant in the class

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -62,25 +82,53 @@ public class TeamMatchmakingMapListController extends NodeController<Pane> {
...
-  private final ObservableValue<List<MapVersion>> sortedMaps = brackets.map(this::getSortedMaps).orElse(List.of());
+  private final ObservableValue<MatchmakerQueueMapPool> currentBracket = Bindings.createObjectBinding(
+      this::getCurrentBracket, sortedMapPools, currentBracketIndex);
+
+  private final ObservableValue<Boolean> hasVetoTokens = currentBracket.map(bracket -> bracket.vetoTokensPerPlayer() > 0)
+                                                                       .orElse(false);
+  private final BooleanBinding showVetoWallet = Bindings.createBooleanBinding(
+      () -> vetoModeEnabled.get() && hasVetoTokens.getValue(),
+      vetoModeEnabled, hasVetoTokens
+  );
```
Sheikah45: convert hasVetoTokens to a BooleanExpression and use the and method

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -89,14 +137,49 @@ protected void onInitialize() {
...
+
+    applyVetoesSvg.setContent(VETO_ICON_SVG_PATH);
+
+    this.sortedMapPools.subscribe(pools -> {
+      this.bracketComboBox.getItems().setAll(pools.stream().map(this::getBracketTitle).toList());
+      this.bracketComboBox.getSelectionModel().select(Optional.ofNullable(this.currentBracketIndex.getValue()).orElse(0));
+    });
+
+    bracketComboBox.getSelectionModel().selectedIndexProperty().subscribe(index -> this.currentBracketIndex.set(index.intValue()));
+    this.currentBracketIndex.addListener(v -> bracketComboBox.getSelectionModel().select((Optional.ofNullable(this.currentBracketIndex.getValue()).orElse(0))));
```
Sheikah45: this can just use subscribe and a null check

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -89,14 +137,49 @@ protected void onInitialize() {
...
+
+    tilesContainer.getChildren().subscribe(() -> this.loadingPane.setVisible(false));
+
+    vetoTokensApplied = Bindings.createObjectBinding(this::calculateVetoTokensApplied, currentBracketMaps,
+                                                     matchmakerPrefs.getAppliedVetoes(), currentBracket);
```
Sheikah45: This is binding to the matchmakerPrefs so has to be applied using when(showing) to prevent memory leaks

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -89,14 +137,49 @@ protected void onInitialize() {
...
+                      .map(MatchmakerQueueMapPool::vetoTokensPerPlayer)
+                      .orElse(0) - Optional.ofNullable(vetoTokensApplied.getValue()).orElse(0),
+        currentBracket, vetoTokensApplied);
+
+    this.currentBracket.subscribe(this::updateContent);
+    this.currentBracketMaps.when(showing).subscribe(this::updateContent);
+    this.currentBracketIndex.subscribe(this::updateContent);
+
+    currentBracket.subscribe(this::updateVetoes);
+    vetoTokensApplied.subscribe(this::updateVetoes);
```
Sheikah45: should use when(showing)

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -62,25 +82,53 @@ public class TeamMatchmakingMapListController extends NodeController<Pane> {
...
   private final PlayerService playerService;
   private final FxApplicationThreadExecutor fxApplicationThreadExecutor;
+  private final I18n i18n;
+
+  private final MatchmakerPrefs matchmakerPrefs;
 
   private final DoubleProperty maxWidth = new SimpleDoubleProperty(0);
   private final DoubleProperty maxHeight = new SimpleDoubleProperty(0);
-  private final ObjectProperty<Map<MatchmakerQueueMapPool, List<MapVersion>>> brackets = new SimpleObjectProperty<>(
+  private final SimpleBooleanProperty vetoModeEnabled = new SimpleBooleanProperty(false);
```
Sheikah45:
```suggestion
  private final BooleanProperty vetoModeEnabled = new SimpleBooleanProperty(false);
```

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -62,25 +82,53 @@ public class TeamMatchmakingMapListController extends NodeController<Pane> {
...
       Map.of());
   private final IntegerProperty playerRating = new SimpleIntegerProperty();
   private final ObjectProperty<MatchmakerQueueInfo> queue = new SimpleObjectProperty<>();
+  private final ObjectProperty<Integer> currentBracketIndex = new SimpleObjectProperty<>(null);
 
   private final ObservableValue<List<MatchmakerQueueMapPool>> sortedMapPools = brackets.map(this::getSortedMapPools)
                                                                                        .orElse(List.of());
-  private final ObservableValue<List<MapVersion>> sortedMaps = brackets.map(this::getSortedMaps).orElse(List.of());
+  private final ObservableValue<MatchmakerQueueMapPool> currentBracket = Bindings.createObjectBinding(
+      this::getCurrentBracket, sortedMapPools, currentBracketIndex);
```
Sheikah45, line 99: You can just use Bindings.valueAt

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -157,17 +248,65 @@ public ObjectProperty<MatchmakerQueueInfo> queueProperty() {
...
+
+    return this.sortedMapPools.getValue().get(v);
+  }
+
+  private String getBracketTitle(MatchmakerQueueMapPool bracket) {
+    if (bracket == null) {
+      return "";
+    }
+
+    Double min = bracket.minRating(), max = bracket.maxRating();
```
Sheikah45: Put these on separate lines

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -157,17 +248,65 @@ public ObjectProperty<MatchmakerQueueInfo> queueProperty() {
...
+    String rating = i18n.get("game.rating");
```
Sheikah45: Rating should be moved below the first if

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -157,17 +248,65 @@ public ObjectProperty<MatchmakerQueueInfo> queueProperty() {
...
+    if (currentBracket.getValue() == null) {
+      return Collections.emptyList();
+    }
+
+    return this.brackets.getValue()
+                        .get(currentBracket.getValue())
+                        .stream()
+                        .sorted((m1, m2) -> MAP_VERSION_COMPARATOR.compare(m1.mapVersion(), m2.mapVersion()))
+                        .toList();
+  }
```
Sheikah45, line 293: This can be done with Bindings.valueAt too

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -157,17 +248,65 @@ public ObjectProperty<MatchmakerQueueInfo> queueProperty() {
...
 
-  private List<MapVersion> getSortedMaps(Map<MatchmakerQueueMapPool, List<MapVersion>> map) {
-    return map.entrySet()
-              .stream()
-              .sorted(Entry.comparingByKey(MAP_POOL_COMPARATOR))
-              .flatMap(entry -> entry.getValue().stream().sorted(MAP_VERSION_COMPARATOR))
-              .distinct()
-              .toList();
+  private Integer calculateVetoTokensApplied() {
+    if (matchmakerPrefs == null || getCurrentBracket() == null) {
```
Sheikah45: matchmakerPrefs won't be null

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -229,16 +351,50 @@ private void resizeToContent() {
...
+
+  private void updateContent() {
+    List<Pane> mapTiles = currentBracketMaps.getValue().stream().map(this::createMapTile).toList();
     fxApplicationThreadExecutor.execute(() -> {
       this.tilesContainer.getChildren().setAll(mapTiles);
       this.resizeToContent();
     });
   }
+
+  public void changeMode() {
```
Sheikah45:
```suggestion
  public void toggleVetoMode() {
```

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
```diff
@@ -1,82 +1,102 @@
...
-    this.relevanceLevel.set(value);
-  }
-  public DoubleProperty relevanceLevelProperty() {
-    return this.relevanceLevel;
-  }
+  protected final ObjectProperty<MapPoolAssignment> assignment = new SimpleObjectProperty<>();
+  @Setter
+  private ObservableValue<Integer> vetoTokensLeft;
+  @Setter
+  private SimpleBooleanProperty vetoModeEnabled;
```
Sheikah45: These should not be externally settable

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
```diff
@@ -1,82 +1,102 @@
...
+  private final SimpleIntegerProperty vetoTokensMax = new SimpleIntegerProperty(0);
+  private final SimpleIntegerProperty tokenCount = new SimpleIntegerProperty(0);
```
Sheikah45:
```suggestion
  private BooleanProperty vetoModeEnabled;
  private final IntegerProperty vetoTokensMax = new SimpleIntegerProperty(0);
  private final IntegerProperty tokenCount = new SimpleIntegerProperty(0);
```

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
```diff
@@ -96,6 +117,58 @@ protected void onInitialize(){
...
+
+    vetoSvg.fillProperty().bind(Bindings.when(tokenCount.greaterThan(0))
+                                        .then(Paint.valueOf(VETO_ICON_ACTIVE_COLOR))
+                                        .otherwise(Paint.valueOf(VETO_ICON_INACTIVE_COLOR)));
+
+    minusButton.visibleProperty().bind(vetoesBox.hoverProperty().and(tokenCount.greaterThan(0)));
+    minusButton.managedProperty().bind(minusButton.visibleProperty());
+
+    vetoButton.setOnAction(event -> {
+      if (assignment.getValue() == null) return;
```
Sheikah45: use brackets for if statement

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
```diff
@@ -96,6 +117,58 @@ protected void onInitialize(){
...
+      int current = tokenCount.get();
+      if (current < vetoTokensMax.get() && vetoTokensLeft.getValue() > 0) {
+        matchmakerPrefs.setVetoData(new VetoData(assignment.getValue().id(), current + 1, assignment.getValue().mapPool().mapPool().id()));
+      }
+    });
+
+    minusButton.setOnAction(event -> {
+      if (assignment.getValue() == null) return;
```
Sheikah45: Use brackets for if statement

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
```diff
@@ -96,6 +117,58 @@ protected void onInitialize(){
...
+      }
+    });
+
+    tokenCount.addListener((obs, oldValue, newValue) -> {
+      if (newValue.intValue() >= vetoTokensMax.get()) {
+        root.getStyleClass().add("banned");
+      } else {
+        root.getStyleClass().remove("banned");
+      }
+    });
```
Sheikah45: Use subscriibe

src/test/java/com/faforever/client/builders/VetoDataBuilder.java
Sheikah45, line 1: This is not needed Instancio should be used instead

src/test/java/com/faforever/client/preferences/MatchmakerPrefsTest.java
Sheikah45, line 1: This is not needed once a map is used

Sheikah45 (changes requested):
Also don't use just v for a variable name give it a proper name or use _

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: in which one? you mean put veto_icon_svg_path into .fxml? im using this path in both .fxml (maptile and maplist)
putting it in both .fxml is code duplication (more than duplication actually). Or do we have some special file for reused svg's?

src/main/java/com/faforever/client/preferences/MatchmakerPrefs.java
K-ETFreeman: eh, linear search is kinda fine, we barely will have dwo-digit array size there
but fine, i just ask: is this two-layered map: appliedVetoes[matchmakerQueueMapPoolId][mapPoolMapVersionId] = vetoTokensApplied, you mean? dont wanna rewrite it wrongly so clarifying

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: I cannot add it on `this.currentBracket.subscribe(this::updateContent);` because it causes whole maplist to forever stuck in loadingPane. updateContent never happens, it seems. 
Ai's version why:
image

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: If i do directly `Bindings.valueAt(sortedMapPools, currentBracketIndex)`, i get

> Cannot resolve method 'valueAt(ObservableValue<List<MatchmakerQueueMapPool>>, ObjectProperty<Integer>)'

image

Instructions unclear

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Actually `currentBracket.subscribe(this::updateVetoes);` also leads to bug when you shift veto modes if i put `.when(showing)` on it. Revering it too.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Same issue as before. The best i can do is
image
Need that?

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Removed left part of "if"

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
K-ETFreeman: Removed @ Setter, made void setVetoTokensLeft and etc. Hope this is what you meant.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
K-ETFreeman: Done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
K-ETFreeman: Done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
K-ETFreeman: Done

src/test/java/com/faforever/client/builders/VetoDataBuilder.java
K-ETFreeman: Done

src/test/java/com/faforever/client/preferences/MatchmakerPrefsTest.java
K-ETFreeman: Wdym? instructions unclear
Delete the file?

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
K-ETFreeman: Done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
Sheikah45: I mean as svg image file

src/main/java/com/faforever/client/preferences/MatchmakerPrefs.java
Sheikah45: You can create a record for a complex key so it is easier. So just a record that is {mappoolid, mapversionid}

src/test/java/com/faforever/client/preferences/MatchmakerPrefsTest.java
Sheikah45: yes the prefs object is just a data holder that doesn't need test since there should be no logic in them

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
Sheikah45: well there are double subscriptions since currentBrack depends on currentBracketIndex so the content is getting updated twice so only one needs to be subscribed to. Also the AI is just wrong. most likely the resulting observable is just getting garbage collected. You can add it to the subscriptions that get removed when the controller is detached and that should stop that.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
Sheikah45: You need to wrap the index in IntegerExpression,integerExpression  then it will work

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
Sheikah45: See same as before

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
Sheikah45: These setters were setting the properties when we should only expose setting the value of the property

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -62,25 +83,53 @@ public class TeamMatchmakingMapListController extends NodeController<Pane> {
...
                                                                                        .orElse(List.of());
-  private final ObservableValue<List<MapVersion>> sortedMaps = brackets.map(this::getSortedMaps).orElse(List.of());
+  private final ObservableValue<MatchmakerQueueMapPool> currentBracket = Bindings.createObjectBinding(
+      this::getCurrentBracket, sortedMapPools, currentBracketIndex);
+
+  private final BooleanBinding hasVetoTokens = Bindings.createBooleanBinding(
+      () -> Optional.ofNullable(currentBracket.getValue())
+                    .map(bracket -> bracket.vetoTokensPerPlayer() > 0)
+                    .orElse(false),
+      currentBracket);
```
Sheikah45: Use the BooleanExpression.booleanExpression and don't wrap it in Optional just use the native orElse of the property

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
```diff
@@ -62,25 +83,53 @@ public class TeamMatchmakingMapListController extends NodeController<Pane> {
...
+  private final BooleanBinding showVetoWallet = vetoModeEnabled.and(hasVetoTokens);
+
+  private final ObservableValue<List<MapPoolAssignment>> currentBracketMaps = Bindings.createObjectBinding(
+      this::getCurrentBracketMaps, currentBracket);
```
Sheikah45, line 106: Same Bindings.valueAt

src/main/java/com/faforever/client/preferences/MatchmakerPrefs.java
Sheikah45: Also this shouldn't be using the commons VetoData class

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: You wont be able to change its color if it will be a file AFAIK
image
And i need this.

src/main/java/com/faforever/client/preferences/MatchmakerPrefs.java
K-ETFreeman: > Also this shouldn't be using the commons VetoData class

what? what then? why not?

Sheikah45: The map is the replacement for the vetodata object. And we want to avoid using commons classes in client internals.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: @Sheikah45

K-ETFreeman: image
currentColor trick is not supported here as expected so image svg is not an option, it limits the control over svg badly

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: This one seems to do the trick
image

K-ETFreeman: expect i restored `addShownSubscription(this.currentBracket.subscribe(this::updateVetoes));` because it actually not redundant it seems. but first 3 lines seemed to be triple-subscription. Well, better safe than sorry!

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: > You need to wrap the index in IntegerExpression,integerExpression then it will work

It doesnt:

> Cannot resolve method 'valueAt(ObservableValue<List<MatchmakerQueueMapPool>>, IntegerExpression)'

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Same as before

> No, Bindings.valueAt() won't work here. The issue is that valueAt() requires an ObservableMap, but brackets is an ObjectProperty<Map<...>>, not an ObservableMap<...>.
> 
>   Looking at the Bindings API:
>   public static <K, V> ObjectBinding<V> valueAt(ObservableMap<K, V> op, ObservableValue<? extends K> key)
> 
>   To use Bindings.valueAt(), you'd need to change line 93-94 from:
>   private final ObjectProperty<Map<MatchmakerQueueMapPool, List<MapPoolAssignment>>> brackets = new SimpleObjectProperty<>(Map.of());
> 
>   to:
>   private final ObservableMap<MatchmakerQueueMapPool, List<MapPoolAssignment>> brackets = FXCollections.observableHashMap();
> 
>   But that's a larger refactor since brackets is set via brackets::setValue at line 169, and you'd need to change all the places that interact with it.
> 
>   The original Bindings.createObjectBinding(this::getCurrentBracketMaps, currentBracket) approach is actually the correct pattern for this case where you have an ObjectProperty<Map> rather than an ObservableMap.
> 
>   Either the maintainer misunderstood the structure, or they're suggesting you refactor brackets to be an ObservableMap instead - which would be more work.
>

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapTileController.java
K-ETFreeman: i see

src/test/java/com/faforever/client/preferences/MatchmakerPrefsTest.java
K-ETFreeman: Done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: Done

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: https://github.com/FAForever/downlords-faf-client/pull/3330#discussion_r2423365027

src/main/java/com/faforever/client/preferences/MatchmakerPrefs.java
K-ETFreeman: Done.

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: final version
```
protected void onShow() {
    addShownSubscription(this.maxWidth.subscribe(this::resizeToContent));
    addShownSubscription(this.maxHeight.subscribe(this::resizeToContent));
    addShownSubscription(this.currentBracketMaps.subscribe(this::updateContent));
    addShownSubscription(this.currentBracket.subscribe(this::updateVetoes));
    addShownSubscription(this.vetoTokensApplied.subscribe(this::updateVetoes));

    if (!currentBracketMaps.getValue().isEmpty()) {
      updateContent();
    }
    if (currentBracket.getValue() != null && vetoTokensApplied.getValue() != null) {
      updateVetoes();
    }
  }
```
  
  otherwise still caused bugs. but hopefully no memory leaks

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
Sheikah45: You can just put the contents into a file

Sheikah45: probably a default theme file

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
Sheikah45: ah yeah because they are not observable maps

src/main/java/com/faforever/client/teammatchmaking/TeamMatchmakingMapListController.java
K-ETFreeman: > probably a default theme file

wdym exactly? 
image
image

Put is as `public static final String` in ThemeService.java?

K-ETFreeman: Or in the `public record Theme`  itself even?

Sheikah45: Put it as a file in the default theme folder and then load it using the theme service

K-ETFreeman: image

 **//TODO** add proper loading of SVG into the codebase
thats out of scope of this PR so i will stick to something not very pleasurable it seems
will store just the path (that long SVG line) and read as text or smth

K-ETFreeman: Done. Had to make it just private instead of private final, because otherwise got NPE on themeService moment

K-ETFreeman:
I also added map details (create game) popup on map tile click, it was requested a long time ago in discord but the issue was never created it seems

I remember you was saying that passing root from MatchmakingController to the children for popup purposes is bad so instead i made summon method in the MatchmakingController itself and then propogated the event to it

K-ETFreeman:
image

cannot fully merge from upstream due to commons

K-ETFreeman:
man why no one told me we have icons.css file for icons
i probably can use it instead of svgpath for more clean approach
10 mins

K-ETFreeman:
man why check randomly fails with timeout, this doesnt seem like my code issue
time for some random commit which changes nothing to reroll

K-ETFreeman:
closes #2319 btw

src/main/java/com/faforever/client/preferences/VetoKey.java
```diff
@@ -0,0 +1,14 @@
...
+
+  @Override
+  public String toString() {
+    return matchmakerQueueMapPoolId + ":" + mapPoolMapVersionId;
+  }
+
+  public static VetoKey fromString(String key) {
+    String[] parts = key.split(":");
+    return new VetoKey(Integer.parseInt(parts[0]), Integer.parseInt(parts[1]));
+  }
```
Sheikah45: These should be in the serializer and deserializer not part of the class itself

src/main/java/com/faforever/client/preferences/MatchmakerPrefs.java
```diff
@@ -13,11 +19,41 @@ public class MatchmakerPrefs {
...
+  public List<VetoData> getVetoesAsList() {
+    return appliedVetoes.entrySet()
+                        .stream()
+                        .filter(entry -> entry.getValue() > 0)
+                        .map(entry -> new VetoData(
+                            entry.getKey().mapPoolMapVersionId(),
+                            entry.getValue(),
+                            entry.getKey().matchmakerQueueMapPoolId()))
+                        .collect(Collectors.toList());
+  }
```
Sheikah45: These should be in the service not the prefs object

Sheikah45 (changes requested):

src/main/java/com/faforever/client/preferences/MatchmakerPrefs.java
K-ETFreeman: > These should be in the service not the prefs object

TeamMatchmakingService you mean?

src/main/java/com/faforever/client/preferences/VetoKey.java
K-ETFreeman: Done (i think?)

src/main/java/com/faforever/client/preferences/MatchmakerPrefs.java
Sheikah45: yes

Sheikah45 (dismissed):

src/main/java/com/faforever/client/remote/FafServerAccessor.java
```diff
@@ -72,6 +76,8 @@ public class FafServerAccessor implements InitializingBean, DisposableBean, Life
       ConnectionState.DISCONNECTED);
 
   private final NotificationService notificationService;
+  @Lazy
+  private final TeamMatchmakingService teamMatchmakingService;
```
K-ETFreeman: If i not add @ Lazy here i get circular dependency issues, not sure is it ok or not tho

src/main/java/com/faforever/client/preferences/MatchmakerPrefs.java
K-ETFreeman: > yes

Done (i think?) see comment below

src/main/java/com/faforever/client/remote/FafServerAccessor.java
```diff
@@ -101,6 +108,12 @@ public void start() {
                                  .retry()
                                  .subscribe();
 
+      getEvents(VetoesChangedInfo.class).publishOn(fxApplicationThreadExecutor.asScheduler())
```
Sheikah45: Since this is team matchmaking specific the team matchmaking service should do the subscribing and have the onVetoesChanged function.

This aalso resolves the circular dependency

Sheikah45 (changes requested):

src/main/java/com/faforever/client/remote/FafServerAccessor.java
K-ETFreeman: yep, done

K-ETFreeman:
If all OK now - need to make the RC from this. Anything required from me? @Sheikah45

Issue #3342, opened 2025-05-17
[Feature]: Use javafx gradle plugin for native library management


pat-mccusker:
I was taking a look at incorporating some additional dependency management beyond spring boot and came across this pretty nifty [gradle plugin for javafx](https://github.com/openjfx/javafx-gradle-plugin) and figured it could be its own issue.

It saves native dependency resolution from having to be done manually in this build.gradle, and there appears to be some more specific platform library resolution than what is currently present.

For example, the following are resolved when building on apple silicon:

![Image](https://github.com/user-attachments/assets/46d78ee5-fcb9-462a-87dd-82e73737a870)

### Describe the solution you'd like

The javafx plugin can be used to handle javafx library versions and native platform

### Describe alternatives you've considered

I had initially thought these libraries could be handled through the spring dependencyManagement plugin, but that doesn't appear to work too well for resolving the correct native libraries.

pat-mccusker:
I possibly should have opened this as a bug; a majority of the tests in `develop` actually fail when running on my apple-silicon machine, but are fixed by the usage of the plugin for resolving the specific apple silicon native libraries.

Sheikah45:
I had tried to use this before but ran into some issues that I forget what they are now but can always look again to see if they may have resolved.

PR #3343, opened 2025-05-17
Feature/#3342 use javafx gradle plugin

pat-mccusker:
Implemented #3342

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,9 +21,26 @@ java {
 
 // These are read from gradle.properties
 version "${version}"
+project.ext.javafxPlatform = "${javafxPlatform}"
+project.ext.useJfxFromProperties = javafxPlatform != "unspecified"
 
 bootJar.enabled = false
 jar.enabled = true
+
+if (javafxPlatform == "unspecified") {
+  switch (OperatingSystem.current()) {
+    case OperatingSystem.LINUX:
+      project.ext.javafxPlatform = "linux"
+      break
+    case OperatingSystem.MAC_OS:
+      project.ext.javafxPlatform = "mac"
+      break
+    case OperatingSystem.WINDOWS:
+      project.ext.javafxPlatform = "win"
+      break
+  }
+}
+println "Platform is: ${javafxPlatform}"
 
 compileJava.dependsOn(processResources)
 processResources {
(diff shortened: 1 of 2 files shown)


pat-mccusker:
The majority of tests in `develop` fail for me on my apple silicon machine, and with this plugin all tests are passing.

build.gradle
```diff
@@ -153,14 +136,14 @@ tasks.register('buildInstall4jMediaFiles', Install4jTask) {
...
       mediaTypes = ['unixInstaller', 'unixArchive']
       break
-    case "mac":
+    case OperatingSystem.MAC_OS:
       mediaTypes = ['macos', 'macosFolder', 'macosArchive', 'macosFolderArchive']
       break
-    case "win":
+    case OperatingSystem.WINDOWS:
       mediaTypes = ['windows', 'windowsArchive']
       break
```
Sheikah45: This won't work for us because we cross compile using the GitHub actions that are running on Linux containers.

Sheikah45 (changes requested):

build.gradle
```diff
@@ -3,6 +3,7 @@ import org.gradle.internal.os.OperatingSystem
 
 plugins {
   id("org.springframework.boot") version "3.4.4"
+  id("io.spring.dependency-management") version "1.1.7"
```
Sheikah45, line 6: The spring boot plugin also applies the dependency management

build.gradle
pat-mccusker: I see, I'll change that back to use a gradle property.  Would you support a property rename given that strictly speaking it wouldn't be used for javaFx anymore?  I can make the update to the actions file as well.

pat-mccusker: Given your point I'm realizing I'll have to rework how the platform is set for the javaFx plugin to support those cross-compilations, while maintaining the support for automatic local platform matching.  I'll keep playing with it.

pat-mccusker: All right, both build contexts should be functional now, with the plugin deciding the jfx platform only when there is not an operating system value read in from config.

Brutus5000:
@pat-mccusker Did you get the client to launch at all? On my MBP M2 it just crashes without any exception on startup (on the dedicated ` mac-running` branch with MacOS spring profile).

pat-mccusker:
@Brutus5000 Looks like I'm still not able to launch the client with these changes either.

On develop my M3 fails with `Caused by: java.lang.RuntimeException: Error initializing QuantumRenderer: no suitable pipeline found` which appears to be related to the javafx library issue.

On this branch I get a little further but fail on `Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.faforever.client.os.OperatingSystem]: Factory method 'runtimeDetection' threw exception with message: 'int org.bridj.Platform.sizeOf_ptrdiff_t()'
`

pat-mccusker:
Looks like apple silicon is still [unsupported by BridJ](https://github.com/nativelibs4java/BridJ/issues/115)

pat-mccusker:
I think given the state of BridJ, it could make sense to move towards expanding our direct JNA usage.  For example, I can get through the `OsConfiguration::runtimeDetection` just by substituting the `com.sun.jna.Platform` static os methods for the `org.bridj.Platform` methods.

There appear to be other issues, but if it makes sense to you guys I think I could open a separate issue to work out a JNA-only replacement for our BridJ usage.

Brutus5000:
I found hopefully suitable replacements for bridj (see https://github.com/FAForever/downlords-faf-client/pull/3344)

pat-mccusker:
@Sheikah45 I think the issue you mentioned is resolved, how do things look now?

Sheikah45 (approved):

PR #3349, opened 2025-05-28
Add hmac to user and api

Sheikah45:


src/main/java/com/faforever/client/api/TokenRetriever.java
```diff
@@ -59,7 +60,17 @@ private Mono<Void> getExpirationMono(OAuth2AccessToken token) {
...
+    return refreshedTokenMono.map(OAuth2AccessToken::getTokenValue).doOnError(this::onTokenError);
+  }
+
+  public Mono<String> getRefreshedHmacValue() {
+    return getRefreshedTokenValue().flatMap(tokenValue -> {
+      try {
+        return Mono.just(JWTParser.parse(tokenValue).getJWTClaimsSet().getJSONObjectClaim("ext").get("hmac").toString());
+      } catch (ParseException e) {
+        return Mono.error(e);
+      }
```
Brutus5000: Why only ParseException? Could in theory also be NullPointerException if the claim is missing

Brutus5000 (approved):

src/main/java/com/faforever/client/api/TokenRetriever.java
Sheikah45: ParseException is a CheckedException. RuntimeExceptions naturally flow through already but the parse exception has to be caught because java

PR #3354, opened 2025-06-02
Bugfix/#3353 apple silicon startup

### LINKED ISSUE #3353: [Bug]: Set java.awt.headless to false

There appear to be a few more blockers for booting up on apple silicon (in the order I encountered them):

1. discord-rpc/libdiscord-rpc.dylib native linker failures
2. java.awt.HeadlessException resulting from the `install4j` `StartupNotification#registerStartupListener` called by this project's `FileOpeningHandler`
3. steamworks4j/1.9.0/libsteamworks4j.dylib native linker failures

I suspect these were all known issues as they're each accounted for in the `mac-running` branch.  However I still get a phantom exception in `mac-running` that appears to be related to point 2 above.

Starting the client on the `mac-running` branch with `mac` profile active only requires setting the system property `java.awt.headless` to `false` (spring boot sets this value to `true` by default).

On `develop` on apple silicon the client will start up with just this property tweak and the mac active profile + `@Profile` annotations placed on the Steam and Discord support beans.  I have also tested the `java.awt.headless` property tweak on a windows machine and am still able to boot up the client.

While on macOS apple silicon the faf client window appears with the browser login prompt, clicking the button has no effect and there are no errors in the logs, so this is not a complete fix.

Was leaving unaltered spring boot's default for headless mode intentional?  It seems strange given this is a GUI but there could be factors I'm not aware of.

I plan on opening a pr against `develop` but it could also make sense to integrate it with `mac-running`.

pat-mccusker:
Adding the `java.awt.headless` run config param and the annotations from `mac-running` allow the client to start up on `develop`.  

~~Browser login button is still not functional/has no effect.~~ URL encoding the oauth scopes fixes the browser login functionality (reaches the success checkmark page) and continues to work on windows.   However, on macos it seems like the websocket connection times out without emitting in the [FafLobbyClient](https://github.com/FAForever/faf-java-commons/blob/a12989f68d8c5d5e461cdc71b5034d5528059da2/lobby/src/main/kotlin/com/faforever/commons/lobby/FafLobbyClient.kt#L138) (Same timeout behavior in mac-running with the fixes in this pr applied)

Fixes #3353

### DIFF
--- a/.idea/runConfigurations/Main.xml
+++ b/.idea/runConfigurations/Main.xml
@@ -7,7 +7,7 @@
     </envs>
     <option name="MAIN_CLASS_NAME" value="com.faforever.client.Main" />
     <module name="faf-client.main" />
-    <option name="VM_PARAMETERS" value="-Dio.netty.allocator.type=unpooled -DnativeDir=build/resources/native -Dprism.forceGPU=true -Xmx1g -XX:MinHeapFreeRatio=15 -XX:MaxHeapFreeRatio=33 -XX:+HeapDumpOnOutOfMemoryError -XX:+UseZGC -XX:+ZGenerational -Djava.net.preferIPv4Stack=true -XX:TieredStopAtLevel=1 --add-opens=java.base/java.lang=ALL-UNNAMED" />
+    <option name="VM_PARAMETERS" value="-Dio.netty.allocator.type=unpooled -DnativeDir=build/resources/native -Dprism.forceGPU=true -Xmx1g -XX:MinHeapFreeRatio=15 -XX:MaxHeapFreeRatio=33 -XX:+HeapDumpOnOutOfMemoryError -XX:+UseZGC -XX:+ZGenerational -Djava.awt.headless=false -Djava.net.preferIPv4Stack=true -XX:TieredStopAtLevel=1 --add-opens=java.base/java.lang=ALL-UNNAMED" />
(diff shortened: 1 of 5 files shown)


Brutus5000:
Did you take a look at the current macos branch? https://github.com/FAForever/downlords-faf-client/tree/mac-running

pat-mccusker:
Yeah I went into more detail with my troubleshooting steps in the issue but the tl;dr is that even in mac running I couldn't get to the login prompt without this `java.awt.headless` property.

And since I only needed that along with some of the annotations already present in mac-running to get `develop` to reach the same point, I thought it worth opening the pr against develop.

Brutus5000:
I cherry-picked parts of your and my branch onto develop.
The remains of the `mac-running` branch are not working as far as I can see, so I will drop it.

pat-mccusker:
Ok, I just pulled develop and I still get the headless exception when starting up.  Are we against cherry-picking [6adba53](https://github.com/FAForever/downlords-faf-client/pull/3354/commits/6adba5385394003449d5bb1ff2e093cfc0802102) as well?  I'm able to start up once I add that param to the run config.

Brutus5000:
I simply didn't have time yet to check this one one my MacBook and my Linux machine =)

Brutus5000:
All merged via cherry-picks

PR #3386, opened 2025-08-18
Deploy releases to Cloudflare R2

Brutus5000:
I want to get rid of my own webspace as the mirror.

Not really sure how to test this...

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -154,5 +154,5 @@ jobs:
         run: |
           aws s3 cp \
             ${{ steps.artifact_paths.outputs.WINDOWS_EXE }} \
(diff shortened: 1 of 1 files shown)


Sheikah45:
We could create a release from this branch to test it. Although will it get a static url?

Brutus5000:
The urls created should be `https://downloads.faforever.com/faf-client/<filename>`

Sheikah45:
Can they be removed or overwritten?

Brutus5000:
Overwrite is possible even though we could disable it. They get deleted after 1 year on the bucket.

Sheikah45:
Hmm so if we don't release for a year then things would break

Brutus5000:
Well yes. But we can change it anytime. I was to lazy to build a worker to keep only the last X releases.

Sheikah45 (approved):

PR #3392, opened 2025-09-05
Add force relay option to ice adapter based on ice session

Brutus5000:


src/main/java/com/faforever/client/fa/relay/ice/IceAdapterImpl.java
```diff
@@ -241,6 +241,11 @@ List<String> buildCommand(Path workDirectory, int adapterPort, int gpgPort, int
 
     cmd.addAll(standardIceOptions);
 
+    if (forceRelay) {
+      cmd.add("--force-relay");
+      log.info("Forcing ice adapter relay connection");
```
Sheikah45:
```suggestion
      log.debug("Forcing ice adapter relay connection");
```

src/main/java/com/faforever/client/game/GameRunner.java
```diff
@@ -391,12 +391,10 @@ private Process launchOnlineGame(GameParameters gameParameters, Integer gpgPort,
...
   private CompletableFuture<Integer> startIceAdapter(int uid) {
-    return iceAdapter.start(uid)
-                     .thenCompose(icePort -> coturnService.getSelectedCoturns(uid)
-                                                          .collectList()
-                                                          .doOnNext(iceAdapter::setIceServers)
-                                                          .thenReturn(icePort)
-                                                          .toFuture());
+    return coturnService.getIceSession(uid).toFuture().thenCompose(session -> {
+      iceAdapter.setIceServers(session.servers());
+      return iceAdapter.start(uid, session.forceRelay());
```
Sheikah45: I don't think the iceadapter will accept the ice servers being set before it starts since that is an rpc call

src/main/java/com/faforever/client/fa/relay/ice/CoturnService.java
```diff
@@ -23,18 +25,34 @@ public class CoturnService {
...
+                                    .filter(coturnServer -> forgedAlliancePrefs.getPreferredCoturnIds()
+                                                                               .contains(coturnServer.getId()))
+                                    .toList();
+
+      if (preferredServers.isEmpty()) {
+        return session;
+      } else {
+        return new IceSession(session.id(), session.forceRelay(), preferredServers);
+      }
+    });
```
Sheikah45:
```suggestion
    Set<String> preferredCoturns = forgedAlliancePrefs.getPreferredCoturnIds();
    return fafApiAccessor.getApiObject("/ice/session/game/" + gameId, IceSession.class).map(session -> {
      List<CoturnServer>preferredServers = session.servers()
                                    .stream()
                                    .filter(coturnServer -> preferredCoturns.contains(coturnServer.getId()))
                                    .toList();

      if (preferredServers.isEmpty()) {
        return session;
      } else {
        return new IceSession(session.id(), session.forceRelay(), preferredServers);
      }
    });
```

Sheikah45 (changes requested):

src/main/java/com/faforever/client/game/GameRunner.java
Brutus5000: Yeah wasn't aware its an rpc call, thought it just sets a property. Please check again.

Sheikah45 (approved):

PR #3394, opened 2025-09-08
Improve local replays tab loading time

### LINKED ISSUE #3270: High loading time every time you go to local replays tab

Every time you select the local replays tab there is a signifant loading time. I guess this is because the client parses all the local replays, which can be a lot. 

The loading time is understandable on the first opening, but then there should be some sort of caching to prevent this

Maybe the client could do some sort of check if a file was added to the folder instead of doing a time-based cache, which seems suboptimal

Rogerstalker:
Fixes #3270  

Some respones to the issue

> Every time you select the local replays tab there is a signifant loading time. I guess this is because the client parses all the local replays, which can be a lot.

Upon selecting the Local replays tab the client does not parse all local replays, only those, which would be shown on the current page, which is 20 by default. I understand this guess, however it just took this much time to load 20.

> The loading time is understandable on the first opening, but then there should be some sort of caching to prevent this

I thought the loading time is too high for parsing and displaying just 20 replays even on the first opening, therefore I investigated if it could be improved. I created a companion issue https://github.com/FAForever/faf-java-commons/issues/250 and a corresponding pull request https://github.com/FAForever/faf-java-commons/pull/251 with my findings and improvements.

> Maybe the client could do some sort of check if a file was added to the folder instead of doing a time-based cache, which seems suboptimal

Since the assumption that all files are parsed turned out to be not true I don't think monitoring file changes is the way to go, so I went with the caching solution with size and time limits. I also parallelized replay loading.

### Description

Implemented replay caching for ReplayService

Parallelized local replay loading to improve performance

Fixed a bug, where two separate subscribed actions prompted the VaultEntityController to load the replays twice on changing the number of replays per page combobox.

### Measurements
I measured the performance of the parser improvements in faf-java-commons and the client with the help of VisualVM through the action of changing the local replay tab per page number combobox form 20 to 100. CPU: AMD Ryzen 9 5950x. I included my local repo of faf-java-commons in the client's gradle build settings, to gain access to my changes faster. These measurements are not that precise, since during measurements new local replays were added, whose parsing times are different than the ones, which were pushed out from the first 100 replays. Also in some cases I changed the faf-java-commons code in conjunction with the client code. Still I think these could serve as an adequate set of snapshots to properly protray the trajectory of the code changes and their results.

See attached measurements:
[Client Local Replay tab load profiling.zip](https://github.com/user-attachments/files/22199666/Client.Local.Replay.tab.load.profiling.zip)

Baseline
We can immediately notice, the bug, which makes the vault controller load the page twice.
**Client original 100.nps**
com.faforever.client.replay.ReplayService.loadLocalReplayPage ()	9 799 ms
com.faforever.client.replay.ReplayService.loadLocalReplayPage ()	9 609 ms

faf-java-commons changes state similar to measurement in companion task: **ByteBuffer full.nps**
(Replaced LittleEndianDataInputStream with ByteBuffer just in the tokenizer and used System.arraycopy() to extract TokenContent bytes
Used ByteBuffer instead of the LittleEndianDataInputStream everywhere throughout the parsing process.)
**Client improved 100.nps**
com.faforever.client.replay.ReplayService.loadLocalReplayPage ()	8 600 ms
com.faforever.client.replay.ReplayService.loadLocalReplayPage ()	8 199 ms

faf-java-commons changes state similar to measurement in companion task: **ByteBuffer full final.nps**
(Reused the initially created ByteBuffer everywhere to avoid buffer creation for individual tokens.)
**Client improved single buffer 100.nps**
com.faforever.client.replay.ReplayService.loadLocalReplayPage ()	7 398 ms
com.faforever.client.replay.ReplayService.loadLocalReplayPage ()	7 200 ms

Fix of double loading in client
**Client improved single buffer single changeload fix 100.nps**
com.faforever.client.replay.ReplayService.loadLocalReplayPage ()	7 096 ms

Implemented parallel loading of replays on a single page
**Client improved single buffer single changeload fix parallel 100.nps**
com.faforever.client.replay.ReplayService.loadLocalReplayPage ()	3 111 ms

Changed `loadLocalReplayPage` to only use one stream for filtering replay results and parallel completing the `CompletableFutures` from `tryLoadingLocalReplay`
**Client improved single buffer single changeload fix parallel 1stream 100.nps**
com.faforever.client.replay.ReplayService.loadLocalReplayPage ()	2 999 ms

### DIFF
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -36,8 +36,11 @@ import com.faforever.commons.api.elide.ElideNavigatorOnId;
 import com.faforever.commons.replay.ReplayDataParser;
 import com.faforever.commons.replay.ReplayMetadata;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.jetbrains.annotations.NotNull;
 import org.springframework.beans.factory.ObjectFactory;
 import org.springframework.cache.annotation.Cacheable;
 import org.springframework.context.annotation.Lazy;
@@ -58,12 +61,18 @@ import java.nio.file.StandardCopyOption;
 import java.nio.file.attribute.FileTime;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -103,6 +112,10 @@ public class ReplayService {
   private final DataPrefs dataPrefs;
   private final ObjectFactory<ReplayDownloadTask> replayDownloadTaskFactory;
   private final ReplayHistoryPrefs replayHistory;
+  private final Cache<@NotNull Path, @NotNull Replay> replayCache = CacheBuilder.newBuilder()
+                                                                                .maximumSize(1000)
+                                                                                .expireAfterWrite(2, TimeUnit.HOURS)
+                                                                                .build();
 
   @VisibleForTesting
   static Integer parseSupComVersion(ReplayDataParser parser) {
@@ -172,25 +185,46 @@ public class ReplayService {
 
       int numPages = filesList.size() / pageSize;
 
-      List<CompletableFuture<Replay>> replayFutures = filesList.stream()
-                                                               .skip(skippedReplays)
-                                                               .limit(pageSize)
-                                                               .map(this::tryLoadingLocalReplay)
-                                                               .filter(e -> !e.isCompletedExceptionally())
-                                                               .toList();
+      List<Replay> replays;
+      try (ExecutorService executor = Executors.newFixedThreadPool(4)) {
+        List<Callable<Replay>> tasks = new ArrayList<>();
+        int limit = Math.min(skippedReplays + pageSize, filesList.size());
+        for (int i = skippedReplays; i < limit; i++) {
+          var filePath = filesList.get(i);
+          var c = new Callable<Replay>() {
+            @Override
+            public Replay call() {
+              return tryLoadingLocalReplay(filePath).join();
+            }
+          };
+          tasks.add(c);
+        }
 
-      return Mono.fromFuture(CompletableFuture.allOf(replayFutures.toArray(new CompletableFuture[0]))
-                                              .thenApply(ignoredVoid -> replayFutures.stream()
-                                                                                     .map(CompletableFuture::join)
-                                                                                     .filter(Objects::nonNull)
-                                                                                     .collect(Collectors.toList())))
-                 .zipWith(Mono.just(numPages));
+        try {
+          replays = executor.invokeAll(tasks).stream().map(result -> {
+            try {
+              return result.get();
+            } catch (InterruptedException | ExecutionException e) {
+              throw new RuntimeException(e);
+            }
+          }).filter(Objects::nonNull).toList();
+        } catch (InterruptedException e) {
+          throw new RuntimeException(e);
+        }
+      }
+
+      return Mono.justOrEmpty(replays).zipWith(Mono.just(numPages));
     }
   }
 
 
   private CompletableFuture<Replay> tryLoadingLocalReplay(Path replayFile) {
     try {
+      final Replay cachedReplay = this.replayCache.getIfPresent(replayFile);
+      if (cachedReplay != null) {
+        return CompletableFuture.completedFuture(cachedReplay);
+      }
+
       ReplayDataParser replayData = replayFileReader.parseReplay(replayFile);
       ReplayMetadata replayMetadata = replayData.getMetadata();
 
@@ -205,7 +239,10 @@ public class ReplayService {
         if (mapVersion == null) {
           log.warn("Could not find map for replay file `{}`", replayFile);
         }
-        return replayMapper.map(replayData, replayFile, featuredMod, mapVersion);
+
+        final Replay replay = replayMapper.map(replayData, replayFile, featuredMod, mapVersion);
+        this.replayCache.put(replayFile, replay);
+        return replay;
       }).exceptionally(throwable -> {
         log.warn("Could not read replay file `{}`", replayFile, throwable);
         moveCorruptedReplayFile(replayFile);
@@ -289,7 +326,10 @@ public class ReplayService {
     List<ChatMessage> chatMessages = replayDataParser.getChatMessages().stream().map(replayMapper::map).toList();
     List<GameOption> gameOptions = Stream.concat(
         Stream.of(new GameOption("FAF Version", String.valueOf(parseSupComVersion(replayDataParser)))),
-        replayDataParser.getGameOptions().stream().map(replayMapper::map).sorted(Comparator.comparing(GameOption::key, String.CASE_INSENSITIVE_ORDER))).toList();
+        replayDataParser.getGameOptions()
+                        .stream()
+                        .map(replayMapper::map)
+                        .sorted(Comparator.comparing(GameOption::key, String.CASE_INSENSITIVE_ORDER))).toList();
 
     String mapFolderName = parseMapFolderName(replayDataParser);
     Map map = new Map(null, mapFolderName, 0, null, false, null, null);
@@ -447,8 +487,7 @@ public class ReplayService {
   @Cacheable(value = CacheNames.REPLAYS_SEARCH, sync = true)
   public Mono<Replay> findById(int id) {
     ElideNavigatorOnId<Game> navigator = ElideNavigator.of(Game.class).id(String.valueOf(id));
-    return fafApiAccessor.getOne(navigator).map(replayMapper::map)
-                         .cache();
+    return fafApiAccessor.getOne(navigator).map(replayMapper::map).cache();
   }
 
   @Cacheable(value = CacheNames.REPLAYS_MINE, sync = true)
@@ -469,10 +508,13 @@ public class ReplayService {
   }
 
   public Flux<LeagueScoreJournal> getLeagueScoreJournalForReplay(Replay replay) {
-    ElideNavigatorOnCollection<com.faforever.commons.api.dto.LeagueScoreJournal> navigator = ElideNavigator.of(com.faforever.commons.api.dto.LeagueScoreJournal.class).collection()
-        .setFilter(qBuilder().intNum("gameId").eq(replay.id()));
-    return fafApiAccessor.getMany(navigator)
-        .map(replayMapper::map)
-        .cache();
+    ElideNavigatorOnCollection<com.faforever.commons.api.dto.LeagueScoreJournal> navigator = ElideNavigator.of(
+                                                                                                               com.faforever.commons.api.dto.LeagueScoreJournal.class)
+                                                                                                           .collection()
+                                                                                                           .setFilter(
+                                                                                                               qBuilder().intNum(
+                                                                                                                             "gameId")
+                                                                                                                         .eq(replay.id()));
+    return fafApiAccessor.getMany(navigator).map(replayMapper::map).cache();
   }
 }
--- a/src/main/java/com/faforever/client/replay/ReplayService.java
+++ b/src/main/java/com/faforever/client/replay/ReplayService.java
@@ -61,17 +61,12 @@ import java.nio.file.StandardCopyOption;
 import java.nio.file.attribute.FileTime;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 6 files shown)


Sheikah45:
I am not sure about the caching of the replays for such a long time since this could lead to larger memory overhead for users after they open the replay tab. If anything it should just cache the page for similar timescales as we cache the other vault page contents and API calls using the configured Caffeine Caches.

Additionally the parallelization of the parsing is a little bit of a minefield since you optimized it for your system it might not port well to all the different kinds of hardware that others might have.

Rogerstalker:
> I am not sure about the caching of the replays for such a long time since this could lead to larger memory overhead for users after they open the replay tab. If anything it should just cache the page for similar timescales as we cache the other vault page contents and API calls using the configured Caffeine Caches.

My thought process was based on how I used this feature. I like to check out the replay right after a game , when the online replay is usually not available yet. I wanted the cache to be alive after a long game, if I accessed it previously, hence the 2 hour limit. If this is excessive we could certainly lower it. I did not look too deep into the used caching solutions, I could try to apply the same Caffeine caching used elsewhere.
 
> Additionally the parallelization of the parsing is a little bit of a minefield since you optimized it for your system it might not port well to all the different kinds of hardware that others might have.

Could you suggest a better way? It is clearly a lot of perf gain. I was a bit stumped when I saw a bunch of out of memory exceptions in my faf-java-commons parallelization tests when working with such small amount of data. Not sure how optimized it is to my system aside from testing on it. I expected the parallelization gains to level out around the number of "threads" supported by the CPU or at least the number of physical cores, not 4.

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -172,25 +185,46 @@ public Mono<Tuple2<List<Replay>, Integer>> loadLocalReplayPage(int pageSize, int
...
-                                                               .toList();
-
-      return Mono.fromFuture(CompletableFuture.allOf(replayFutures.toArray(new CompletableFuture[0]))
-                                              .thenApply(ignoredVoid -> replayFutures.stream()
-                                                                                     .map(CompletableFuture::join)
-                                                                                     .filter(Objects::nonNull)
-                                                                                     .collect(Collectors.toList())))
-                 .zipWith(Mono.just(numPages));
+      List<Replay> replays;
+      try (ExecutorService executor = Executors.newFixedThreadPool(4)) {
```
GodFuper: It's unnecessary. Try 

```
public Mono<Tuple2<List<Replay>, Integer>> loadLocalReplayPage(int pageSize, int page) throws IOException {
    String replayFileGlob = clientProperties.getReplay().getReplayFileGlob();

    Path replaysDirectory = dataPrefs.getReplaysDirectory();
    if (Files.notExists(replaysDirectory)) {
      Files.createDirectories(replaysDirectory);
    }

    int skippedReplays = pageSize * (page - 1);
    List<Path> filesList;
    try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(replaysDirectory, replayFileGlob)) {
      filesList = StreamSupport.stream(directoryStream.spliterator(), false).sorted(Comparator.comparing(path -> {
        try {
          return Files.getLastModifiedTime((Path) path);
        } catch (IOException e) {
          log.warn("Could not get last modified time of file {}", path, e);
          return FileTime.from(Instant.EPOCH);
        }
      }).reversed()).parallel().toList();
    }

    return Mono.fromFuture(CompletableFuture.supplyAsync(() -> filesList.stream()
                                                                        .skip(skippedReplays)
                                                                        .limit(pageSize)
                                                                        .map(this::tryLoadingLocalReplay)
                                                                        .filter(e -> !e.isCompletedExceptionally())
                                                                        .map(CompletableFuture::join)
                                                                        .filter(Objects::nonNull)
                                                                        .collect(Collectors.toList())))
               .zipWith(Mono.just(filesList.size() / pageSize));
  }
```

`(Timed) Mono com.faforever.client.replay.ReplayService.loadLocalReplayPage(int,int), took 5 ms`

Rogerstalker: I'm not sure how you timed this. What I did was, I commented out this part of the caching: 
```
//      final Replay cachedReplay = this.replayCache.getIfPresent(replayFile);
//      if (cachedReplay != null) {
//        return CompletableFuture.completedFuture(cachedReplay);
//      }
```
from `tryLoadingLocalReplay` to not be influenced by it. I also copy pasted your code. Navigated to the Local Replays tab. Started CPU sampling with VisualVM. Changed the ItemsPerPage combobox to 100. Stopped sampling. What I got is this: com.faforever.client.replay.ReplayService.tryLoadingLocalReplay ()	7 804 ms. Essentially back to the single threaded performance on first load.
[Client improved single buffer single changeload fix GodFuper parallel stream 100.zip](https://github.com/user-attachments/files/22242122/Client.improved.single.buffer.single.changeload.fix.GodFuper.parallel.stream.100.zip)

Rogerstalker:
I looked into the CaffeineCache usage, if I correctly understand how they work we can't directly apply the cache to the `loadLocalReplayPage` because it only has the page size and page parameters, which are not enough to deremine if the replays should be read again as there could be new files added to the folder since caching. Applying the cache to `tryLoadingLocalReplay` would not work because it is only called from its class, which circumvents the interception of calls if we don't implement workarounds or refactor this part of the class into a new one. https://stackoverflow.com/questions/16899604/spring-cache-cacheable-not-working-while-calling-from-another-method-of-the-s/. Maybe I am not seeing an easy solution, so let me know if there is a better way to do this. In the meantime I reduced the cache and hopefully improved the parallelization a bit.

Rogerstalker:
I also updated the code according to the faf-java-commons changes, but that means we would need a new `commonsVersion`. I updated it too, but seems like just changing it to the commit hash is not enough. Please point me in the right direction if I can do anything about this.

Sheikah45:
The multi threading should be removed since it can be a little harmful across different machines due to the core count and setup.

Additionally you might have to change the repo/coordinates of commons since it uses maven central now I believe. Or you will have to start the build on jitpack.

Brutus5000:
https://mvnrepository.com/artifact/com.faforever.commons 

The last merge is already released

src/main/java/com/faforever/client/replay/ReplayService.java
GodFuper: I did not put it quite correctly. If you move all the load in the `Future`, the application window will not hang. It will be processed in Async mode. 

Now, if you specify the download of 100 files, the application freezes for 5-10 seconds.
If you move it to Async, the window will not freeze, but the loading animation will be shown and you can return to another window at any time.

Rogerstalker:
> The multi threading should be removed since it can be a little harmful across different machines due to the core count and setup.
> 
> Additionally you might have to change the repo/coordinates of commons since it uses maven central now I believe. Or you will have to start the build on jitpack.

I limited the number of threads for the threadpool to 4, or the number of logical cores, whichever is lower. What kind of harm should we be aware of?

src/main/java/com/faforever/client/replay/ReplayService.java
Rogerstalker: I think it would be worthwile to apply this, even to other pages, because the application usually hangs on most interactions. That said, it won't make the replay tab loading time shorter. I'd suggest a separate issue for this.

Sheikah45:
The harm is just the cpu usage spike and consumption of resources with a static threadpool like that across a variety of cpu architectures. The threadpool should be removed and it can just load serially with the caching

GodFuper:
I conducted an experiment. 
Without caching:
With `Executors.newFixedThreadPool(10)`, with `Executors.newVirtualThreadPerTaskExecutor()` and with what was originally. I always load 100 replays stably in 4-4.5 seconds. 

With caching for the first time in 4-4.5 seconds, then correspondingly faster in 5-30ms. But there is a flaw with mapgen maps, since initially there is no map, but it will still put replay in the cache. Even after the map is generated, the map image will not appear, meaning it needs to be cleared on a trigger.

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -172,25 +187,45 @@ public Mono<Tuple2<List<Replay>, Integer>> loadLocalReplayPage(int pageSize, int
...
-                                                               .toList();
-
-      return Mono.fromFuture(CompletableFuture.allOf(replayFutures.toArray(new CompletableFuture[0]))
-                                              .thenApply(ignoredVoid -> replayFutures.stream()
-                                                                                     .map(CompletableFuture::join)
-                                                                                     .filter(Objects::nonNull)
-                                                                                     .collect(Collectors.toList())))
-                 .zipWith(Mono.just(numPages));
+      List<Replay> replays;
+      try (ExecutorService executor = Executors.newFixedThreadPool(nThreads)) {
```
GodFuper: In addition, you can get Executor to work without changing the `loadLocalReplayPage()` function. 

To do this, add `@Async` before the `private CompletableFuture<Replay> tryladinglocalreplay(Path replayFile) {...}` function and move it to a separate bean accordingly (for the proxy to work correctly).

Again, as I said, this doesn't help much, since all operations already take place in Async. At least on my CPU and my SSD.

Caching helps, of course, but there are still problems when the replay has not loaded fully, for example, if it is a mapgen replay.

Rogerstalker:
> I conducted an experiment. Without caching: With `Executors.newFixedThreadPool(10)`, with `Executors.newVirtualThreadPerTaskExecutor()` and with what was originally. I always load 100 replays stably in 4-4.5 seconds.
> 
> With caching for the first time in 4-4.5 seconds, then correspondingly faster in 5-30ms. But there is a flaw with mapgen maps, since initially there is no map, but it will still put replay in the cache. Even after the map is generated, the map image will not appear, meaning it needs to be cleared on a trigger.

See PR of the companion task(https://github.com/FAForever/faf-java-commons/pull/251) on why I settled on the max 4 threads. 10 is way too much. My best guess is that this could be a storage or memory speed limitation.

Mapgen map images don't load in the release version either. I'd suggest a separate issue for this.

Rogerstalker:
> https://mvnrepository.com/artifact/com.faforever.commons
> 
> The last merge is already released

@Brutus5000 The build fails locally and the pr check build fails too with it not being able to find the common packages: https://github.com/FAForever/downlords-faf-client/actions/runs/18295067833/job/52091409431?pr=3394. Any idea why the packages are unavailable?

Brutus5000:
That is still old jitpack builds. You have to switch to the mvn repository artifacts.

Rogerstalker:
> The harm is just the cpu usage spike and consumption of resources with a static threadpool like that across a variety of cpu architectures. The threadpool should be removed and it can just load serially with the caching

That is kinda the point of parallelization, to use more resources but in a shorter amount of time. Anyways, I give up, I restored the original single threaded code.

GodFuper:
> See PR of the companion task([FAForever/faf-java-commons#251](https://github.com/FAForever/faf-java-commons/pull/251)) on why I settled on the max 4 threads. 10 is way too much. My best guess is that this could be a storage or memory speed limitation.

Well, I meant that why use an Executor with 4 or 10 threads if the processing speed does not increase. It is better then to use virtual threads have no restrictions at all.

And then I compared the original code, the code with 10 threads and virtual threads and realized that the speed is the same.

src/main/java/com/faforever/client/replay/ReplayService.java
GodFuper: Since you've returned the code to its original state, why don't you do as I suggested? It's a little better optimized here.

src/main/java/com/faforever/client/replay/ReplayService.java
```diff
@@ -84,6 +89,7 @@ public class ReplayService {
   public static final String SUP_COM_REPLAY_FILE_ENDING = ".scfareplay";
   private static final String TEMP_SCFA_REPLAY_FILE_NAME = "temp.scfareplay";
   private static final Pattern invalidCharacters = Pattern.compile("[?@*%{}<>|\"]");
+  private static final int nThreads = Math.min(Runtime.getRuntime().availableProcessors(), 4);
```
GodFuper: not used

src/main/java/com/faforever/client/replay/ReplayService.java
Rogerstalker: It would require adjusting unit tests and manual retesting and it is not relevant to performance improvement, which is the goal of this task. Feel free to create a new issue and implement it.

src/main/java/com/faforever/client/replay/ReplayService.java
Rogerstalker: removed

Rogerstalker:
> > See PR of the companion task([FAForever/faf-java-commons#251](https://github.com/FAForever/faf-java-commons/pull/251)) on why I settled on the max 4 threads. 10 is way too much. My best guess is that this could be a storage or memory speed limitation.
> 
> Well, I meant that why use an Executor with 4 or 10 threads if the processing speed does not increase. It is better then to use virtual threads have no restrictions at all.
> 
> And then I compared the original code, the code with 10 threads and virtual threads and realized that the speed is the same.

I'm quoting here from the companion PR I linked you:

> Reused the initially created ByteBuffer everywhere to avoid buffer creation for individual tokens.
> ByteBuffer full final.nps: com.faforever.commons.replay.ReplayDataParser. () 17 898 ms
> 
> Used this measurement to determine how parallelization of different replay parsing is scaling. By not limiting the number of used threads I ran into out of memory exceptions due to heap constraints, so I elected to limit them according to my measurements and saw the best performance with a limit of 4. Up to 4 threads there were measurable performance improvements, however more than 4 threads made the parsing actually slower.
> ByteBuffer full final parallel.nps: com.faforever.commons.replay.ReplayLoaderTableParserTest.ExecuteParallel () 13 299 ms

As you can see I stated that the the processing speed did increase with increasing the threadpool thread count from 1 until 4, but after 4 the performance was deteriorating.

src/main/java/com/faforever/client/vault/VaultEntityController.java
```diff
@@ -134,7 +134,6 @@ protected void onInitialize() {
                        ((observable, oldValue, newValue) -> onPerPageCountChanged(oldValue == null ? 0 : oldValue,
                                                                                   newValue)));
     perPageComboBox.setValue(20);
-    perPageComboBox.setOnAction((event -> changePerPageCount()));
```
Sheikah45, line 137: Why was this removed?

Rogerstalker: > Fixed a bug, where two separate subscribed actions prompted the VaultEntityController to load the replays twice on changing the number of replays per page combobox.

https://github.com/FAForever/downlords-faf-client/pull/3394/commits/fd25229df9f64c11a9e21e54f865d4df2b15e5c4

Sheikah45: Where is the second loading of the replays happeniing? iit looks like now the changePerPageCount function is unused in the VaultEntityController

Rogerstalker: perPageComboBox.valueProperty()
                   .addListener(
                       ((observable, oldValue, newValue) -> onPerPageCountChanged(oldValue == null ? 0 : oldValue,
                                                                                  newValue)));

Sheikah45: Got it then we can remove the unused function

Sheikah45 (approved):

src/main/java/com/faforever/client/vault/VaultEntityController.java
Rogerstalker: Removed and adjusted unit test

PR #3396, opened 2025-09-15
Split Replay logs into own set

### LINKED ISSUE #3352: [Feature]: Separate replay and game logs

When you watch a replay it overwrites the old log file, which can have useful user-side data from when you played the game. The replay only has sim side data and less useful observer user data.

Watching replays can also delete your game logs rather quickly because it is easy to watch many replays, even for a short moment.

Change the prefix for replay log files from `game_` to `replay_`.

Replay files should have a separate maximum count (game logs have a max of 10).

Putting replays in a different folder seems unnecessarily difficult to navigate because they match the purpose of game logs in analyzing game issues.

kais58:
Resolves: https://github.com/FAForever/downlords-faf-client/issues/3352

### DIFF
--- a/src/main/java/com/faforever/client/logging/LoggingService.java
+++ b/src/main/java/com/faforever/client/logging/LoggingService.java
@@ -32,6 +32,8 @@ import java.util.stream.Stream;
 public class LoggingService implements InitializingBean {
 
   private static final Pattern GAME_LOG_PATTERN = Pattern.compile("game(_\\d*)?.log");
+  private static final Pattern REPLAY_LOG_PATTERN = Pattern.compile("replay(_\\d*)?.log");
+  private static Pattern LOG_PATTERN;
   private static final int NUMBER_GAME_LOGS_STORED = 10;
 
   private final OperatingSystem operatingSystem;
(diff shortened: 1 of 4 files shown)


src/main/java/com/faforever/client/logging/LoggingService.java
```diff
@@ -69,10 +71,17 @@ public void afterPropertiesSet() throws IOException, InterruptedException, Joran
...
+  public Path getNewGameLogFile(int gameUID, boolean replay) {
+    if (replay) {
+      LOG_PATTERN = REPLAY_LOG_PATTERN;
+    } else {
+      LOG_PATTERN = GAME_LOG_PATTERN;
+    }
     try (Stream<Path> listOfLogFiles = Files.list(operatingSystem.getLoggingDirectory())) {
       listOfLogFiles
-          .filter(logPath -> GAME_LOG_PATTERN.matcher(logPath.getFileName().toString()).matches())
+          .filter(logPath -> LOG_PATTERN.matcher(logPath.getFileName().toString()).matches())
```
Sheikah45: We shouldn't be assigning a static variable here as this makes it not thread safe when people can run a replay and a game at the same time.

We should instead have two public methods with one private method that takes the pattern as a parameter

Sheikah45 (changes requested):

src/main/java/com/faforever/client/logging/LoggingService.java
```diff
@@ -69,10 +70,13 @@ public void afterPropertiesSet() throws IOException, InterruptedException, Joran
     JavaFxUtil.addAndTriggerListener(developerPrefs.logLevelProperty(), new WeakChangeListener<>(logLevelChangeListener));
   }
 
-  public Path getNewGameLogFile(int gameUID) {
+  public Path getNewGameLogFile(int gameUID) {return getNewLogFile(gameUID, GAME_LOG_PATTERN);}
+  public Path getNewReplayLogFile(int gameUID) {return getNewLogFile(gameUID, REPLAY_LOG_PATTERN);}
+
+  private Path getNewLogFile(int gameUID, Pattern LOG_PATTERN) {
```
Sheikah45: Here we can just proper case the variable

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #3431, opened 2025-11-03
Added checkboxes to include various command line switches

### LINKED ISSUE #3425: [Feature]: Add checkboxes to include various command line switches

No, it is not related to a problem.

Add a few additional settings to automatically include some command line switches that may be useful to some players. As two examples:

- `/exitongameover` to immediately exit the game when it is over.
- `/nosound` to make it more convenient to play without sound. This can be useful when you have sound issues.

The alternative is to populate the command line format yourself:

Image

But perhaps we should not provide players with a setting that is this broad. It allows you to override other command line switches, for example. If we can find all the useful switches then perhaps we can remove it.

Marc-Spector:
Closes #3425 

paint

## Summary by CodeRabbit

* **New Features**
  * Added "Skip Statistics" setting to prevent recording game statistics after matches.
  * Added "Disable Game Sound" setting to launch games without audio output.
  * Both toggles are now available in the preferences interface and take effect on game launch.


Marc-Spector:
@coderabbitai review

src/main/java/com/faforever/client/fa/GameLaunchLocalParametersResolver.java
Sheikah45, line 1: I don't quite understand the purpose of the game resolver and actively subscribing to the prefs object.

We only need to care what the preferences are when we launch the game right? I would think more that the prefs object would just be an observable set of the enums. Then we can just get that.

Then in the settings controller we can have the add remove logic since it is really more a UI translation rather than embedding that into the preferences.

It also gives a little more flexibility and less proliferation of preferences fields.

You can also inject the ForgedAlliancePrefs directly

src/main/java/com/faforever/client/fa/GameLaunchLocalParameter.java
Sheikah45, line 1: I don't think these enums need parameters. In the launcher they could just be in a switch statement and should only be used in one place.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/fa/LaunchCommandBuilder.java
```diff
@@ -207,10 +213,11 @@ public List<String> build() {
...
 
     command.addAll(split(String.format(executableDecorator, "\"" + executable.toAbsolutePath() + "\"")));
-    command.addAll(Arrays.asList(
-        "/init", ForgedAlliancePrefs.INIT_FILE_NAME,
-        "/nobugreport"
-    ));
+
+    List<GameLaunchLocalParameter> defaultGameParameters = Lists.newArrayList(INIT_FILE, NO_BUG_REPORT);
+    defaultGameParameters.addAll(gameLaunchLocalParameters != null ? gameLaunchLocalParameters : List.of());
+    defaultGameParameters.forEach(parameter -> command.addAll(parameter.getCommand()));
```
Sheikah45, line 219: Also I wouldn't add always present options to the enums. It should just be optional options.

Sheikah45 (changes requested):

src/main/java/com/faforever/client/fa/GameLaunchLocalParametersResolver.java
Marc-Spector: > We only need to care what the preferences are when we launch the game right?

That is right, you are right! I wrote the code, which was the first thing that came to mind. My implementation of the code is crap. I will redo it.

PR #3438, opened 2025-11-14
Correction of the loss of the current matchmaking status.

GodFuper:
There are cases when the UI and what is happening on the server are out of sync. You think you're on the queue, but in fact you're not even not in.

## Summary by CodeRabbit

* **Bug Fixes**
  * Matchmaking queue selection is now reliably tracked and preserved across UI changes and server updates.
  * Queue status and info stay synchronized with live server/game changes.

* **New Features**
  * Queue selections are auto-applied when joining matchmaking and queue game counts update in real time.


Sheikah45 (commented):
The prefs already store the non sselected queues. So adding the selectedQueueIds doesn't add any information. Rather than adding that the reason why the unselected queues are workiing should be fixed instead.

Also the matchmaker games already update as game messages come in

PR #3464, opened 2026-02-02
Mask JWT in logs

mchiefsauce:
## Summary by CodeRabbit

* **New Features**
  * Authentication secrets (access tokens, refresh tokens, HMACs) are automatically masked in application logs to prevent exposure.
  * Masking is applied consistently to all token-related log output.

* **Chores**
  * Logging infrastructure enhanced to support comprehensive token masking.
  * Test coverage extended to verify masking of refresh tokens.

### DIFF
--- a/src/main/java/com/faforever/client/preferences/LoginPrefs.java
+++ b/src/main/java/com/faforever/client/preferences/LoginPrefs.java
@@ -1,5 +1,7 @@
 package com.faforever.client.preferences;
 
+import com.faforever.client.util.LogMaskingRegistry;
+import com.faforever.client.util.LogMaskingRegistry.SensitiveValueType;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.property.SimpleStringProperty;
@@ -9,8 +11,13 @@ import javafx.beans.property.StringProperty;
 
 public class LoginPrefs {
   private final StringProperty endpoint = new SimpleStringProperty();
-  private final StringProperty refreshToken = new SimpleStringProperty();
+  private final StringProperty refreshToken;
   private final BooleanProperty rememberMe = new SimpleBooleanProperty(true);
+
+  public LoginPrefs() {
+    refreshToken = new SimpleStringProperty();
+    refreshToken.addListener((_, _, newValue) -> LogMaskingRegistry.update(SensitiveValueType.REFRESH_TOKEN, newValue));
+  }
 
   public String getEndpoint() {
     return endpoint.get();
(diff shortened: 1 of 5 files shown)


Sheikah45:
Rather than doing this inline it should be added to the MaskPatternLayout that handles all log messages.

mchiefsauce:
> Rather than doing this inline it should be added to the MaskPatternLayout that handles all log messages.

Good idea! Did the change.

src/main/java/com/faforever/client/preferences/LoginPrefs.java
```diff
@@ -9,9 +11,14 @@
 
 public class LoginPrefs {
   private final StringProperty endpoint = new SimpleStringProperty();
-  private final StringProperty refreshToken = new SimpleStringProperty();
+  private final StringProperty refreshToken;
   private final BooleanProperty rememberMe = new SimpleBooleanProperty(true);
 
+  public LoginPrefs() {
+    refreshToken = new SimpleStringProperty();
+    refreshToken.addListener((_, _, newValue) -> LogMaskingRegistry.update(SensitiveValueType.REFRESH_TOKEN, newValue));
```
Sheikah45: We don't need a listener here and in the TokenRetriever on the refreshToken since it is bound.  I would just add it in the TokenRetriever since that is the source of the token

src/main/java/com/faforever/client/util/LogMaskingRegistry.java
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Masks log message
+   */
+  static String maskMessage(String message) {
+    if (message == null || message.isEmpty()) {return message;}
+
+    String maskedMessage = message;
+    for (LogMaskingRegistry.SensitiveValueType type : LogMaskingRegistry.SensitiveValueType.values()) {
+      List<String> secrets = LogMaskingRegistry.STORE.get(type).get();
+      if (secrets.isEmpty()) {continue;}
```
Sheikah45: The isEmpty check isn't necessary since that s handled for us by the enhanced for loop

src/main/java/com/faforever/client/util/LogMaskingRegistry.java
```diff
@@ -0,0 +1,62 @@
...
+      }
+      return List.of(newValue, existing.getFirst());
+    });
+  }
+
+  /**
+   * Masks log message
+   */
+  static String maskMessage(String message) {
+    if (message == null || message.isEmpty()) {return message;}
```
Sheikah45:
```suggestion
    if (message == null || message.isBlank()) {
        return message;
    }
```

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #3504, opened 2026-04-18
Add macOS (Apple Silicon) support

jfuruness:
Enables running the FAF client natively on Apple Silicon, with Supreme Commander launched through               
  `wine-crossover` + Rosetta 2.

  ## Why Rosetta 2

  Previous attempts to play FAF on Apple Silicon used Parallels + Windows-on-ARM, which desyncs in multiplayer    
  because Microsoft's x86 emulator uses 64-bit floats instead of proper 80-bit x87 precision. Rosetta 2 includes a
   full software x87 FPU with 80-bit extended precision; at `_PC_24` (single precision, which Supreme Commander   
  sets via `_controlfp`) IEEE 754 exactly specifies add/sub/mul/div/sqrt, making results bit-identical to x86     
  hardware and prevents desync (which has prevented Mac/Arm support in the past).

  ## Changes

  - **`DiscordEventHandler`** — adds `@Profile("!" + PROFILE_MAC)`. The class extends `DiscordEventHandlers` from 
  the `discord-rpc` JAR, whose JNA native has no arm64 macOS binary; without the annotation Spring context startup
   fails. `DiscordRichPresenceService` and `SteamService` already follow this pattern.                            

  - **`scripts/mac/faf-client.sh`** — launcher that filters `discord-rpc-*`, `steamworks4j-*`, and                
  `faf-ice-adapter.jar` from the classpath (native components lack arm64 macOS binaries / ice-adapter runs as a   
  subprocess), sets `-Djava.awt.headless=false` (Spring Boot sets `true` otherwise, which prevents JavaFX         
  startup), and activates the `prod,mac` profile.

  - **`scripts/mac/faf-run.sh`** — thin Wine wrapper used as the client's `executableDecorator`.                  

  - **`scripts/mac/README.md`** — setup (Rosetta 2, wine-crossover, winetricks, JDK 25 aarch64, game files,       
  `installDist -PjavafxPlatform=mac-aarch64`) and client configuration.

  - **`readme.md`** — new `### macOS (Apple Silicon)` section matching the existing Linux section's style.        

  ## Companion PR

  Depends on FAForever/uid#17, which adds macOS to the uid build matrix. Once that merges and a release is  
  tagged, mac users can download `faf-uid-macos` directly from the uid releases page.                             

  ## Known follow-up

  `installDist` currently bundles the Linux `faf-uid` binary on all platforms (`downloadUnixUid`,                 
[...]

PR #3506, opened 2026-04-20
Add mac profile via Gradle

Brutus5000:
## Summary by CodeRabbit

* **Chores**
  * Added macOS-specific runtime configuration to optimize application behavior on macOS systems.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -127,6 +127,12 @@ bootRun {
             // Needed for getting process handle via reflection
             "--add-opens=java.base/java.lang=ALL-UNNAMED",
     ]
+    if (OperatingSystem.current().isMacOsX()) {
+        // Fix HeadlessException in install4j's StartupNotification (FileOpeningHandler)
+        jvmArgs += "-Djava.awt.headless=false"
+        // Activate 'mac' profile so @Profile("!mac") beans (e.g. DiscordRichPresenceService) are excluded
+        environment "SPRING_PROFILES_ACTIVE", "prod,mac"
+    }
 }
 
 jar {


build.gradle
```diff
@@ -127,6 +127,12 @@ bootRun {
             // Needed for getting process handle via reflection
             "--add-opens=java.base/java.lang=ALL-UNNAMED",
     ]
+    if (OperatingSystem.current().isMacOsX()) {
+        // Fix HeadlessException in install4j's StartupNotification (FileOpeningHandler)
+        jvmArgs += "-Djava.awt.headless=false"
```
Sheikah45, line 132: Could this be wanted across oses?

Brutus5000: Benefit is a big word. I asked the AI

>This is a system property that tells the JVM there's no display, keyboard, or mouse available. When set to true, the AWT subsystem initializes in a restricted mode that skips native windowing system dependencies.
>
>**The general rule**
If your app is JavaFX-only and doesn't intentionally use AWT windows, setting -Djava.awt.headless=true is a reasonable default on all platforms. It makes your intent explicit — "AWT is not driving the UI here" — and prevents accidental AWT initialization from causing subtle issues. The only reason to not set it is if you deliberately mix AWT/Swing components with JavaFX (e.g. via SwingNode or JFXPanel), in which case AWT must be fully initialized.

jfuruness:
Do we want to work this into here? https://github.com/FAForever/downlords-faf-client/pull/3504
I don't think I've gotten a review for it yet

Brutus5000:
I looked at your PR but I was busy fixing my faf-uid rewrite in Go that was unfortunately already 98% done when you started working on the original uid repo.

I haven't worked with wine-crossover yet, so I will need some more time. But I am heavily invested in getting this running, since I am using an M2 Pro almost exclusively in the last 2 years.

But to answer your questions: this PR works standalone, it just fixes the bootRun script so the client launches at all.

jfuruness:
Thanks a bunch! I really appreciate you putting the time into this and take your time, no rush. Maybe it's the classic "it works on my machine" but I hope it's easy to use :)

build.gradle
Brutus5000: @Sheikah45 Do we use any AWT features? I was thinking of the traybar...

Sheikah45: The tray icon and the imgur uploader use awt

Sheikah45: and IdenticonUtil

Brutus5000: Then I would only do it for macos

PR #3519, opened 2026-05-16
Map blacklist toggle

Deribus:
Lets you reveal hidden maps without having to delete the blacklist entries and add them back in later

Before:
image

After:
image

## Summary by CodeRabbit

* **New Features**
  * Added an option to enable or disable the map folder name blacklist in custom games filtering.
* **Bug Fixes**
  * When the map folder name blacklist is disabled, filtering now correctly passes all entries without applying the blacklist.
* **Refactor**
  * Improved how filter updates propagate to keep predicate refreshes responsive.
* **Chores**
  * Updated i18n text and development configuration/ignore rules.

### DIFF
--- a/src/main/java/com/faforever/client/filter/MutableListFilterController.java
+++ b/src/main/java/com/faforever/client/filter/MutableListFilterController.java
@@ -7,11 +7,9 @@ import com.faforever.client.preferences.ui.RemovableListCell;
 import com.faforever.client.theme.UiService;
 import com.faforever.client.ui.list.NoFocusModelListView;
 import com.faforever.client.ui.list.NoSelectionModel;
-import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.ListProperty;
 import javafx.beans.property.SimpleListProperty;
 import javafx.collections.FXCollections;
-import javafx.scene.control.CheckBox;
 import javafx.scene.control.ListView;
 import javafx.scene.control.MenuButton;
 import javafx.scene.control.TextField;
@@ -22,7 +20,6 @@ import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
 import java.util.List;
-import java.util.function.BiFunction;
 
 @Component
 @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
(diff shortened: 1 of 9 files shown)


src/main/java/com/faforever/client/filter/MutableListFilterController.java
```diff
@@ -50,6 +56,25 @@ protected void onInitialize() {
     listView.setCellFactory(param -> new RemovableListCell<>(uiService, fxApplicationThreadExecutor));
   }
 
+  @Override
+  public void registerListener(BiFunction<List<String>, T, Boolean> filter) {
+    this.filterFunction = filter;
+    JavaFxUtil.addAndTriggerListener(valueProperty(), observable -> updatePredicate());
+    JavaFxUtil.addAndTriggerListener(enableToggle.selectedProperty(), observable -> updatePredicate());
+  }
```
Sheikah45: This method doesn't make any sense. Why are we setting a filter in a method called register listener? Additionally this will register the listeners multiple times when it only should happen once. And the new subscribe methods should be used instead of the JavaFxUtil.addAndTriggerlistener.

src/main/java/com/faforever/client/filter/MutableListFilterController.java
```diff
@@ -50,6 +56,25 @@ protected void onInitialize() {
...
+    if (enableToggle.isSelected()) {
+      predicateProperty().set(item -> filterFunction.apply(getValue(), item));
+    } else {
+      predicateProperty().set(item -> true);
+    }
+  }
+
+  public BooleanProperty enabledProperty() {
+    return enableToggle.selectedProperty();
+  }
```
Sheikah45: Also it doesn't make sense for the enable to be done at this level. The whole filter is turned off or on so that enablement should be at the CustomGameFilter level.

.claude/settings.local.json
Sheikah45, line 1: settings.local files shouldn't be committed as they are per developer overrides not project based overrides

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #3523, opened 2026-05-30
Fixing the TMM search after reconnect

GodFuper:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved team matchmaking reconnection: after connecting, the client now re-attempts to (re)join the previously selected queues only when the party is actively searching and the client is the party owner.
* **Refactor**
  * Centralized the “party owner” check to ensure consistent queue leave/join behavior across matchmaking state updates.


Sheikah45:
Why the call too leave queues?

PR #3534, opened 2026-06-22
Add multiple map generation

GodFuper:
https://github.com/user-attachments/assets/3472922f-a6e6-426a-9c82-f465aaeae093

## Summary by CodeRabbit

* **New Features**
  * Added multi-map generation with a generated-map selection dialog featuring preview cards and metadata tooltips.
  * Introduced a configurable default map-count for multi-map generation, with progress feedback during generation.
  * Added localized seed prompting and map-selection UI text, including improved empty-state messaging.
* **Bug Fixes**
  * Generated-map selection now reliably updates the active map choice.
  * Improved handling of generation results, including clearer progress/termination behavior and warnings when no maps are returned.


Sheikah45 (changes requested):
There are a few things wrong here that need to be fixed before I do a deeper review.

1. Platform.runLater should not be used directly. We have the fxThreadExecutor that should be used. Additionally it should be used as a dedicated publisher in a reactor chain for the fx parts or an executor in the complete able future chain.

2. Any text shown to the user in the client needs to be retrieved using proper i18n.

3. The map generator natively supports multiple map generation with the --num-to-generate option this should be used. Also parallelizing map generation doesn't make sense since the map generation process itself already is parallel and will use all available cores.

4. There should be no inline styling everything should be done with class names and put into the CSS files.

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -61,9 +67,12 @@ public class GenerateMapController extends NodeController<Pane> {
   private final MapGeneratorService mapGeneratorService;
   private final I18n i18n;
   private final GeneratorPrefs generatorPrefs;
+  private final UiService uiService;
+  private final FxApplicationThreadExecutor fxApplicationThreadExecutor;
 
   public Pane generateMapRoot;
   public Button generateMapButton;
+  public Button generateMultipleMapsButton;
```
Sheikah45: There shouldn't be two buttons for generating we can just take action based on if map count is greater than 1

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -83,7 +92,13 @@ public class GenerateMapController extends NodeController<Pane> {
   public RangeSlider reclaimDensitySlider;
   public RangeSlider resourcesDensitySlider;
 
+  // UI for multiple map generation
+  public Spinner<Integer> mapCountSpinner;
+  public Label progressLabel;
+
   private Runnable onCloseButtonClickedListener;
+  @Setter
+  private CreateGameController createGameController;
```
Sheikah45: This shouldn't hold a direct reference to the create game controller a listener should be used instead.

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -242,6 +259,17 @@ private void initMapSizeSpinner() {
                       previousMapName.textProperty().isNotEmpty().or(commandLineArgsText.textProperty().isNotEmpty()));
   }
 
+  private void initMapCountSpinner() {
+    int defaultMapCount = generatorPrefs.getDefaultMapCount();
+    mapCountSpinner.setValueFactory(new IntegerSpinnerValueFactory(1, 10, defaultMapCount));
```
Sheikah45: Should allow more than 10

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -496,6 +528,51 @@ public void onSeedRerollButtonClicked() {
...
+    progressLabel.setVisible(true);
+
+    Long seed = baseOptions.seed() != null ? Long.parseLong(baseOptions.seed()) : new Random().nextLong();
+    int mapCount = baseOptions.seed() != null ? 1 : mapCountSpinner.getValue();
+    log.debug("Starting multiple map generation: {} maps", mapCount);
+
+    mapGeneratorService.generateMultipleMapsWithResults(baseOptions, mapCount, seed)
+                       .publishOn(fxApplicationThreadExecutor.asScheduler())
+                       .doOnSuccess(results -> progressLabel.setVisible(false))
+                       .doOnError(throwable -> progressLabel.setVisible(false))
```
Sheikah45: There is a doOnComplete or similar that handles success or failure

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -496,6 +528,51 @@ public void onSeedRerollButtonClicked() {
...
+                         }
+                         showMapSelectionDialog(results);
+                       }, this::handleGenerationException);
+  }
+  
+  private void showMapSelectionDialog(List<MapGenerationResult> results) {
+    MapSelectionController selectionController = uiService.loadFxml("theme/play/generate_map_selection.fxml");
+    selectionController.setMapResults(results);
+    
+    Dialog dialog = uiService.showInDialog((javafx.scene.layout.StackPane) generateMapRoot.getParent(), selectionController.getRoot(), i18n.get("game.generateMap.selection.dialogTitle"));
```
Sheikah45: This shouldn't be a blind cast no guarantee the parent is a stack pane

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -496,6 +528,51 @@ public void onSeedRerollButtonClicked() {
...
+    
+    selectionController.setOnCancelButtonClickedListener(dialog::close);
+    
+    selectionController.setOnOkButtonClickedListener(() -> {
+      dialog.close();
+      onCloseButtonClicked();
+    });
+    
+    dialog.addEventHandler(Dialog.DialogEvent.CLOSED, event -> {
```
Sheikah45: This shouldn't be on when the dialog is closed but rather when the selection controller fires some selection event

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,214 @@
...
+
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+@Component
+@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+@RequiredArgsConstructor
+@Slf4j
+@Getter
+@Setter
```
Sheikah45: Not every field in this should be accessible by getter or setter

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,214 @@
...
+  protected void onInitialize() {
+    updateMapDisplay();
+
+    okButton.disableProperty().bind(selectedResult.isNull());
+  }
+
+  public void setMapResults(List<MapGenerationResult> results) {
+    if (results == null) {
+      results = List.of();
+    }
```
Sheikah45: Just don't allow null to be passed

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,214 @@
...
+    selectedResult.set(null);
+    mapResults.setAll(results);
+    log.info("Setting {} map results for selection", results.size());
+    updateMapDisplay();
```
Sheikah45: Update map display should be a subscription on the observable list

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,214 @@
...
+    VBox card = new VBox(8);
+    card.getStyleClass().add("map-card");
+
+    ImageView previewImageView = new ImageView();
+    previewImageView.setFitWidth(PREVIEW_WIDTH);
+    previewImageView.setFitHeight(PREVIEW_HEIGHT);
+    previewImageView.setPreserveRatio(true);
+    previewImageView.setSmooth(true);
+
+    SimpleObjectProperty<Image> previewImageProperty = new SimpleObjectProperty<>();
```
Sheikah45: Why are we creating a local property to bind instead of just setting the image directly

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,214 @@
...
+    previewImageProperty.set(mapService.loadPreview(mapName, PreviewSize.LARGE));
+
+    previewImageView.setOnMouseClicked(event -> {
+      if (event.getButton() == javafx.scene.input.MouseButton.PRIMARY) {
+        Image image = previewImageProperty.get();
+        if (image != null) {
+          PopupUtil.showImagePopup(image);
+        }
+      }
+    });
```
Sheikah45: There should be a utility method for this

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,214 @@
...
+
+    card.getChildren().add(previewImageView);
+
+    if (selectedResult.get() == result) {
+      card.getStyleClass().add("selected");
+    }
```
Sheikah45: This should be a binding

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,214 @@
...
+    if (!result.isSuccess() && result.getErrorMessage().isPresent()) {
+      sb.append("\n\n").append(i18n.get("mapSelection.tooltip.error", result.getErrorMessage().get()));
+    }
+
+    return sb.toString();
+  }
+
+  public void onCancelButtonClicked() {
+    if (onCancelButtonClickedListener != null) {
+      onCancelButtonClickedListener.run();
```
Sheikah45: Can make this default to an empty runnable

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,214 @@
...
+  private final GeneratorPrefs generatorPrefs;
+  private final MapService mapService;
+
+  public Pane selectionRoot;
+  public GridPane mapsGrid;
+  public Label statusLabel;
+  public Button okButton;
+
+  private final ObjectProperty<MapGenerationResult> selectedResult = new SimpleObjectProperty<>();
+  private final ObjectProperty<MapGenerationResult> result = new SimpleObjectProperty<>();
```
Sheikah45: What is result vs selected result they don't seem different

src/main/java/com/faforever/client/map/generator/GenerateMultipleMapsTask.java
Sheikah45, line 1: This doesn't need to be a separate task type it can just be added to the main generate map task

src/main/java/com/faforever/client/map/generator/GeneratorCommand.java
```diff
@@ -39,6 +38,9 @@ public List<String> getCommand() {
     if (version.compareTo(new ComparableVersion("1")) >= 0) {
       if (commandLineArgs != null) {
         command.addAll(Arrays.asList(commandLineArgs.split(" ")));
+        if (numToGenerate != null && numToGenerate > 1) {
+          command.addAll(Arrays.asList("--num-to-generate", String.valueOf(numToGenerate)));
```
Sheikah45: Command line args are standalone all other options should be ignored or not provided.

src/main/java/com/faforever/client/map/generator/GeneratorCommand.java
```diff
@@ -101,6 +105,12 @@ public List<String> getCommand() {
         command.addAll(Arrays.asList("--reclaim-density", String.valueOf(reclaimDensity)));
       }
 
+      if (commandLineArgs != null) {
+        command.addAll(Arrays.asList(commandLineArgs.split(" ")));
+      } else if (generatorOptions != null && generatorOptions.commandLineArgs() != null) {
+        command.addAll(Arrays.asList(generatorOptions.commandLineArgs().split(" ")));
+      }
```
Sheikah45: Why was this added?

src/main/java/com/faforever/client/map/generator/GeneratorCommand.java
```diff
@@ -64,7 +66,9 @@ public List<String> getCommand() {
         return command;
       }
 
-      if (seed != null) {
+      if (numToGenerate != null && numToGenerate > 1) {
+        command.addAll(Arrays.asList("--num-to-generate", String.valueOf(numToGenerate)));
+      } else if (seed != null) {
```
Sheikah45: This doesn't need to be an if else. The check is done by the UI and this is a valid command to have seed and num to generate.

src/main/java/com/faforever/client/map/generator/MapGenerationResult.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.nio.file.Path;
+import java.util.Optional;
+
+/**
+ * Represents the result of a map generation operation.
+ * Used to track generated maps and their status.
+ */
+@Value
+public class MapGenerationResult {
```
Sheikah45: Use a record and maybe a sealed interface to represent success vs failure

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
Sheikah45, line 1: I don't really understand alot of the changes to this class with creating new options. We should be able to just use the options as is.

Additionally the base task can be used to get the maps created and maybe watch the file system. But just parsing the logs should work.

src/main/java/com/faforever/client/preferences/GeneratorPrefs.java
```diff
@@ -210,4 +211,17 @@ public IntegerProperty resourceDensityMaxProperty() {
...
+  public int getDefaultMapCount() {
+    return defaultMapCount.get();
+  }
+
+  public void setDefaultMapCount(int defaultMapCount) {
+    this.defaultMapCount.set(defaultMapCount);
+  }
+
+  public IntegerProperty mapCountProperty() {
+    return defaultMapCount;
```
Sheikah45: Make the naming consistent. This isn't the default it is the stored count used

src/main/resources/i18n/messages.properties
```diff
@@ -1338,3 +1338,16 @@ iceAdapter.connecting = Connecting to ICE adapter
...
+game.generateMap.selection.description = Choose the map
+game.generateMap.selection.noMaps = No maps generated yet
+
+mapSelection.tooltip.map=Map: {0}
+mapSelection.tooltip.seed=Seed: {0}
+mapSelection.tooltip.teams=Teams: {0}
+mapSelection.tooltip.spawnCount=Spawn Count: {0}
+mapSelection.tooltip.error=Error: {0}
+game.generateMap.seedPrompt=seed
+game.generateMap.mapNamePrompt=neroxis_map_generator_(version)_(seed)_(options)
```
Sheikah45: This isn't internationalizable since it is a pattern for the map name

Sheikah45 (changes requested):

src/main/java/com/faforever/client/game/GenerateMapController.java
GodFuper: I did it through the `Consumer<MapGenerationResult>`. If need to use another listener, then tell me an example, if it's not difficult.
I haven't had any reactive projects before, so I can be dumb. It seemed to me that Spring Events is not very useful here.

src/main/java/com/faforever/client/game/GenerateMapController.java
GodFuper: I've done 50 so far, but that's a long time. The generation takes at least 20-30 seconds.

src/main/java/com/faforever/client/game/MapSelectionController.java
GodFuper: I don't quite understand what you mean.

src/main/java/com/faforever/client/map/generator/GenerateMultipleMapsTask.java
GodFuper: GenerateMultipleMapsTask returns a List<String>, while GenerateMapTask returns only one String.
It will have to fix this functionality in many places, and PR will become even more important.

src/main/java/com/faforever/client/map/generator/GeneratorCommand.java
GodFuper: Autotests added this section of code.. I accidentally skipped this section of the code when agent(Tester) add this.

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
GodFuper: Removed the excess.

We get the list of maps from the console (When the map generator is running).

src/main/java/com/faforever/client/game/MapSelectionController.java
Sheikah45: we can create a binding for wheen the selected result is equal to the result and then have a subscription on that which add or removes the style class

src/main/java/com/faforever/client/map/generator/GenerateMultipleMapsTask.java
Sheikah45: The generateMapsTask can bee updated to return a list

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -83,7 +95,15 @@ public class GenerateMapController extends NodeController<Pane> {
   public RangeSlider reclaimDensitySlider;
   public RangeSlider resourcesDensitySlider;
 
-  private Runnable onCloseButtonClickedListener;
+  // UI for multiple map generation
```
Sheikah45: This comment isn't needed

src/main/java/com/faforever/client/game/GenerateMapController.java
```diff
@@ -488,14 +535,32 @@ public Pane getRoot() {
...
-  }
-
   public void onSeedRerollButtonClicked() {
     seedTextField.setText(String.valueOf(new Random().nextLong()));
   }
 
+  private void showMapSelectionDialog(List<MapGenerationResult> results) {
+    if (CollectionUtils.isEmpty(results)) {
+      return;
+    }
```
Sheikah45:
```suggestion
  private void showMapSelectionDialog(Collection<MapGenerationResult> results) {
    if (results.isEmpty()) {
      return;
    }
```

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,193 @@
...
+
+    if (mapResults.isEmpty()) {
+      Label noMapsLabel = new Label(i18n.get("game.generateMap.selection.noMaps"));
+      noMapsLabel.getStyleClass().add("no-maps-label");
+      mapsGrid.add(noMapsLabel, 0, 0);
+      return;
+    }
+
+    AtomicInteger rowIndex = new AtomicInteger(0);
+    AtomicInteger columnIndex = new AtomicInteger(0);
```
Sheikah45: Why are these atomicIntegers instead of just ints?

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,193 @@
...
+        rowIndex.incrementAndGet();
+      }
+
+      mapsGrid.add(mapCard, columnIndex.get(), rowIndex.get());
+      columnIndex.incrementAndGet();
+
+      mapCard.setOnMouseClicked(event -> {
+        selectedResult.set(result);
+        updateMapDisplay();
+      });
```
Sheikah45: Does this recreate the grid each time a map is clicked? There is not need to do that we just need to update the style of the one that is selected. Which we can do when we create the map card.

src/main/java/com/faforever/client/map/generator/MapGeneratorService.java
```diff
@@ -244,4 +246,44 @@ public Path getGeneratorExecutablePath(ComparableVersion defaultGeneratorVersion
...
+    log.debug("Starting multiple map generation with results: {} maps with seed {}", mapCount, seed);
+
+    return downloadGeneratorIfNecessary(defaultGeneratorVersion).then(Mono.defer(() -> {
+      GenerateMultipleMapsTask task = generateMultipleMapsTaskFactory.getObject();
+      Path generatorExecutablePath = getGeneratorExecutablePath(defaultGeneratorVersion);
+      task.setVersion(defaultGeneratorVersion);
+      task.setGeneratorExecutableFile(generatorExecutablePath);
+      task.setBaseOptions(baseOptions);
+      task.setSeed(seed);
+      task.setMapCount(mapCount);
```
Sheikah45: This can be done outside of the defer

src/main/java/com/faforever/client/preferences/GeneratorPrefs.java
Sheikah45: This still needs to be fixed

src/main/java/com/faforever/client/map/generator/GenerateMultipleMapsTask.java
Sheikah45: This still needs to be fixed

Sheikah45 (changes requested):

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,225 @@
...
+  private Runnable onOkButtonClickedListener = () -> {};
+  @Setter
+  private Runnable onCancelButtonClickedListener = () -> {};
+
+  private final ObservableMap<MapGenerationResult, VBox> mapCards = FXCollections.observableHashMap();
+
+  @Override
+  protected void onInitialize() {
+    mapResults.addListener((ListChangeListener<? super MapGenerationResult>) _ -> updateMapDisplay());
+    selectedResult.addListener((observable, oldValue, newValue) -> {
```
Sheikah45: Use subscribe not addListener

src/main/java/com/faforever/client/game/MapSelectionController.java
```diff
@@ -0,0 +1,225 @@
...
+  @Setter
+  private Runnable onOkButtonClickedListener = () -> {};
+  @Setter
+  private Runnable onCancelButtonClickedListener = () -> {};
+
+  private final ObservableMap<MapGenerationResult, VBox> mapCards = FXCollections.observableHashMap();
+
+  @Override
+  protected void onInitialize() {
+    mapResults.addListener((ListChangeListener<? super MapGenerationResult>) _ -> updateMapDisplay());
```
Sheikah45: This can be subscribe as well I believe. Or an invalidation listener.

src/main/java/com/faforever/client/map/generator/GenerateMapTask.java
```diff
@@ -112,11 +114,21 @@ protected String call() throws Exception {
...
+
+      // Extract all generated map names from log lines
+      List<String> generatedMapNames = allLogLines.stream()
+                                                  .map(MapGeneratorService.GENERATED_MAP_PATTERN::matcher)
+                                                  .filter(Matcher::find)
+                                                  .map(Matcher::group)
+                                                  .distinct()
+                                                  .toList();
+
+      log.info("Successfully generated {} map(s): {}", generatedMapNames.size(), generatedMapNames);
```
Sheikah45: You can add to the list inside of the gobble lines call instead of storing all log lines.

Sheikah45 (changes requested):
Getting close just a few more small changes

Sheikah45 (approved):

Issue #1146, opened 2019-04-04
Background image not properly displayed


micheljung:
Note the top and bottom

![image](https://user-images.githubusercontent.com/1864183/55575633-99532600-570f-11e9-9482-3ee79a73c7b7.png)

Probably introduced by #1059

1-alex98:
defenitely introduced by #1059

micheljung:
I meant #1089 which fixes #1059 

It changed from `-fx-background-size: cover, auto;` to `BackgroundSize.DEFAULT` and I am 99% sure it's caused by this

PR #3536, opened 2026-06-27
Select avatars via the REST API instead of the lobby server

Brutus5000:
## What

Avatar listing and selection now go through the FAF REST API (Elide JSON:API) instead of the legacy lobby socket commands. The API is the single writer of the selected avatar (`login.avatar_id` → `player.currentAvatar`) and validates that the chosen avatar is actually assigned to the player.

## Changes

- **`AvatarService.getAvailableAvatars()`** queries the current player's `AvatarAssignment`s via `FafApiAccessor` (with the `avatar` relationship included) and maps them.
- **`AvatarService.changeAvatar()`** PATCHes `player.currentAvatar` through a new `FafApiAccessor.patchToOneRelationship(...)` helper that targets the JSON:API relationship endpoint (`PATCH /data/player/{id}/relationships/currentAvatar`). This is used for both setting and clearing — clearing sends `{"data": null}`, which a DTO PATCH cannot express because the serializer is configured with `Include.NON_NULL`.
- **Avatar picker** preselects the active avatar by `url` rather than full record equality, because the current avatar may originate from the lobby feed (no `id`) while available avatars now come from the API (with `id`). This also makes "no avatar" preselect correctly.
- Removed the now-dead lobby avatar list/select code (`FafServerAccessor`, `AvatarMapper.mapDtos`) and their tests.
- Bumped `faf-commons` to `20260626-2e8a5a8`, the release exposing `Player.currentAvatar` / deprecating `AvatarAssignment.selected`.

## Testing

- `./gradlew compileJava` clean (JDK 25).
- `AvatarServiceTest` rewritten against a mocked `FafApiAccessor` (list filter on `player.id`, set + clear PATCH); picker and `ServerAccessorTest` pass. 51 tests green.

Depends on the API-side change (faf-java-api PR #1146) that makes `Player.currentAvatar` PATCH-able and server-validated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Enabled requesting the avatar relationship include and added support for retrieving the player’s currently equipped avatar to drive UI selection.

* **Bug Fixes**
  * Improved avatar picker reliability by loading available avatars and the current avatar together, selecting the correct option, and safely falling back to “no avatar” on missing data or errors.
  * Clearing an avatar now reliably removes the current selection via the appropriate relationship update flow; selecting an avatar updates the player’s current avatar remotely.

* **Chores**
  * Updated the bundled shared library version to keep dependencies current.

### DIFF
--- a/src/main/java/com/faforever/client/api/FafApiAccessor.java
+++ b/src/main/java/com/faforever/client/api/FafApiAccessor.java
@@ -225,13 +225,15 @@ public class FafApiAccessor implements InitializingBean {
 
   /**
    * Removes the named member from a relationship via the JSON:API relationship endpoint
-   * ({@code DELETE /data/{type}/{id}/relationships/{name}}). Elide requires the related resource to be named
-   * in the body, so for a to-one relationship this clears it only when {@code relatedId} is the current value.
+   * ({@code DELETE /data/{type}/{id}/relationships/{name}}). The owning resource path is built from
+   * {@code navigator}; the relationship segment is appended since the navigator only models the
+   * related-resource path. Elide requires the related resource to be named in the body, so for a
+   * to-one relationship this clears it only when {@code relatedId} is the current value.
    */
-  public Mono<Void> deleteToOneRelationship(String type, String id, String relationshipName,
-                                            String relatedType, String relatedId) {
+  public Mono<Void> deleteFromRelationship(ElideNavigatorOnId<?> navigator, String relationshipName,
+                                           String relatedType, String relatedId) {
     RelationshipDocument body = new RelationshipDocument(new ResourceIdentifier(relatedType, relatedId));
-    String endpointPath = "/data/" + type + "/" + id + "/relationships/" + relationshipName;
+    String endpointPath = navigator.build() + "/relationships/" + relationshipName;
     return retrieveMonoWithErrorHandling(Void.class, apiWebClient.method(HttpMethod.DELETE)
         .uri(endpointPath)
         .contentType(MediaType.parseMediaType(JSONAPI_MEDIA_TYPE))
--- a/src/main/java/com/faforever/client/api/FafApiAccessor.java
+++ b/src/main/java/com/faforever/client/api/FafApiAccessor.java
@@ -33,8 +33,10 @@ import com.faforever.commons.api.elide.ElideEndpointBuilder;
 import com.faforever.commons.api.elide.ElideEntity;
 import com.faforever.commons.api.elide.ElideNavigatorOnCollection;
 import com.faforever.commons.api.elide.ElideNavigatorOnId;
+import com.faforever.commons.api.elide.ElideNavigatorOnRelationshipLink;
 import com.faforever.commons.io.ByteCountListener;
 import com.github.jasminb.jsonapi.JSONAPIDocument;
+import com.github.jasminb.jsonapi.annotations.Type;
 import com.github.jasminb.jsonapi.exceptions.ResourceParseException;
 import com.github.rutledgepaulv.qbuilders.builders.QBuilder;
 import com.github.rutledgepaulv.qbuilders.conditions.Condition;
@@ -225,15 +227,14 @@ public class FafApiAccessor implements InitializingBean {
 
   /**
    * Removes the named member from a relationship via the JSON:API relationship endpoint
-   * ({@code DELETE /data/{type}/{id}/relationships/{name}}). The owning resource path is built from
-   * {@code navigator}; the relationship segment is appended since the navigator only models the
-   * related-resource path. Elide requires the related resource to be named in the body, so for a
+   * ({@code DELETE /data/{type}/{id}/relationships/{name}}). The related resource type is derived from the
+   * navigator's entity type. Elide requires the related resource to be named in the body, so for a
    * to-one relationship this clears it only when {@code relatedId} is the current value.
    */
-  public Mono<Void> deleteFromRelationship(ElideNavigatorOnId<?> navigator, String relationshipName,
-                                           String relatedType, String relatedId) {
+  public Mono<Void> deleteFromRelationship(ElideNavigatorOnRelationshipLink<?> relationshipLink, String relatedId) {
+    String relatedType = relationshipLink.getDtoClass().getAnnotation(Type.class).value();
     RelationshipDocument body = new RelationshipDocument(new ResourceIdentifier(relatedType, relatedId));
-    String endpointPath = navigator.build() + "/relationships/" + relationshipName;
+    String endpointPath = relationshipLink.build();
     return retrieveMonoWithErrorHandling(Void.class, apiWebClient.method(HttpMethod.DELETE)
         .uri(endpointPath)
         .contentType(MediaType.parseMediaType(JSONAPI_MEDIA_TYPE))
--- a/src/main/java/com/faforever/client/api/FafApiAccessor.java
+++ b/src/main/java/com/faforever/client/api/FafApiAccessor.java
@@ -5,6 +5,7 @@ import com.faforever.client.config.ClientProperties.Api;
 import com.faforever.client.io.CountingFileSystemResource;
 import com.faforever.client.login.TokenRetrievalException;
 import com.faforever.commons.api.dto.ApiException;
+import com.faforever.commons.api.dto.AvatarAssignment;
 import com.faforever.commons.api.dto.Clan;
 import com.faforever.commons.api.dto.CoopResult;
 import com.faforever.commons.api.dto.CoopScenario;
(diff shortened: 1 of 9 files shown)


src/main/java/com/faforever/client/mapstruct/AvatarMapper.java
```diff
@@ -17,6 +15,4 @@ public interface AvatarMapper {
   com.faforever.commons.api.dto.Avatar map(Avatar bean);
 
   Avatar map(Player.Avatar dto);
-
-  List<Avatar> mapDtos(List<Player.Avatar> dtos);
```
Sheikah45, line 21: Why was this removed?

src/main/java/com/faforever/client/api/FafApiAccessor.java
```diff
@@ -220,6 +222,26 @@ public <T extends ElideEntity> Mono<Void> patch(ElideNavigatorOnId<T> navigator,
...
+   * DTO serializer (configured with {@code Include.NON_NULL}) cannot express.
+   *
+   * @param relatedId the related resource id, or {@code null} to clear the relationship
+   */
+  public Mono<Void> patchToOneRelationship(String type, String id, String relationshipName,
+                                           String relatedType, String relatedId) {
+    String body = relatedId == null
+        ? "{\"data\":null}"
+        : "{\"data\":{\"type\":\"" + relatedType + "\",\"id\":\"" + relatedId + "\"}}";
+    String endpointPath = "/data/" + type + "/" + id + "/relationships/" + relationshipName;
```
Sheikah45: This should change we don't want to be concatenating strings and should use the jackson and built in serialization and change that if needed instead of this.

Sheikah45: Also why are we changing player instead of the avatar assignment object?

Sheikah45 (changes requested):

src/main/java/com/faforever/client/api/FafApiAccessor.java
Sheikah45: Also for removing the current avatar shouldn't that be handled by issuing a delete instead of a patch anyway?

Brutus5000: Done in 1844dafc — removed the string building. Setting now goes through the existing typed `fafApiAccessor.patch()` with a `Player` DTO carrying `currentAvatar` (jasminb serialises the relationship), and clearing uses a small typed JSON:API record serialised by Jackson instead of a hand-built string.

Brutus5000: Because the selected avatar is now the `login.avatar_id` FK, exposed as `player.currentAvatar`, and that's the single source of truth (merged in #1146). The old `avatarAssignment.selected` flag is deprecated and no longer drives anything — its update/publish hook lives on `player.currentAvatar` — so patching the assignment wouldn't change what's displayed. The assignment is still used for *which* avatars a player owns (the available list).

Brutus5000: Switched to DELETE in 1844dafc. One caveat worth noting: Elide's relationship DELETE requires the member to be named in the body (`{"data":{"type":"avatar","id":...}}`) — a null/empty body is rejected — so the service resolves the current avatar first and then DELETEs that id. (A `PATCH {"data":null}` would clear without knowing the id, but DELETE is the more conventional verb here as you noted.)

src/main/java/com/faforever/client/mapstruct/AvatarMapper.java
Brutus5000: `mapDtos` only mapped the lobby `Player.Avatar` *list*, which was used solely by the old lobby-based `getAvailableAvatars()`. That path is gone — the available list now comes from the API as `commons.api.dto.Avatar` (mapped by the existing `map(...)`). The single `map(Player.Avatar)` is kept on purpose, since chat still shows other players' avatars from the lobby feed.

src/main/java/com/faforever/client/api/FafApiAccessor.java
```diff
@@ -220,6 +225,27 @@ public <T extends ElideEntity> Mono<Void> patch(ElideNavigatorOnId<T> navigator,
...
+   * Removes the named member from a relationship via the JSON:API relationship endpoint
+   * ({@code DELETE /data/{type}/{id}/relationships/{name}}). The related resource type is derived from the
+   * navigator's entity type. Elide requires the related resource to be named in the body, so for a
+   * to-one relationship this clears it only when {@code relatedId} is the current value.
+   */
+  public Mono<Void> deleteFromRelationship(ElideNavigatorOnRelationshipLink<?> relationshipLink, String relatedId) {
+    String relatedType = relationshipLink.getDtoClass().getAnnotation(Type.class).value();
+    RelationshipDocument body = new RelationshipDocument(new ResourceIdentifier(relatedType, relatedId));
+    String endpointPath = relationshipLink.build();
+    return retrieveMonoWithErrorHandling(Void.class, apiWebClient.method(HttpMethod.DELETE)
```
Sheikah45, line 238: There is a delete method on the Web client so you don't need to use method.

src/main/java/com/faforever/client/api/FafApiAccessor.java
```diff
@@ -220,6 +225,27 @@ public <T extends ElideEntity> Mono<Void> patch(ElideNavigatorOnId<T> navigator,
...
+        .uri(endpointPath)
+        .contentType(MediaType.parseMediaType(JSONAPI_MEDIA_TYPE))
+        .bodyValue(body)).doOnSuccess(aVoid -> log.trace("Deleted relationship member {} at {}", relatedId, endpointPath));
+  }
+
+  /** Minimal JSON:API relationship payload (`{"data": {"type": ..., "id": ...}}`) for relationship endpoints. */
+  record RelationshipDocument(ResourceIdentifier data) {}
+
+  record ResourceIdentifier(String type, String id) {}
+
```
Sheikah45, line 248: Couldn't this be an jsonApiDocument?

src/main/java/com/faforever/client/avatar/AvatarService.java
```diff
@@ -35,11 +43,60 @@ public Image loadAvatar(Avatar avatar) {
...
+
+  public CompletableFuture<Avatar> getCurrentAvatar() {
+    String playerId = String.valueOf(playerService.getCurrentPlayer().getId());
+    ElideNavigatorOnId<Player> navigator = ElideNavigator.of(Player.class)
+        .id(playerId)
+        .addInclude("currentAvatar");
+    return fafApiAccessor.getOne(navigator)
+        .mapNotNull(Player::getCurrentAvatar)
+        .map(avatarMapper::map)
+        .toFuture();
```
Sheikah45: Why are we getting the current avatar from the API here?

Elsewhere we are waiting for the server to tell us so why not use that?

src/main/java/com/faforever/client/api/FafApiAccessor.java
Brutus5000: The delete method does not allow adding a body. For whatever reason it doesn't work without a body on Elide side.

src/main/java/com/faforever/client/api/FafApiAccessor.java
Brutus5000: It's a generic JasminB type, you would need to pass a valid JasminB registered object to make it work. So, no it does not really work.

src/main/java/com/faforever/client/avatar/AvatarService.java
Brutus5000: There seems to be a timing issue on login. When I login, the avatar selection combobox is always empty, regardless of whether I have an avatar selected or not. This fixes it.

Sheikah45: How does it work now? wouldn't it be the same thing?

src/main/java/com/faforever/client/api/FafApiAccessor.java
Sheikah45: Bah that is annoying

src/main/java/com/faforever/client/api/FafApiAccessor.java
Sheikah45: Oh limitation of spring

src/main/java/com/faforever/client/avatar/AvatarService.java
Sheikah45: I think it might have more to do with an avatar from the api and avatar from the lobby server not being equal to one another due to the id. The selected avatar is shown by comparing the available ones to the current one with an equality check which matches with what you were saying of it always being blank.

Sheikah45: Which should be fixed

PR #3542, opened 2026-07-17
Correcting sorting after initializing games

GodFuper:
Problem:
When the "Custom games" tab is opened for the first time, sorting is not applied. 
Check:
1. Open the "Custom games" tab.
2. Select the sort by any column.
3. Open any other tab.
4. Open the "Custom games" tab again
**The list is not sorted in the first seconds**

This fix fixes it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved game list initialization so the saved sorting order is reliably applied when the games table loads.
  * Added a deferred refresh of the table sorting on the UI thread after restoring the last sort settings, ensuring the displayed order matches the remembered state.

### DIFF
--- a/src/main/java/com/faforever/client/game/GamesTableController.java
+++ b/src/main/java/com/faforever/client/game/GamesTableController.java
@@ -113,6 +113,7 @@ public class GamesTableController extends NodeController<Node> {
     gamesTable.setItems(sortedList);
 
     applyLastSorting(gamesTable);
+    fxApplicationThreadExecutor.runLater(gamesTable::sort);
     gamesTable.setOnSort(this::onColumnSorted);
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/faforever/client/game/GamesTableController.java
```diff
@@ -113,6 +113,7 @@ public void initializeGameTable(ObservableList<GameInfo> games, Function<String,
     gamesTable.setItems(sortedList);
 
     applyLastSorting(gamesTable);
+    fxApplicationThreadExecutor.runLater(gamesTable::sort);
```
Sheikah45: I would just move this into the applyLastSort function

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #3556, opened 2026-08-23
Proposal: crash-containment fix for xactengine2_9.dll access violations (draft, seeking feedback)

baumannalexj:
## What this is

A community-found and root-caused fix for a long-standing crash: repeated `ACCESS_VIOLATION` crashes during play (and occasionally on exit), several minutes into a match, since forever. Confirmed the root cause via static analysis (Ghidra) and live testing — this fix has caught and contained real crashes during actual play sessions today.

## Root cause

Component: `xactengine2_9.dll`, Microsoft's XACT audio engine (DirectX Aug 2007 redistributable, not an OS file — same version regardless of Windows version). Three known fault sites, all traced to one wave-bank reload function (`FUN_0041d60e`):

- A missing-lock **race condition**: the reload function correctly frees+nulls a resource pointer under its own lock, but two other functions read that same pointer without ever acquiring that lock — depending on timing this manifests as a **use-after-free** or a **null-pointer dereference**.
- A separate **iterator-invalidation** bug: a linked-list walk calls the same reload function on every node, which can free/mutate nodes later in the list that the walk hasn't reached yet.

## The fix

A proxy `BugSplat.dll` (the client already imports this by name from its own bin directory — confirmed via `GameBinariesUpdateTaskImpl.java`'s `BINARIES_TO_COPY` list) forwards all 19 real BugSplat exports transparently, and additionally installs 3 targeted hooks (MinHook) at the known fault offsets inside `xactengine2_9.dll` once it loads. Each hook wraps the risky call in a recovery point (GCC doesn't support MSVC's `__try`/`__except`, so this uses `setjmp`/`longjmp` from inside a Vectored Exception Handler instead) — a caught fault means that one sound/cue operation gets skipped and logged, instead of crashing the whole game.

The actual protection logic is split out (`xact_guards.c`) from the injection vector (`dllmain.c`) — a standalone `xact_guards.dll` build target also exists with zero BugSplat dependency, in case another loader (e.g. [FADeepProbe](https://github.com/faforever/fadeepprobe), which already launches the game as a debugged child process) ever wanted to inject it directly instead.

## What's NOT done here (deliberately)

This is added under `contrib/xact-crash-fix/` as a standalone, buildable proposal — **not wired into the client's actual binary-update path.** See `contrib/xact-crash-fix/INTEGRATION.md` for the concrete integration point I found (a small addition to `GameBinariesUpdateTaskImpl.java`) and why I didn't just implement it: it needs an opt-in mechanism (this patches behavior around a Microsoft-owned DLL, not something in this codebase, and shouldn't be silently forced on everyone), and ideally a runtime version check against the on-disk `xactengine2_9.dll` so a mismatched DLL version fails safe instead of hooking wrong offsets — that's not yet verified across different players' machines, only confirmed on mine plus one matching community bug report.

## Why draft

I don't have visibility into whether a native DLL hook is something this project wants to take on at all, or how you'd want it gated/shipped if so — opening this as a concrete starting point for that conversation rather than presenting it as finished/mergeable. Happy to iterate based on feedback, including moving this out of the Java repo entirely if that's preferred (standalone tool + a link from here, for example).

### DIFF
--- a/contrib/xact-crash-fix/src/flags.h
+++ b/contrib/xact-crash-fix/src/flags.h
@@ -1,5 +1,7 @@
 #ifndef XACT_FLAGS_H
 #define XACT_FLAGS_H
+#include <windows.h>
+#include "proxylog.h"
 
 /* Strategy pattern: each guarded site's runtime behavior is one of three
  * interchangeable strategies, selected via an external config file (no
@@ -17,14 +19,23 @@ typedef struct {
     GuardStrategy hook1D751;  /* use-after-free / dangling-pointer teardown site (§2.7) */
     GuardStrategy hook1D212;  /* null pointer-chain-hop site (§2.7) */
     GuardStrategy hook1DE46;  /* linked-list iterator-invalidation site (§2.7c) */
+
+    /* Logging - see proxylog.h's ProxyLogConfigure(), which these get
+     * passed to verbatim after parsing. */
+    LogLevel logLevel;             /* LOG_LEVEL= */
+    BOOL     logDebugStringEnabled; /* LOG_DEBUGSTRING= (true/false) */
+    char     logFileName[MAX_PATH]; /* LOG_FILE= - bare filename, "" = file sink off
+                                      * (unless /logxactguards overrides it, see proxylog.c) */
 } XactFlags;
 
 extern XactFlags g_flags;
 
-/* Reads C:\Users\admin\Desktop\supreme commander crash debug\xact_proxy_flags.ini
- * if present; any missing/unrecognized key keeps its default (STRAT_CONTAIN -
- * i.e. full protection - so a missing or malformed config file never silently
- * turns protection off). Writes out a template file if none exists yet. */
+/* Reads "<dir this module loaded from>\xact_proxy_flags.ini" if present;
+ * any missing/unrecognized key keeps its default (STRAT_CONTAIN for the
+ * hooks - i.e. full protection - so a missing or malformed config file
+ * never silently turns protection off). Writes out a template file if
+ * none exists yet. Also calls ProxyLogConfigure() with the parsed logging
+ * settings before returning. */
 void LoadXactFlags(void);
 
 #endif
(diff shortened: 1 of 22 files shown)


contrib/xact-crash-fix/src/flags.h
```diff
@@ -0,0 +1,30 @@
+#ifndef XACT_FLAGS_H
+#define XACT_FLAGS_H
+
+/* Strategy pattern: each guarded site's runtime behavior is one of three
+ * interchangeable strategies, selected via an external config file (no
+ * rebuild needed to switch) - read once per game launch. */
+typedef enum {
+    STRAT_DISABLED    = 0, /* hook not installed at all - identical to pre-wrapper behavior for this site */
+    STRAT_PASSTHROUGH = 1, /* hook installed, but detour just calls the original with no try/catch -
```
baumannalexj, line 11: rename these, full descriptive names enum XACT_GUARD_DISABLED, XACT_GUARD_ENABLED, XACT_GUARD_ENABLED_WITH_LOGGING

contrib/xact-crash-fix/src/flags.h
```diff
@@ -0,0 +1,30 @@
...
+                             * useful for isolating "does the hook itself change anything" from
+                             * "does the recovery logic work", since a crash here still crashes
+                             * the process same as if unhooked, but still gets logged by the VEH */
+    STRAT_CONTAIN     = 2  /* full setjmp/longjmp recovery - the original wrapper behavior */
+} GuardStrategy;
+
+typedef struct {
+    GuardStrategy hook1D751;  /* use-after-free / dangling-pointer teardown site (§2.7) */
+    GuardStrategy hook1D212;  /* null pointer-chain-hop site (§2.7) */
+    GuardStrategy hook1DE46;  /* linked-list iterator-invalidation site (§2.7c) */
```
baumannalexj, line 21: what is guard strategy? instead of comments can we name the guard strategy something more descriptive you can keep the hook address in the name

4z0t:
Consider looking into here. https://github.com/FAForever/FA-Binary-Patches

baumannalexj:
> Consider looking into here. https://github.com/FAForever/FA-Binary-Patches

Thanks!

Sheikah45:
Yes ass @4z0t mentioned this repo is not the place for dll hooks. That would be the binary patches repo or a separate repository like we did with the optional debugger that gets runs. https://github.com/FAForever/FADeepProbe

